PR #2914, opened 2022-08-19
add client_id to initialize/identify logs

mayberryzane:
- add clientId to initialize/identify logs
- log on processPayload once for every 100 requests (1, 101, etc)


Vadman97:
stale

PR #3123, opened 2022-10-04
dedupe env var declarations

et:


Vadman97:
@et do we still need this? we get the env vars from `turbo.json` now

et:
Definitely not! Let's close.

PR #3334, opened 2022-11-19
separate web worker from client bundle

Vadman97:
## Summary

A customer requested that our CSP was more strict than `blob:` for the `worker-src` which would
allow any of their js code to create web workers.
Have the client bundle the web worker as another file hosted on static.highlight.io.

Local deploy continues to inline the web worker because of CORS.
Preview environment should use 5.1.3 highlight.run and use the separated `web-worker-0.js` file.

Bumped version and updated docs.


Vadman97:
inlined in #5282

PR #3578, opened 2023-01-23
fix(HIG-3688): popover for sessions and errors

aptlin:
## Summary

Screenshot 2023-01-23 at 2 40 29 PM

local click test

no


frontend/src/components/KeyboardShortcutsEducation/KeyboardShortcutsEducation.tsx
```diff
@@ -377,11 +377,11 @@ export const ErrorsKeyboardShortcuts: ShortcutItem[] = [
 	},
 	{
 		description: `Next error`,
-		shortcut: ['j'],
+		shortcut: ['k'],
 	},
 	{
 		description: `Previous error`,
-		shortcut: ['k'],
+		shortcut: ['j'],
```
ccschmitz, line 384: What's the reason for swapping these? `j` for moving down a list and `k` for moving up are pretty standardized. I try them in almost every app I use that presents a list of items I can interact with and there's a lot of muscle memory built up from apps like Gmail, GitHub, Linear, etc. This makes them behave the opposite of what I would expect.

ccschmitz (changes requested):
Code looks great, but I'm wondering about the behavior swap for `j` and `k` - it's the opposite of all the muscle memory I've developed using these shortcuts in other apps. Would you mind sharing more about why we're changing it?

frontend/src/components/KeyboardShortcutsEducation/KeyboardShortcutsEducation.tsx
aptlin: Thanks for catching this! Fixed in [4438ce2](https://github.com/highlight/highlight/pull/3578/commits/4438ce23f4c83dd42bbdcc81ccdb15ded550d9fd)

ccschmitz: Nice! Do we still need to revert the changes to these keyboard shortcut docs?

aptlin:
Got confused by the vim-like `down j - up k` vs our `up - down` buttons that do not match with the QWERTY / Dvorak ordering of j and k. 
Screenshot 2023-01-26 at 9 35 51 AM
Fixed now

ccschmitz (approved):

PR #3597, opened 2023-01-25
Update how we capture referral data

ccschmitz:
## Summary

* Changes logic for reading/writing referrer data to a cookie instead of localStorage. This allows us to write a cookie that can be read on the top-level domain (the marketing site) and the `app` subdomain.
* Updates the logic to fall back to `document.referrer` if there is no `ref` query param.
* Sends the referrer as a custom property to Hubspot so we can use it in reporting there as well.

Note that there is another PR (https://github.com/highlight/highlight.io/pull/263) for writing the cookie on the marketing site.

I confirmed cookies were being written/read correctly by removing the `domain` parameter, but I wasn't able to test the handoff from the marketing site to the app locally. Will test the full flow once both PRs are deployed to production.

Will need to test the handoff from marketing site to app, but shouldn't impact anything.


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -4,14 +4,14 @@ about: Create a report to help us improve
 title: ''
 labels: ''
 assignees: ''
-
```
ccschmitz, line 7: There are a few changes to `.md` files here that I believe are from prettier running after a merge commit.

aptlin (approved):
Thank you!

Vadman97:
looks like this broke the backend build but it wasn't obvious because only the first push to this PR showed the failing check
https://github.com/highlight/highlight/actions/runs/3991224557/jobs/6845787250
Screenshot 2023-01-25 at 11 19 54 AM

will improve the CI workflow and fix this in https://github.com/highlight/highlight/pull/3604

PR #3602, opened 2023-01-25
Add ability to opt out of client integrations

### LINKED ISSUE #4069: support disabling intercom integration

ccschmitz:
## Summary

Updates our tracking integrations so you can disable them and opt out of sending `track` and `identify` calls to Intercom, Mixpanel, and Amplify.

Resolves #4069.

I tested that Mixpanel didn't get called when I opted out of the integration, but I did not test Amplify or Intercom.

Will need to finalize and merge https://github.com/highlight/highlight.io/pull/277 after this lands.


sdk/firstload/package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "highlight.run",
-	"version": "5.2.3",
+	"version": "6.0.0",
```
ccschmitz, line 3: @Vadman97 do you think we should do a major version bump since this will technically break existing functionality (Intercom integration wouldn't be enabled by default)?

Vadman97: probably, but what's the reasoning for having these integrations opt-in? given no other configuration is required, i'd say that the integrations should be opt-out for specific cases where a user does not want to send their analytics metadata to highlight.

ccschmitz: That's cool with me. @aptlin surfaced this and we thought it made more sense to make them opt-in, but I can get behind making them opt-out instead 👍 

Have any opinions on how to specify it? Maybe set `H.options.integrations.[integrationName].disabled` to `true`?

ccschmitz: @Vadman97 I ended up making it so you can opt out 👍

sdk/firstload/src/index.tsx
```diff
@@ -283,23 +285,29 @@ export const H: HighlightPublicInterface = {
...
+				window.mixpanel.identify(identifier)
+			}
 		}
-		if (window.amplitude?.getInstance) {
-			window.amplitude.getInstance().setUserId(identifier)
 
-			if (Object.keys(metadata).length > 0) {
-				const amplitudeUserProperties = Object.keys(metadata).reduce(
-					(acc, key) => {
+		if (!H.options?.integrations?.mixpanel?.disabled) {
```
Vadman97, line 294: nit: same if block as 288?

sdk/firstload/src/index.tsx
```diff
@@ -217,30 +217,32 @@ export const H: HighlightPublicInterface = {
...
 
-			if (window.amplitude?.getInstance) {
-				window.amplitude.getInstance().logEvent(event, {
-					...metadata,
-					highlightSessionURL: highlightUrl,
-				})
-			}
-			if (window.Intercom) {
-				window.Intercom('trackEvent', event, metadata)
+			if (!H.options?.integrations?.intercom?.disabled) {
```
Vadman97, line 242: nit: same if as 233?

Vadman97 (approved):

sdk/firstload/src/index.tsx
ccschmitz: This one is for `intercom` instead of `amplitude`. Is that incorrect? Or do you see a refactor that could improve this?

sdk/firstload/package.json
Vadman97: sounds good!

sdk/firstload/src/index.tsx
Vadman97: oh my bad!

PR #3606, opened 2023-01-25
Send console messages to clickhouse

et:
## Summary

This PR intercepts the flow to process a console message and parses + sends the data to clickhouse.

See the note in #3531 about how to get `console.*` messages to be emitted in development.

* Verified the data landed in Clickhouse:

Screenshot 2023-01-25 at 3 02 53 PM

* Added some light unit tests around parsing

* `CLICKHOUSE_*` environment variables were added to Doppler.
* Currently, this goes to my test Clickhouse db for a proof of concept. I've only enabled it for the Highlight project. I'd like to monitor this on production data before rolling this out more broadly.
* Because this is in the kafka flow, I opted to log errors such that we wouldn't reprocess message. You'll notice in the PR that I am unmarshalling the console message byte array into a struct and I'm not totally sure if the format will always look like this.


backend/clickhouse/logs.go
```diff
@@ -0,0 +1,109 @@
...
+	Body            string
+	ProjectId       uint32
+	SecureSessionID string
+}
+
+// func (client *Client) CreateLogsTable(ctx context.Context) error {
+// 	client.conn.Exec(ctx, "DROP TABLE IF EXISTS logs") //nolint:errcheck
+
+// 	return client.conn.Exec(ctx, `
+// 	CREATE TABLE IF NOT EXISTS logs (
```
et, line 24: I manually ran this through Clickhouse cloud. 
I suspect we'll be tweaking this design (e.g. adding a trace id) so I'm leaving this here as a comment for future reference.

backend/clickhouse/logs.go
```diff
@@ -0,0 +1,109 @@
...
+	batch, err := client.conn.PrepareBatch(ctx, "INSERT INTO logs")
+
+	if err != nil {
+		return e.Wrap(err, "failed to create logs batch")
+	}
+
+	for _, message := range messagesParsed.Messages {
+		logRow, err := makeLogRow(projectID, sessionSecureID, message)
+		if err != nil {
+			continue
```
et, line 75: If parsing fails, we just skip this log.

backend/public-graph/graph/resolver.go
```diff
@@ -2671,6 +2673,7 @@ func (r *Resolver) ProcessPayload(ctx context.Context, sessionSecureID string, e
 				return e.Wrap(err, "error saving messages data")
 			}
 		} else {
+			// TODO - this code path is no longer hit.
```
Vadman97, line 2676: remove dead code?

backend/clickhouse/logs.go
```diff
@@ -0,0 +1,112 @@
...
+	} `json:"trace"`
+	Value []string `json:"value"`
+	Time  int64    `json:"time"`
+}
+
+type Messages struct {
+	Messages []Message `json:"messages"`
+}
+
+func (client *Client) BatchWriteMessagesForSession(ctx context.Context, projectID int, sessionSecureID string, messages string) error {
```
Vadman97, line 57: nit: i would prefer to have a separate function here for parsing / unmarshalling the `messages` string and doing the clickhouse batched write itself. that way, if we have a structured (typed) source of the messages, that can use the `BatchWrite` here and reuse this code path.

maybe there is a `BatchWriteMessages` that takes `[]Messages` and `BatchWriteMessagesForSession` calls `BatchWriteMessages` after unmarshalling?

Vadman97 (approved):
awesome!

backend/public-graph/graph/resolver.go
et: Was mostly leaving this as a comment to follow the logic. I can sync up with @mayberryzane and we can remove this in a follow up if he feels comfortable removing it now.

backend/clickhouse/logs.go
et: I see what you're saying. I'm just not sure when we'd be using `[]Message` outside the context of a session.
Let's see what this looks like when we add additional logging ingestions and we can tweak the design as needed.

PR #3609, opened 2023-01-26
Fix focus styles on buttons

ccschmitz:
## Summary

We've been adding styles with a `:focus` selector but these are having some undesirable side effects (see Linear issue linked below). It also results in losing the visual indicator of the active element in many of our interactive elements.

This PR restores the default styles for focusable elements which fixes the originally reported issue of confusing styles after closing a menu and restores the visual indicator to most of our interactive elements.

This video shows the changes in this PR first and then loads production and tabs through the page.

https://www.loom.com/share/fe0199d54754416fbbfbbf153974e22c

Local click test.

N/A


aptlin:
Thanks!

aptlin (approved):

PR #3613, opened 2023-01-26
command to generate clickhouse databases

et:
## Summary

We need dedicated Clickhouse cloud databases for developer usage.

This PR operates similar to how we handle [initializing](https://www.notion.so/runhighlight/Running-the-App-8dd48f388f12443692340b5a0db1c717#3f6cd8e255474546a0a0b3187331c060) dedicated opensearch indexes.

Ran `cd backend && make init-clickhouse` and confirmed I have a database.

Screenshot 2023-01-26 at 12 37 15 PM

Ran `cd backend && make init-clickhouse-logs` and confirmed I have a `logs` table.
Screenshot 2023-01-26 at 12 37 43 PM

* I will update the internal [docs](https://www.notion.so/runhighlight/Running-the-App-8dd48f388f12443692340b5a0db1c717) and ping the team that they need to run these `make` commands.
* Because this requires running a manual `make` command, I've opted to leave the prod environment to use the existing `CLICKHOUSE_DATABASE` (`default`) database to ensure there's no downtime.
* We should remove `CLICKHOUSE_DATABASE` from the `dev` config as it doesn't make sense anymore.


Vadman97 (commented):
any reason to do the database creation as a separate `handler` from the table creation? it would be easier to use if there is one clickhouse handler that is idempotent and just knows if the table has already been created, and does everything necessary

et:
Deprecating in favor of #4104 and #4103

PR #3625, opened 2023-01-29
change

### LINKED ISSUE #3931: Retention work
* Set a default retention limit for new customers
* for historical customers retention-wise
  * for free customers, set a retention limit
  * for paid, set a retenion limit and email them about it?

jay-khatri:
resolves   #3931


Vadman97 (changes requested):
why is this reintroducing sub modules?

Issue #4095, opened 2023-01-30
Search URL params aren't consistent


jay-khatri:
URL is very long / complicated and doesn't fully represent the session (ex: if I click on top users to view all their sessions, that is not a sharable URL, if I search for identifier = abc, and then click on that user's session, when I share the link the search is not there).

from discord: https://discord.com/channels/1026884757667188757/1052834249868324895/1068966443519049770

jay-khatri:
Curious if this is specific to the home page, or if there's something broken beyond that. @ccschmitz if you don't mind taking a gander.

ccschmitz:
@jay-khatri I don't think it's specific to homepage. It's happening anywhere we try to link to the session page and pass `user_properties`.

PR #4101, opened 2023-01-31
Fix deep linking to session search

### LINKED ISSUE #4098: "Top Users" deep link doesn't work.
"When I click one of our top users in the dashboard, I'm sent to a list of everyone's completed sessions – not their's."

ccschmitz:
## Summary

Deep linking to sessions with query parameters is broken in a number of places. This PR updates many of these links to stop passing any parameters besides `query` and adds some helpers for building a `query` string param that can be used in the URL and `searchParams`.

There is also some refactoring and consolidation of logic in various `useEffect`s for query builder components. This fixed a bug with URL state not being preserved on deep links and (hopefully) makes the logic a little easier to follow.

Resolves #4098, #4095, #4120

Local click test.

Want to make sure all existing behavior remains in tact and that we didn't introduce any issues with the consolidation of the logic in various `useEffect`s.


frontend/src/pages/Home/components/ActiveUsersTable/ActiveUsersTable.tsx
```diff
@@ -100,20 +100,16 @@ const ActiveUsersTable = ({
...
-									value: record.identifier,
-								},
-							],
+
+						const userParam = validateEmail(record.identifier)
+							? 'email'
+							: 'identifier'
+						history.push({
+							pathname: `/${projectIdRemapped}/sessions`,
+							search: `?query=and%7C%7Cuser_${userParam}%2Cis%2C${record.identifier}`,
```
ccschmitz, line 110: Might be something in session query builder that can generate this string.

ccschmitz: Use email address when available instead of `record.identifier`.

PR #4108, opened 2023-01-31
Add Dependabot entry for frontend directory

ccschmitz:


Vadman97:
i think [yarn v3 monorepo packages](https://github.blog/changelog/2022-10-20-dependabot-can-now-generate-security-and-version-updates-for-yarn-v2-and-v3/) should be picked up via the top level `npm` scanning 

closing, re-open if this is still relevant

Issue #4103, opened 2023-01-31
Handle migrations


et:
We need to handle clickhouse migrations.

[migrate](https://github.com/golang-migrate/migrate) looks like a decent tool and they have [support](https://github.com/golang-migrate/migrate/tree/master/database/clickhouse) for clickhouse.

PR #4109, opened 2023-01-31
add migrations for clickhouse

et:
## Summary

This PR introduces migrations for Clickhouse (#4103). 

Previously, when we were testing out logs for Highlight, we manually created a table called `logs`. This was mostly for a prof of concept.

Long term, we want our `logs` schema match the [OTEL spec](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/clickhouseexporter#logs-1) with the only modification being the Highlight specific attributes, namely `ProjectId` and `SecureSessionId`.

To ensure we can have a seamless rollout, this PR creates a new table, `logs_new` that writes and reads to this new schema (see rollout plan with how we'll transition the old data).

### Verifying migrations

Dropped data locally via http://localhost:8123/play:

```sql
drop table if exists logs_new;
drop table if exists schema_migrations;
```

Started backend via `cd backend && make start`

Observe logs:

Screenshot 2023-01-31 at 3 10 22 PM

```sql
show tables
=> 
logs_new
schema_migrations
```

### Verifying new logs schema

(See #3606 for how to get local console logs to propagate)

Verified that the new logs table is populated
Screenshot 2023-02-01 at 2 11 32 PM

Observe that even though we aren't writing all data (e.g. `TraceFlags`) and don't have a default set for that column, Clickhouse set a default for us. This is because Clickhouse has [default values](https://clickhouse.com/docs/en/sql-reference/statements/create/table/#default-values) (similar to go):

`If an expression for the default value is not defined, the default values will be set to zeros for numbers, empty strings for strings, empty arrays for arrays, and 1970-01-01 for dates or zero unix timestamp for DateTime, NULL for Nullable.`

`Map` also works (had to horizontally scroll from the above screenshot) defaulting to `{}`:
Screenshot 2023-02-01 at 2 28 05 PM

### Unit testing

Additionally, I added a basic test to verify the data passed back to the frontend in #3611. Some helpers have been added to assist with setting up a `test` Clickhouse db and handle running migrations. 

### Dev

Currently, everyone shares the same clickhouse database. Going forward, we need distinct ones. Devs can either run clickhouse [locally](https://www.notion.so/runhighlight/Environment-Setup-c8a54ced26a844afb623726e21f20af4#40d7e291d5cc408aaa4dce4273339740) or through Docker. 

In the dev Doppler config, the following changes will be made:
* `CLICKHOUSE_ADDRESS` change from `dxeqfquwoc.us-west-2.aws.clickhouse.cloud:9440` to `localhost:9000`
* `CLICKHOUSE_PASSWORD` change from "REDACTED" to "" (empty string)

### Prod

There already is a `logs` table on prod which is why the new table is named `logs_new`.

The rollout for this should be:
* Merge PR to auto-run migrations on prod. After which `logs`, `logs_new`, and `schema_migrations` should exist.
* New data should be written and read to `logs_new`. `logs` should not be used.
* Manually drop the `logs` table through Clickhouse cloud 
* Create a new PR with a new migration that renames `logs_new` to `logs` and update code references appropriately.


backend/clickhouse/logs.go
```diff
@@ -2,91 +2,36 @@ package clickhouse
...
-	} `json:"trace"`
-	Value []string `json:"value"`
-	Time  int64    `json:"time"`
-}
-
-type Messages struct {
-	Messages []Message `json:"messages"`
-}
-
-func ParseConsoleMessages(projectID int, sessionSecureID string, messages string) ([]*LogRow, error) {
```
et, line 61: I moved this to `console_messages_parser.go` to keep this top level interface light weight. If we add additional parsing for other sdks, they can go in a respective file (e.g. `python_logs_parser.go`.

backend/clickhouse/console_messages_parser.go
```diff
@@ -0,0 +1,73 @@
...
+			continue
+		}
+		logRows = append(logRows, logRow)
+	}
+	return logRows, nil
+}
+
+func makeLogRow(projectID int, sessionSecureID string, message Message) (*LogRow, error) {
+
+	// Try unquoting data from console.log
```
et, line 52: The previous [code](https://github.com/highlight/highlight/blob/59afaa33977fc81bb42c5867a81fd28e445cbfee/backend/clickhouse/logs.go#L141-L147) was trying to be smart and pluck out the timestamp.

This isn't very robust since there's no real guarantee on the order or length of arguments passed to `console.log`.

Now we just join the messages together and stuff it into `Body`.

The final output looks like this:
Screenshot 2023-02-01 at 2 14 10 PM

That `[timestamp]` prefix is coming from here:
https://github.com/highlight/highlight/blame/8311ff5036d19d6de74d80a546281f5fb92cdec9/sdk/client/src/logger.ts#L12-L16

I'm not really sure why we even do this (since something like `1675285833795` isn't really readable by a human) but it only happens when running the client with the `debug` option.

backend/clickhouse/migrations/000001_create_logs_table.up.sql
```diff
@@ -0,0 +1,27 @@
...
+    INDEX idx_res_attr_key      mapKeys(ResourceAttributes) TYPE bloom_filter(0.01) GRANULARITY 1,
+    INDEX idx_res_attr_value    mapValues(ResourceAttributes) TYPE bloom_filter(0.01) GRANULARITY 1,
+    INDEX idx_log_attr_key      mapKeys(LogAttributes) TYPE bloom_filter(0.01) GRANULARITY 1,
+    INDEX idx_log_attr_value    mapValues(LogAttributes) TYPE bloom_filter(0.01) GRANULARITY 1,
+    INDEX idx_body              Body TYPE tokenbf_v1(32768, 3, 0) GRANULARITY 1
+)
+    ENGINE = MergeTree
+        PARTITION BY toDate(Timestamp)
+        ORDER BY (ProjectId, toUnixTimestamp(Timestamp), TraceId)
+        TTL toDateTime(Timestamp) + toIntervalDay(3)
```
Vadman97, line 26: should this be 30 days?

Vadman97 (approved):

backend/clickhouse/migrations/000001_create_logs_table.up.sql
et: Good catch. Fixed in 99bfb0e81

PR #4119, opened 2023-02-01
fix(#3903): error to session ux

### LINKED ISSUE #3903: UX for viewing a session from an error
When i click into a session from an error:
- open up the session in a new url (for now, long term, we want to do a popup)
- don't render the left search feed
- specifically in the session UI, open up:
  * error in the timeline bar
  * error in the devtools window
  * error in the right panel
  * take the user to the correct time stamp

aptlin:
## Summary

Closes #3903 

Go to the error page, click on "View session" and make sure that the left panel is closed, the session is set to the error timestamp, and the right panel shows the active error

No


ccschmitz (approved):
Code + click test LGTM ✨ 

One quick thought on the UX. What do you think about opening the right panel when you click an error from the dev tools? I got confused because I was clicking errors but I couldn't see the info because the panel was closed.

https://www.loom.com/share/9cbf6d47ab374e1e8925906c2a665d30

PR #4129, opened 2023-02-02
New sign up form

ccschmitz:
## Summary

* Adds a new version of the sign up form that can be visited from the `/sign_up` route. Nothing redirects to this page yet, that will come in a future PR. 
* Updates the component generator templates to reflect our current best practices using vanilla extract.
* Adds some tracking on the about you page.
* Adds a `screen` option for `height` and `width` which will set the height and width to `100vh` and `100wh`.
* Adds support for the `type` prop on buttons so we can put them within a form without causing them to submit it.
* Fixes the line height on headings by using `leading` instead of `lineGap`, which maps more cleanly to the values we see for `line-height` in Figma.

image

Click test locally and in PR preview.

Should make sure some of the component library style changes don't throw off anything in the UI, but all the new code shouldn't really be getting any traffic yet.


frontend/src/index.scss
```diff
@@ -222,7 +222,6 @@ h4,
 h5,
 h6 {
 	font-family: var(--header-font-family) !important;
-	line-height: initial !important;
```
ccschmitz, line 225: This was causing issues with getting the correct `line-height` on headings.

packages/ui/src/components/Form/Form.tsx
```diff
@@ -50,9 +50,9 @@ export const NamedSection = ({
 	)
 }
 
-type Props = React.PropsWithChildren<{ state: AriaKitFormState }>
-export const Form: FormComponent = ({ children, state }: Props) => {
-	return <AriaKitForm state={state}>{children}</AriaKitForm>
+type Props = AriaKitFormProps
```
et, line 53: Nice clean up.

frontend/src/pages/AboutYou/AboutYouCard.tsx
```diff
@@ -136,19 +138,6 @@ const AboutYouPage = ({ onSubmitHandler }: Props) => {
 							/>
 						</div>
 					</section>
-					<section className={styles.section}>
```
et, line 139: Thank you. I hate when websites ask this.

et (approved):

PR #4133, opened 2023-02-03
add dummy simple auth for docker

### LINKED ISSUE #3596: Get rid of firebase (at least for self-hosted)

Vadman97:
## Summary

Since we don't want to require firebase to be set up for folks to self-host hobby highlight,
this change provides a dummy auth used when hosting in docker that does not require a username/password.

Local deploy in docker and outside with the new env vars set.

New backend middleware + frontend code path is only enabled if `IN_DOCKER=true` and `REACT_APP_AUTH_MODE=simple`.

Closes #3596


docker/.env
```diff
@@ -36,7 +37,3 @@ REDIS_ADDRESS=redis:6379
 REDIS_EVENTS_STAGING_ENDPOINT=redis:6379
 RENDER_PREVIEW=true
 SESSION_FILE_PATH_PREFIX=/tmp/
-
-# secrets expected to be set by the user
```
et, line 40: Nice!

et (approved):

mayberryzane (approved):

PR #4137, opened 2023-02-03
Fix snoozing errors

### LINKED ISSUE #4034: Snoozing an error + submitting an issue, should give me feedback
Right now if I click the snooze option an select a time, there's no message/pop-p saying "you've snoozed this error for 10 days" or whatever. Can we do that?

ccschmitz:
## Summary

Snoozing an error doesn't currently work because of a regression introduced in https://github.com/highlight/highlight/pull/3521/files#diff-1118372a92bc179eb5799ac9af7dfea83da9368881356f012fbf52736af7305bR65 which is resolved in this PR.

image

Resolves #4034 

Local click test.

N/A


et:
@ccschmitz - It seems unlikely but can you see if the cypress test failure is related?

Vadman97:
> @ccschmitz - It seems unlikely but can you see if the cypress test failure is related?

seems like it's broken everywhere... i'm looking into it

Vadman97 (approved):

PR #4139, opened 2023-02-03
Fix inconsistent page width on errors

### LINKED ISSUE #4064: Horizontal jankiness
[https://www.loom.com/share/a4784b40b3bf49cc85a85fb498a6cfa2](https://www.loom.com/share/a4784b40b3bf49cc85a85fb498a6cfa2)

See the last error that I view.

ccschmitz:
## Summary

The width of the error page content can be inconsistent if we are viewing an error that has a short title ([example](https://app.highlight.io/1/errors/mQVMpaHJjI3Kfvo275T4B51xifET?page=1)). This PR updates the `Container` component so it will always take up the default width (940px) when available. Once the container size gets smaller than the viewport width + padding (980ox) it goes back to `width: auto` to be responsive.

Resolves #4064

Local click test.

N/A


aptlin (approved):
Great find!

PR #4158, opened 2023-02-06
turn on logging for all projects

et:


backend/public-graph/graph/resolver.go
```diff
@@ -2702,19 +2702,17 @@ func (r *Resolver) ProcessPayload(ctx context.Context, sessionSecureID string, e
...
+		if err != nil {
+			log.WithError(err).Error("failed to parse console messages")
+		} else {
+			if err := r.BatchedQueue.Submit(&kafka_queue.Message{
+				Type: kafka_queue.PushLogs,
+				PushLogs: &kafka_queue.PushLogsArgs{
+					SessionSecureID: sessionSecureID,
+					LogRows:         logRows,
+				}}, sessionSecureID); err != nil {
+				log.WithError(err).Error("error writing console messages to clickhouse")
```
Vadman97, line 2715: in a future pr, could we move this into the public graph PushPayload handler since we are no longer depending on the project id lookup?

Vadman97 (approved):

backend/public-graph/graph/resolver.go
et: I might be misunderstanding what you're suggesting but we are still using the project id (see the first param to `ParseConsoleMessages`).

PR #4181, opened 2023-02-07
Replace integration modal with empty state CTAs

ccschmitz:
## Summary

Replaces the integration modal with a CTA in the empty state to complete setup for new accounts.

image

Note that the ENV vars referenced in this PR have been set in our dev and prod configs in Doppler.

Local click test.

This might be the last place we are linking to the demo account. We should consider cleaning up the demo account + logic for it.


et (approved):
Awesome!

PR #4182, opened 2023-02-07
fix(#3925): clean up the live mode for sessions

### LINKED ISSUE #3925: Changes to live session experience.
Make some changes to live sessions:
1) Linking to a live session should put you into "non-live" mode by default
2) Change the copy of the button from "Live" to "Switch to live mode"
3) Julianify the live loading modal

Customer feedback / background.
"- it took me a few minutes to figure out clicking on "Live" takes you to the session navigator view for live sessions"

How can we make the the Live indicator more obvious? For some reason it doesn't obviously look like a button, it just looks like an indicator. 

cc @vadim

aptlin:
## Summary

Resolves #3925:

Linking to a live session should put you into "non-live" mode by default
Change the copy of the button from "Live" to "Switch to live mode"
Julianify the live loading modal

Local click test

no


frontend/src/pages/Player/Toolbar/DevToolsWindowV2/NetworkPage/NetworkPage.tsx
```diff
@@ -419,7 +419,6 @@ const ResourceRow = ({
 						className={styles.timingBar}
 						style={{
 							width: `${actualPercent}%`,
-							zIndex: 100,
```
aptlin, line 422: To avoid this:
Screenshot 2023-02-07 at 9 08 05 AM

jay-khatri:
two comments:
- on live sessions, in the timeline, i'm often in a position where I get this sorta error (`NaN`). Is this because the session is too short?
image

- can we make this button primary in both the live and non-live state? 
image

- I could be wrong, but is the text in that ^ button vertically ccentered?

Vadman97 (approved):
nice!

PR #4185, opened 2023-02-07
improve kafka log batching and highlight-py sdk

Vadman97:
## Summary

* Allow highlight python sdk to override OTLP endpoint.
* Batch messages from multiple partitions into a shared in-memory buffer.
    * currently, each public-worker container runs 8 goroutines for the batched topic. each goroutine gets a set of subset of partitions assigned to it. as messages are consumed by the goroutines, we should aggregate the LogRows together across these partitions so that the batching can be more substantial. without this change, each goroutine had its own in-memory buffer which would not get many messages (and actually could lead to dropping messages because a goroutine would flush its partition's messages as the front of the topic, while in reality there may be other unflushed messages)
* Don't write log messages as an attribute.

Local deploy with debugger.

New highlight python sdk version.

### DIFF
--- a/backend/otel/otel.go
+++ b/backend/otel/otel.go
@@ -197,7 +197,7 @@ func (o *Handler) HandleTrace(w http.ResponseWriter, r *http.Request) {
 						eventAttributesMap := make(map[string]string)
 						for k, v := range eventAttributes {
 							vStr := castString(v, "")
-							if vStr != "" {
+							if vStr != "" && k != string(hlog.LogMessageKey) {
 								eventAttributesMap[k] = castString(v, "")
 							}
 						}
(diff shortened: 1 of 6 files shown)


backend/otel/otel.go
```diff
@@ -197,7 +197,7 @@ func (o *Handler) HandleTrace(w http.ResponseWriter, r *http.Request) {
 						eventAttributesMap := make(map[string]string)
 						for k, v := range eventAttributes {
 							vStr := castString(v, "")
-							if vStr != "" {
+							if vStr != "" && k != string(hlog.LogMessageKey) {
```
et, line 200: Can we filter out `hlog.LogSeverityKey` as well?

et (commented):
Can you add a little bit more context on `Batch messages from multiple partitions into a shared in-memory buffer.`?
Why do we need to do this again?

Vadman97:
> Can you add a little bit more context on `Batch messages from multiple partitions into a shared in-memory buffer.`? Why do we need to do this again?

good point, added to the pr description

et (approved):
Thanks! Makes sense.

PR #4192, opened 2023-02-08
add freetext searchg

et:


frontend/src/pages/LogsPage/SearchForm/SearchForm.tsx
```diff
@@ -1,29 +1,47 @@
...
 type Props = {
 	onFormSubmit: (query: string) => void
-	query: string
+	initialQuery: string
+	startDate: string
+	endDate: string
 }
 
-const SearchForm = ({ query, onFormSubmit }: Props) => {
-	const form = useFormState({
```
et, line 10: I found that `useFormState` was freezing when the passed in `query` was changing. I don't know if it doesn't like the default value getting constantly updated but I opted to remove it for now and track the state myself.

backend/clickhouse/logs.go
```diff
@@ -86,11 +87,48 @@ func (client *Client) ReadLogs(ctx context.Context, projectID int, params modelI
...
+			SeverityText:  makeSeverityText(SeverityText),
 			Body:          Body,
 			LogAttributes: gqlLogAttributes,
 		})
 	}
 	rows.Close()
 	return logLines, rows.Err()
 }
+
+func makeSeverityText(severityText string) modelInputs.SeverityText {
```
et, line 99: Ensures that every log severity text is present and fits into one of the following [types](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#severity-fields).

The client sdk sends in the types as things like `log`, `console.error`, etc: https://github.com/highlight/highlight/blob/ae909c5aa610b1f1dac9223adc423a77941e26bc/sdk/client/src/types/shared-types.tsx#L6-L12

But these have no meaning in OTEL and we want the log viewing to be consistent on the frontend.

backend/clickhouse/query_builder.go
```diff
@@ -28,16 +28,28 @@ func buildLogAttributes(userQuery string) string {
...
+		} else if len(parts) == 2 {
 			key := parts[0]
 			value := parts[1]
 			logAttributes = append(logAttributes, fmt.Sprintf("LogAttributes['%s'] = '%s'", key, value))
 		}
 	}
 
+	if len(bodyParams) > 0 {
+		wildcardBodyParams := "%" + strings.Join(bodyParams, " ") + "%"
+		logAttributes = append(logAttributes, fmt.Sprintf("Body ILIKE %s", quoteString(wildcardBodyParams)))
```
et, line 47: A little suspect here but trying to avoid a sql injection. Ideally, we'd switch to an ORM but I haven't found one for clickhouse that works with our current setup.

backend/clickhouse/logs.go
Vadman97, line 99: nice to structure this now

Vadman97 (approved):
🚢 awesome

PR #4202, opened 2023-02-08
New UI for auth screens

ccschmitz:
## Summary

Revamps the UI for signing up, creating your workspace, and signing in. Backend changes to support this were merged in #4224.

https://www.loom.com/share/7611d7a2e2b54b71b6b8bd5fbc38e0d7

```[tasklist]
### Tasks
- [x] Fix flash after submitting sign up form.
- [x] Make a separate endpoint for creating admins (currently happens as part of querying admins).
- [x] New endpoint for submitting about you page. Handles updating admin, creating workspace, and creating project.
```

Local click test + having the team click test in the PR preview.

Not really as it's all front end changes, but we should make sure that things are looking good for different states you could be within the app to ensure all the router changes are working as expected. We may also want to monitor the number of sign ups and logins we have after this deploy as a health metric.


frontend/src/components/AutoJoinEmailsInput/index.tsx
```diff
@@ -0,0 +1,117 @@
+import { useAuthContext } from '@authentication/AuthContext'
```
ccschmitz, line 1: Most of this code was copied from the `AutoJoinForm` component with some tweaks to get the UI looking the way we want.

backend/private-graph/graph/schema.resolvers.go
```diff
@@ -6224,14 +6224,12 @@ func (r *queryResolver) Admin(ctx context.Context) (*model.Admin, error) {
 			Phone:                 &firebaseUser.PhoneNumber,
 			AboutYouDetailsFilled: &model.F,
 		}
-		if err := r.DB.Create(newAdmin).Error; err != nil {
+		if err := r.DB.Where(newAdmin).FirstOrCreate(&admin).Error; err != nil {
```
ccschmitz, line 6227: I was hitting an issue where there was only 1 network request I could find hitting this method, but somehow it was being invoked 3 times and trying to call `Create` with the same attributes more than once. I'm still not sure what the culprit is there, but I figured this change couldn't hurt to make it more resilient if it gets called quickly in succession.

backend/private-graph/graph/schema.resolvers.go
```diff
@@ -6212,37 +6243,21 @@ func (r *queryResolver) WorkspaceForProject(ctx context.Context, projectID int)
...
-			Phone:                 &firebaseUser.PhoneNumber,
-			AboutYouDetailsFilled: &model.F,
-		}
-		if err := r.DB.Create(newAdmin).Error; err != nil {
-			spanError := e.Wrap(err, "error creating new admin")
-			adminSpan.Finish(tracer.WithError(spanError))
-			return nil, spanError
-		}
-		firebaseSpan.Finish()
+	firebaseSpan, _ := tracer.StartSpanFromContext(ctx, "resolver.getAdmin", tracer.ResourceName("db.createAdminFromFirebase"),
```
aptlin, line 6249: Do we need two separate spans here?

aptlin (commented):
Did the first pass, awesome work, thanks for this! Noticed a couple of issues: creating a new account redirects to a workspace, not the email verification page

https://www.loom.com/share/b41d2ed70d9348218cac68c44dcf307b

The rest are mostly nits, see comments

frontend/src/pages/Auth/SignIn.tsx
```diff
@@ -0,0 +1,141 @@
...
+			</AuthHeader>
+			<AuthBody>
+				<Stack gap="12">
+					<Form.Input
+						name={formState.names.email}
+						label="Email"
+						type="email"
+						autoFocus
+					/>
+					<Form.Input
```
aptlin, line 90: Should we add `autocomplete="new-password"` here and `autocomplete="current-password"` elsewhere? Chrome issues a warning:
Screenshot 2023-02-17 at 12 07 46 PM

frontend/src/pages/Auth/AdminForm.tsx
```diff
@@ -0,0 +1,214 @@
...
+import analytics from '@util/analytics'
+import { getAttributionData } from '@util/attribution'
+import { message } from 'antd'
+import React, { useEffect, useState } from 'react'
+import { useNavigate } from 'react-router-dom'
+
+import * as styles from './AdminForm.css'
+import * as authRouterStyles from './AuthRouter.css'
+
+const COMMON_EMAIL_PROVIDERS = ['gmail', 'yahoo', 'hotmail']
```
aptlin, line 32:
```suggestion
const COMMON_EMAIL_PROVIDERS = ['gmail', 'yahoo', 'hotmail', 'fastmail', 'protonmail', 'hey.com']
```

frontend/src/routers/AppRouter/AppRouter.tsx
```diff
@@ -19,18 +21,92 @@ import { DefaultWorkspaceRouter } from '@routers/OrgRouter/DefaultWorkspaceRoute
...
+				id,
+				name,
+				email,
+			}
+
+			if (admin.photo_url) {
+				identifyMetadata.avatar = admin.photo_url
+			}
+
+			// TODO: Move this to another router
```
aptlin, line 83: TODO

frontend/src/pages/Auth/VerifyEmail.tsx
```diff
@@ -0,0 +1,95 @@
...
+import * as styles from './AuthRouter.css'
+
+export const VerifyEmail: React.FC = () => {
+	const { setLoadingState } = useAppLoadingContext()
+	const { data, stopPolling } = useGetAdminQuery({
+		pollInterval: 1000,
+	})
+	const [loading, setLoading] = useState(false)
+	const navigate = useNavigate()
+	const isEmailVerified = data?.admin?.email_verified || false
```
aptlin, line 24:
```suggestion
	const isEmailVerified = !!data?.admin?.email_verified
```

frontend/src/pages/Auth/VerifyEmail.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+	useEffect(() => {
+		if (isEmailVerified) {
+			stopPolling()
+			navigate('/about_you')
+		} else {
+			// Show the Intercom message after 5 seconds in case the user needs help.
+			setTimeout(() => {
+				window.Intercom('update', {
```
aptlin, line 33: Should we use the Intercom helper instead?
https://github.com/highlight/highlight/blob/main/frontend/src/util/window.ts#L17

ccschmitz:
> Did the first pass, awesome work, thanks for this! Noticed a couple of issues: creating a new account redirects to a workspace, not the email verification page

Thanks for click testing locally, Sasha! Did you happen to catch what the error was that you were hitting? I click tested a lot locally and didn't have the same issue, but will see if I can repro. Also, is the error in the Look a different error than you mentioned with being redirect to the wrong page after sign up?

frontend/src/pages/Auth/AdminForm.tsx
```diff
@@ -0,0 +1,214 @@
...
+						allowed_auto_join_email_origins:
+							formState.values.autoJoinDomains,
+						promo_code: formState.values.promoCode || undefined,
+						...attributionData,
+					},
+				},
+			})
+
+			message.success(
+				`Nice to meet you ${formState.values.firstName}, let's get started!`,
```
aptlin, line 92: Can we put that inside onComplete of `updateAdminAndCreateWorkspace` and handle errors inside onError to avoid `catch (e:any)` ?

aptlin:
@ccschmitz tested again, the issue from the Loom is gone! Looks like a stale cache is to blame. Would be cool to add the backend changes in a separate PR for us to test the full sign up flow in the preview!

frontend/src/pages/Auth/AdminForm.tsx
```diff
@@ -0,0 +1,214 @@
...
+import analytics from '@util/analytics'
+import { getAttributionData } from '@util/attribution'
+import { message } from 'antd'
+import React, { useEffect, useState } from 'react'
+import { useNavigate } from 'react-router-dom'
+
+import * as styles from './AdminForm.css'
+import * as authRouterStyles from './AuthRouter.css'
+
+const COMMON_EMAIL_PROVIDERS = ['gmail', 'yahoo', 'hotmail', 'fastmail', 'protonmail', 'hey.com']
```
aptlin, line 32: Just noticed that our CI fails because this file is not formatted

```suggestion
const COMMON_EMAIL_PROVIDERS = ['gmail', 'yahoo', 'hotmail', 'fastmail', 'protonmail', 'hey.com'] as const
```

aptlin (approved):
Looks so much better, thank you!

backend/private-graph/graph/schema.resolvers.go
ccschmitz: Good catch - I don't think we do.

frontend/src/pages/Auth/SignIn.tsx
ccschmitz: Set this on all the email and password fields here. Good call!

frontend/src/pages/Auth/AdminForm.tsx
ccschmitz: I looked at this, but wouldn't we still end up with an `any` type on the error?

image

ccschmitz:
> Would be cool to add the backend changes in a separate PR for us to test the full sign up flow in the preview!

I agree! I actually thought about this ([#4318](https://github.com/highlight/highlight/pull/4318)) but we can't ship that change in isolation because of how we move the create admin logic. We could ship the new endpoint without any changes to the `Admin` resolver, but we would still be hitting that old code path in the PR preview click test.

PR #4204, opened 2023-02-08
make otel process stacktraces into structured format

Vadman97:
## Summary

* Makes our otel errors ingest parse stack traces into a structured format that the frontend can render.
* Resolves a noisy log `PushMetricsImpl called without secureID` from session-less metrics (not yet supported).

New unit test, full stack deploy.

No.

### DIFF
--- a/backend/otel/parse_test.go
+++ b/backend/otel/parse_test.go
@@ -30,6 +30,7 @@ func TestFormatStructureStackTrace(t *testing.T) {
 				assert.NotNil(t, frame)
 				assert.NotNil(t, frame.FileName)
 				assert.GreaterOrEqual(t, len(*frame.FileName), 1)
+				assert.Equal(t, input.expectedFrameError, *frame.Error)
 			}
 		})
 	}
--- a/backend/otel/parse_test.go
+++ b/backend/otel/parse_test.go
@@ -1,0 +1,36 @@
+package otel
+
+import (
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func FuzzFormatStructureStackTrace(f *testing.F) {
+	f.Fuzz(func(t *testing.T, stackTrace string) {
+		formatStructureStackTrace(stackTrace)
+	})
+}
+
+func TestFormatStructureStackTrace(t *testing.T) {
+	var inputs = []struct {
+		language           string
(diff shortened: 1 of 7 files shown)


backend/otel/parse_test.go
```diff
@@ -0,0 +1,36 @@
...
+	f.Fuzz(func(t *testing.T, stackTrace string) {
+		formatStructureStackTrace(stackTrace)
+	})
+}
+
+func TestFormatStructureStackTrace(t *testing.T) {
+	var inputs = []struct {
+		language           string
+		stacktrace         string
+		expectedFrameError string
```
et, line 18: What are we using this for? Should this be something we're asserting on down below?

backend/otel/parse_test.go
```diff
@@ -0,0 +1,36 @@
+package otel
+
+import (
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func FuzzFormatStructureStackTrace(f *testing.F) {
+	f.Fuzz(func(t *testing.T, stackTrace string) {
```
et, line 9: This is really cool!

backend/public-graph/graph/resolver.go
```diff
@@ -2047,10 +2047,12 @@ func (r *Resolver) AddLegacyMetric(ctx context.Context, sessionID int, name stri
 
 func (r *Resolver) PushMetricsImpl(ctx context.Context, sessionSecureID string, metrics []*publicModel.MetricInput) error {
 	span, _ := tracer.StartSpanFromContext(ctx, "public-graph.PushMetricsImpl", tracer.ResourceName("go.push-metrics"))
+	span.SetTag("SessionSecureID", sessionSecureID)
+	span.SetTag("NumMetrics", len(metrics))
 	defer span.Finish()
 
 	if sessionSecureID == "" {
-		return e.New("PushMetricsImpl called without secureID")
+		return nil
```
et, line 2055: Thanks for fixing this.

et (approved):
One q about testing but not blocking.

backend/otel/parse_test.go
Vadman97: whoops! accidently removed the `expectedFrameError` check while testing. good catch!

PR #4246, opened 2023-02-13
Optimistically update UI for error state changes

ccschmitz:
## Summary

Makes the error state dropdown update optimistically and revert back to the original state if there is an error.

https://www.loom.com/share/02b46a8b685244dea2e7419f2eb618bc

Local click test.

N/A


frontend/src/pages/ErrorsV2/ErrorStateSelect/ErrorStateSelect.tsx
```diff
@@ -51,18 +52,16 @@ const ErrorStateSelectImpl: React.FC<Props> = ({
 	const [menuState, setMenuState] = React.useState<MenuState>(
 		MenuState.Default,
 	)
-	const [errorState, setErrorState] = useState<ErrorState>(initialErrorState)
```
aptlin, line 54: At the moment quick consecutive changes of the state will result in some flicker:

https://www.loom.com/share/b1f251265c634bc8a04c799e319dd5df

Wonder if we should have a local state to avoid it, or perhaps there's a cleaner way

aptlin (approved):
Works great! Left a non-blocking comment, curious what you think

frontend/src/pages/ErrorsV2/ErrorStateSelect/ErrorStateSelect.tsx
ccschmitz: @aptlin hmm... what did you have in mind when you said "we should have a local state to avoid it"? I think I'm OK with this limitation, but would be curious to hear how you think we could work around it. I'd think this would always be a problem with [Apollo's optimistic mutation updates](https://www.apollographql.com/docs/react/performance/optimistic-ui/).

ccschmitz: @aptlin I'm going to go ahead and merge this, but happy to follow up on this if you have thoughts on how to avoid this. Thanks!

Issue #4211, opened 2023-02-09
Load more results for logs page


et:
When we hit N number of logs, load the next N number of logs.

et:
We'll need to add cursor pagination. In order to do this, we need to add a new UUID column to ensure [stable pagination](https://morningcoffee.io/stable-pagination.html). 

I did a spike on this and confirmed that cursor pagination does indeed work correctly with clickhouse with this strategy:

![image](https://user-images.githubusercontent.com/58678/220434697-a668fb31-ae92-4818-91b8-af351bcc3741.png)

laojianzi:
@et Thank you for sharing, this looks great :+1:

PR #4253, opened 2023-02-14
add total count to logs page

et:
## Summary

Currently, we throw an arbitrary limit on our logs query: https://github.com/highlight/highlight/blob/7a6090f5f5d54768aaddafb46fce52b678a5bf08/backend/clickhouse/logs.go#L54-L58

At some point, we'll be adjusting to add infinite scroll (#4211) but we'll still need to show the total count that a query matches. 

I pulled the script from #4173 to generate a lot of data. 

Confirmed the total count is present using the same formatting as the sessions view:
Screenshot 2023-02-14 at 9 18 43 AM

N/A

### DIFF
--- a/backend/clickhouse/seeds/main.go
+++ b/backend/clickhouse/seeds/main.go
@@ -1,0 +1,138 @@
+package main
+
+import (
+	"context"
+	"math/rand"
+	"strconv"
+	"time"
+
+	"github.com/highlight-run/highlight/backend/clickhouse"
+	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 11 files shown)


backend/clickhouse/seeds/main.go
```diff
@@ -0,0 +1,138 @@
+package main
```
et, line 1: Pulled this script from #4173 so I could inject many logs quickly.

aptlin (approved):
LGTM!

PR #4258, opened 2023-02-14
fix(#4227): `go to` behavior for errors

### LINKED ISSUE #4227: Timestamp for session <> error linking is incorrect.
In most cases, when I click "go to session" on an error, the timestamp is incorrect:

Example (see the second session that I tested out):
https://www.loom.com/share/1403bce54c0247798668e2b597258d35

Session link:
https://app.highlight.io/1/sessions/MuASuqup02h2Im32NtKoEK5gFOeD?page=1&query=and%7C%7Ccustom_processed%2Cis%2Ctrue%7C%7Ccustom_created_at%2Cbetween_date%2C30%20days&tsAbs=2023-02-12T14%3A46%3A21.664Z

I think the issue is that the "go to error" button is using a different timetamp maybe? Or just defaulting to zero? Because when I use the actual popup (see below), there's no issues. 

image

aptlin:
## Summary

Resolves #4227.


jay-khatri:
Sorta related to this, but if I click "go to" on an error when the devtools window is closed, the devtools window opens and the indicator window stays in the sam eplace. See the screenshot below. 

image

Vadman97 (approved):

PR #4259, opened 2023-02-14
Fix query param rendering via use-query-params

et:
## Summary

See the revert of #4257. 

Basically, when we call `set*` with use-query-params, the updated value wasn't triggering a re-render. 

I discovered that the fix for this was two things:
* upgrading [use-query-params](https://github.com/pbeshai/use-query-params/tree/master/packages/use-query-params) to the next major version (see [migration plan](https://github.com/pbeshai/use-query-params/blob/master/packages/use-query-params/CHANGELOG.md#migrating-from-v1))
* correctly nesting `QueryParamProvider` under `BrowserRouter`

I undid the revert of #4257 and confirmed that updating the logs query correctly re-renders (triggers a new request to load logs). 

It seems like my bug occurs only when we are using a `set*` but most times we just pluck off the query value and then pipe it into something else. For example, the `sign_up` param on the login form is simply piped into some form state: https://github.com/highlight/highlight/blob/07d3e88369aa29ee538291216508362197e024b6/frontend/src/pages/Login/Login.tsx#L106-L112

I did find some examples of the `set*` in our pagination: 
https://github.com/highlight/highlight/blame/07d3e88369aa29ee538291216508362197e024b6/frontend/src/components/SearchPagination/SearchPagination.tsx#L100-L113

and in the query builder:
https://github.com/highlight/highlight/blob/07d3e88369aa29ee538291216508362197e024b6/frontend/src/pages/ErrorsV2/ErrorQueryBuilder/components/QueryBuilder/QueryBuilder.tsx#L1371-L1380

Doing some click testing around there (back/forward buttons, deep linking, etc) looks to be working as expected.

N/A

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -83,7 +83,7 @@
 		"stacktrace-js": "^2.0.2",
 		"subscriptions-transport-ws": "^0.11.0",
 		"use-http": "^1.0.26",
-		"use-query-params": "^1.2.0",
+		"use-query-params": "^2.2.0",
 		"validator": "13.7.0",
 		"xss": "^1.0.14"
 	},
--- a/frontend/src/index.tsx
+++ b/frontend/src/index.tsx
@@ -159,24 +159,24 @@ const App = () => {
 			}}
 		>
 			<ApolloProvider client={client}>
-				<QueryParamProvider>
-					<SkeletonTheme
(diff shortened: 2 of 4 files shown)


aptlin (approved):
Looks good!

PR #4261, opened 2023-02-14
add sales observability for error monitoring

Vadman97:
## Summary

Track error group views and add hubspot instrumentation for reporting error group view counts.

Local deploy tracing whether a session view is reported.

No.

### DIFF
--- a/backend/model/model.go
+++ b/backend/model/model.go
@@ -618,6 +618,12 @@ type Session struct {
 	AvoidPostgresStorage bool
 }
 
+type SessionAdminsViews struct {
+	SessionID int       `gorm:"primaryKey"`
+	AdminID   int       `gorm:"primaryKey"`
+	ViewedAt  time.Time `gorm:"default:NOW()"`
+}
+
 type EventChunk struct {
 	Model
 	SessionID  int `gorm:"index"`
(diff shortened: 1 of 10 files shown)


backend/model/model.go
```diff
@@ -937,6 +938,10 @@ type ErrorGroup struct {
 	} `gorm:"-"`
 	FirstOccurrence *time.Time `gorm:"-"`
 	LastOccurrence  *time.Time `gorm:"-"`
+
+	// Represents the admins that have viewed this session.
+	ViewedByAdmins []Admin `json:"viewed_by_admins" gorm:"many2many:error_group_admins_views;"`
```
ccschmitz, line 943: What do you think about storing a timestamp on this table as well? I've wanted that a few times on the `session_admins_views` table and I'm guessing we'll want it at some point as well. Might require using a [custom join table](https://gorm.io/docs/many_to_many.html#Customize-JoinTable) but can still leverage the `many2many` helper.

ccschmitz (approved):
Looks great, but one quick question on customizing the join table to track an additional property.

backend/model/model.go
Vadman97: good point! we added that for session_admins_views in prod, but i'll make it more explicit here and also do the same for the error group views.

PR #4273, opened 2023-02-15
add context to all logrus usage

Vadman97:
## Summary

Adds context to all logrus log usage so that logs can be associated with sessions.
Adds a CI check to ensure that all new log usage is accompanied with `WithContext()` 

Local deploy and CI.

No


backend/alerts/integrations/discord/messages_test.go
```diff
@@ -34,7 +34,7 @@ func (suite *DiscordChannelsTestSuite) SetupTest() {
 
 	bot, err := NewDiscordBot(guildID)
 	if err != nil {
-		log.Error("Failed to initialize Bot")
+		log.WithContext(ctx).Error("Failed to initialize Bot")
```
et, line 37: Should we add a fake context here?

et (approved):
Awesome! There's a type error but it should be an easy fix.

PR #4274, opened 2023-02-15
add datepicker to logs page

et:
## Summary

Using the datepicker from #4245, this PR integrates it into the logs page.

## How did you test this change?

Verified the default time is "Last 15 minutes"

Screenshot 2023-02-14 at 9 52 23 PM

Verified that selecting a preset works:
Screenshot 2023-02-14 at 9 52 52 PM

Verified the calendar works as expected:
Screenshot 2023-02-14 at 9 53 26 PM

For each of these examples, the start/end dates successfully update the query params as expected.

## Are there any deployment considerations?

N/A

PR #4278, opened 2023-02-15
use strict typing in ui package

### LINKED ISSUE #4277: packages/ui does not use strict mode in tsconfig
We aren't using `strict: true` in the `packages/ui` [tsconfig.json](https://github.com/highlight/highlight/blob/eric/add-datepicker-to-logs-page/packages/ui/tsconfig.json).

This can cause problems with our frontend repo which does use `strict: true`.

As an example, the following function's return value should be `string | undefined` but it is currently `string`
Screenshot 2023-02-15 at 8 52 09 AM

We should add `strict: true` in the `package/ui` directory and fix any type errors.

et:
## Summary

Not using `strict: true` everywhere can cause undesirable type mismatches (see [comment](https://github.com/highlight/highlight/pull/4274#issuecomment-1431596631)).

Fixes #4277

Build passes

N/A

### DIFF
--- a/packages/ui/tsconfig.json
+++ b/packages/ui/tsconfig.json
@@ -3,6 +3,7 @@
 	"include": ["src"],
 	"compilerOptions": {
 		"allowSyntheticDefaultImports": true,
+		"strict": true,
 		"incremental": true,
 		"rootDir": "src",
 		"outDir": "dist",


aptlin (approved):
Thank you!

ccschmitz:
Thanks, @et!

PR #4300, opened 2023-02-16
otel ingest should mark backend setup

Vadman97:
## Summary

Makes the otel data ingest mark a project as having their backend integration setup.
For otel data coming from the frontend (frontend console logs), the flag should not be set.

Local deploy.

No.


aptlin (approved):
🚢

PR #4308, opened 2023-02-16
add an endpoint to fetch values for a given key

et:
## Summary

When the user types into the log input and has a filter selected, we should provide possible values that might facilliate their search. This will work similar to datadog's search:

Screenshot 2023-02-16 at 4 17 49 PM

This is the backend endpoint. A follow up PR will handle wiring this up to the frontend.

Unit test added

N/A

### DIFF
--- a/backend/clickhouse/logs.go
+++ b/backend/clickhouse/logs.go
@@ -117,7 +117,7 @@ func (client *Client) LogsKeys(ctx context.Context, projectID int) ([]*modelInpu
 		`
 		SELECT arrayJoin(LogAttributes.keys) as key, count() as cnt
 		FROM logs
-		WHERE ProjectId = ? AND Timestamp < now() AND Timestamp >= (now() - toIntervalDay(30))
+		WHERE ProjectId = ?
 		GROUP BY key
 		ORDER BY cnt DESC;`,
 		projectID,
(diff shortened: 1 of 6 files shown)


backend/clickhouse/logs.go
```diff
@@ -117,7 +117,7 @@ func (client *Client) LogsKeys(ctx context.Context, projectID int) ([]*modelInpu
 		`
 		SELECT arrayJoin(LogAttributes.keys) as key, count() as cnt
 		FROM logs
-		WHERE ProjectId = ? AND Timestamp < now() AND Timestamp >= (now() - toIntervalDay(30))
```
et, line 120: I removed the time constraints because this is functionally equivalent to the [TTL](https://github.com/highlight/highlight/blob/1800d58201aca21876d5fd7d02f802e5e133a579/backend/clickhouse/migrations/000001_create_logs_table.up.sql#L26).

backend/clickhouse/logs_test.go
```diff
@@ -116,3 +116,36 @@ func (suite *ReadLogsTestSuite) TestLogsKeys(t *testing.T) {
 	}
 	assert.Equal(t, expected, keys)
 }
+
+func (suite *ReadLogsTestSuite) TestLogKeys(t *testing.T) {
```
mayberryzane, line 120: nice looking tests!

backend/clickhouse/logs.go
```diff
@@ -148,6 +148,39 @@ func (client *Client) LogsKeys(ctx context.Context, projectID int) ([]*modelInpu
 
 }
 
+func (client *Client) LogsKeyValues(ctx context.Context, projectID int, keyName string) ([]string, error) {
```
mayberryzane, line 151: should there be a limit for how many results are returned?

mayberryzane (approved):

backend/clickhouse/logs.go
et: Good call! Added in 266a0f40273c29b8e3517377cd81cd985e5cdca1

PR #4318, opened 2023-02-17
Backend changes for auth flow updates

ccschmitz:
## Summary

Backend changes to support #4202. Want to merge these in so the PR preview can be used for click testing.


ccschmitz:
Merged in #4202.

PR #4319, opened 2023-02-17
Wire up infinite scroll for logs page

et:
## Summary

This PR adds infinite scroll to the logs page.

_This PR assumes the reviewer is familiar with offset vs cursor pagination (see [here](https://bun.uptrace.dev/guide/cursor-pagination.html) if not)._

#### Auto-incrementing ids

The easiest way to implement cursor pagination is with an auto-incrementing id since to find the next cursor, you can just use the next id for the cursor. We do this for finding the next error instance:

https://github.com/highlight/highlight/blob/2abc14b9d8f064a42116f6dfc670a6ebfd79820b/backend/private-graph/graph/schema.resolvers.go#L4008-L4018

However, Clickhouse has no concept of an auto-incrementing id so we can't use this approach.

#### Timestamps

In Clickhouse, each record has a timestamp, so in theory that could be the cursor. However, while we have nano-second precision on this column (e.g. `2023-02-24 11:29:14.789873000`). It's not guaranteed to be unique. While it may be unlikely (since logs are injected through our SDKs), two logs could share the same timestamp. 

While we don't support it now, if we allow logs to be manually injected (e.g. through `curl`), it opens up the possibility of a two logs sharing the same timestamp since the user could set whatever they want.

#### Timestamp + ?

Concatenating the timestamp with another column could be an option for a cursor. At first glance, `TraceID` seems like a good candidate since each log should (in theory) have this. However, two logs at the same time can belong to the same trace.

That really only leaves one option and that is to create a new UUID column that is guaranteed to be unique for each log.

On their own, UUIDs are not ordered but when concatenated with a timestamp, we can achieve stable ordering:

Let's look at an example whereby this is in our DB:

| Timestamp                     | UUID                                  |
|-------------------------------|---------------------------------------|
| 2023-02-24 11:29:15.000000000 | uuid                                  |
| 2023-02-24 11:29:14.000000000 | aaaaaaaaa-3de1-4458-a306-3d4fd406de88 |
| 2023-02-24 11:29:14.000000000 | bbbbbbbb-3de1-4458-a306-3d4fd406de88 |
| 2023-02-24 11:29:13.000000000 | uuid                                  |

In this example, the first and last rows are in the correct order (since we order time in descending order). The second and third row share the exact same timestamp so we rely on the second part of the cursor `(Timestamp + UUID)` to figure out what should come next. To be consistent with the timestamp, the UUID is also in descending order so the output will flip these rows:

| Timestamp                     | UUID                                  |
|-------------------------------|---------------------------------------|
| 2023-02-24 11:29:15.000000000 | uuid                                  |
| 2023-02-24 11:29:14.000000000 | bbbbbbbb-3de1-4458-a306-3d4fd406de88 |
| 2023-02-24 11:29:14.000000000 | aaaaaaaaa-3de1-4458-a306-3d4fd406de88 |
| 2023-02-24 11:29:13.000000000 | uuid                                  |

### API spec

As mentioned, the cursor is `Timestamp + UUID`. Passing these values directly as params has some downsides:

* Passing around a timestamp is likely to get botched somewhere (imagine the differences between time parsing in Go and Javascript)
* UUID is not something the user is aware of (we don't show UUID anywhere in the UI)
* It doesn't allow flexibility to change the API later (less important for us since it's a private API but could be if we ever make logs a public API).

Fortunately, this is a solved problem and I stumbled across Relay's [opinionated](https://relay.dev/graphql/connections.htm) way of structuring this in GraphQL. What's nice is that I only need to implement a subset of their spec (e.g. I didn't handle paging backwards). Wiring this up into Apollo's `InMemoryCache` also has built-in functions ([docs](https://www.apollographql.com/docs/react/pagination/cursor-based/#relay-style-cursor-pagination)) (if we rolled our own API spec, we'd have to write custom logic to figure this out). 

### Generating the cursor

Finally, I pulled some code from this [blog article](https://dev.to/sadhakbj/implementing-cursor-pagination-in-golang-go-fiber-mysql-gorm-from-scratch-2p60) on how to encode/decode the cursor (Timestamp + UUID) on the backend. 

Lots and lots of unit tests. 
I consider pagination to be one of [three hard problems in computer science](https://dev.to/sadhakbj/implementing-cursor-pagination-in-golang-go-fiber-mysql-gorm-from-scratch-2p60). I feel pretty confident I got stable pagination, but please let me know if you think I'm missing something.

Confirmed infinite scroll works (see [Loom](https://www.loom.com/share/3efca1b6d1944dadbcdea5b2a52d9447)).

Yes. This does recreate the clickhouse table schema (which will drop the data). See inline comment for justification.

This PR also does not cover some known follow up issues:
* #4388
* #4387


backend/private-graph/graph/schema.graphqls
```diff
@@ -494,13 +494,28 @@ type S3File {
...
 }
 
+type LogEdge {
+	cursor: String!
+	node: Log!
+}
+
+type PageInfo {
+	hasNextPage: Boolean!
+	endCursor: String!
```
et, line 511: We could have gotten this on the frontend by just looking at the last item in `edges` but it is nice to just have this explicit to reduce the amount of work the frontend has to do.

backend/private-graph/graph/schema.graphqls
```diff
@@ -494,13 +494,28 @@ type S3File {
...
+}
+
+type LogsPayload {
```
et, line 514: See the [relay spec](https://relay.dev/graphql/connections.htm) for why things are called what they are called.

frontend/src/pages/LogsPage/LogsPage.tsx
```diff
@@ -73,6 +75,7 @@ const LogsPage = () => {
 			},
 		},
 		skip: !project_id,
+		nextFetchPolicy: 'cache-first', // See: https://stackoverflow.com/a/66489126
```
et, line 78: Thank you @ccschmitz for the tip.

backend/clickhouse/cursor.go
```diff
@@ -0,0 +1,55 @@
...
+	"fmt"
+	"strings"
+	"time"
+
+	e "github.com/pkg/errors"
+
+	modelInputs "github.com/highlight-run/highlight/backend/private-graph/graph/model"
+)
+
+func getLogsPayload(logs []*modelInputs.LogEdge, limit uint64) *modelInputs.LogsPayload {
```
Vadman97, line 15: would it be possible to make this bi-directional? we might have some usecases for infinite scrolling in the other direction as well

backend/clickhouse/logs.go
```diff
@@ -43,9 +48,11 @@ func (client *Client) BatchWriteLogRows(ctx context.Context, logRows []*LogRow)
 	return batch.Send()
 }
 
-func (client *Client) ReadLogs(ctx context.Context, projectID int, params modelInputs.LogsParamsInput) ([]*modelInputs.LogLine, error) {
-	query := makeSelectQuery("Timestamp, SeverityText, Body, LogAttributes", projectID, params)
-	query = query.Limit(100)
+func (client *Client) ReadLogs(ctx context.Context, projectID int, params modelInputs.LogsParamsInput, after *string) (*modelInputs.LogsPayload, error) {
+	var limit uint64 = 100
```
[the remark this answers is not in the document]
Vadman97: nit: bring the constant out to the file-level

backend/clickhouse/logs.go
```diff
@@ -222,12 +234,25 @@ func makeSeverityText(severityText string) modelInputs.SeverityText {
...
+		Where(sq.Eq{"ProjectId": projectID})
+
+	if after != nil && len(*after) > 1 {
+		timestamp, uuid, err := decodeCursor(*after)
+		if err != nil {
+			fmt.Print("error decoding cursor")
+		}
+
+		// See https://dba.stackexchange.com/a/206811
+		query = query.Where("toUInt64(toDateTime(Timestamp)) <= ?", uint64(timestamp.Unix())).
```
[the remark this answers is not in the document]
Vadman97: could this work in the other direction too if the `<` and `<=` are inverted?

backend/clickhouse/migrations/000004_create_new_logs_table.up.sql
```diff
@@ -0,0 +1,28 @@
...
+    ProjectId          UInt32 CODEC (ZSTD(1)),
+    SecureSessionId    String CODEC (ZSTD(1)),
+    INDEX idx_trace_id          TraceId TYPE bloom_filter(0.001) GRANULARITY 1,
+    INDEX idx_secure_session_id SecureSessionId TYPE bloom_filter(0.001) GRANULARITY 1,
+    INDEX idx_res_attr_key      mapKeys(ResourceAttributes) TYPE bloom_filter(0.01) GRANULARITY 1,
+    INDEX idx_res_attr_value    mapValues(ResourceAttributes) TYPE bloom_filter(0.01) GRANULARITY 1,
+    INDEX idx_log_attr_key      mapKeys(LogAttributes) TYPE bloom_filter(0.01) GRANULARITY 1,
+    INDEX idx_log_attr_value    mapValues(LogAttributes) TYPE bloom_filter(0.01) GRANULARITY 1,
+    INDEX idx_body              Body TYPE tokenbf_v1(32768, 3, 0) GRANULARITY 1
+)
```
Vadman97, line 23: while we're changing this, could we try the clickhouse default of `.025`? some [reading here](https://clickhouse.com/docs/en/guides/improving-query-performance/skipping-indexes/#bloom-filter-types) suggest better performance with a greater false positive rate if the values are less unique?

could omit the false positive rate value entirely to use the cc default

Vadman97 (approved):
🎉 amazing!

backend/clickhouse/cursor.go
et: Yes. I think we should definitely support backwards scrolling and I see no reason the logic can't support it (just flip the operators). 
Let's tackle that when we do #4396

backend/clickhouse/migrations/000004_create_new_logs_table.up.sql
et: Good idea! Would rather stick with clickhouse defaults as much as possible and we can modify later once we have a better baseline.

PR #4338, opened 2023-02-21
Improve key filtering

et:
## Summary

Currently, we can only do a direct match for key attributes (e.g. `workspace_id:1`).

This PR improves the matching, namely:

* Add wildcard support (`service:*foo*` will match `service:barfoobaz`)
* Add multi-word support `service:'image processor'` will match `service:image processor`)

Additionally this PR improves code quality:

* Use a true query builder instead of rolling our code. I chose to use [squirrel](https://github.com/Masterminds/squirrel). This makes conditional queries a lot easier and removes usages of `fmt.Sprintf` which is just ripe for a sql injection.
* Drop support for [testify suite](https://github.com/stretchr/testify#suite-package). This is currently unusable with vscode (https://github.com/golang/vscode-go/issues/2414).

Verified space search works
Screenshot 2023-02-21 at 10 50 49 AM

Verified wildcard search works
Screenshot 2023-02-21 at 10 56 00 AM

Verified we can search with body as well
Screenshot 2023-02-21 at 10 56 36 AM

N/A


Vadman97 (approved):
nice, and great tests! 🎉

PR #4346, opened 2023-02-22
report otel ingested errors as logs

Vadman97:
## Summary

Frameworks such as gofiber only support reporting errors for now.
When we encounter an error, write a log row with `ERROR` level so that the error is easy to find in the context of other logs.

Local deploy.

No


backend/otel/otel.go
```diff
@@ -137,51 +137,99 @@ func (o *Handler) HandleTrace(w http.ResponseWriter, r *http.Request) {
...
+								return
+							} else if excType == "" && excMessage == "" {
+								log.WithContext(ctx).WithField("Span", span).WithField("EventAttributes", eventAttributes).Warn("otel received exception with no type and no message")
+								return
+							}
+							stackTrace = formatStructureStackTrace(ctx, stackTrace)
+							err := &model.BackendErrorObjectInput{
+								SessionSecureID: &sessionID,
+								RequestID:       &requestID,
+								TraceID:         pointy.String(span.TraceID().String()),
```
Vadman97, line 202: the `TraceID` should be consistent here with the `LogRow`

jay-khatri (approved):

PR #4362, opened 2023-02-22
update git submodule url

Vadman97:
## Summary

`.gitmodules` pointed to old repository url (referencing old org name) which currently 301s but may break in the future.

Local `git submodule init --update`; CI

No

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "rrweb"]
 	path = rrweb
-	url = git@github.com:highlight-run/rrweb.git
+	url = https://github.com/highlight/rrweb.git


.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "rrweb"]
 	path = rrweb
-	url = git@github.com:highlight-run/rrweb.git
+	url = https://github.com/highlight/rrweb.git
```
aptlin, line 3:
```suggestion
	url = git@github.com:highlight/rrweb.git
```

aptlin (approved):

.gitmodules
Vadman97: the `ssh` reference was necessary previously when our rrweb fork was private. now that it is public, we should be able to clone over https. this avoids problems for folks who do not have an ssh key configured.

PR #4365, opened 2023-02-22
fix quoting logic for filters

### LINKED ISSUE #4359: Adding multiple log key filters results in a panic
Query logs with `service_name:backed workspace_id:75 user_id:81` and observe panic error:

Screenshot 2023-02-22 at 1 58 00 PM

et:
## Summary

This PR solves two things:
* Fix our quoting logic that results in a panic with multiple filters (#4359)
* Changes the quote delimiter from single quote (`'`) to double quote (`"`). Per a conversation with @ccschmitz, the latter seems more logical to support (and matches the way datadog works).

Fixes #4359

I red/greened this with a unit test.

I also click tested the query from #4359 to validate there is no panic error anymore:
Screenshot 2023-02-22 at 3 32 27 PM

N/A

### DIFF
--- a/backend/clickhouse/logs.go
+++ b/backend/clickhouse/logs.go
@@ -294,16 +294,16 @@ func splitQuery(query string) []string {
 	inquote := false
 	i := 0
 	for j, c := range query {
-		if c == '\'' {
+		if c == '"' {
 			inquote = !inquote
 		} else if c == ' ' && !inquote {
 			result = append(result, unquoteAndTrim(query[i:j]))
-			i = j + i
+			i = j + 1
 		}
 	}
 	return append(result, unquoteAndTrim(query[i:]))
 }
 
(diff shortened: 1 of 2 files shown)


backend/clickhouse/logs.go
```diff
@@ -294,16 +294,16 @@ func splitQuery(query string) []string {
 	inquote := false
 	i := 0
 	for j, c := range query {
-		if c == '\'' {
+		if c == '"' {
 			inquote = !inquote
 		} else if c == ' ' && !inquote {
 			result = append(result, unquoteAndTrim(query[i:j]))
-			i = j + i
```
et, line 301: This was the bug that was causing #4359 because `i` would end up being too big when we tried to capture the rest of the string on line 304 below (`query[i:]`).

Vadman97 (approved):
🎉

PR #4366, opened 2023-02-23
New UI for auth screens - take 2

ccschmitz:
## Summary

This PR reverts #4363 bringing back the new auth UI + a few fixes. Here is a list of updates in this PR:

* Fixes a bug where admin polling wasn't working on the verify email screen
* Fixes behavior of Intercom modal on the email verification page
* Fixes a UI bug where the sign up form would get stuck in the loading state indefinitely
* Updates the UI to handle workspace invitations in the new sign up/in flows

Refer to the [original PR](https://github.com/highlight/highlight/pull/4202) for more details.

Local click test. The changes can't be tested in the PR preview, so it would be great if the reviewer could click test locally as well. Here are the flows that need to be tested:

* [x] Creating an account + filling out about you with an invite
* [x] Creating an account + filling out about you without an invite
* [x] Sign in without an invite
* [x] Sign in with an invite
* [x] Verify email address

Need to make sure users are successfully signing up and signing in after this goes out. Should do some click testing ourselves and monitor first time sessions to confirm behavior is what we expect.


backend/private-graph/graph/schema.resolvers.go
```diff
@@ -826,33 +866,6 @@ func (r *mutationResolver) DeleteProject(ctx context.Context, id int) (*bool, er
 	return &model.T, nil
 }
 
-// SendAdminProjectInvite is the resolver for the sendAdminProjectInvite field.
-func (r *mutationResolver) SendAdminProjectInvite(ctx context.Context, projectID int, email string, baseURL string) (*string, error) {
```
ccschmitz, line 830: This method didn't appear to be used. This threw me because I thought you could be added to a project OR a workspace, but it looks like we only allow joining a workspace.

backend/private-graph/graph/schema.resolvers.go
```diff
@@ -402,11 +401,15 @@ func (r *mutationResolver) UpdateAdminAboutYouDetails(ctx context.Context, admin
...
 	admin.UserDefinedPersona = &adminDetails.UserDefinedPersona
-	admin.Phone = adminDetails.Phone
+	admin.Phone = pointy.String("")
 	admin.AboutYouDetailsFilled = &model.T
 
 	if !util.IsDevEnv() {
 		r.PrivateWorkerPool.SubmitRecover(func() {
+			// Delay this because we want the DB transaction creating the workspace
+			// (UpdateAdminAndCreateWorkspace) to complete before this executes.
+			time.Sleep(10 * time.Second)
```
ccschmitz, line 411: This is kind of janky. I'd like to figure out a better way of triggering these. I explored using GORM callbacks, but that would have been a larger refactor because we don't have access to the resolver there. Open to ideas how to improve this 😅

frontend/src/components/AutoJoinEmailsInput/index.tsx
```diff
@@ -0,0 +1,98 @@
+import { useAuthContext } from '@authentication/AuthContext'
```
ccschmitz, line 1: This was a copy/paste mostly from the original component. Designs are just Julianified.

backend/hubspot/hubspot.go
```diff
@@ -16,13 +16,11 @@ import (
 
 type HubspotApi struct {
 	hubspotClient hubspot.Client
-	db            *gorm.DB
```
ccschmitz, line 19: Ended up requiring this to be passed in as an argument to each method rather than creating a new instance. Having this in a separate instance was problematic now that we wrap a bunch of the calls for creating/updating admins/workspaces in a transaction, which this new instance would have no knowledge of.

Vadman97 (approved):
awesome, looks great now! just clicktested with @jay-khatri 

couple of small things we found
* deleting an admin in the workspace settings needs a refresh to show
* reset password form does not carry over the email from the login screen

ccschmitz:
@Vadman97 thank you for click testing! I'll follow up with a quick fix for carrying over the email to password reset. I am tracking the other issue with another bug we found on the user management page yesterday: https://github.com/highlight/highlight/issues/4465

PR #4378, opened 2023-02-23
filter out empty strings when fetching values for a key

et:
## Summary

We can get an empty string when fetching key values. 

Example:

```sql
CREATE TABLE table_map (a Map(String, String)) ENGINE=Memory;
INSERT INTO table_map VALUES ({'foo':'bar'})
INSERT INTO table_map VALUES ({'baz':'buzz'})
```

```sql
SELECT a['foo'] as value, count() as cnt FROM table_map
GROUP BY value
ORDER BY cnt DESC
```

| value | count |
|-------|-------|
|       | 1     |
| bar   | 1     |

This is seems to be because of some weird clickhouse quirk whereby the `baz` row is saying I have nothing that matches the `a['foo']` column. We can filter it out with a `mapContains`

```
SELECT a['foo'] as value, count() as cnt FROM table_map
where mapContains(a, 'foo')
GROUP BY value
ORDER BY cnt DESC
```

| value | count |
|-------|-------|
| bar   | 1     |

Unit tests FTW

N/A

### DIFF
--- a/backend/clickhouse/logs.go
+++ b/backend/clickhouse/logs.go
@@ -148,6 +148,7 @@ func (client *Client) LogsKeyValues(ctx context.Context, projectID int, keyName 
 	query := sq.Select("LogAttributes[?] as value, count() as cnt").
 		From("logs").
 		Where(sq.Eq{"ProjectId": projectID}).
+		Where("mapContains(LogAttributes, ?)", keyName).
 		GroupBy("value").
 		OrderBy("cnt DESC").
 		Limit(50)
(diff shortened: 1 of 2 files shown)


Vadman97 (approved):

backend/clickhouse/logs.go
```diff
@@ -148,6 +148,7 @@ func (client *Client) LogsKeyValues(ctx context.Context, projectID int, keyName
 	query := sq.Select("LogAttributes[?] as value, count() as cnt").
 		From("logs").
 		Where(sq.Eq{"ProjectId": projectID}).
+		Where("mapContains(LogAttributes, ?)", keyName).
```
Vadman97, line 151: :tada: TIL that you can do `WHERE mapContains(LogAttributes, 'foo')` and clickhouse/sql will check that it is truthy, without you having to explicitly say `WHERE mapContains(LogAttributes, ?) = 1`

Vadman97 (approved):

PR #4392, opened 2023-02-24
hotfix: commandbar polish

aptlin:
## Summary

Jay's suggestions:

1. At least one option is always selected, which avoids the confusion about the behavior on `enter` press with no explicit selection.
2. Adds `enter` hint for the selected option.

Screenshot 2023-02-24 at 3 45 42 PM

cmd-k or /

no

### DIFF
--- a/frontend/src/components/CommandBar/CommandBar.tsx
+++ b/frontend/src/components/CommandBar/CommandBar.tsx
@@ -107,6 +107,7 @@ const CommandBar = () => {
 		'cmd+k, ctrl+k, /',
 		(e) => {
 			e.preventDefault()
+			form.reset()
 			setCurrentAttribute(ATTRIBUTES[0])
 			commandBarDialog.toggle()
 		},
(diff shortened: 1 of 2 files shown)


frontend/src/components/CommandBar/CommandBar.tsx
```diff
@@ -231,15 +225,12 @@ const SearchBar = ({ form }: { form: FormState<CommandBarSearch> }) => {
...
-											currentAttribute ?? ATTRIBUTES[0],
-											{
-												newTab: true,
-												withDate:
-													selectedDates[0].getTime() !==
-													last30Days.startDate.getTime(),
-											},
-										)
+										searchAttribute(currentAttribute, {
+											newTab: true,
```
Vadman97, line 229: noticed that the new-tab search does not clear the main tab's command bar search term.
eg. if you cmd+k, type sasha, shift+enter, come back to original tab, cmd+k, still shows sasha

Vadman97 (approved):

frontend/src/components/CommandBar/CommandBar.tsx
aptlin: This is how Notion does it too. Jay mentioned this as confusing as well though, so I'll reset the form on open.

PR #4417, opened 2023-02-27
Handle backward pagination and permalinking logs (backend)

et:
## Summary

This PR extends #4319 to add backward pagination and support to permalink to log. A frontend PR will follow that wires up these changes.

### Backward pagination

This continues with the [Relay spec](https://relay.dev/graphql/connections.htm#sec-Backward-pagination-arguments) adding a `before` argument.

Similar to handling the `after` cursor logic:

```sql
WHERE Timestamp <= x
  AND ( Timestamp < x OR UUID < y )
ORDER BY Timestamp DESC, UUID DESC
```

we need to invert the logic to handle going "backwards" ("backward" in flying quotes because it's going forward in time to get the newer logs)

```sql
WHERE Timestamp >= x
  AND ( Timestamp > x OR UUID > y )
ORDER BY Timestamp ASC, UUID ASC # Note the flipped order here
```

If we simply returned this directly though, the result would not match the existing descending order, hence, we wrap this in a sub-select to return the order correctly.

```sql
SELECT * FROM logs_window (
  SELECT * FROM logs
    WHERE Timestamp >= x
      AND ( Timestamp > x OR UUID > y )
    ORDER BY Timestamp ASC, UUID ASC
 AS logs_window
) ORDER BY Timestamp DESC, UUID DESC
```

### Permalinking

The relay spec has no concept of permalinking so we leverage that.

A naive approach would be:
* given a cursor, return the log that matches this cursor
* frontend can make two more requests to get the logs before/after this cursor

This is obviously unperformant and is pretty annoying for the frontend to have to make all these requests.

Hence, I chose to get a "window" of logs (see [stackoverflow](https://stackoverflow.com/questions/71699810/it-is-possible-in-cursor-based-pagination-to-get-the-prev-and-next-cursor-in-the/71738696#71738696)) around this cursor. Specifically loading `Limit / 2` logs before and after this cursor (`Limit` is 100 so with the log matching, the result is 101 logs).

Unit tests were added (specifically around ensuring [PageInfo](https://relay.dev/graphql/connections.htm#sec-undefined.PageInfo) is correct as the frontend needs this to be accurate to determine if it should load additional requests).

I also verified this behavior with a simple frontend (https://github.com/highlight/highlight/pull/4486/files#diff-96c1192170afa8f77ed262c7c42ab64f29d1f31091d907d211563db845369e54R103) that will be more complete in a follow up PR.

N/A

### DIFF
--- a/backend/clickhouse/logs.go
+++ b/backend/clickhouse/logs.go
@@ -89,7 +89,7 @@ func (client *Client) ReadLogs(ctx context.Context, projectID int, params modelI
 	sb := sqlbuilder.NewSelectBuilder()
 	var err error
 	var args []interface{}
-	selectStr := "Timestamp, UUID, SeverityText, Body, LogAttributes"
+	selectStr := "Timestamp, UUID, SeverityText, Body, LogAttributes, TraceId, SpanId, SecureSessionId"
 
 	if pagination.At != nil && len(*pagination.At) > 1 {
 		// Create a "window" around the cursor
@@ -123,8 +123,6 @@ func (client *Client) ReadLogs(ctx context.Context, projectID int, params modelI
 	}
 
 	sql, args := sb.Build()
-	test := sb.String()
-	fmt.Print(test)
 
 	rows, err := client.conn.Query(ctx, sql, args...)
 
--- a/backend/clickhouse/logs.go
+++ b/backend/clickhouse/logs.go
@@ -72,21 +72,59 @@ func (client *Client) BatchWriteLogRows(ctx context.Context, logRows []*LogRow) 
 	return batch.Send()
 }
 
-const Limit int = 100
+const LogsLimit int = 100
 const KeyValuesLimit int = 50
 
-func (client *Client) ReadLogs(ctx context.Context, projectID int, params modelInputs.LogsParamsInput, after *string) (*modelInputs.LogsPayload, error) {
-	sb, err := makeSelectBuilder("Timestamp, UUID, SeverityText, Body, LogAttributes, TraceId, SpanId, SecureSessionId", projectID, params, after)
-	if err != nil {
-		return nil, err
-	}
+const OrderBackward = "Timestamp ASC, UUID ASC"
+const OrderForward = "Timestamp DESC, UUID DESC"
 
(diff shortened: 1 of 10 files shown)


backend/private-graph/graph/schema.graphqls
```diff
@@ -522,10 +522,12 @@ type LogEdge {
 
 type PageInfo {
 	hasNextPage: Boolean!
+	hasPreviousPage: Boolean!
+	startCursor: String!
 	endCursor: String!
 }
 
-type LogsPayload {
+type LogsConnection {
```
et, line 530: Renamed to `LogsConnection` to match the relay spec.

backend/clickhouse/logs.go
```diff
@@ -72,29 +72,65 @@ func (client *Client) BatchWriteLogRows(ctx context.Context, logRows []*LogRow)
...
-	if err != nil {
-		return nil, err
-	}
+const OrderBackward = "Timestamp ASC, UUID ASC"
+const OrderForward = "Timestamp DESC, UUID DESC"
 
-	sb.OrderBy("Timestamp DESC, UUID DESC").Limit(Limit + 1)
+type Pagination struct {
+	After          *string
+	AfterOrEqualTo *string // currently only used internally
```
Vadman97, line 83: looks like this overlap with the `before` results (the row at the cursor is included in both `afterOrEqual` and `before` results. guessing that there is some other logic that removes the duplicate row, but just wanted to double check.

Vadman97 (approved):
looks great!

backend/clickhouse/logs.go
et: Good catch. I think in my testing that apollo was handling deduplication. 
I think the union below should be change from `UNION ALL` to `UNION DISTINCT`

I'll add a test to confirm.

et: Added 8ee38e023cec42da092964b113f5ec49c9d737ee

It removes the `afterOrEqual` flow and added an equal sub select to the union. It's effectively:

SELECT (Before Cursor) UNION ALL SELECT (Equal to cursor) UNION ALL SELECT (After cursor)

This removes the need to use distinct.

PR #4424, opened 2023-02-28
fix refetching logs

et:
## Summary

This PR fixes two frontend issues (see inline comments below).

See inline comments below.

N/A

### DIFF
--- a/frontend/src/pages/LogsPage/LogsPage.tsx
+++ b/frontend/src/pages/LogsPage/LogsPage.tsx
@@ -63,7 +63,7 @@ const LogsPage = () => {
 
 	const [endDate, setEndDate] = useQueryParam('end_date', EndDateParam)
 
-	const { data, loading, fetchMore } = useGetLogsQuery({
+	const { data, loading, fetchMore, refetch } = useGetLogsQuery({
 		variables: {
 			project_id: project_id!,
 			params: {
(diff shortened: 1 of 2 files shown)


frontend/src/pages/LogsPage/LogsTable/LogsTable.tsx
```diff
@@ -108,7 +108,7 @@ export const LogsTable = ({ data, loading, query }: Props) => {
 	useEffect(() => {
 		// Collapse all rows when search changes
 		table.toggleAllRowsExpanded(false)
-	}, [logEdges])
+	}, [logEdges, table])
```
et, line 111: See https://github.com/highlight/highlight/pull/4393#discussion_r1119002239

Prior to this change, we had an infinite loop:

Screenshot 2023-02-27 at 4 43 45 PM

After this change, the error goes away.

frontend/src/pages/LogsPage/LogsPage.tsx
```diff
@@ -75,7 +75,7 @@ const LogsPage = () => {
 			},
 		},
 		skip: !project_id,
-		nextFetchPolicy: 'cache-first', // See: https://stackoverflow.com/a/66489126
+		fetchPolicy: 'cache-and-network',
```
et, line 78: Prior to this change, when we updated the query input or set the time range, the request to fetch new logs wouldn't be run. 

There's not too much written about `cache-and-network` but it is what is recommended to use from the [docs](https://www.apollographql.com/docs/react/v2/caching/cache-interaction/#incremental-loading-fetchmore).

et: I confirmed that forward scrolling works _and_ it's not making a double request (what the previous stackoverflow comment was addressing).

Vadman97 (approved):
nice find

PR #4427, opened 2023-02-28
support non-string structured log attributes

Vadman97:
## Summary

support non-string log attributes such as ints and floats

local docker deploy writing to clickhouse

no


backend/otel/otel.go
```diff
@@ -224,6 +236,7 @@ func (o *Handler) HandleTrace(w http.ResponseWriter, r *http.Request) {
 								RequestID:       &requestID,
 								TraceID:         pointy.String(traceID),
 								SpanID:          pointy.String(spanID),
+								LogCursor:       pointy.String(clickhouse.EncodeCursor(ts, uuid)),
```
et, line 239: This seems fine for now, but it does feel a bit like we're leaking private logic.
Maybe in the future we could have an accessor like "cursorFor(logRow)"

Mostly just trying to prevent the logic from spreading out. We can chat about this in our next 1-1z

et (approved):

backend/otel/otel.go
et: Actually, this would be pretty elegant:

```go
func (logRow LogRow) Cursor() string {
	return encodeCursor(logRow.Timestamp, logRow.UUID)
}
```

backend/otel/otel.go
```diff
@@ -171,10 +180,12 @@ func (o *Handler) HandleTrace(w http.ResponseWriter, r *http.Request) {
...
 					if event.Name() == semconv.ExceptionEventName {
-						ts := event.Timestamp().AsTime()
-						traceID := castString(requestID, span.TraceID().String())
+						traceID := cast(requestID, span.TraceID().String())
 						spanID := span.SpanID().String()
-						excMessage := castString(eventAttributes[string(semconv.ExceptionMessageKey)], "")
+						excMessage := cast(eventAttributes[string(semconv.ExceptionMessageKey)], "")
+
+						ts := event.Timestamp().AsTime()
+						uuid := uuid.New().String()
```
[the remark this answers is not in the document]
Vadman97: great idea! added a `NewLogRow` helper

backend/otel/otel.go
Vadman97: makes sense! added a `LogRow.Cursor()` method for this

backend/clickhouse/logs.go
```diff
@@ -29,6 +30,24 @@ type LogRow struct {
 	SecureSessionId    string
 }
 
+func NewLogRow(ts time.Time, projectID int, sessionID, traceID, spanID string) *LogRow {
```
et, line 33: Nice!

The only downside I see to this is that it would be pretty easy to mix up the params for `projectID`, `sessionID`, etc.

I wish Go had named parameters (without needing a struct) since that could solve this pretty elegantly.

One quick and dirty way would be for the parameter to accept the same struct:

```go
func NewLogRow(attrs LogRow) *LogRow {
	uuid := uuid.New().String()
	return &LogRow{
		UUID:            uuid,
		Timestamp:       attrs.Timestamp,
		TraceId:         attrs.TraceId,
		SpanId:          attrs.SpanId,
		SeverityText:    "INFO",
		SeverityNumber:  int32(log.InfoLevel),
		ProjectId:       uint32(attrs.ProjectId),
		SecureSessionId: attrs.SecureSessionId,
	}
}
```


Another idea is to make the `logRow` struct private and create a new struct that defines what someone can actually set:

```go
type LogRowAttrs struct {
	Timestamp       time.Time
	ProjectId       uint32
	TraceId         string
	SpanId          string
	SecureSessionId string
}

func NewLogRow(attrs LogRowAttrs) *logRow {
	uuid := uuid.New().String()
	return &logRow{
		UUID:            uuid,
		Timestamp:       attrs.Timestamp,
		TraceId:         attrs.TraceId,
		SpanId:          attrs.SpanId,
		SeverityText:    "INFO",
		SeverityNumber:  int32(log.InfoLevel),
		ProjectId:       uint32(attrs.ProjectId),
		SecureSessionId: attrs.SecureSessionId,
	}
}
```

backend/clickhouse/logs.go
```diff
@@ -29,6 +30,24 @@ type LogRow struct {
...
+		TraceId:         traceID,
+		SpanId:          spanID,
+		SeverityText:    "INFO",
+		SeverityNumber:  int32(log.InfoLevel),
+		ProjectId:       uint32(projectID),
+		SecureSessionId: sessionID,
+	}
+}
+
+func (l *LogRow) Cursor() string {
```
et, line 47: Thanks!

Can we make `encode/decode` cursor private again?

et (approved):

backend/clickhouse/logs.go
Vadman97: that's fair. i think a base struct that the `LogRow` inherits is a pretty clean way, with that base struct being the arg for `NewLogRow`. just changed that.

PR #4428, opened 2023-02-28
lint in the CI

et:
## Summary

Adds an eslint check to CI. This can catch subtle bugs for us.

* Validated that we are running `eslint` as part of the turborepo step
* Fixed eslint errors that are an `error`
* Bumped `react-hooks/exhaustive-deps` from `warn` -> `error` (5e19fbd3b811f9072e3c7b860841729e3056091b). We want this at `error` to actually block CI as it can cause an infinite loop (see https://github.com/highlight/highlight/pull/4393/files#r1119002239 and #4424) 
* Fixed `react-hooks/exhaustive-deps` in ac7930f1bc9b23781b97e3d661d6eef33300958e

N/A


.github/workflows/turbo.yml
```diff
@@ -47,6 +47,9 @@ jobs:
             - name: Check formatting
               run: yarn format-check
 
+            - name: Lint
+              run: yarn lint
```
Vadman97, line 51: could we just update `turbo.json` to run `lint` as a dependency of `test`?

turbo.json
```diff
@@ -59,6 +59,7 @@
 		"enforce-size": {
 			"dependsOn": ["build"]
 		},
+		"lint": {},
```
Vadman97, line 62: could we make `test` depend on `lint` to have it run automatically in all workspaces?

package.json
```diff
@@ -43,6 +43,7 @@
 		"frontend:start": "yarn dev:frontend",
 		"g": "yarn generate",
 		"generate": "hygen",
+		"lint": "yarn workspace @highlight-run/frontend lint",
```
Vadman97, line 46: could this just be a `yarn turbo run lint` to make this forward-compatible for other workspaces adding the command?

et:
Hey @Vadman97 - good suggestions! I tried to adjust to make this more turborepo compatible (c198371dc7712ee9f78d80aa66369e226e6856cf) but it fails on linting the nextjs package. It wasn't immediately obvious to me how to fix that so I filed a ticket so that we can follow up on this (#4431)

Vadman97 (approved):

PR #4430, opened 2023-02-28
add log ingest support for node and next sdks

Vadman97:


sdk/highlight-node/src/hooks.ts
```diff
@@ -0,0 +1,65 @@
...
+
+let consoleHooked = false
+
+export function hookConsole(cb: (cb: ConsolePayload) => void) {
+	if (consoleHooked) return
+	consoleHooked = true
+	const levels = [
+		'debug',
+		'info',
+		'log',
```
et, line 36: Here's where we normalize the logs to the frontend:
https://github.com/highlight/highlight/blame/main/backend/clickhouse/logs.go#L202-L234

`log` will get normalized to `info` when it is shown the to the end user on the frontend. Hence, to find this log, they expect to search for `level:info`.

However, I think we should normalize this _before_ we persist to the database. The reason being is that if this is stored as `SeverityText=log` in the db, then we can't filter for it via `level:info`.

Realistically, every language/sdk is going to add these arbitrary levels  (other things in this list including `count`, `assert`, `dir`, etc.) and we need to land on some agreed standard across the board.

If we don't normalize here, can we normalize at the OTEL level (I don't think anything exists but correct me if I'm wrong).

et: I just noticed that we have a `consoleMethodsToRecord` [option](https://www.highlight.io/docs/sdk/client#consoleMethodsToRecord). We should respect that option when deciding what to monkeypatch.

Vadman97: great point on both! making those improvements now

Vadman97: using the `consoleMethodsToRecord` option now when set, and mapping the levels to their go parsed values.

sdk/highlight-node/src/hooks.ts
```diff
@@ -0,0 +1,74 @@
...
+	consoleHooked = true
+	const levels = {
+		debug: 'debug',
+		info: 'info',
+		log: 'info',
+		count: 'info',
+		dir: 'info',
+		warn: 'warn',
+		assert: 'warn',
+		error: 'panic',
```
et, line 45: Is this mapping `console.error`'s level to `panic`? Can it be `error: 'error'`?

et (approved):
Looks great!

PR #4440, opened 2023-03-01
fix about you forms getting stuck on completion

Vadman97:
## Summary

`AboutYou` page would get stuck on completion because the Admin query would be cached by indexeddb, and we were not awaiting the mutation to complete. Submitting the `AboutYou` page would not change the UI until the page was refreshed. This bug would only surface on production frontend builds.

Fixes a backend race condition with `Admin` row creation waiting for firebase data.

Local production deploy of frontend: `doppler run -- yarn vite build; doppler run -- yarn vite preview --port 3000`
`make debug` backend gdb tracing
Testing with new account creation.

No

### DIFF
--- a/backend/private-graph/graph/resolver_test.go
+++ b/backend/private-graph/graph/resolver_test.go
@@ -112,13 +112,13 @@ func TestMutationResolver_AddAdminToWorkspace(t *testing.T) {
 			errorExpected: false,
 		},
 		"same email different case": {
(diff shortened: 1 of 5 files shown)


ccschmitz:
Changes in https://github.com/highlight/highlight/pull/4366 would also help with this bug as it pulls the admin creation and about you update to a separate request.

PR #4445, opened 2023-03-01
Filter by log level

et:
## Summary

This PR adds support to filter by the log level (internally called `SeverityText`).

Will follow up for adding support for trace id and other non-custom attributes.

This PR also replaces [squirrel](https://github.com/Masterminds/squirrel) with [go-sqlbuilder](https://github.com/huandu/go-sqlbuilder). There's a couple reasons for this:

* squirrel is in [maintenance](https://github.com/Masterminds/squirrel#squirrel-is-complete) mode where as go-sqlbuilder has active development.
* squirrel lacks `UNION` support. We'll need this for permalinking ([example](https://github.com/highlight/highlight/pull/4417/files#diff-e913069aa4c9c7492c6180d43d70b4d9200d25b074a1b346c87319b510b0327fR72-R95)). Per the previous point, it doesn't seem like squirrel will add union support even with a [PR ready](https://github.com/Masterminds/squirrel/pull/320#issuecomment-1367329550). 
* seems like squirrel has some design pitfalls (see [FAQ](https://github.com/huandu/go-sqlbuilder#whats-the-difference-between-this-package-and-squirrel)).

Unit tests added for testing level filtering. Our existing tests should cover the squirrel -> go-sqlbuilder transition

Click test:

Kapture 2023-03-01 at 16 25 41

N/A

### DIFF
--- a/backend/clickhouse/cursor.go
+++ b/backend/clickhouse/cursor.go
@@ -12,12 +12,12 @@ import (
 	modelInputs "github.com/highlight-run/highlight/backend/private-graph/graph/model"
 )
 
-func getLogsPayload(logs []*modelInputs.LogEdge, limit uint64) *modelInputs.LogsPayload {
-	hasNextPage := uint64(len(logs)) == limit+1
+func getLogsPayload(logs []*modelInputs.LogEdge) *modelInputs.LogsPayload {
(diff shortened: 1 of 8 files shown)


Vadman97 (approved):
nice, this looks great!

PR #4453, opened 2023-03-02
feat: added traceloop exporter

nirga:
## Summary

Added [Traceloop](https://www.traceloop.dev/) exporter support so tests can be automatically generated for Highlight.

We should enable it in staging first.


Vadman97 (approved):
🚢 🎉

PR #4468, opened 2023-03-02
add filtering for trace id, span id, and secure session id

et:
## Summary

This PR adds support for filtering by `trace_id`, `span_id`, and `secure_session_id`. 

The naming of `"secure"_session_id` may be confusing (I could see an argument for calling this `session_id`). Jay and I chatted about this and we are opting to leave it as for now. People won't really be using this search directly (they'll only see it because they they were on a session page and clicked View logs which populated that filter). We can revisit this if it seems confusing.

Observe that values are only visible when a non-empty string:
Screenshot 2023-03-02 at 3 44 15 PM

Trace id filtering works:
Screenshot 2023-03-02 at 3 45 44 PM

Span id filtering works:
Screenshot 2023-03-02 at 3 44 55 PM

Secure session id filtering works:

Screenshot 2023-03-02 at 3 46 13 PM

### DIFF
--- a/backend/clickhouse/logs.go
+++ b/backend/clickhouse/logs.go
@@ -73,6 +73,7 @@ func (client *Client) BatchWriteLogRows(ctx context.Context, logRows []*LogRow) 
 }
 
 const Limit int = 100
+const KeyValuesLimit int = 50
 
 func (client *Client) ReadLogs(ctx context.Context, projectID int, params modelInputs.LogsParamsInput, after *string) (*modelInputs.LogsPayload, error) {
 	sb, err := makeSelectBuilder("Timestamp, UUID, SeverityText, Body, LogAttributes, TraceId, SpanId, SecureSessionId", projectID, params, after)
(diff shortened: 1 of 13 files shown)


backend/clickhouse/logs.go
```diff
@@ -185,16 +194,40 @@ func (client *Client) LogsKeys(ctx context.Context, projectID int) ([]*modelInpu
...
 			Limit(50)
-	} else {
+	case modelInputs.ReservedLogKeySecureSessionID.String():
+		sb.Select("SecureSessionId secure_session_id, count() as cnt").
+			From("logs").
+			Where(sb.Equal("ProjectId", projectID)).
+			Where(sb.NotEqual("secure_session_id", "")).
+			GroupBy("secure_session_id").
+			OrderBy("cnt DESC").
+			Limit(50)
```
Vadman97, line 213: nit: limit value could be `Limit / 2`?

Vadman97 (approved):

backend/clickhouse/logs.go
et: Good idea. I made a separate `const` since the existing `Limit` should probably only apply to log lines.
4ba4931279db508261a368b1007dd6b93815e3f8

PR #4472, opened 2023-03-02
fix: removed groupbytrace otel processor

nirga:
## Summary
Since we're not using the `otel-contrib` image, the `groupby` processor is missing, so we remove it.


Vadman97:
before
Screenshot 2023-03-02 at 3 31 51 PM


after
image

Vadman97 (approved):

PR #4490, opened 2023-03-03
try to fix logs tests

et:
## Summary

Each logs test is creating the test db and running migrations. Let's use `TestMain` so we only run this once.


aptlin (approved):
Looks great!

PR #4491, opened 2023-03-03
fix admin creation

Vadman97:
## Summary

Admin resolver no longer creates an admin, which may be necessary because the CreateAdmin mutation is not guaranteed to be called in some edge cases.

Local deploy.

No


ccschmitz:
@Vadman97 what are the edge cases you were seeing where an admin wasn't being created? Thank you for jumping on this and fixing. Would love to figure out a way to decouple the mutation logic from query logic still. I can take a look at this next week.

PR #4496, opened 2023-03-05
enable console and error recording on localhost

Vadman97:
## Summary

Console recording was disabled on localhost because we would have print messages on the console message handler
causing infinite recursion. Since we no longer emit any messages in the console message handler, we do not need to disable this recording.

Local deploy of nextjs 13 app.

New highlight.run client.


aptlin (approved):
Thank you!

PR #4501, opened 2023-03-06
Auth and invite flow revisions

ccschmitz:
## Summary

* Handles errors when attempting to join a workspace from an invite
* Allows users to use the back button to get back to the join workspace screen after ignoring an invitation
* Passes the email between the sign up/in and password reset pages
* Fixes an issue where we were creating multiple listeners on the Firebase auth changes events
* Renames the `OrgRouter` to `ProjectRouter`

Local click test.

Will need to keep an eye out for issues encountered by users signing up. Trying to watch sessions for first time users to monitor.


Vadman97 (commented):
thanks chris! doing a bit of a click test, i noticed the app seems to stutter between the sign up screen and the email verification modal

https://www.loom.com/share/5999125dbe0a42959a4736f868f95b3d

ccschmitz:
@Vadman97 thanks for click testing! Will look into that issue and follow. Thanks for reporting. Just curious, it sounds like you're using IndexedDB locally, is that correct?

Vadman97:
@ccschmitz yup, this was in the render.com preview which has indexeddb enabled by default

ccschmitz:
@Vadman97 this is ready for another look whenever you have time! I figured out the culprit of the redirects was that we were instantiating multiple listeners for the auth change events. All the listeners would trigger the admin creating/fetching logic multiple times and cause the app to get in weird states + a loop of redirects. The fix was removing all the dependencies from the `useEffect` that sets up the listener.

https://github.com/highlight/highlight/pull/4501/commits/8a358d346169e1f621b603128484b2445f1ea66b#diff-5d954d9a2e243edfbe0547bea7eb29755d24446dd2b527818d1f3fb6b4373b31L310-R312

Vadman97 (approved):
🚢

PR #4507, opened 2023-03-06
add a date range when choosing log key values

et:
## Summary

Fetching values for keys takes too long. This is for two reasons:

* we are using an aggregation query `COUNT`
* we are not using a time range

This PR simplifies the query to fetch values to fix this issues. Per the linked issue, this dramatically speeds up the query.

Clicktested locally to confirm query still works.

We should monitor in prod to confirm this results in the queries passing.


aptlin (approved):
Thank you!

Issue #4471, opened 2023-03-02
Copy on log line includes too much information


et:
```[tasklist]
### Tasks
- [x] Add copy button next to each value
- [x] Add button to left of log attribute keys to apply as a filter
- [x] Clean up copy data
```

Clicking `Copy` on a log line includes too much information. For example, this log:
![Screenshot 2023-03-02 at 3 34 27 PM](https://user-images.githubusercontent.com/58678/222575251-28012c83-cff1-4653-a277-c0cec720c0e1.png)

results in this copied text:

```json
{
  "__typename": "LogEdge",
  "cursor": "MjAyMy0wMy0wMlQyMjozMzowMC45ODU3Njc5NjFaLDE3OTJhYWI5LWIwZDUtNDU2YS1hYjJjLTc1N2QxMjQ5YjgxNQ==",
  "node": {
    "__typename": "Log",
    "timestamp": "2023-03-02T22:33:00.985767961Z",
    "severityText": "ERROR",
    "body": "public-graph graphql request failed",
    "logAttributes": {
      "code": {
        "filepath": "/build/backend/util/logging.go",
        "function": "github.com/highlight-run/highlight/backend/util.GraphQLErrorPresenter.func1",
        "lineno": "19"
      },
      "container": {
        "id": "387925026"
      },
      "host": {
        "name": "ip-172-31-0-233.us-east-2.compute.internal"
      },
      "log": {
        "message": "public-graph graphql request failed",
        "severity": "ERROR"
      },
      "os": {
        "description": "Alpine Linux 3.17.2 (Linux ip-172-31-0-233.us-east-2.compute.internal 5.10.162-141.675.amzn2.aarch64 #1 SMP Mon Jan 9 22:45:17 UTC 2023 aarch64)",
        "type": "linux"
      },
      "path": "initializeSession",
      "process": {
        "executable": {
          "name": "backend",
          "path": "/bin/backend"
        },
        "owner": "root",
        "pid": "14",
        "runtime": {
          "description": "go version go1.20.1 linux/arm64",
          "name": "go",
          "version": "go1.20.1"
        }
      }
    }
  }
}
```

There's a couple issues here:
* We are including graphql stuff (e.g. `__typename`)
* `logAttributes` is shown when all of those keys should be lifted to the root level

The copied text should be whatever is under `node` without any graphql information and without showing `logAttributes`

In this example, I expect the following to be copied to my clipboard:

```json
{
  "__typename": "Log",
[...]

PR #4509, opened 2023-03-06
rename SeverityText -> level and Body -> message

et:
## Summary

For #4471, we intend to add copy buttons for each log property (to prefill the query). To prepare for that, we should rename OTEL lingo to a more product friendly lingo, specifically:

* SeverityText -> level (to coincide with #4445)
* Body -> message (since every other logging app calls it this)

Additionally, these should sit as top level keys like `trace_id`, `span_id`, and `secure_session_id` (#4468)

Verified that `level` and `message` are top level keys in the JSON view:

Screenshot 2023-03-06 at 1 22 50 PM

A follow up PR will come that will add support for searching `message:value` (#4510)


Vadman97 (approved):
🚢

PR #4513, opened 2023-03-07
add a way to query the error object for a log cursor

Vadman97:
## Summary

* Stores the log cursor of error logs on the error object.
* Adds a resolver to query the error object for a given log cursor.

Local deploy.

Will monitor the db migration for index creation blocking.

### DIFF
--- a/backend/clickhouse/logs.go
+++ b/backend/clickhouse/logs.go
@@ -32,7 +32,6 @@ type LogRow struct {
 	ServiceName    string
 	Body           string
 	LogAttributes  map[string]string
-	ErrorObjectID  uint64
 }
 
 func NewLogRow(attrs LogRowPrimaryAttrs) *LogRow {
--- a/backend/clickhouse/logs.go
+++ b/backend/clickhouse/logs.go
@@ -32,6 +32,7 @@ type LogRow struct {
 	ServiceName    string
 	Body           string
 	LogAttributes  map[string]string
+	ErrorObjectID  uint64
 }
 
 func NewLogRow(attrs LogRowPrimaryAttrs) *LogRow {
@@ -54,6 +55,9 @@ func (l *LogRow) Cursor() string {
 }
 
 func (client *Client) BatchWriteLogRows(ctx context.Context, logRows []*LogRow) error {
+	if len(logRows) == 0 {
+		return nil
+	}
 	batch, err := client.conn.PrepareBatch(ctx, "INSERT INTO logs")
 
 	if err != nil {
--- a/backend/private-graph/graph/schema.graphqls
+++ b/backend/private-graph/graph/schema.graphqls
@@ -1284,6 +1284,7 @@ type Query {
 	): ErrorsHistogram!
 	error_group(secure_id: String!): ErrorGroup
 	error_object(id: ID!): ErrorObject
+	error_object_for_log(log_cursor: String!): ErrorObject
 	error_instance(
 		error_group_secure_id: String!
 		error_object_id: ID
--- a/frontend/src/pages/ErrorsV2/ErrorLogCursor/ErrorLogCursorRedirect.tsx
+++ b/frontend/src/pages/ErrorsV2/ErrorLogCursor/ErrorLogCursorRedirect.tsx
@@ -1,10 +1,6 @@
 import { useGetErrorObjectForLogQuery } from '@graph/hooks'
 import { useParams } from '@util/react-router/useParams'
-import React, { useEffect } from 'react'
-import {
-	AppLoadingState,
-	useAppLoadingContext,
-} from '@context/AppLoadingContext'
+import React from 'react'
 import { Navigate } from 'react-router'
 import { ErrorState } from '@components/ErrorState/ErrorState'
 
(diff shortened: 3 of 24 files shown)


backend/private-graph/graph/schema.graphqls
```diff
@@ -1281,6 +1281,7 @@ type Query {
 	): ErrorsHistogram!
 	error_group(secure_id: String!): ErrorGroup
 	error_object(id: ID!): ErrorObject
+	error_object_for_log(log_cursor: String!): ErrorObject
```
et, line 1284: Trying to figure this out over slack but how would one use this endpoint? We can't use it from the logs page since that would be an n+1 query (and request).
I'm hoping we can have some type of error id on the logs table.
Let me know if you had something else in mind!

Vadman97: updated in slack - https://highlightcorp.slack.com/archives/C04NUD29BT3/p1678203728066969
enpoint works by having frontend redirect to the existing `/errors/<group>/instances/<instance>` url

backend/clickhouse/logs.go
```diff
@@ -54,6 +54,9 @@ func (l *LogRow) Cursor() string {
 }
 
 func (client *Client) BatchWriteLogRows(ctx context.Context, logRows []*LogRow) error {
+	if len(logRows) == 0 {
```
et, line 57: Is this really needed?

frontend/src/pages/ErrorsV2/ErrorsV2.tsx
```diff
@@ -129,6 +145,28 @@ const ErrorsV2: React.FC<React.PropsWithChildren<{ integrated: boolean }>> = ({
 		// eslint-disable-next-line react-hooks/exhaustive-deps
 	}, [error_secure_id])
 
+	useEffect(() => {
+		if (logCursor) {
+			if (errorObjectForLogLoading) {
+				setLoadingState(AppLoadingState.LOADING)
+			} else {
+				setLoadingState(AppLoadingState.LOADED)
+				console.log('vadim', errorObjectForLogData)
```
et, line 154: Remove

frontend/src/pages/LogsPage/LogsTable/LogDetails.css.ts
```diff
@@ -7,3 +8,15 @@ export const line = style({
 		},
 	},
 })
+
+export const linkButton = style({
```
et, line 12: Do we have to override the styling from the component library? Can we move this code into the component library if we feel like the styling is important?
cc @ccschmitz on what we should be doing here.

frontend/src/pages/ErrorsV2/ErrorsV2.tsx
```diff
@@ -129,6 +145,28 @@ const ErrorsV2: React.FC<React.PropsWithChildren<{ integrated: boolean }>> = ({
...
+				if (errorObjectForLogData?.error_object_for_log) {
+					navigate(
```
et, line 156: Can this be done at the routing level? Or at least in the parent component of this. Doing this in a `useEffect` seems like it's going to have unintended side effects especially since we're using `navigate`

frontend/src/pages/LogsPage/LogsTable/LogDetails.css.ts
ccschmitz: It looks like this is a standard secondary tag in the mocks. I'd be curious to know why this doesn't work the way we want it to in this context. Maybe we need to update the styles for the `kind="secondary"` + `emphasis="medium"` compound variant?

https://github.com/highlight/highlight/blob/24a224728391315e9a374e68fd1dd44bf7f9c00a/packages/ui/src/components/Tag/styles.css.ts#L214-L236

frontend/src/pages/ErrorsV2/ErrorsV2.tsx
ccschmitz: What do you think about adding a new route + component for handling this logic? e.g. add a new route like `/errors/logs/:cursor` that renders a new `<ErrorsLogsRedirect />` component which contains the logic for looking up the error and handling the redirect.

Vadman97: i like that idea, certainly feels cleaner to leverage react router. the original reason to go with qs params was the thought that the url pattern of `/errors?log=<cursor>` looks nicer than `/errors/logs/:cursor`, but i don't have a strong preference. cc @jay-khatri

backend/clickhouse/logs.go
Vadman97: shouldn't we avoid creating an empty batch if there are no rows? i'd rather support this implementation-specific logic here than wrap calls to `BatchWriteLogRows` with the check

frontend/src/pages/ErrorsV2/ErrorsV2.tsx
Vadman97: updated!

frontend/src/pages/LogsPage/LogsTable/LogDetails.css.ts
Vadman97: updated the compound tag style

backend/clickhouse/logs.go
Vadman97: gotcha, in https://github.com/highlight/highlight/blob/93cc114c6d5e70f76abacb2200bad7a9358c67b5/backend/worker/kafka_worker.go#L79 our batch processing will batch different message types together adn invoke the clickhouse write if the logs slice is nil (could be no logs in a batch but batch still has `markbackendsetup` messages)

frontend/src/pages/ErrorsV2/ErrorLogCursor/ErrorLogCursorRedirect.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { useParams } from '@util/react-router/useParams'
+import React, { useEffect } from 'react'
+import {
+	AppLoadingState,
+	useAppLoadingContext,
+} from '@context/AppLoadingContext'
+import { Navigate } from 'react-router'
+import { ErrorState } from '@components/ErrorState/ErrorState'
+
+const ErrorLogCursorRedirect: React.FC = () => {
```
et, line 11: This is awesome! Good separation of concerns.

et (approved):
Looks great! Thanks for figuring this out!

PR #4519, opened 2023-03-07
fix loading states for logs view

et:


frontend/src/pages/LogsPage/LogsPage.tsx
```diff
@@ -63,6 +63,7 @@ const LogsPage = () => {
 
 	const [endDate, setEndDate] = useQueryParam('end_date', EndDateParam)
 
+	const [loadingAfter, setLoadingAfter] = useState(false)
```
et, line 66: We can add another loading state when we wire up scrolling backwards.

ccschmitz (approved):

PR #4526, opened 2023-03-07
fix searching log body with a space

et:
## Summary

This PR ensures we persist a space when searching a log body.

Confirmed that searching with a space works.

Before:
Screenshot 2023-03-07 at 1 40 06 PM

After:
Screenshot 2023-03-07 at 1 39 24 PM

N/A


aptlin (approved):
Thank you!

PR #4527, opened 2023-03-07
skip loading message values

et:
## Summary

The frontend is sending a request for fetching the values for the `message` key. This results in the backend actually searching for `LogAttributes['message']` which doesn't make sense:

Screenshot 2023-03-07 at 4 12 33 PM

Try on the render preview as that takes longer to fetch keys which is how I got into this state.

N/A

### DIFF
--- a/frontend/src/pages/LogsPage/SearchForm/SearchForm.tsx
+++ b/frontend/src/pages/LogsPage/SearchForm/SearchForm.tsx
@@ -1,5 +1,6 @@
 import { useGetLogsKeysQuery, useGetLogsKeyValuesLazyQuery } from '@graph/hooks'
 import { GetLogsKeysQuery } from '@graph/operations'
+import { ReservedLogKey } from '@graph/schemas'
 import {
 	Badge,
 	Box,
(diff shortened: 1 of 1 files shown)


frontend/src/pages/LogsPage/SearchForm/SearchForm.tsx
```diff
@@ -142,16 +142,18 @@ const Search: React.FC<{
...
 		}
 
-		getLogsKeyValues({
-			variables: {
-				project_id: project_id!,
-				key_name: activeTerm.key,
-				date_range: {
-					start_date: moment(startDate).format(FORMAT),
-					end_date: moment(endDate).format(FORMAT),
+		if (activeTerm.key !== 'message') {
```
et, line 145: Tried to use the enum and the build failed...
https://github.com/highlight/highlight/actions/runs/4359439593/jobs/7621231851

Not sure why since this works for me locally

Vadman97 (approved):

Vadman97:
stale

PR #4530, opened 2023-03-08
add manual log methods to highlight go sdk

Vadman97:
## Summary

Enhances the go sdk with manual log functions that can be used without a logging library.

Local deploy.
Screenshot 2023-03-08 at 2 16 06 PM

Will release new highlight-go tag.

### DIFF
--- a/backend/main.go
+++ b/backend/main.go
@@ -48,7 +48,7 @@ import (
 	"github.com/highlight-run/highlight/backend/zapier"
 	"github.com/highlight-run/workerpool"
 	H "github.com/highlight/highlight/sdk/highlight-go"
-	"github.com/highlight/highlight/sdk/highlight-go/log"
+	hlog "github.com/highlight/highlight/sdk/highlight-go/log"
 	highlightChi "github.com/highlight/highlight/sdk/highlight-go/middleware/chi"
 	e "github.com/pkg/errors"
 	"github.com/rs/cors"
--- a/backend/main.go
+++ b/backend/main.go
@@ -236,6 +236,7 @@ func main() {
 	H.Start()
 	defer H.Stop()
 	H.SetDebugMode(log.StandardLogger())
+	hlog.Info("welcome to highlight.io")
 	// setup highlight logrus hook
 	log.AddHook(hlog.NewHook(hlog.WithLevels(
 		log.PanicLevel,
--- a/backend/main.go
+++ b/backend/main.go
@@ -228,10 +228,12 @@ func main() {
 	log.SetReportCaller(true)
 	// setup highlight
 	H.SetProjectID("1jdkoe52")
-	if util.IsDevOrTestEnv() && !util.IsInDocker() {
+	if util.IsDevOrTestEnv() {
 		log.WithContext(ctx).Info("overwriting highlight-go graphql client address...")
 		H.SetGraphqlClientAddress("https://localhost:8082/public")
-		H.SetOTLPEndpoint("http://collector:4318")
+		if util.IsInDocker() {
+			H.SetOTLPEndpoint("http://collector:4318")
+		}
 	}
 	H.Start()
 	defer H.Stop()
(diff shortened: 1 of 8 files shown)


sdk/highlight-go/highlight.go
```diff
@@ -255,6 +255,20 @@ func SetProjectID(id string) {
 	projectID = id
 }
 
+func SetLogOut(print bool) {
+	logOut = print
```
et, line 259: I don't see why anyone would want this. If anything, this should be setting the min log level. For example, here's how logrus does it:
https://github.com/sirupsen/logrus#level-logging

sdk/highlight-go/highlight.go
```diff
@@ -255,6 +255,20 @@ func SetProjectID(id string) {
...
+}
+
+func Print(message string) {
```
et, line 262: Per the above, I would expect we'd set the config in the logging sdk, not here. Which would mean, we don't need this wrapper around `fmt.Print`

sdk/highlight-go/log/log.go
```diff
@@ -52,6 +55,70 @@ type VercelLog struct {
...
+	var tagsStr string
+	if len(tags) > 0 {
+		s, _ := json.Marshal(tags)
+		tagsStr = fmt.Sprintf(" %s", s)
+	}
+	highlight.Print(fmt.Sprintf("[%s] %s%s\n", level, message, tagsStr))
+	LogWithContext(context.TODO(), level, message, tags...)
+}
+
+func Trace(message string, tags ...attribute.KeyValue) {
```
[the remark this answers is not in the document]
et: Alternatively, we could also provide a `WithContext` so one can do `hlog.WithContext(ctx).Info(....)`

sdk/highlight-go/highlight.go
Vadman97: this came from a customer request to use highlight side by side with their existing logger that prints to console. i'll adjust this to customize a max level that logs will be emitted for.

sdk/highlight-go/log/log.go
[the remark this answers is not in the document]
Vadman97: i like that idea - will add a `WithContext` provider

sdk/highlight-go/highlight.go
Vadman97: yup, makes sense to move this and the above logic to the `hlog` package

sdk/highlight-go/log/log.go
```diff
@@ -52,6 +55,70 @@ type VercelLog struct {
 	Proxy       VercelProxy `json:"proxy"`
 }
 
+func HighlightTags(projectID, sessionID, requestID string) []attribute.KeyValue {
```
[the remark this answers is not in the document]
Vadman97: i'll be consistent with the public `WithContext` and expose this as an `hlog.WithSession(...).Log()` function to customers

backend/main.go
```diff
@@ -228,14 +228,19 @@ func main() {
...
 		H.SetGraphqlClientAddress("https://localhost:8082/public")
-		H.SetOTLPEndpoint("http://collector:4318")
+		if util.IsInDocker() {
+			H.SetOTLPEndpoint("http://collector:4318")
+		}
 	}
 	H.Start()
 	defer H.Stop()
 	H.SetDebugMode(log.StandardLogger())
+	hlog.SetOutput(true)
```
et, line 241: This isn't actually needed, correct?

backend/main.go
```diff
@@ -228,14 +228,19 @@ func main() {
...
+	hlog.SetOutputLevel(hlog.DebugLevel)
```
et, line 242: Nice!

et (approved):
Looks great!

backend/main.go
Vadman97: right, `true` is the default. keeping it in as an example though.

PR #4536, opened 2023-03-08
hotfix: commandbar editing and search experience

aptlin:
## Summary

1. Switches the default from identifier to email and back based on the command bar query.
2. At the moment editing the query in the command bar is unusable because of the cursor jumping to the end. This PR fixes this by using a hook for a `keydown` event listener rather than overwriting onKeyDown on the Input. Also sneaks in refactoring as a result of the investigation into the unstable cursor issue.

1. cmd + k 
2. Editing the query and navigating back and forth
3. Trying out email-like queries

no


Vadman97 (approved):
works great!

PR #4540, opened 2023-03-09
Fix infinite workspace redirect

ccschmitz:
## Summary

Fixes an issue where users we stuck in a redirect loop when visiting the root path for their workspace (`/w/:workspace_id`).

Local click test. Should be able to confirm in the PR preview as well.

N/A

### DIFF
--- a/frontend/src/pages/WorkspaceTabs/WorkspaceTabs.tsx
+++ b/frontend/src/pages/WorkspaceTabs/WorkspaceTabs.tsx
@@ -5,7 +5,7 @@ import WorkspaceTeam from '@pages/WorkspaceTeam/WorkspaceTeam'
 import analytics from '@util/analytics'
 import React, { Suspense, useEffect } from 'react'
 import { Helmet } from 'react-helmet'
-import { useLocation, useMatch, useNavigate } from 'react-router-dom'
+import { useLocation, useMatch, useNavigate, useParams } from 'react-router-dom'
 
 import styles from './WorkspaceTabs.module.scss'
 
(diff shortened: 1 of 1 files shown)


frontend/src/pages/WorkspaceTabs/WorkspaceTabs.tsx
```diff
@@ -36,8 +36,8 @@ export const WorkspaceTabs = () => {
 	const location = useLocation()
 	const navigate = useNavigate()
 
+	const { workspace_id: workspaceId } = useParams<{ workspace_id: string }>()
 	const workspaceMatch = useMatch('/w/:workspace_id/:page_id')
```
ccschmitz, line 40: The problem is that this match doesn't work without the `page_id` segment of the route and it returns `null`. We then get redirected to `/w/undefined/team`, but that causes the [`DefaultWorkspaceRouter` to be rendered](https://github.com/highlight/highlight/blob/dc126bcc276dc81b5f218e8ca602c6fb0df878e1/frontend/src/routers/AppRouter/AppRouter.tsx#L228) and [redirect them back to `/w/:workspace_id`](https://github.com/highlight/highlight/blob/dc126bcc276dc81b5f218e8ca602c6fb0df878e1/frontend/src/routers/OrgRouter/DefaultWorkspaceRouter.tsx#L37).

Vadman97: aha! awesome find chris, thank you 🎉 would another option be to use optional params eg. `"errors/:error_secure_id?/:error_tab_key?/:error_object_id?"`? trying that locally doesn't work for me so the `useParams` is changing the logic in a way that fixes this.

Vadman97 (approved):

frontend/src/pages/WorkspaceTabs/WorkspaceTabs.tsx
ccschmitz: Yeah, I tried that locally as well, but I couldn't get it to work with `useMatch`. Looking at it again, I'm guessing we could get it working if we changed the route definition [here](https://github.com/highlight/highlight/blob/e2f78ab275dfd2d7039dade1dcf6d234ef52e109/frontend/src/routers/OrgRouter/WorkspaceRouter.tsx#L109), but I was kind of rushing to get the fix out.

PR #4547, opened 2023-03-09
fix empty state flashing

### LINKED ISSUE #4546: Empty state is flashing again
Kapture 2023-03-09 at 09 20 28

et:
## Summary

Regression from #4535. It's more noticeable on prod where the initial logs request takes longer than development.

Fixes #4546

Confirmed on render preview that flashing doesn't occur.

N/A


aptlin (approved):
Thank you!

PR #4549, opened 2023-03-09
add WithTag helper for logging sdk

Vadman97:
Allows reporting string tags on log statements.

### DIFF
--- a/backend/main.go
+++ b/backend/main.go
@@ -240,7 +240,7 @@ func main() {
 	H.SetDebugMode(log.StandardLogger())
 	hlog.SetOutput(true)
 	hlog.SetOutputLevel(hlog.DebugLevel)
-	hlog.WithContext(ctx).Info("welcome to highlight.io")
+	hlog.WithContext(ctx).WithTag("hello", "world").Info("welcome to highlight.io")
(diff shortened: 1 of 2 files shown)


backend/main.go
```diff
@@ -240,7 +240,7 @@ func main() {
 	H.SetDebugMode(log.StandardLogger())
 	hlog.SetOutput(true)
 	hlog.SetOutputLevel(hlog.DebugLevel)
-	hlog.WithContext(ctx).Info("welcome to highlight.io")
+	hlog.WithContext(ctx).WithTag("hello", "world").Info("welcome to highlight.io")
```
et, line 243: Can this be called `WithField`? 

logrus and zap both call it a field

Also, what's the longer vision for supporting multiple key/value pairs? i.e. what do you see that API looking like?

Vadman97: just had a call with a customer that wanted this and they are ok with setting up logrus as the way to log. agreed that maintaining our own sdk here doesn't make much sense, so i've rewritten the implementation accordingly. will update highlight.io docs after releasing this package

et (approved):
Thanks!

Issue #4009, opened 2023-01-30
Create alerts for logs


jay-khatri:

julian-highlight:
Current design here: https://www.figma.com/file/rdBPNzn7Klk6RG1LHUCE5B/Screen-Designs?node-id=6105%3A318502&t=fEdWJspKqSHAFNTu-11

julian-highlight:
I'll update you @aptlin as soon as this design enters a final stage. Probably will be done by tomorrow!

PR #4554, opened 2023-03-10
feat(#4009): add a log alert page

aptlin:
## Summary

Adds a new alert page for logs.

https://frontend-pr-4554.onrender.com/alerts/logs/new

no


frontend/src/pages/LogsPage/SearchForm/SearchForm.tsx
```diff
@@ -180,13 +183,6 @@ const Search: React.FC<{
 		// eslint-disable-next-line react-hooks/exhaustive-deps
 	}, [initialQuery])
 
-	useEffect(() => {
```
aptlin, line 183: Re-use of the Search component on the log alert page does not work well if we always have the form and combobox state in sync (since we send a request to fetch the histogram based on the form state)

PR #4555, opened 2023-03-10
Log attribute buttons and keyword match highlighting

ccschmitz:
## Summary

Adds the following features for log attributes:

* Highlighting for matching values
* Copy to clipboard action
* Apply as filter action

https://www.loom.com/share/a1409e02a3e74fb087bcf3291ae2e97c

Local click test

N/A - all client changes


et:
Looking good @ccschmitz!

Nested keys work for filtering but the top level keys to not
Kapture 2023-03-10 at 12 53 29

Is this something you could look into?

ccschmitz:
My bad, @et - thanks for catching that! Just pushed a fix in https://github.com/highlight/highlight/pull/4555/commits/8b13fa15f69c336a3ea210eb76aac57ea1bbec4d.

et (approved):

PR #4559, opened 2023-03-11
make python logging opt-in

Vadman97:
## Summary

Switches the python sdk to not record logs by default for consistency across SDKs.

Python unit tests.

New python version released. We don't have any customers actively using python for logging, so this should not have impact on existing users.


Vadman97:
python sdk logging was already logged in 🤦

PR #4560, opened 2023-03-11
add reflame endpoint for cors

Vadman97:
## Summary

Allows hosting our frontend on https://reflame.app/
cc @lewisl9029

Checking that local public / private graph work.

No


lewisl9029 (commented):
Thanks for the ping. Looks great! 👍

PR #4561, opened 2023-03-11
add links from errors to logs

Vadman97:
## Summary

Enables the logs button in the header for all as we are starting to link the logs page in the docs and from errors.
Errors now link to logs as described in the cohesion doc.
The log cursor is linked and a query is populated with the `host.name`, `secure_session_id`, and `trace_id`,
all that are defined on an error.
For example, the following link is generated: /1/logs/MjAyMy0wMy0xMVQwMTo1MToxMC44MzY2ODQ4NzhaLDk2NGI5MzJkLWNjYTctNDVkNy1hN2Q2LTA5YzVkNjU2NzljNw==?query=host.name:07c1855195ee%20secure_session_id:jmkO7NeeZ6McS0v6HYeE0tBHiUKO%20trace_id:th8ddI7y2Q

Screenshot 2023-03-10 at 6 48 01 PM

Screenshot 2023-03-10 at 6 47 45 PM

Local deploy.

No


frontend/src/pages/ErrorsV2/ErrorInstance/ErrorInstance.tsx
```diff
@@ -302,6 +317,57 @@ const ErrorInstance: React.FC<Props> = ({ errorGroup }) => {
...
+										value: errorObject.session?.secure_id,
+										offsetStart: 2,
+									})
+								}
+								if (errorObject.trace_id) {
+									queryParams.push({
+										key: 'trace_id',
+										operator: DEFAULT_LOGS_OPERATOR,
+										value: errorObject.trace_id,
+										offsetStart: 3,
```
et, line 342: Does `offsetStart` need to be adjusted based on whether the conditional actually applies? i.e. if the two above conditionals (`errorObject.source` and `errorObject.session?.secure_id`) return false, should `offsetStart` be 1 here?

et: cc @ccschmitz - thoughts on the above question?
I saw this comment and I'm not sure if we're using `offsetStart`:
https://github.com/highlight/highlight/blob/d099fff437f8772f380fa2ca29272a386418487b/frontend/src/pages/LogsPage/LogsTable/LogDetails.tsx#L385

et (commented):
This is great! 
I had one q about `offsetStart`. Can you take a look? Thanks

frontend/src/pages/ErrorsV2/ErrorInstance/ErrorInstance.tsx
Vadman97: good catch - just updated that to increment `offsetStart` in a way that will keep them sequential.

et (approved):
Thanks for getting this through!

frontend/src/pages/ErrorsV2/ErrorInstance/ErrorInstance.tsx
ccschmitz: `offsetStart` is used to specify the index the term starts at in the query string, not the index in the array of query terms. Sorry, this logic is kind of confusing, but it's used with the cursor's index to determine what the active query term is, but the active query term isn't relevant when you are clicking in the logs table to append a filter because the input loses focus. That's why I left that comment that it wasn't used and just made it `0`.

I don't think `offsetStart` matters in the context of this PR for the same reasons and it's going to be [recalculated on each render](https://github.com/highlight/highlight/blob/e69897a25356eabe1fcd62f68c992e08e4e41a7e/frontend/src/pages/LogsPage/SearchForm/SearchForm.tsx#L127-L129). I'd say just set it to `0` here. Let me know if that makes sense!

PR #4563, opened 2023-03-11
Aaron i4556

aaronmaxcarver:
## Summary

Issue #4556 
- For player console Network rows and Error rows:
- "Goto" button should only navigate to the correct timestamp, not open the right panel

## How did you test this change?
- Dev mode click test
- https://www.loom.com/share/e3d006d50fa04832bc6ff4fab8300e25

## Are there any deployment considerations?

- None

PR #4564, opened 2023-03-11
Revert "Player console Network and Error rows "goto" to not open right panel"

aaronmaxcarver:
Reverts highlight/highlight#4563


Vadman97:
stale, go-to buttons are working

PR #4567, opened 2023-03-12
ensure frontend does not show new workspace creation flow

Vadman97:


ccschmitz (approved):
Thank you for catching and fixing this, @Vadman97!

PR #4576, opened 2023-03-13
use randomized key for project-level kafka messages, show timestamp for error instance

mayberryzane:
## Summary

- randomize partition key for project-level kafka messages to avoid unbalanced partitions
- show timestamp instead of created_at in error instance (time error was recorded rather than received)
- add leading 0 to seconds in timestamp
Screen Shot 2023-03-13 at 3 37 18 PM

- click tested locally

- no


Vadman97 (approved):
nice!

Issue #4575, opened 2023-03-13
Body searches can be slow


et:
Searching for `STRIPE_INTEGRATION_ERROR` is slow.

This appears to be because we aren't using the `tokenbf_v1` as visible in our EXPLAIN:

```sql
 EXPLAIN indexes = 1
                    SELECT Timestamp, Body
                    FROM
                      logs
                    WHERE
                      ProjectId = 1
                      AND Timestamp >= (now() - toIntervalDay(15))
                      AND Body ILIKE '%STRIPE_INTEGRATION_ERROR%'
                    ORDER BY
                      toUnixTimestamp(Timestamp) DESC
                    LIMIT
                      100

┌─explain──────────────────────────────────────────────────────────────────────────────────────────────────┐
│ Expression (Projection)                                                                                  │
│   Limit (preliminary LIMIT (without OFFSET))                                                             │
│     Sorting (Sorting for ORDER BY)                                                                       │
│       Expression (Before ORDER BY)                                                                       │
│         Filter (WHERE)                                                                                   │
│           ReadFromMergeTree (default.logs)                                                               │
│           Indexes:                                                                                       │
│             MinMax                                                                                       │
│               Keys:                                                                                      │
│                 Timestamp                                                                                │
│               Condition: and((Timestamp in ['1677450254', +Inf)), (Timestamp in ['1677450254', +Inf)))   │
│               Parts: 143/164                                                                             │
│               Granules: 124210/124231                                                                    │
│             Partition                                                                                    │
│               Keys:                                                                                      │
│                 toDate(Timestamp)                                                                        │
│               Condition: and((toDate(Timestamp) in [19414, +Inf)), (toDate(Timestamp) in [19414, +Inf))) │
│               Parts: 143/143                                                                             │
│               Granules: 124210/124210                                                                    │
│             PrimaryKey                                                                                   │
│               Keys:                                                                                      │
│                 ProjectId                                                                                │
│                 toUnixTimestamp(Timestamp)                                                               │
│               Condition: and((ProjectId in [1, 1]), (toUnixTimestamp(Timestamp) in [1677450254, +Inf)))  │
│               Parts: 105/143                                                                             │
│               Granules: 34729/124210                                                                     │
└──────────────────────────────────────────────────────────────────────────────────────────────────────────┘
```

### Using hasToken

```sql
EXPLAIN indexes = 1
                    SELECT Timestamp, Body
                    FROM
                      logs
                    WHERE
                      ProjectId = 1
[...]

PR #4578, opened 2023-03-14
speed up body searches by utilizing bloom filter index

et:
## Summary

When we search on the log body, we aren't utilizing the `tokenbf_v1` index (see #4575). This particular index basically takes a log body such as `log.Info("some message that we emit")` and splits it into tokens `["some", "message", "that", "we", "emit"])`.

In order to leverage it properly, we need to switch to actually use the `hasToken` function to see if one of those tokens contains our string. This does require us to split up the body string even further:

For example: searching for "some message that we emit" would construct this query:

```sql
SELECT Timestamp, Body
FROM 
  logs 
WHERE 
  ProjectId = 1 
  AND Timestamp >= (now() - toIntervalDay(15))  
  AND (hasToken(Body, 'some'))
  AND (hasToken(Body, 'message'))
  AND (hasToken(Body, 'that'))
  AND (hasToken(Body, 'we'))
  AND (hasToken(Body, 'emit'))
ORDER BY 
  toUnixTimestamp(Timestamp) DESC
LIMIT 
  100 
```

This does mean that we aren't searching for a direct match of `AND Body = 'some message that we emit'` but the performance gain is worth it.

When we first noticed this, we observed that searching for "STRIPE_INTEGRATION_ERROR" is slow ([something we log a lot of](https://github.com/highlight/highlight/search?q=STRIPE_INTEGRATION_ERROR)). A caveat with using `tokenbf_v1` is that tokens are split by non-alphanumeric characters ([docs](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/mergetree/#token-bloom-filter)). Hence, the `_` is split similar to whitespace.

With all that being said, searching for `STRIPE_INTEGRATION_ERROR` for the last 15 days:

Previous:
```sql
SELECT Timestamp, Body
FROM 
  logs 
WHERE 
  ProjectId = 1 
  AND Timestamp >= (now() - toIntervalDay(15))  
  AND Body ILIKE '%STRIPE_INTEGRATION_ERROR%' 
ORDER BY 
  toUnixTimestamp(Timestamp) DESC
LIMIT 
  100 

Elapsed: 18.954s
Read: 44,413,741 rows (4.05 GB)
```

```sql
SELECT Timestamp, Body
FROM 
  logs 
WHERE 
  ProjectId = 1 
  AND Timestamp >= (now() - toIntervalDay(15))  
  AND (hasTokenCaseInsensitive(Body, 'STRIPE'))
  AND (hasTokenCaseInsensitive(Body, 'INTEGRATION'))
  AND (hasTokenCaseInsensitive(Body, 'ERROR'))
ORDER BY 
  toUnixTimestamp(Timestamp) DESC
LIMIT 
  100 
Elapsed: 4.730s
Read: 22,468,784 rows (2.18 GB)
```

So a fairly significant speed up. More importantly is that we are using our `tokenbf_v1` index

```sql
EXPLAIN indexes = 1
SELECT Timestamp, Body
FROM 
  logs 
WHERE 
  ProjectId = 1 
  AND Timestamp >= (now() - toIntervalDay(15))  
  AND (hasToken(Body, 'STRIPE'))
  AND (hasTokenCaseInsensitive(Body, 'INTEGRATION'))
  AND (hasTokenCaseInsensitive(Body, 'ERROR'))
ORDER BY 
  toUnixTimestamp(Timestamp) DESC
LIMIT 
  100 

...truncated

│             Skip                                                                                         │
│               Name: idx_body                                                                             │
│               Description: tokenbf_v1 GRANULARITY 1                                                      │
│               Parts: 47/116                                                                              │
│               Granules: 165/35779
```

N/A

### DIFF
--- a/backend/clickhouse/logs.go
+++ b/backend/clickhouse/logs.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"strings"
 	"time"
+	"unicode"
 
 	log "github.com/sirupsen/logrus"
 
@@ -488,8 +489,12 @@ func makeSelectBuilder(selectStr string, projectID int, params modelInputs.LogsP
 
 	filters := makeFilters(params.Query)
 
-	if len(filters.body) > 0 {
-		sb.Where("Body ILIKE" + sb.Var(filters.body))
+	for _, body := range filters.body {
+		if strings.Contains(body, "%") {
+			sb.Where("Body ILIKE" + sb.Var(body))
+		} else {
+			sb.Where("hasTokenCaseInsensitive(Body, " + sb.Var(body) + ")")
+		}
 	}
 
 	if len(filters.level) > 0 {
@@ -537,7 +542,7 @@ func makeSelectBuilder(selectStr string, projectID int, params modelInputs.LogsP
 }
 
 type filters struct {
-	body              string
+	body              []string
 	level             string
 	trace_id          string
 	span_id           string
@@ -547,7 +552,6 @@ type filters struct {
 
 func makeFilters(query string) filters {
 	filters := filters{
-		body:       "",
 		attributes: make(map[string]string),
 	}
 
@@ -558,13 +562,13 @@ func makeFilters(query string) filters {
 
 		if len(parts) == 1 && len(parts[0]) > 0 {
 			body := parts[0]
+
 			if strings.Contains(body, "*") {
 				body = strings.ReplaceAll(body, "*", "%")
-			}
-			if len(filters.body) == 0 {
-				filters.body = body
+				filters.body = append(filters.body, body)
 			} else {
-				filters.body = filters.body + " " + body
+				splitBody := strings.FieldsFunc(body, isSeparator)
+				filters.body = append(filters.body, splitBody...)
 			}
 		} else if len(parts) == 2 {
 			key, value := parts[0], parts[1]
@@ -586,11 +590,11 @@ func makeFilters(query string) filters {
 		}
 	}
 
-	if len(filters.body) > 0 && !strings.Contains(filters.body, "%") {
-		filters.body = "%" + filters.body + "%"
-	}
-
 	return filters
+}
+
+func isSeparator(r rune) bool {
+	return !unicode.IsLetter(r) && !unicode.IsDigit(r)
 }
 
 // Splits the query by spaces _unless_ it is quoted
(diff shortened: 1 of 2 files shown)


backend/clickhouse/logs.go
```diff
@@ -488,8 +489,12 @@ func makeSelectBuilder(selectStr string, projectID int, params modelInputs.LogsP
 
 	filters := makeFilters(params.Query)
 
-	if len(filters.body) > 0 {
-		sb.Where("Body ILIKE" + sb.Var(filters.body))
+	for _, body := range filters.body {
+		if strings.Contains(body, "%") {
+			sb.Where("Body ILIKE" + sb.Var(body))
```
et, line 494: `hasToken` can only operate like a wildcard if the search is at the beginning of the token. For other cases, we fall back to using `ILIKE`, like this test case:

https://github.com/highlight/highlight/blob/998afb633ae33ad31689b30f2da8121e4abde451/backend/clickhouse/logs_test.go#L543-L548

backend/clickhouse/logs_test.go
```diff
@@ -539,6 +553,20 @@ func TestReadLogsWithBodyFilter(t *testing.T) {
 	}, Pagination{})
 	assert.NoError(t, err)
 	assert.Len(t, payload.Edges, 1)
+
+	payload, err = client.ReadLogs(ctx, 1, modelInputs.LogsParamsInput{
+		DateRange: makeDateWithinRange(now),
+		Query:     "STRIPE_INTEGRATION_ERROR", // uses reserved separator character
+	}, Pagination{})
+	assert.NoError(t, err)
+	assert.Len(t, payload.Edges, 2)
```
et, line 562: This matches both logs, (`STRIPE_INTEGRATION_ERROR` and `STRIPE-INTEGRATION-ERROR`). This is an artifact of how `hasToken` works. 

The trade off is basically: gain perf win but at the possibility of matching more than expected.

I think we should optimize for this performance gain (since that gets the most value of searches) and if people dislike the overmatching, we can advise them to take advantage of structured logging more.

backend/clickhouse/logs.go
Vadman97: should line 493 be checking for `Contains(body, "*")`?

Vadman97 (approved):
significant perf improvement!

backend/clickhouse/logs.go
et: Good question. 

We sub the `*` for `%` earlier on:
https://github.com/highlight/highlight/blob/998afb633ae33ad31689b30f2da8121e4abde451/backend/clickhouse/logs.go#L566-L567

So that's why we check for `%` here.

But this definitely is a good call out that the logic is kind of scattered. I'll look into cleaning this up so it's more explicit about what is going on.

PR #4579, opened 2023-03-14
add tracing for querying clickhouse

et:
## Summary

We are currently optimizing slow queries for things we know about it but It would be great to have insight on other slow queries we don't know about.

This PR adds tracing data for our read logs queries.

Will observe in datadog.

N/A


backend/clickhouse/logs.go
```diff
@@ -134,8 +135,18 @@ func (client *Client) ReadLogs(ctx context.Context, projectID int, params modelI
 
 	sql, args := sb.Build()
 
+	span, _ := tracer.StartSpanFromContext(ctx, "logs", tracer.ResourceName("ReadLogs"))
+	query, err := sqlbuilder.ClickHouse.Interpolate(sql, args)
+	if err != nil {
+		return nil, err
```
Vadman97, line 141: `span.Finish(tracer.WithError(err))` before this return

Vadman97 (approved):

Issue #4456, opened 2023-03-02
Add new filters/fields to network requests


jay-khatri:
- filtering based on request status and type. 
- show request latency in network request panel

```
Add a section to the network request details that gives the user "latency" or how long the network request took to resolve.
discord convo to follow up w/: https://discord.com/channels/1026884757667188757/1076574340939788420/1076579008969441400

```
Please react to show interest and comment w/ any thoughts/questions/suggestions.

aaronmaxcarver:
Hey @julian-highlight - @jay-khatri and I have a couple design questions for you. Here's a loom - https://www.loom.com/share/5ae5f95776a14900bc7f7279baa1281a

- how can we allow users to clear filters for multi-select?

- with 2 different filters and multi-select, what to show on the button's face?
e.g. here
![Screenshot 2023-03-15 at 6 10 28 PM](https://user-images.githubusercontent.com/59704103/225463890-da7e2889-3539-445b-8c6a-12aeaa74f04a.png)

julian-highlight:
Thanks @aaronmaxcarver for collecting these thoughts! 
We can easily add a button inside the menu to delete filters! Will serve you with designs for that asap! 

To combat the other issue we have 2 options: 
1. Have a filters dropdown that acts as a hub for filters. The button itself could use 'X filters' as its copy whenever filters are applied. This will probably the easiest fix to the problem.
2. Commit to a full line of space for filters, where individual filters are visualized. This might be relevant if we plan to extend the list of filters soon! 
e.g. stripe: 
<img width="1405" alt="image" src="https://user-images.githubusercontent.com/114743131/225642125-fa7d45c7-c568-4cf9-862a-785df5687ddf.png">

PR #4580, opened 2023-03-14
Add latency field and http status filter to network requests

aaronmaxcarver:
## Summary

Issue #4456, "Add new filters/fields to network requests "
  - adds latency row
  - adds network request filter based on status

-Dev mode click test, Render URL click test

-None

### DIFF
--- a/frontend/src/pages/Player/Toolbar/DevToolsWindowV2/DevToolsWindowV2.tsx
+++ b/frontend/src/pages/Player/Toolbar/DevToolsWindowV2/DevToolsWindowV2.tsx
@@ -115,7 +115,7 @@ const DevToolsWindowV2: React.FC<
 			'3XX': 0,
 			'4XX': 0,
 			'5XX': 0,
-			'???': 0,
+			Unknown: 0,
 		}
 
 		parsedResources.forEach((request) => {
@@ -141,14 +141,13 @@ const DevToolsWindowV2: React.FC<
 						count['5XX'] += 1
 						break
 					default:
-						count['???'] += 1
+						count['Unknown'] += 1
 						break
 				}
 			}
 		})
 
 		return count
-		// eslint-disable-next-line react-hooks/exhaustive-deps
 	}, [parsedResources])
 
 	if (!showDevTools || isPlayerFullscreen) {
--- a/frontend/src/pages/Player/Toolbar/DevToolsWindowV2/DevToolsWindowV2.tsx
+++ b/frontend/src/pages/Player/Toolbar/DevToolsWindowV2/DevToolsWindowV2.tsx
@@ -24,10 +24,14 @@ import {
 import {
 	LogLevel,
 	LogLevelVariants,
+	RequestStatus,
 	RequestType,
 	Tab,
 } from '@pages/Player/Toolbar/DevToolsWindowV2/utils'
-import { ICountPerRequestType } from '@pages/Player/Toolbar/DevToolsWindowV2/utils'
+import {
+	ICountPerRequestStatus,
+	ICountPerRequestType,
+} from '@pages/Player/Toolbar/DevToolsWindowV2/utils'
 import useLocalStorage from '@rehooks/local-storage'
 import clsx from 'clsx'
 import React, { useMemo } from 'react'
@@ -51,6 +55,10 @@ const DevToolsWindowV2: React.FC<
 	const [requestType, setRequestType] = React.useState<RequestType>(
 		RequestType.All,
 	)
+	const [requestStatus, setRequestStatus] = React.useState<RequestStatus>(
+		RequestStatus.All,
+	)
+
 	const [searchShown, setSearchShown] = React.useState<boolean>(false)
 	const [logLevel, setLogLevel] = React.useState<LogLevel>(LogLevel.All)
 	const form = useFormState({
@@ -70,6 +78,7 @@ const DevToolsWindowV2: React.FC<
 
 	const { resources: parsedResources } = useResourcesContext()
 
+	/* Count request per type (XHR, etc.) */
 	const countPerRequestType = useMemo(() => {
 		const count: ICountPerRequestType = {
 			All: 0,
@@ -87,7 +96,7 @@ const DevToolsWindowV2: React.FC<
 			const requestType =
 				request.initiatorType as keyof ICountPerRequestType
 
-			//-- Only count request types defined in ICountPerRequestType, e.g. skip 'beacon' --//
+			/* Only count request types defined in ICountPerRequestType, e.g. skip 'beacon' */
 			if (count.hasOwnProperty(request.initiatorType)) {
 				count['All'] += 1
 				count[requestType] += 1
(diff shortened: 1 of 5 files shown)


aaronmaxcarver:
@jay-khatri this is the draft PR from last Tuesday. There's a render URL above to try it out

jay-khatri:
https://www.loom.com/share/e9994de2d2b442938268cd71acb6be66

- text in error callout should wrap
- each filter should have a key:val (i.e. "Type: All")
- in the selected state, the filter should be purple

@julian-highlight do you mind supporting with design? In particular, what style should @aaronmaxcarver use for the selected state?

julian-highlight:
Heyo, 
I've updated the [designs](https://www.figma.com/file/rdBPNzn7Klk6RG1LHUCE5B/Screen-Designs?node-id=6701-621859&t=8r4qnW9nJWRxl26E-11) and recorded a little [loom].(https://www.loom.com/share/eca52ce80bb5446f9ce0450257ce33ef)

Remarks: 
- We don't need to show 'Type: All' as that's the default state. Let's just use 'Type' in that case.
- We need to truncate filter stacks as we need to prioritize space for the tabs and search in the dev tools for now

@jay-khatri @aaronmaxcarver

frontend/src/pages/Player/Toolbar/DevToolsWindowV2/utils.ts
```diff
@@ -148,6 +148,26 @@ export interface ICountPerRequestType {
...
+	All = 'All',
+	'1XX' = '1XX',
+	'2XX' = '2XX',
+	'3XX' = '3XX',
+	'4XX' = '4XX',
+	'5XX' = '5XX',
+	'???' = '???',
+}
+
+export interface ICountPerRequestStatus {
```
et, line 161: Pretty sure we can simplify this to be:

```ts
export type ICountPerRequestStatus = {
    [key in RequestStatus]: number
};
```

frontend/src/pages/Player/Toolbar/DevToolsWindowV2/NetworkPage/NetworkPage.tsx
```diff
@@ -446,6 +470,9 @@ const ResourceRow = ({
 						  })
 						: MillisToMinutesAndSeconds(resource.startTime)}
 				</Text>
+				<Text size="small" weight={showingDetails ? 'bold' : 'medium'}>
+					{formatTime(resource.responseEnd - resource.startTime)}
```
et, line 474: Correct me if I"m wrong, but shouldn't `resource.startTime` be `resource.responseStart`?

frontend/src/pages/Player/Toolbar/DevToolsWindowV2/NetworkPage/NetworkPage.tsx
```diff
@@ -102,6 +105,23 @@ export const NetworkPage = ({
 						requestType === RequestType.All ||
 						requestType === request.initiatorType,
 				)
+				.filter((request) => {
```
et, line 108: There's a bug with how the request type and status interact with each other.
Here's a loom to describe the issue: https://www.loom.com/share/109f9d3fe14d493db7f97700ff8bb4aa

Happy to talk through how we might fix this.

et (commented):
Nice work @aaronmaxcarver! 
Left a couple comments. Let me know if you'd like to pair on getting this finished. Happy to help!

frontend/src/pages/Player/Toolbar/DevToolsWindowV2/DevToolsWindowV2.tsx
```diff
@@ -93,6 +102,51 @@ const DevToolsWindowV2: React.FC<
...
+	/* Count request per http status (200, etc.) */
+	const countPerRequestStatus = useMemo(() => {
+		const count: ICountPerRequestStatus = {
+			All: 0,
+			'1XX': 0,
+			'2XX': 0,
+			'3XX': 0,
+			'4XX': 0,
+			'5XX': 0,
+			'???': 0,
```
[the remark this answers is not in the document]
et: > This is not for ranges outside of 500. It's for network requests we weren't able to record (given that our snippet loads async).

Thanks for explaining this. But if we weren't able to record this, how will we have the data? To put it another way, can you link a session where this is a non-zero value?

aaronmaxcarver:
The code in PR is pretty janky and I personally wouldn't recommend shipping it. I think we should:
-use UI components that have the functionality we want
-then make filtering logic nice
-then make the component look nice

frontend/src/pages/Player/Toolbar/DevToolsWindowV2/NetworkPage/NetworkPage.tsx
jay-khatri: @aaronmaxcarver , yes, a new column sounds good. Though, maybe this is a better fit for a new PR? Will defer to @Vadman97 on exact definitions.

aaronmaxcarver:
Notes on special cases encountered when filtering network resources:
- Only `fetch` and `xhr` requests have https statuses being recorded. These statuses are in the property `request.requestResponsePairs.response.status`. They are recorded by watching the Fetch API.
- For requests without a status:
  - If it's a `fetch` or `xhr` request, then we didn't record the request so we call its status `Unknown`. This is because the async snippet watching the Fetch API didn't load in time to see the request happen, but we still have the request's timing data from the Performance API
  - If it's a different request type (e.g. `css`), we render its status as `200` in rows and such

frontend/src/pages/Player/Toolbar/DevToolsWindowV2/NetworkPage/NetworkPage.tsx
aaronmaxcarver: @jay-khatri Yeah new column was requested per the original issue, https://github.com/highlight/highlight/issues/4456#issue-1605911039. 

The question is do we want `latency` or do we want `duration`? What do you want to show the user?

They are not the same value, so I don't know which one to calculate and display. Can you provide guidance on which one you want to show users? Then I'll follow @Vadman97 guidance on the proper definition.

Side note -the value `resource.responseEnd - resource.startTime` is already being display in the right hand panel (visible after clicking a request row). We call it 'time' currently. I'll update its name to 'Duration'.

Vadman97: for latency, i suggest `resource.responseEnd - resource.responseStart`
for duration, `resource.responseEnd - resource.startTime` is what we want since that captures the entire actual request send/response (ignoring having the request data getting rendered via the dom events)

i think it would be a bit busy to include both `latency` and `duration` though. i think we should just have a column for `duration` and revisit a more granular breakdown in the future (if we want to mimic the full google chrome dev tools flame graph at some point)

frontend/src/pages/Player/Toolbar/DevToolsWindowV2/DevToolsWindowV2.tsx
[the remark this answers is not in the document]
Vadman97: Screenshot 2023-03-31 at 1 55 19 PM

eg https://app.highlight.io/1/sessions/TO1AVhFocWzCAFD2133SMlwDU8QH

this happens when the `window.performance` api records a network request (the browser should have *all* network requests recorded by window.performance), but highlight doesn't catch it. the `window.performance` api has network request timing info but does not have request/response data.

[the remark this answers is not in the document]
Vadman97: updating this text to `Unknown`

frontend/src/pages/Player/Toolbar/DevToolsWindowV2/NetworkPage/NetworkPage.tsx
Vadman97: fixed this in #2398c86 thanks for catching this!

Vadman97 (approved):

PR #4581, opened 2023-03-14
add @highlight-run/nest sdk for nest.js api errors and logs

Vadman97:


sdk/highlight-node/src/sdk.ts
```diff
@@ -25,7 +25,8 @@ export interface HighlightInterface {
 		tags?: { name: string; value: string }[],
 	) => void
 	flush: () => Promise<void>
-	log: (...data: any[]) => void
+	log: (message: any, level: string, ...optionalParams: any[]) => void
```
et, line 28: What do you think about exposing this `level` as a string union?
https://github.com/highlight/highlight/blob/702bc392acb78af5416bd06584d1db87e139aae0/sdk/highlight-node/src/hooks.ts#L37-L47

Something like:

```ts
const Levels = {
		debug: 'debug',
		info: 'info',
		log: 'info',
		count: 'info',
		dir: 'info',
		warn: 'warn',
		assert: 'warn',
		error: 'error',
		trace: 'trace',
	} as { [k in keyof Console]: string }

type Level = keyof Levels
```

And replacing `level: string` with `level: Level`

sdk/highlight-node/src/sdk.ts
```diff
@@ -84,6 +85,30 @@ export const H: HighlightInterface = {
...
 	},
+	log: (
+		message: any,
+		level: string,
+		secureSessionId?: string | undefined,
+		requestId?: string | undefined,
+	) => {
+		const o: { stack: any } = { stack: {} }
+		const saveTraceLimit = Error.stackTraceLimit
+		Error.stackTraceLimit = 3
```
et, line 96: Out of curiosity, why 3?

et (approved):
Awesome 🚢

sdk/highlight-node/src/sdk.ts
Vadman97: good catch - we shouldn't be doing a limit on the sdk side but rather limit how we render long frames if necessary on the frontend. i'll remove this here.

PR #4583, opened 2023-03-14
Don't allow integration setup when disabled

ccschmitz:
## Summary

Ensures we don't initialize Mixpanel or Amplitude integrations when they are explicitly disabled. Also removes a Mixpanel warning which wasn't accurate and makes the tokens/keys optional in the integration objects optional so you don't need to specify them when trying to set `disabled: true`.

Local click test.

Could possibly change tracking behavior if projects were not configured correctly in the past (were passing the token/API key + `disabled`).

### DIFF
--- a/sdk/firstload/package.json
+++ b/sdk/firstload/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "highlight.run",
-	"version": "5.4.0",
+	"version": "5.4.1",
 	"scripts": {
 		"build": "yarn typegen && rollup -c",
 		"dev": "rollup -c -w",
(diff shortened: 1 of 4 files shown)


sdk/firstload/package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "highlight.run",
-	"version": "5.4.0",
+	"version": "5.4.1",
```
Vadman97, line 3: could you also please update `CHANGELOG.md` in this directory describing the change?

Vadman97 (approved):
thanks for jumping on this @ccschmitz !

PR #4586, opened 2023-03-14
Enable Reflame for Highlight web app

lewisl9029:
## Summary

This PR finally enables [Reflame](https://reflame.app/) [previews](https://preview.highlight.io/~r/start-preview/?mode=production&variantId=01GY11FVTK9FMXX2DSCR6T4VRD&variantName=git%7Enew-reflame-app-1&userId=01FQZZ7XJFDA799Z1Z9DRCFXWA) for the Highlight web app. 🥳 

### What does this mean for me?

First you'll need to sign up at https://reflame.app, connect your GitHub account, and ask @Vadman97 for an invite to the highlight organization.

Then, once this PR is merged, every time you push up a change to the web app (in /frontend) or any of its dependencies, you'll get a Reflame preview link on your commit immediately, usually within 3 seconds, instead of however many minutes it used to take:
Screenshot 2023-04-13 at 5 40 57 PM

In addition, you'll have access to the Reflame [VSCode extension](https://marketplace.visualstudio.com/items?itemName=reflame.agent) for development, which deploys your changes usually within ~500ms of a file save, and then reflects your changes instantly with a full browser refresh in production mode, or with state-preserving React fast refresh in development mode. See development mode in action here:

https://user-images.githubusercontent.com/6934200/231914985-3679e955-ddcf-4ad1-9c7e-1c7451dc3ef5.mp4

It's worth noting that Reflame is actually _deploying your changes to the internet_ every time, so you can send these links to yourself to check your changes on another device (even multiple devices simultaneously), or share them with teammates or customers to give them a sneak peak of what you're working on, iterate with their feedback, and have those changes reflected on their browsers in real time (even if they're on the other side of the world)!

### How do I even review this? There's like 500 files here?!

I'd recommend reviewing either commit by commit and skipping the 2nd commit, or by viewing all files changed and skipping over everything in `__generated` folders, since they only contain files generated using the newly added build scripts. This should leave only about 30 files to review, and most of the changes are a only a couple of lines long.

These scripts and generated files are a temporary stop-gap to support features that don't have first-class support in Reflame yet, specifically:

- Reading version from package.json
- CSS/SCSS modules
- Tailwind
- SVGR
- Git Submodules
- Vanilla Extract

These are roughly ranked in order of how quickly I think they will get first-class support in Reflame, at which point we'll be able to remove the associated scripts and generated files. Notably, Vanilla Extract is as far down the list as it is because it requires executing user-supplied code as part of its build process, which is going to take quite a bit of work to enable safely in a multitenant system like Reflame (but I do plan on tackling this eventually as I get closer to building features like testing and backend APIs support). Though we may still be able to get rid of the build script sooner than that by building it into the VSCode extension if there's enough demand.

Outside of the script and configuration changes, there are a few additional source changes for compatibility. I left comments on the PR for every one of these explaining the motivation behind them. I've also split most of them out into separate PRs so they can be reviewed, tested and shipped independently:

- https://github.com/highlight/highlight/pull/5086
- https://github.com/highlight/highlight/pull/5087
- https://github.com/highlight/highlight/pull/5088
- https://github.com/highlight/highlight/pull/5089 

A lot of care has gone into making sure your existing local dev workflow works exactly as you're used to (just with a few more scripts running than before), and that the production deployment process remains untouched as well. If you notice any material differences in any of your day to day workflows while trying out this PR, or in the Render preview deploys, please let me know and I'll try to address it ASAP.

I've tried following the [docker dev guide here](https://www.highlight.io/docs/getting-started/self-host/dev-deployment-guide) and running `yarn dev:frontend` (without the `doppler run --` part), and both seem to be working identically as on main as far as I can tell, though for the latter I'm missing a few env vars from doppler so couldn't verify past the login screen, will need your help to make sure everything works as expected there.

If you want to try out Reflame before this PR gets merged, just make a branch off of this PR (previews are not working for this PR because it's coming from a fork, and the appId in the config is for the highlight/highlight repo, but it should work if you cherry pick these commits into another branch within this repo). The VSCode extension should also just work once you get a VSCode account. [Here's](https://preview.highlight.io/~r/start-preview/?mode=production&variantId=01GY11FVTK9FMXX2DSCR6T4VRD&variantName=git%7Enew-reflame-app-1&userId=01FQZZ7XJFDA799Z1Z9DRCFXWA) the preview I'm using to test my changes.

Would love to make Reflame better with your feedback! Just leave a comment here, [shoot me an email](mailto:lewis@reflame.app), or ask for an invite to the #highlight-reflame channel in Slack to chat there or send me a DM.

Definitely would be helpful to get a Render preview for this to poke around in.


frontend/src/components/Avatar/Avatar.tsx
```diff
@@ -3,12 +3,10 @@ import React, { useMemo } from 'react'
 import ReactNiceAvatar, { genConfig } from 'react-nice-avatar'
 
 import { generateRandomColor } from '../../util/color'
-import {
-	userAvatar,
-	userAvatarBorder,
-	userAvatarText,
-	userAvatarWrapper,
-} from './Avatar.module.scss'
+import styles from './Avatar.module.scss'
```
lewisl9029, line 6: The esbuild plugin I'm using for generating js modules for scss doesn't seem to support named exports. Luckily there's only a handful of named export usages across the codebase, so I've manually refactored them to use the default export.

frontend/src/components/SearchEmptyState/SearchEmptyState.tsx
```diff
@@ -3,14 +3,11 @@ import clsx from 'clsx'
...
 import { ReactComponent as EmptyState } from '../../static/empty-state.svg'
-import {
-	emptyStateSection,
-	emptyStateWrapper,
-	emptySubTitle,
-	emptyTitle,
-} from './SearchEmptyState.module.scss'
 import styles from './SearchEmptyState.module.scss'
 
+const { emptyStateSection, emptyStateWrapper, emptySubTitle, emptyTitle } =
```
lewisl9029, line 8: See https://github.com/highlight/highlight/pull/4586/commits/21110fc9c5d9ed3a9f6d573816330f1e53426f39#r1167356999

frontend/src/index.tsx
```diff
@@ -51,6 +51,8 @@ import { BrowserRouter, Route, Routes, useLocation } from 'react-router-dom'
 import { QueryParamProvider } from 'use-query-params'
 import { ReactRouter6Adapter } from 'use-query-params/adapters/react-router-6'
 
+document.body.className = 'highlight-light-theme'
```
lewisl9029, line 54: We don't currently support arbitrary html customizations (we currently generate the html from scratch using the config and dependency graph in order to save a network roundtrip by not having to fetch a template on every deploy). This is a workaround for not being able to just set a class attribute on body. Under Vite this should basically be a noop.

frontend/index.html
```diff
@@ -83,40 +83,7 @@
...
-					}
-					if (w.attachEvent) {
-						w.attachEvent('onload', l)
-					} else {
-						w.addEventListener('load', l, false)
-					}
-				}
-			})()
-		</script>
+		<script src="./src/intercom.js" defer></script>
```
lewisl9029, line 86: Moved intercom and canny scripts out of html and into their own scripts in src folder because we can't add arbitrary HTML like this in Reflame, but we can add additional entry point scripts. A side benefit is this should result in a slightly smaller document payload, and slightly less network traffic overall since these scripts will stay cached on repeat visits as they rarely ever change.

Also added defer to keep sequential execution order while keeping network requests parallelized.

frontend/index.html
```diff
@@ -142,8 +109,8 @@
 			src="//js.hs-scripts.com/20473940.js"
 		></script>
 		<!-- End of HubSpot Embed Code -->
-
-		<script type="module" src="./src/index.tsx"></script>
+		<script src="./src/canny.js" defer></script>
```
lewisl9029, line 112: Moved this from body to up here and switched from async to defer to make sure this is executed before the index bundle. Otherwise if index ever gets executed first, references to Canny will fail (unlikely because index is so much bigger, but still possible when cached).

frontend/src/pages/Player/PlayerHook/utils/usePlayerConfiguration.tsx
```diff
@@ -6,7 +6,7 @@ import { Tab } from '@pages/Player/Toolbar/DevToolsWindowV2/utils'
 import useLocalStorage from '@rehooks/local-storage'
 import { useCallback, useMemo } from 'react'
 
-import { EventsForTimeline } from '.'
+import { EventsForTimeline } from './index'
```
lewisl9029, line 9: Had to change this because `'.'` isn't a valid relative URL specifier according to the spec (needs to start with `./` or `../`).

Reflame unlike most other tools tries to avoid module specifier rewriting as much as possible, because specifier rewriting can be extremely wasteful in terms of write amplification when combined with content-addressed modules.

frontend/src/util/attribution.ts
```diff
@@ -1,7 +1,6 @@
+import { GenerateSecureID } from '@highlight-run/client/src/utils/secure-id'
 import Cookies from 'js-cookie'
 
-import { GenerateSecureID } from '../../../sdk/client/src/utils/secure-id'
```
lewisl9029, line 3: Changed this because relative imports outside the package root is incompatible with Reflame's stricter package isolation, and as far as I could see, there's nothing that necessitates their use here.

frontend/index.html
```diff
@@ -83,40 +83,7 @@
...
-					}
-					if (w.attachEvent) {
-						w.attachEvent('onload', l)
-					} else {
-						w.addEventListener('load', l, false)
-					}
-				}
-			})()
-		</script>
+		<script type="module" src="./src/intercom.js" defer></script>
```
lewisl9029, line 86: Moved intercom and canny scripts out of html and into their own scripts in src folder because we can't add arbitrary HTML like this in Reflame, but we can add additional entry point scripts. A side benefit is this should result in a slightly smaller document payload, and slightly less network traffic overall since these scripts will stay cached on repeat visits as they rarely ever change. (UPDATE: just found out that this not true for Vite, since all of these scripts just end up getting bundled into a single index.js file...)

Also added defer to keep sequential execution order while keeping network requests parallelized.

frontend/index.html
```diff
@@ -142,8 +109,8 @@
 			src="//js.hs-scripts.com/20473940.js"
 		></script>
 		<!-- End of HubSpot Embed Code -->
-
-		<script type="module" src="./src/index.tsx"></script>
+		<script type="module" src="./src/canny.js" defer></script>
```
lewisl9029, line 112: Moved this from body to up here and switched from async to defer to make sure this is executed before the index bundle. Otherwise if index ever gets executed first, references to Canny will fail (unlikely because index is so much bigger, but still possible when cached).

packages/ui/src/components/DatePicker/Calendar/Calendar.tsx
```diff
@@ -1,5 +1,7 @@
 import React from 'react'
-import { Box, Stack, Text } from '@highlight-run/ui'
+import { Box } from '../../Box/Box'
+import { Stack } from '../../Stack/Stack'
+import { Text } from '../../Text/Text'
```
lewisl9029, line 4: Had to do a bunch of untangling of circular dependencies here.

Reflame's development mode makes use of [es-module-shims](https://github.com/guybedford/es-module-shims)' shim mode to be able to hook into module loading and update import maps at runtime. It generally does a good job of supporting circular imports transparently, but it seems to run into some trouble with this specific pattern where index.ts is full of reexports. I'll try to find a minimal repro and file an issue there.

packages/ui/src/components/DatePicker/Calendar/DatePicker.tsx
```diff
@@ -1,9 +1,8 @@
 import React from 'react'
-import {
-	Box,
-	IconSolidCheveronLeft,
-	IconSolidCheveronRight,
-} from '@highlight-run/ui'
+import { Box } from '../../Box/Box'
+import { IconSolidCheveronLeft } from '../../icons/IconSolidCheveronLeft'
+import { IconSolidCheveronRight } from '../../icons/IconSolidCheveronRight'
```
lewisl9029, line 4: See https://github.com/highlight/highlight/pull/4586/files#r1167367592

packages/ui/src/components/DatePicker/Calendar/Day.tsx
```diff
@@ -1,5 +1,5 @@
 import React from 'react'
-import { Box } from '@highlight-run/ui'
+import { Box } from '../../Box/Box'
```
lewisl9029, line 2: See https://github.com/highlight/highlight/pull/4586/files#r1167367592

packages/ui/src/components/Tabs/Tabs.tsx
```diff
@@ -2,7 +2,8 @@ import React from 'react'
 import { Box } from '../Box/Box'
 
 import * as styles from './styles.css'
-import { Button, Text } from '../../components'
+import { Button } from '../../components/Button/Button'
+import { Text } from '../../components/Text/Text'
```
lewisl9029, line 6: See https://github.com/highlight/highlight/pull/4586/files#r1167367592

sdk/firstload/src/browserExtension/extensionListener.ts
```diff
@@ -1,4 +1,4 @@
-import { H } from '..'
+import { H } from '../index'
```
lewisl9029, line 1: Similar to https://github.com/highlight/highlight/pull/4586/files#r1167365950, `..` is also not a valid relative specifier.

sdk/firstload/src/index.tsx
```diff
@@ -22,6 +21,7 @@ import type {
 } from '@highlight-run/client/src/types/types'
 import HighlightSegmentMiddleware from './integrations/segment'
 import configureElectronHighlight from './environments/electron'
+import firstloadVersion from './__generated/version'
```
lewisl9029, line 24: I decided to refer to the generated module directly here instead of import mapping around it, because this allows us to stop bundling the entire package.json into the bundle just to access the version in the Rollup build.

packages/ui/src/components/Form/styles.css.ts
```diff
@@ -41,8 +41,10 @@ export const inputVariants = recipe({
 				width: 0,
 				padding: 0,
 				border: 'none',
-				'&:focus, &:active, &:placeholder-shown:hover': {
```
lewisl9029, line 44: This was ending up in the final CSS output apparently (screenshot from prod):

Screenshot 2023-04-17 at 5 42 56 PM


Not valid syntax outside of browsers with draft CSS nesting support, though support is actually surprisingly good for a draft: https://caniuse.com/css-nesting

Vadman97 (approved):
excited to start trying reflame out!

Issue #4464, opened 2023-03-02
Normalize severity text


et:
We can do this in `NewLogRow`: https://github.com/highlight/highlight/blob/main/backend/clickhouse/logs.go#L36-L49

Should lowercase and default to `info`

et:
Fixed by #4622

PR #4597, opened 2023-03-14
skip writing log attributes if these are internal attributes

et:
## Summary

`log.severity` and `log.message` aren't being excluded from `LogAttributes`.

This PR fixes that. It also moves some data parsing logic down out of the OTEL code and into the logs code. This makes it easier to unit test and it sets up a better interface for building a log row.

Confirmed `log.severity` and `log.message`  aren't present in log attributes anymore.

Screenshot 2023-03-14 at 10 42 54 PM

N/A

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -13,7 +13,7 @@
 			"port": 2345,
 			"substitutePath": [
 				{ "from": "${workspaceFolder}", "to": "/build" },
-			],
+			]
 		}
 	]
 }
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -11,9 +11,7 @@
 			"debugAdapter": "dlv-dap",
 			"mode": "remote",
 			"port": 2345,
-			"substitutePath": [
-				{ "from": "${workspaceFolder}", "to": "/build" },
-			]
+			"substitutePath": [{ "from": "${workspaceFolder}", "to": "/build" }]
 		}
 	]
 }
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -8,8 +8,12 @@
 			"name": "Attach to Process",
 			"type": "go",
 			"request": "attach",
-			"mode": "local",
-			"processId": "main"
+			"debugAdapter": "dlv-dap",
+			"mode": "remote",
+			"port": 2345,
+			"substitutePath": [
+				{ "from": "${workspaceFolder}", "to": "/build" },
+			],
 		}
 	]
 }
--- a/backend/clickhouse/log_row.go
+++ b/backend/clickhouse/log_row.go
@@ -6,7 +6,7 @@ import (
 
 	"github.com/google/uuid"
 	"github.com/highlight/highlight/sdk/highlight-go"
-	"github.com/influxdata/influxdb-client-go/v2/log"
+	log "github.com/sirupsen/logrus"
 )
 
 type LogRowPrimaryAttrs struct {
--- a/backend/clickhouse/log_row.go
+++ b/backend/clickhouse/log_row.go
@@ -78,7 +78,7 @@ func getAttributesMap(resourceAttributes, eventAttributes map[string]any) map[st
 			for _, attr := range highlight.InternalAttributes {
 				if k == attr {
 					shouldSkip = true
-					continue
+					break
 				}
 			}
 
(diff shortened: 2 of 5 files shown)


backend/otel/otel.go
```diff
@@ -82,32 +82,6 @@ func projectToInt(projectID string) (int, error) {
 	return 0, e.New(fmt.Sprintf("invalid project id %s", projectID))
 }
 
-func getAttributesMap(resourceAttributes, eventAttributes map[string]any) map[string]string {
-	attributesMap := make(map[string]string)
-	for _, m := range []map[string]any{resourceAttributes, eventAttributes} {
-		for k, v := range m {
-			for _, attr := range highlight.InternalAttributes {
-				if k == attr {
-					continue
```
et, line 91: This was the bug...we were doing a `continue` for the inner loop (`for _, attr := range highlight.InternalAttributes`) but we needed to `continue` out of the outer loop `for k, v := range m {`

backend/clickhouse/log_row.go
```diff
@@ -0,0 +1,104 @@
...
+
+func (l *LogRow) Cursor() string {
+	return encodeCursor(l.Timestamp, l.UUID)
+}
+
+type LogRowOption func(*LogRow)
+
+func WithLogAttributes(resourceAttributes, eventAttributes map[string]any) LogRowOption {
+	return func(h *LogRow) {
+		h.LogAttributes = getAttributesMap(resourceAttributes, eventAttributes)
```
et, line 60: Ideally, we can make things like `LogAttributes` private (`logAttributes`) and external code can only use these `With*` functions to set the value.

For example, this would help with something like #4464 where we could write code like:

```go
logRow1 := NewLogRow(
  WithSeverityText("INFO")
)
logRow2 := NewLogRow(
  WithSeverityText("info")
)
```

and the underlying code in `WithSeverityText` would handle normalizing to some agreed format.

.vscode/launch.json
```diff
@@ -8,8 +8,10 @@
 			"name": "Attach to Process",
 			"type": "go",
 			"request": "attach",
-			"mode": "local",
-			"processId": "main"
+			"debugAdapter": "dlv-dap",
```
et, line 11: Modified our vscode settings to work with docker.

Vadman97: nice!

backend/clickhouse/log_row.go
```diff
@@ -0,0 +1,104 @@
...
+		LogRowPrimaryAttrs: LogRowPrimaryAttrs{
+			Timestamp:       attrs.Timestamp,
+			TraceId:         attrs.TraceId,
+			SpanId:          attrs.SpanId,
+			ProjectId:       attrs.ProjectId,
+			SecureSessionId: attrs.SecureSessionId,
+		},
+		UUID:           uuid.New().String(),
+		SeverityText:   "INFO",
+		SeverityNumber: int32(log.InfoLevel),
```
Vadman97, line 42: this imported `github.com/influxdata/influxdb-client-go/v2/log` - is that consistent with the otel value for the info log level?

backend/clickhouse/log_row.go
```diff
@@ -0,0 +1,104 @@
...
+func getAttributesMap(resourceAttributes, eventAttributes map[string]any) map[string]string {
+	attributesMap := make(map[string]string)
+	for _, m := range []map[string]any{resourceAttributes, eventAttributes} {
+		for k, v := range m {
+			shouldSkip := false
+
+			for _, attr := range highlight.InternalAttributes {
+				if k == attr {
+					shouldSkip = true
+					continue
```
Vadman97, line 81: can `break` here, looks like that was my bad.

Vadman97 (approved):

backend/clickhouse/log_row.go
et: Good catch. The auto-import caught the wrong import.

et: fe8345b0b05acb8bc69a471bb9ede7dc0d054dfa

backend/clickhouse/log_row.go
et: +1 
d6da312bdf67cf64d28e9dd670d3c399eaf02510

PR #4610, opened 2023-03-15
fix filtering by log level

et:
## Summary

Filtering by log level is sometimes broken because we haven't normalized it before writing to the db (something that will have a more robust fix in #4464).

To compensate for the discrepancy, in [#4555,](https://github.com/highlight/highlight/pull/4555/files#diff-44251d85d1baf341cd7313aaa7aa61bd8d4832267460557e8cbb6b215ba3b52fR373-R378) the frontend is lowercasing the level but this is something the backend should just be figuring out.

This PR ensures that the backend always sends the frontend lowercase log levels and that we always write it to lowercase before hitting the clickhouse db.

* Confirmed the backend is sending lowercase log levels to the frontend
* Confirmed we can filter by log level via the frontend

N/A

### DIFF
--- a/backend/clickhouse/logs.go
+++ b/backend/clickhouse/logs.go
@@ -69,6 +69,8 @@ func (client *Client) BatchWriteLogRows(ctx context.Context, logRows []*LogRow) 
 		if len(logRow.UUID) == 0 {
 			logRow.UUID = uuid.New().String()
 		}
+		// TODO (et) - move this logic to a builder function (#4464)
+		logRow.SeverityText = strings.ToLower(logRow.SeverityText)
 		err = batch.AppendStruct(logRow)
 		if err != nil {
 			return err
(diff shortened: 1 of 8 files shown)


backend/clickhouse/logs.go
```diff
@@ -69,6 +69,8 @@ func (client *Client) BatchWriteLogRows(ctx context.Context, logRows []*LogRow)
 		if len(logRow.UUID) == 0 {
 			logRow.UUID = uuid.New().String()
 		}
+		// TODO (et) - move this logic to a builder function (#4464)
+		logRow.SeverityText = strings.ToLower(logRow.SeverityText)
```
et, line 73: This partially fixes #4464. The true fix though should also limit the number of levels we support:

```sql
SELECT distinct(SeverityText) from logs;



log
--
warn
debug
ERROR
WARN
INFO
info
group
assert
timeEnd
time
trace
 
groupCollapsed
stdout
stderr
dir
error
edge-function-invocation
fatal
unknown
table
Info
groupEnd
middleware-invocation
```

We need to default everything that's not in `LogLevel` to be just `info`

ccschmitz (approved):

Vadman97 (approved):

PR #4611, opened 2023-03-15
add backfill script for hubspot contact associations

Vadman97:
## Summary

* Adds a hubspot contact / company backfilling script for populating hubspot with existing admins and workspaces.
* Fixes bug in hubspot that would avoid creating hubspot contacts for most users.

Running backfill script locally.

No

### DIFF
--- a/backend/migrations/cmd/backfill-hubspot/main.go
+++ b/backend/migrations/cmd/backfill-hubspot/main.go
@@ -1,0 +1,95 @@
+package main
+
+import (
+	"context"
+	hubspotApi "github.com/highlight-run/highlight/backend/hubspot"
+	"github.com/highlight-run/highlight/backend/model"
+	public "github.com/highlight-run/highlight/backend/public-graph/graph"
+	"github.com/leonelquinteros/hubspot"
+	log "github.com/sirupsen/logrus"
+	"gorm.io/gorm"
+	"os"
+	"sync"
+)
+
+const BatchSize = 64
+
+func main() {
+	ctx := context.Background()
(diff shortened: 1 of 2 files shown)


backend/migrations/cmd/backfill-hubspot/main.go
```diff
@@ -0,0 +1,95 @@
...
+	wg := sync.WaitGroup{}
+	inner := func(tx *gorm.DB, batch int) error {
+		wg.Add(1)
+		func() {
+			for _, admin := range admins {
+				if admin.Email == nil || admin.UserDefinedRole == nil || admin.UserDefinedPersona == nil || admin.FirstName == nil || admin.LastName == nil || admin.Phone == nil || admin.Referral == nil {
+					log.WithContext(ctx).WithField("AdminID", admin.ID).Error("admin is not filled out")
+					continue
+				}
+				hubspotContactId, err := r.HubspotApi.CreateContactForAdmin(
```
et, line 41: Is this hubspot call (and the rest of the hubspot calls) part of some sign up flow? Can we call some existing function that does this work? Mostly thinking that keeping the hubspot logic for new users in sync would be difficult if it's duplicate.

et (approved):

backend/migrations/cmd/backfill-hubspot/main.go
Vadman97: good idea, but it's part of a more complicated workflow using about you form data. this seems pretty specific to the backfilling logic so i'm going to keep this separate.

Screenshot 2023-03-15 at 1 29 15 PM

PR #4618, opened 2023-03-16
ensure error logs create error objects

Vadman97:
## Summary

OTEL error level logs should create error objects, so that our UI can successfully link logs to errors.

Local deploy.

No

### DIFF
--- a/backend/clickhouse/log_row.go
+++ b/backend/clickhouse/log_row.go
@@ -72,9 +72,9 @@ func (l *LogRow) Cursor() string {
 
 type LogRowOption func(*LogRow)
 
-func WithLogAttributes(resourceAttributes, eventAttributes map[string]any) LogRowOption {
+func WithLogAttributes(resourceAttributes, eventAttributes map[string]any, isFrontendLog bool) LogRowOption {
 	return func(h *LogRow) {
-		h.LogAttributes = getAttributesMap(resourceAttributes, eventAttributes)
+		h.LogAttributes = getAttributesMap(resourceAttributes, eventAttributes, isFrontendLog)
 	}
 }
 
@@ -115,7 +115,7 @@ func cast[T string | int64 | float64](v interface{}, fallback T) T {
 	return c
 }
 
-func getAttributesMap(resourceAttributes, eventAttributes map[string]any) map[string]string {
+func getAttributesMap(resourceAttributes, eventAttributes map[string]any, isFrontendLog bool) map[string]string {
 	attributesMap := make(map[string]string)
 	for _, m := range []map[string]any{resourceAttributes, eventAttributes} {
 		for k, v := range m {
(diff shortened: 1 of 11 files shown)


backend/otel/otel.go
```diff
@@ -80,6 +81,64 @@ func projectToInt(projectID string) (int, error) {
 	return 0, e.New(fmt.Sprintf("invalid project id %s", projectID))
 }
 
+func getLogRow(ctx context.Context, ts time.Time, lvl, projectID, sessionID, traceID, spanID string, excMessage string, resourceAttributes, eventAttributes map[string]any) *clickhouse.LogRow {
```
et, line 84: Would it be helpful to add more things to `NewLogRow`? This seems like a lot parameters that would be easy to mix the order.

Was thinking something like 

```go
clickhouse.NewLogRow(
  WithTimestamp(ts),
  WithProjectId(projectId),
  WithTraceId(traceId),
 ...
)
```

We can have it also handle errors too (for validating that the project id is valid).

et (commented):
Can you check to see if `severityText=log` should be an error? Everything else is not-blocking.

backend/clickhouse/log_row.go
```diff
@@ -115,6 +115,12 @@ func getAttributesMap(resourceAttributes, eventAttributes map[string]any) map[st
 
 func makeLogLevel(severityText string) modelInputs.LogLevel {
 	switch strings.ToLower(severityText) {
+	case "log":
+		return modelInputs.LogLevelError
```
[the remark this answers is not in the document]
Vadman97: good catch!

backend/otel/otel_test.go
```diff
@@ -0,0 +1 @@
+package otel
```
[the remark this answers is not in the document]
Vadman97: added a basic test here that i'll improve in the future

backend/otel/otel.go
Vadman97: cleaned this up!

backend/otel/otel_test.go
```diff
@@ -0,0 +1,34 @@
...
+
+func (m *MockResponseWriter) WriteHeader(statusCode int) {
+	return
+}
+
+func TestHandler_HandleLog(t *testing.T) {
+	w := &MockResponseWriter{}
+	r, _ := http.NewRequest("POST", "", strings.NewReader(""))
+	h := Handler{}
+	h.HandleLog(w, r)
```
et, line 27: Nice. This will make it easier to trace through the code with various scenarios.

docker/compose.yml
```diff
@@ -228,6 +228,7 @@ services:
             retries: 5
         ports:
             - 3000:3000
+            - 6006:6006
```
et, line 231: Is this to make storybook accessible?

et (approved):
Nice work! Thank you.

docker/compose.yml
Vadman97: yup

PR #4621, opened 2023-03-16
use retryables table in opensearch-update job to reduce DB I/O

mayberryzane:
## Summary

- instead of querying all updated records for the past 30 mins, use the `retryables` table to get only the records needed to update

- mocked some `retryables` records locally to make sure documents were pushed to opensearch for all data types and the records were marked with `deleted_at` after updating

- no, this modifies the existing `update-opensearch` job, and I already created this index 
```CREATE INDEX idx_to_retry ON public.retryables USING btree (type, payload_type, payload_id, created_at) WHERE (deleted_at IS NULL)```

### DIFF
--- a/backend/worker/opensearch.go
+++ b/backend/worker/opensearch.go
@@ -2,7 +2,9 @@ package worker
 
 import (
 	"context"
+	"fmt"
 	"reflect"
+	"time"
 
 	"github.com/highlight-run/highlight/backend/model"
 	"github.com/highlight-run/highlight/backend/opensearch"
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: backend/worker/opensearch.go [hunk 1]
```diff
@@ -48,25 +49,54 @@ func (w *Worker) IndexSessions(ctx context.Context, isUpdate bool) {
 		return nil
 	}
 
-	whereClause := "True"
+	start := time.Now()
```
Vadman97, line 52: cool, so this is using the time to make the fetch, processed, update atomic, right?

Vadman97 (approved):

PR #4624, opened 2023-03-16
fix clickhouse tracing calls to correctly match the query time

et:
## Summary

The span to measure a clickhouse read query is incorrectly returning early. I noticed this when the app is taking a couple seconds to load, yet datadog says the query took a couple milliseconds:

Screenshot 2023-03-16 at 1 47 18 PM

What's happening is that we're using `rows.Next()` ([source](https://github.com/ClickHouse/clickhouse-go/blob/b3e5d32dcbc36d6abea48a5383cb3f9a6f08adc0/clickhouse_rows.go#L38-L69)) which is a stream. So the original span was measuring a call to clickhouse with the query, but not actually waiting for all the results to load.

This PR waits until after the loop in `rows.Next()` is complete before finishing the span. 

N/A

### DIFF
--- a/backend/clickhouse/logs.go
+++ b/backend/clickhouse/logs.go
@@ -110,9 +110,8 @@ func (client *Client) ReadLogs(ctx context.Context, projectID int, params modelI
 
 	rows, err := client.conn.Query(ctx, sql, args...)
 
-	span.Finish(tracer.WithError(err))
-
 	if err != nil {
+		span.Finish(tracer.WithError(err))
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


backend/clickhouse/logs.go
```diff
@@ -148,6 +147,7 @@ func (client *Client) ReadLogs(ctx context.Context, projectID int, params modelI
 	}
 	rows.Close()
 
+	span.Finish(tracer.WithError(rows.Err()))
```
Vadman97, line 150: should we finish after the `getLogsConnection` call? you can change this to
```suggestion
	defer span.Finish(tracer.WithError(rows.Err()))
```

Vadman97 (approved):

PR #4625, opened 2023-03-16
do less time conversions at query time

et:
## Summary

While doing some testing of individual production queries, I noticed that running the time conversions using the built-in clickhouse functions can be slower than just comparing against the raw timestamp. 

So basically instead of this:

```sql
    WHERE 
      ProjectId = 3060 
      AND toUInt64(
        toDateTime(Timestamp)
      ) <= 1678970298 
```

we should do this:

```sql
    WHERE 
      ProjectId = 3060 
      AND Timestamp <= '2023-03-16 01:08:40.186000000' 
```

You can see this in this [analysis](https://gist.github.com/et/c048f850ca1632d84e309472604ad0b1).

Two things to look at:
* without clickhouse time functions, we actually using `toUnixTimestamp(Timestamp)` in the primary key
* without clickhouse time functions, the number of granules (and parts which I still don't get) is less which implies that the primary key filtered out more. The number of granules only goes down from 66 -> 64 so I don't think this would be hugely significant perf boost but it should help a little bit.

* Existing tests should pass.
* Verified through the app that the ordering of logs is in the correct order.

N/A

### DIFF
--- a/backend/clickhouse/logs.go
+++ b/backend/clickhouse/logs.go
@@ -345,8 +345,8 @@ func (client *Client) LogsKeyValues(ctx context.Context, projectID int, keyName 
 			Limit(KeyValuesLimit)
 	}
 
-	sb.Where(sb.LessEqualThan("toUInt64(toDateTime(Timestamp))", uint64(endDate.Unix()))).
-		Where(sb.GreaterEqualThan("toUInt64(toDateTime(Timestamp))", uint64(startDate.Unix())))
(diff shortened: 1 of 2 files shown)


Vadman97 (approved):
nice!

et:
This PR actually breaks some tests that I believe proves this isn't a safe change. Opting for a different approach in #4652 that I think is where we'll get more of our speed improvements.

PR #4628, opened 2023-03-16
Track setup events for sessions, errors, and logs

ccschmitz:
## Summary

* Creates a new `setup_events` table for storing data about
* Starts populating the table with records when we process `MarkBackendSetup` events

The UI using this will come in another PR.

Local click test.

Will monitor to ensure events are created as expected.


backend/model/model.go
```diff
@@ -335,6 +337,23 @@ type Project struct {
...
+	// Generic is temporary and can be removed once all messages are processed.
+	MarkBackendSetupTypeGeneric MarkBackendSetupType = "generic"
+	MarkBackendSetupTypeSession MarkBackendSetupType = "session"
+	MarkBackendSetupTypeError   MarkBackendSetupType = "error"
+	MarkBackendSetupTypeLogs    MarkBackendSetupType = "logs"
+)
+
+type SetupEvent struct {
+	ID        int                  `gorm:"primary_key;type:serial" json:"id" deep:"-"`
+	CreatedAt time.Time            `json:"created_at" deep:"-"`
```
ccschmitz, line 352: I didn't use `Model` because we don't need `updated_at` or `deleted_at`.

Vadman97 (approved):
nice! just a comment about inserting the `SetupEvent` rows given he unique constraint

backend/otel/otel.go
```diff
@@ -428,6 +431,7 @@ func (o *Handler) submitProjectLogs(ctx context.Context, projectLogs map[string]
 				Type: kafkaqueue.MarkBackendSetup,
 				MarkBackendSetup: &kafkaqueue.MarkBackendSetupArgs{
 					ProjectID: projectIDInt,
+					Type:      model2.MarkBackendSetupTypeLogs,
```
Vadman97, line 434: @ccschmitz we should only record setup events here if this is not a frontend log. the `SubmitFrontendConsoleMessage` function adds a source attribute to differentiate it

Screenshot 2023-03-20 at 11 18 47 AM

ccschmitz: Perfect - thanks, @Vadman97!

Vadman97 (approved):

PR #4646, opened 2023-03-20
Make debugging work out of the box for docker setup

et:
## Summary

This PR starts the Go debugger ([delve](https://github.com/go-delve/delve)) when running with Docker in development mode. My reasoning is that we should be providing the best "works out of the box" experience for new developers at Highlight.

Additionally, I've started a developer onboarding page ([view rendered](https://github.com/highlight/highlight/blob/634447c50c2b5e112f9dd1fda15d2f8c74cc7667/DEVELOPER_ONBOARDING.md)) that can serve as a place to migrate our internal documentation to a public home. At some point, I think we should replace the [existing docs page](https://www.highlight.io/docs/getting-started/self-host/dev-deployment-guide) with this. 

* Confirmed that backend debugging works in the Docker development mode

N/A

### DIFF
--- a/docker/.env
+++ b/docker/.env
@@ -1,10 +1,6 @@
 # docker compose config
 COMPOSE_PATH_SEPARATOR=:
 COMPOSE_PROJECT_NAME=highlight
-# use if you want to run the backend with remote gdb debugging
-# COMPOSE_FILE=compose.yml:compose.dev.yml:compose.debug.yml
-# use if you want the src to be mounted in with hot reloading for backend + frontend
-# COMPOSE_FILE=compose.yml:compose.dev.yml
 COMPOSE_FILE=compose.yml
 # environment variables
 CLICKHOUSE_ADDRESS=clickhouse:9000


backend/.air.toml
```diff
@@ -1,3 +1,6 @@
 [build]
+cmd = "go build -gcflags='all=-N -l' -o ./tmp/main ."
+bin = "tmp/main"
+full_bin = "dlv exec --accept-multiclient --log --headless --continue --listen :2345 --api-version 2 ./tmp/main"
```
et, line 4: I got this setup from this [blog post](https://dev.to/bruc3mackenzi3/debugging-go-inside-docker-using-vscode-4f67). A couple things to mention:

* We install delve already in our backend Dockerfile ([source](https://github.com/highlight/highlight/blob/02c12243018162d6012853a21d14e5764d620c48/docker/backend.Dockerfile#L3))
* Port 2345 is what we run the debugger on ([source](https://github.com/highlight/highlight/blob/02c12243018162d6012853a21d14e5764d620c48/docker/compose.dev.yml#L15)
* VSCode is configured for remote debugging ([source](https://github.com/highlight/highlight/blob/02c12243018162d6012853a21d14e5764d620c48/.vscode/launch.json#L6-L15))

The one thing new is the `-gcflags='all=-N -l'` which per the blog post:

> We've added the flag -gcflags "all=-N -l" to the build command. This is to disable inlining and optimizations that can interfere with debugging.

I distinctly remember having a problem with our old setup whereby some variables would say something like "code is unreachable". I don't remember the exact message but I remember it happening when you were deep in a function. The only way around it was to manually print the variable name (super annoying).

From my brief testing, all variables are defined in the debugger regardless of where I set the breakpoint so this looks like a huge developer productivity boost:

Screenshot 2023-03-20 at 9 56 06 AM

PR #4653, opened 2023-03-21
add related logs to session devtools

et:
## Summary

[Figma](https://www.figma.com/file/rdBPNzn7Klk6RG1LHUCE5B/Screen-Designs?node-id=40-3089&t=lrQbbI0AVltkH5CZ-0)

This adds a Related logs button to the session's devtools when the Console tab is selected.

Screenshot 2023-03-20 at 9 47 24 PM

Clicking on it loads the logs page when the secure session id filled in with a time range of (session.created_at + 1 day) to capture all the logs while also optimizing the clickhouse query. Setting a range is required so we don't use the default state of "Last 15 minutes". 

Verified that clicking Related logs loads the logs page correctly with the secure session id in the filter box.

Kapture 2023-03-20 at 21 51 26

N/A

### DIFF
--- a/frontend/src/pages/ErrorsV2/ErrorInstance/ErrorInstance.tsx
+++ b/frontend/src/pages/ErrorsV2/ErrorInstance/ErrorInstance.tsx
@@ -12,10 +12,11 @@ import {
 	useGetErrorInstanceQuery,
 } from '@graph/hooks'
 import { GetErrorGroupQuery, GetErrorObjectQuery } from '@graph/operations'
-import type {
+import {
 	ErrorInstance as ErrorInstanceType,
 	ErrorObject,
 	Maybe,
+	ReservedLogKey,
 } from '@graph/schemas'
 import {
 	Box,
@@ -321,7 +322,7 @@ const ErrorInstance: React.FC<Props> = ({ errorGroup }) => {
 								let offsetStart = 1
 								if (errorObject.session?.secure_id) {
 									queryParams.push({
-										key: 'secure_session_id',
+										key: ReservedLogKey.SecureSessionId,
 										operator: DEFAULT_LOGS_OPERATOR,
 										value: errorObject.session?.secure_id,
 										offsetStart: offsetStart++,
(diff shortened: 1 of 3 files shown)


frontend/src/pages/ErrorsV2/ErrorInstance/ErrorInstance.tsx
```diff
@@ -329,15 +331,15 @@ const ErrorInstance: React.FC<Props> = ({ errorGroup }) => {
 								}
 								if (errorObject.session?.secure_id) {
 									queryParams.push({
-										key: 'secure_session_id',
+										key: ReservedLogKey.SecureSessionId,
```
et, line 334: This is referencing these enums: 
https://github.com/highlight/highlight/blob/756e9ad1ee1446bd04437ba7eba18fe9a1af165f/backend/private-graph/graph/schema.graphqls#L538-L546

frontend/src/pages/LogsPage/SearchForm/utils.ts
```diff
@@ -84,3 +89,35 @@ export const stringifyLogsQuery = (params: LogsSearchParam[]) => {
...
+		offsetStart: 0,
+	})
+
+	const sessionStartDate = new Date(session.created_at)
+
+	// Instead of doing `new Date()` (equivalent to the time now), we set an arbitrary end date of one day after the session was created.
+	// We have found that Clickhouse doesn't perform well with long time ranges (see https://github.com/highlight/highlight/pull/4652)
+	// Rarely will a session last longer than a day. If users need to handle this case, we should figure out a way for the backend to optimize
+	// long time ranges.
+	const sessionEndDate = moment(sessionStartDate).add(1, 'day').toDate()
```
Vadman97, line 108: could even set this to 4 hours which is the limit to session length upon processing and in the browser client

Vadman97 (approved):

frontend/src/pages/LogsPage/SearchForm/utils.ts
et: Nice tip. Fixed in 981b8ae728e7e11879cd86b7db79d34fcec384da

PR #4661, opened 2023-03-21
attach a time frame when fetching log keys

et:
## Summary

Fetching log key queries is slow because we don't specify a time range (see #4652 and #4507). 

* Added unit test
* Verified log keys still loads via the frontend

A more robust solution would be #4662 but we'll look into that after our launch.


Vadman97 (approved):
nice!

PR #4672, opened 2023-03-23
report frontend log source

Vadman97:
## Summary

Differentiate frontend logs by setting a `highlight.source` attribute value.

searching for frontend logs
image

frontend

backend (source set explicitely on a span)
Screenshot 2023-03-22 at 6 14 23 PM

No


backend/clickhouse/log_row.go
```diff
@@ -86,6 +86,15 @@ func WithSeverityText(severityText string) LogRowOption {
 	}
 }
 
+func WithSource(source string) LogRowOption {
+	return func(h *LogRow) {
+		if source == highlight.SourceAttributeFrontend {
+			source = "frontend"
+		}
+		h.LogAttributes[highlight.SourceAttribute] = source
```
et, line 94: What do you think about writing this to `ServiceName` instead? We'd have to do some additional wiring up (which I'm happy to handle) but I think it would be a lot more optimized to use a top level column over adding more things to the log attributes map.

Vadman97: makes sense, especially when the otel semconv `ServiceName` attribute is not set (i'll also check that when it is, we pipe it thru to this top-level column). what do I need to change on the frontend to wire this?

et: I think adding a new column makes a lot of sense! Let's do that.

`ServiceName` - `highlight-js-sdk`
`Source` - `frontend`

But long term, we could have `Source` be something the user could actually customize. Would be really helpful if they had multiple frontend apps. 

cc @ccschmitz - per our convo last week about having a way to name sources in the setup page.

Vadman97: updated!

frontend/src/pages/LogsPage/LogsTable/LogDetails.tsx
```diff
@@ -133,6 +141,24 @@ export const LogDetails = ({ row, queryTerms }: Props) => {
 					/>
 				</Box>
 			)}
+			{source && (
```
et, line 144: Heads up @ccschmitz - this will conflict with #4711. I'll leave a suggestion on #4711 to how to dedupe some logic.

et (approved):
Awesome work! This is really going to help with searching.

PR #4678, opened 2023-03-23
add backend webhook sending for alerts payloads

Vadman97:
## Summary

Adds backend logic to store webhook destinations for alerts and send alerts payloads to set destinations as json via a retryable http client.

Payload currently sends the full body of the `integrations.` payload objects, though as a v2 we could support customizing the formats of the payload body.

Frontend to update webhook URLs coming in a separate PR. Implements backend for #4550

Local deploy with manually configured alert webhooks sending to webhook.site

Screenshot 2023-03-23 at 1 03 17 AM

Will monitor alerts db migration.


backend/alerts/integrations/webhook/webhook.go
```diff
@@ -0,0 +1,141 @@
...
+	"net/http"
+)
+
+type ErrorAlertWebhook struct {
+	Event string
+	*integrations.ErrorAlertPayload
+}
+
+func sendWebhookData(destination string, body []byte) error {
+	resp, err := retryablehttp.Post(destination, "application/json", body)
```
et, line 19: I'm trying to find it in their [docs](https://github.com/hashicorp/go-retryablehttp) but it's not really standing out -- what is the retry logic this uses? i.e. how many times does it retry and what is the duration between each retries?

backend/alerts/alerts.go
```diff
@@ -39,25 +41,38 @@ func SendErrorAlert(event SendErrorAlertEvent) error {
 		VisitedURL:      event.VisitedURL,
 	}
 
-	if !isWorkspaceIntegratedWithDiscord(*event.Workspace) {
+	var g errgroup.Group
+	g.Go(func() error {
+		for _, wh := range event.ErrorAlert.WebhookDestinations {
```
et, line 46: Looks like you have to do the same thing as I did with discord and have to dupe all this looping. I thought about having some type of entry point like

```go
func SendAlert(webhookDestination event.ErrorAlert.WebhookDestinations, payload interface{}, alertType AlertType)
```

and then ideally, it would delegate to the correct function given the `alertType`. But I couldn't figure out a way to dynamically call a function in Go. 

Not sure if you know how to pull that off or a simpler way but it would be nice to tackle this as we continue to add more integrations.

et (approved):

backend/alerts/integrations/webhook/webhook.go
Vadman97: here's the spec for their default client - thought it was pretty reasonable.

Screenshot 2023-03-24 at 12 38 43 PM

et: One more thing --- have we considered any authentication?

backend/alerts/alerts.go
Vadman97: good point eric, on the webhook package side, it's just a matter of abstracting the json marshalling out to dedup all the `Send...` functions. let me try that

backend/alerts/integrations/webhook/webhook.go
Vadman97: i feel like that would be more of a V2 for #4550 - allowing customizing the webhook payload and sending a request with authentication

Vadman97: as discussed offline, i'll make the model for webhooks an object so that we can store an auth token and other payload configuration in the future.

backend/alerts/alerts.go
Vadman97: looked into this a bit - we could dynamically call functions by having them be in a `map[string]func()` but since we have different payload types here that becomes more complex with getting the types to match up. going to leave this for now as it would require a lot of refactoring here to the payloads (rewriting the payloads to be an interface implementing a `Send()` method that could be called, for instance)

PR #4682, opened 2023-03-23
teardown old logs table

et:
## Summary

This PR tears down the old logs table now that our data has been fully migrated to the new schema (see #4667) and renames the `logs_new` table to `logs`

* Validated `show tables;` only shows a `logs` table
* Validated the data of `logs_new` still exists, just in a table called `logs` now.

Per the query in #4667, this will free up 125 GB of data

### DIFF
--- a/backend/clickhouse/logs.go
+++ b/backend/clickhouse/logs.go
@@ -16,7 +16,7 @@ import (
 	e "github.com/pkg/errors"
 )
 
-const LogsTable = "logs_new"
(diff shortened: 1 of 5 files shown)


et:
> this migration might cause a brief logs query outage since older containers will still be pointing to logs_new while this container is rolling out and renaming the table. don't think that's a concern but just wanted to point that out.

Good point! I'll think of a zero downtime way for post launch in case we have to do something like this again.

PR #4684, opened 2023-03-23
ensure errors created from a log pass structured metadata

Vadman97:
## Summary

Error created from error logs should have the logattributes as the payload. 
For example, the following [error](https://app.highlight.io/1/errors/pkQAv9oKfgBVYBO99aZ0ocJ3qk1w/instances/162208765?page=1) has an associated [log](https://app.highlight.io/logs/MjAyMy0wMy0yM1QxNToxOTowMVosMjc5OTYzYzgtOGMzNS00NjBjLWJhYTctYjhhMmI1NmQ3MDlm) that has structured attributes that are not passed to the created error.

The PR also captures span attributes that were not being processed on logs and errors (for example, the go sdk can report attributes on a span with the `span.SetAttribute` method exposed via the opentelemetry span interface.

image

Screenshot 2023-03-23 at 10 17 00 PM

After the change, errors show useful tags that can help with debugging an error. In the above example, the `path` attribute is a custom tag on the log line that shows the graphql operation field.

Screenshot 2023-03-23 at 10 22 33 PM

No

### DIFF
--- a/backend/clickhouse/log_row.go
+++ b/backend/clickhouse/log_row.go
@@ -72,9 +72,9 @@ func (l *LogRow) Cursor() string {
 
 type LogRowOption func(*LogRow)
 
-func WithLogAttributes(resourceAttributes, eventAttributes map[string]any, isFrontendLog bool) LogRowOption {
+func WithLogAttributes(resourceAttributes, spanAttributes, eventAttributes map[string]any, isFrontendLog bool) LogRowOption {
 	return func(h *LogRow) {
-		h.LogAttributes = getAttributesMap(resourceAttributes, eventAttributes, isFrontendLog)
+		h.LogAttributes = GetAttributesMap(resourceAttributes, spanAttributes, eventAttributes, isFrontendLog)
 	}
 }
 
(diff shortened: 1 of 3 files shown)


backend/clickhouse/log_row.go
```diff
@@ -72,9 +72,9 @@ func (l *LogRow) Cursor() string {
 
 type LogRowOption func(*LogRow)
 
-func WithLogAttributes(resourceAttributes, eventAttributes map[string]any, isFrontendLog bool) LogRowOption {
+func WithLogAttributes(resourceAttributes, spanAttributes, eventAttributes map[string]any, isFrontendLog bool) LogRowOption {
```
et, line 75: As a thought, since we really don't make a distinction between the attributes, we might just want to make a single param `attributes` and force the callers to pass one map in.

et (approved):
Very nice! This is super helpful.

backend/clickhouse/log_row.go
Vadman97: i guess there could be another method that brings the maps together before they are passed here, but i think i'd rather keep them as separate arguments in case we want to nest them under different keys in the future, or do different processing depending on the attribute source.

PR #4698, opened 2023-03-26
cleanup error loading

Vadman97:
## Summary

Tweaks local error group loading per some feedback.
* errors feed refresh icon does not refresh instances (only refreshes error group)
* error instance / user count not working correctly for new errors
* clicking show session from error opens session in weird state (does not move player to right time
* updates `channels to notify` -> `slack channels to notify` to be more specific

Local deploy.

Correct instances for new error.

Screenshot 2023-03-27 at 12 45 13 AM

No.

### DIFF
--- a/backend/private-graph/graph/resolver.go
+++ b/backend/private-graph/graph/resolver.go
@@ -645,6 +645,17 @@ func (r *Resolver) SetErrorFrequenciesInflux(ctx context.Context, projectID int,
 			if err != nil {
 				log.WithContext(ctx).Error(err)
 			}
+			endDate := time.Now().UTC().AddDate(0, 0, -1)
+			startDate := endDate.AddDate(0, 0, -1*(lookbackPeriod-1))
+			for curDate := startDate; !curDate.After(endDate); curDate = curDate.AddDate(0, 0, 1) {
+				eg.ErrorFrequency = append(eg.ErrorFrequency, 0)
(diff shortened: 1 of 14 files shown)


backend/private-graph/graph/resolver.go
```diff
@@ -636,9 +635,44 @@ func (r *Resolver) SetErrorFrequenciesInflux(ctx context.Context, projectID int,
...
+				if r.Name == "count" {
+					eg.ErrorFrequency = append(eg.ErrorFrequency, r.Value)
+				}
+				eg.ErrorMetrics = append(eg.ErrorMetrics, &struct {
+					ErrorGroupID int
+					Date         time.Time
+					Name         string
+					Value        int64
+				}{ErrorGroupID: eg.ID, Date: r.Date, Name: r.Name, Value: r.Value})
+			}
```
et, line 669: It looks like the above block of code is setting unsampled data if the error group was created in the last hour.
Is there a reason why we let the code continue from here? Why not return early? 
I _think_ it might not matter because `oldErrorGroupIDs` will be empty but wouldn't we be able to scope that variable better if we returned early?

Vadman97: since `errorGroups` may consist of more than one error group of varying age, we still want to build up the `oldErrorGroupIDs` for ones that are older than an hour that can leverage the other aggregation method. the final result could be a set of frequencies for both old and new error groups.

PR #4700, opened 2023-03-26
make python sdk flush more frequently

Vadman97:
## Summary

Adjusts python SDK flushing settings to ensure SDK does not increase azure function memory usage significantly.

`poetry run build; poetry run flask run`

New python SDK version bump.

### DIFF
--- a/sdk/highlight-py/highlight_io/sdk.py
+++ b/sdk/highlight-py/highlight_io/sdk.py
@@ -4,6 +4,7 @@ import typing
 
 from opentelemetry import trace, _logs
 from opentelemetry._logs.severity import std_to_otel
+from opentelemetry.exporter.otlp.proto.http import Compression
 from opentelemetry.exporter.otlp.proto.http._log_exporter import OTLPLogExporter
 from opentelemetry.exporter.otlp.proto.http.trace_exporter import OTLPSpanExporter
 from opentelemetry.instrumentation.logging import LoggingInstrumentor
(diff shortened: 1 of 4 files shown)


sdk/highlight-py/highlight_io/sdk.py
```diff
@@ -59,14 +60,28 @@ def __init__(
 
         self._trace_provider = TracerProvider()
         self._trace_provider.add_span_processor(
-            BatchSpanProcessor(OTLPSpanExporter(f"{self._otlp_endpoint}/v1/traces"))
+            BatchSpanProcessor(
+                OTLPSpanExporter(
+                    f"{self._otlp_endpoint}/v1/traces", compression=Compression.Gzip
+                ),
+                schedule_delay_millis=1000,
+                max_export_batch_size=64,
```
ccschmitz, line 68: Just curious, how died you land on `64` here? It [looks like the default is `512`](https://github.com/open-telemetry/opentelemetry-python/blob/2e0d8eece473c0b191cdbfb856edc8ce3da0a028/opentelemetry-sdk/src/opentelemetry/sdk/trace/export/__init__.py#L166 - just curious why we decided to decrease this.

ccschmitz (approved):

sdk/highlight-py/highlight_io/sdk.py
Vadman97: `64` is arbitrary, but just a decrease in all the values to ensure flushed payloads are smaller. this pr came from a customer POC where their deployment started to hit some potential memory limits

PR #4701, opened 2023-03-26
exclude exception tags and limit log attributes

Vadman97:
## Summary

Removes `exception.` attributes from logs since they are bloating `LogAttributes` and are not useful. 
For error stacktraces, we parse and structure them anyway.
Adds a limit to tag values at 64K characters. 
This is a pretty high limit but prevents arbitrarily-large values from being sent.

Screenshot 2023-03-26 at 10 10 33 PM

No

### DIFF
--- a/backend/clickhouse/log_row.go
+++ b/backend/clickhouse/log_row.go
@@ -15,7 +15,7 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-const LogAttributeValueLengthLimit = 2 << 15
+const LogAttributeValueLengthLimit = 2 << 10
 
 func ProjectToInt(projectID string) (int, error) {
 	i, err := strconv.ParseInt(projectID, 10, 32)
@@ -122,7 +122,7 @@ func WithProjectIDString(projectID string) LogRowOption {
 
 func GetAttributesMap(ctx context.Context, resourceAttributes, spanAttributes, eventAttributes map[string]any, isFrontendLog bool) map[string]string {
 	attributesMap := make(map[string]string)
-	for _, m := range []map[string]any{resourceAttributes, spanAttributes, eventAttributes} {
+	for mIdx, m := range []map[string]any{resourceAttributes, spanAttributes, eventAttributes} {
 		for k, v := range m {
 			prefixes := highlight.InternalAttributePrefixes
 			if isFrontendLog {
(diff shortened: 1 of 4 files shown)


backend/clickhouse/log_row.go
```diff
@@ -107,50 +110,37 @@ func WithProjectIDString(projectID string) LogRowOption {
...
-
 			if shouldSkip {
 				continue
 			}
 
-			vStr := cast(v, "")
-			if vStr != "" {
+			if vStr, ok := v.(string); ok {
+				if len(vStr) > LogAttributeValueLengthLimit {
+					log.WithContext(ctx).Warnf("attribute value for %s is too long %d", k, len(vStr))
```
et, line 135: It might be helpful to get an idea of which of `resourceAttributes, spanAttributes, eventAttributes` actually passed in a large value. As it is now, it would be really challenging to reverse engineer the offending source.

backend/clickhouse/log_row.go
```diff
@@ -14,6 +15,8 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+const LogAttributeValueLengthLimit = 2 << 15
```
et, line 18: Can we set to a lower limit now? I pulled a random stacktrace and it was 3455 characters ([source](https://go.dev/play/p/4FNtQ-cd0Sh)) so 64K characters is far too lenient.

I don't see a reason why anyone needs more than 500 characters. And we can choose to simply trim (instead of drop) those attributes if someone complains. 

Particularly, I think we should put a more aggressive guard in now given that our pricing model is based on number of rows and not by storage.

sdk/highlight-go/otel.go
```diff
@@ -46,6 +46,8 @@ var InternalAttributes = []string{
 	SourceAttribute,
 	LogMessageAttribute,
 	LogSeverityAttribute,
+	// exception should be parsed as structured and not included as part of log attributes
+	"exception.",
```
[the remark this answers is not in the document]
Vadman97: good idea!

backend/clickhouse/log_row.go
Vadman97: improved the log here

backend/clickhouse/log_row.go
Vadman97: added a limit on attribute values and the body as per https://highlightcorp.slack.com/archives/C04NUD29BT3/p1679944459103359

et (approved):
🥳 

Thanks for doing this!

PR #4711, opened 2023-03-27
Clean up JSON written to clipboard for log

ccschmitz:
## Summary

Copying to clipboard includes some unnecessary details of a log. Also, everything under `logAttributes` should be included as top-level keys, similar to what the user sees in the UI.

Local click test.

N/A


frontend/src/pages/LogsPage/LogsTable/LogDetails.tsx
```diff
@@ -57,6 +58,11 @@ export const LogDetails = ({ row, queryTerms }: Props) => {
 	const expandable = Object.values(logAttributes).some(
 		(v) => typeof v === 'object',
 	)
+	const attributesShownWhenPresent = {
```
et, line 61: @ccschmitz - per #4672, can adjust this to `ReservedLogKey`: https://github.com/highlight/highlight/blob/main/backend/private-graph/graph/schema.graphqls#L533-L541

et: We might need to add `message` to that enum.

Vadman97: with #4672, we'll want `Source` and `ServiceName` as part of this as well.

Vadman97 (approved):

ccschmitz:
@Vadman97 @et thanks for the heads up on that. I revised this logic in https://github.com/highlight/highlight/pull/4711/commits/437502e990936f6ea806f7da23a26557708b096b and used the `ReservedLogKey` type to enforce the object I created and iterated over stays in sync with what's on the server.

frontend/src/pages/LogsPage/LogsTable/LogDetails.tsx
```diff
@@ -57,6 +57,15 @@ export const LogDetails = ({ row, queryTerms }: Props) => {
 	const expandable = Object.values(logAttributes).some(
 		(v) => typeof v === 'object',
 	)
+	const reservedLogAttributes: {
+		[key in ReservedLogKey]: Maybe<string> | undefined
```
et, line 61: Awesome 👏

et (approved):

PR #4715, opened 2023-03-27
Fix handling of quotes in logs queries

ccschmitz:
## Summary

Updates the behavior for handling strings in logs queries.

* Does not strip/transform quotes on user input anymore. Instead, we perform a transformation before sending the query string to the server to convert everything to double quotes.
* Adds quotes around values that need it when being applied outside the search input (e.g. "Apply as filter" button or selecting an item in the dropdown list).

Local and PR preview click testing.

N/A


et (approved):
Did a couple tests on the preview -- works great! Thanks!

PR #4674, opened 2023-03-23
add related sessions button

Vadman97:
## Summary

Adds a `Related Session` button for logs that have a session id.
Updates the `Related Error` button design per figma.

## How did you test this change?

Screenshot 2023-03-22 at 6 38 32 PM

https://www.loom.com/share/51ef1bcf3b8f412ea3a90fc09f138880

## Are there any deployment considerations?

No

PR #4719, opened 2023-03-27
ensure logs query string is used when provided

Vadman97:
## Summary

Clicking `Show Logs` from errors would populate the query string that would be replaced with the default value of the form.
Ensure that the search form only updates the query when there is an actual input.

https://www.loom.com/share/12625c23205348369aa2d745bf2a14e7

No

### DIFF
--- a/frontend/src/pages/ErrorsV2/ErrorInstance/ErrorInstance.tsx
+++ b/frontend/src/pages/ErrorsV2/ErrorInstance/ErrorInstance.tsx
@@ -350,7 +350,6 @@ const ErrorInstance: React.FC<Props> = ({ errorGroup }) => {
 								if (logCursor) {
 									navigate(
 										`/${projectId}/logs/${logCursor}?${params}`,
-										{},
 									)
 								} else {
 									navigate(`/${projectId}/logs?${params}`)
--- a/frontend/src/pages/ErrorsV2/ErrorInstance/ErrorInstance.tsx
+++ b/frontend/src/pages/ErrorsV2/ErrorInstance/ErrorInstance.tsx
@@ -344,18 +344,23 @@ const ErrorInstance: React.FC<Props> = ({ errorGroup }) => {
 						>
 							<KeyboardShortcut label="Next" shortcut="]" />
 						</Tooltip>
-						<Button
+						<LinkButton
 							kind="primary"
 							emphasis="high"
+							to={getLogsLink(errorObject)}
 							disabled={!isLoggedIn || !errorObject.session}
-							onClick={() => {
-								navigate(getLogsLink(errorObject))
-							}}
-							iconLeft={<IconSolidViewList />}
-							trackingId="errorInstanceShowLogs"
+							trackingId="logs-related_session_link"
 						>
-							Show logs
-						</Button>
+							<Box
+								display="flex"
+								alignItems="center"
+								flexDirection="row"
+								gap="4"
+							>
+								<IconSolidViewList />
+								Show logs
+							</Box>
+						</LinkButton>
 						<Button
 							kind="primary"
 							emphasis="high"
--- a/frontend/src/pages/ErrorsV2/ErrorInstance/ErrorInstance.tsx
+++ b/frontend/src/pages/ErrorsV2/ErrorInstance/ErrorInstance.tsx
@@ -67,6 +67,41 @@ const METADATA_LABELS: { [key: string]: string } = {
 	id: 'ID',
 } as const
 
+const getLogsLink = (errorObject: ErrorObject): string => {
+	const queryParams: LogsSearchParam[] = []
+	let offsetStart = 1
+	if (errorObject.session?.secure_id) {
+		queryParams.push({
+			key: ReservedLogKey.SecureSessionId,
+			operator: DEFAULT_LOGS_OPERATOR,
+			value: errorObject.session?.secure_id,
+			offsetStart: offsetStart++,
+		})
+	}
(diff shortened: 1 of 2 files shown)


frontend/src/pages/ErrorsV2/ErrorInstance/ErrorInstance.tsx
```diff
@@ -350,7 +350,6 @@ const ErrorInstance: React.FC<Props> = ({ errorGroup }) => {
 								if (logCursor) {
 									navigate(
```
et, line 351: @Vadman97 - what happens if make this a proper link instead (see [comment](https://github.com/highlight/highlight/pull/4674#discussion_r1145623572))?

et: I don't understand the issue then. Is the issue a problem with linking from the errors page or something specific about the logs page?

If it's the former, shouldn't we update the `navigate` statement right below this one as well?
If it's the latter, can you show a loom of what you mean about the `?query=` being the issue?

Vadman97: here's a simple example of the problem:
* visit https://app.highlight.io/1/logs?query=foo
* the query `foo` is not in the search
hence, the problem doesn't have to do with this navigation itself. the issue is how we update the state from the search form with the query string, resetting the initial value even if one is set.
what's happening with the PR
1. the `query` value is parsed from the URL by `LogsPage`
2. the `query` value is passed to `SearchForm` as the`initialQuery`
3. on first render, `SearchForm` calls `setQuery('')` resetting the query and the query string to `query=''`

Vadman97: switched this to use a `LinkButton` - shout out to @ccschmitz for making this so easy

et: > on first render, SearchForm calls setQuery('') resetting the query and the query string to query=''

Sorry, where is this exactly?
I'm seeing this but I'm not sure if that's what you're talking about: https://github.com/highlight/highlight/blob/014061ae28ef8cda077e7771f2ad925f6c02219e/frontend/src/pages/LogsPage/SearchForm/SearchForm.tsx#L195-L201

Sorry if I'm digging in too much. Really want us to avoid the complexity difficulties we have with the session/error page because we have too much state and useEffects updating things.

Vadman97: no worries! the only change is https://github.com/highlight/highlight/blob/796462c79015b95be4310b18e019dafcf1858ce4/frontend/src/pages/LogsPage/SearchForm/SearchForm.tsx#L71-L75

as i found with some debugging, `formState.useSubmit` triggers on initial render of the ariakit form.

before the change
```
	formState.useSubmit(() => {
		onFormSubmit(formState.values.query)
	})
```
would call `onFormSubmit('')` since the search form is not filled out, even when the `initialQuery` was provided

after the change
```
	formState.useSubmit(() => {
		if (formState.submitting) {
			onFormSubmit(formState.values.query)
		}
	})
```
checking for `formState.submitting` only calls the `onFormSubmit` when the ariakit form detects a 'submission' - aka a real input followed by `Enter` key or the input losing focus.

et: Thanks for sharing the write up @Vadman97! Appreciate the writeup.

> as i found with some debugging, formState.useSubmit triggers on initial render of the ariakit form.

If I understand this correctly, we are saying `useSubmit` is being triggered when the component mounts? And just to be clear, it's that we are manually calling this -- I'm fairly certain ariakit's `useSubmit` doesn't do this by default (here's a [codesandbox](https://codesandbox.io/s/8sw6dv?module=src/form.tsx))

With that being said, I think there's something else strange going on here with these two `useEffect`s:
Here's a loom to describe what I mean:

https://www.loom.com/share/94fd8b66468f487ca8e6b7195a090c7e

et: Actually...removing this `useEffect` seems to fix the issue:

https://github.com/highlight/highlight/blob/dae0318e3b4ea23bd06244364ca26324493798e2/frontend/src/pages/LogsPage/SearchForm/SearchForm.tsx#L195-L201

I'm not sure why we have this exactly. Sasha added it here but there's not a whole lot of context as to why:

https://github.com/highlight/highlight/pull/4524/files#diff-f847a8411b2f43cb4537fcb5260ba0125fbd6de14f055f91b3779703844825a5R187-R193

et: @Vadman97 @ccschmitz - what do you think about removing that `useEffect` I listed above? I'm happy to do some rigorous click testing on this.

Vadman97: oh i see! nice find eric - removing that `useEffect` definitely makes more sense. with that `useEffect`, `submitQuery('')` causes a state update, which triggers the `formState.useSubmit`. my method of checking `formState.submitting` determines whether this `submitQuery` comes from this manual call vs the form being updated by an input, but that additional indirection is certainly a worse approach. updating this

et (approved):
Click tested on the preview and things are looking good!
🚢

PR #4738, opened 2023-03-28
add error state for logs

et:


frontend/src/components/Link/styles.css.ts
```diff
@@ -0,0 +1,13 @@
+import { vars } from '@highlight-run/ui/src/css/vars'
+import { style } from '@vanilla-extract/css'
+
+
+export const link = style({
```
et, line 5: I pulled these styles from the Header component
https://github.com/highlight/highlight/blob/4720-handle-backend-errors/frontend/src/components/Header/styles.css.ts#L4-L13

Per the above file, I think we should make a generic component for handling links similar to `<LinkButton>` but for regular links.

Vadman97 (commented):
🚢

Vadman97 (approved):

PR #4746, opened 2023-03-29
Use github actions to run a vercel deploy

### LINKED ISSUE #4733: Only run the vercel build when `docs-content/` or `highlight.io/` is changed.
Probably want to use this: https://vercel.com/guides/how-can-i-use-github-actions-with-vercel

jay-khatri:
## Summary

Resolves #4733 

This will only run the preview deployment or prod deployment if `docs-content/**` or `highlight.io/**` is changed. Allows us to have full control of whether a deploy happens on a given branch / file path. 

Note that we're using `vercel`, not `vercel build`, so all build work is done in Vercel's build suite for now. @Vadman97 , I figured this was a bit simpler, but long term, might be more economical to run the build in github actions.

After we confirm this works w/ a deploy to `main`, we can turn off the vercel integration and use this exclusively.

Run in CI.

### DIFF
--- a/.github/workflows/landing-preview.yml
+++ b/.github/workflows/landing-preview.yml
@@ -7,6 +7,7 @@ on:
         paths:
             - 'highlight.io/**'
             - 'docs-content/**'
+            - '.github/workflows/landing-preview.yml'
         branches-ignore:
             - main
 jobs:
--- a/.github/workflows/landing-preview.yml
+++ b/.github/workflows/landing-preview.yml
@@ -16,9 +16,7 @@ jobs:
             - uses: actions/checkout@v2
               with:
                   submodules: true
-            - name: Install Vercel CLI
-              run: npm install --global vercel@latest
             - name: Pull Vercel Environment Information
-              run: vercel pull --yes --environment=preview --token=${{ secrets.VERCEL_TOKEN }}
+              run: yarn dlx vercel pull --yes --environment=preview --token=${{ secrets.VERCEL_TOKEN }}
             - name: Deploy Project Artifacts to Vercel
-              run: vercel --token=${{ secrets.VERCEL_TOKEN }}
+              run: yarn dlx vercel --token=${{ secrets.VERCEL_TOKEN }}
--- a/.github/workflows/landing-preview.yml
+++ b/.github/workflows/landing-preview.yml
@@ -4,6 +4,9 @@ env:
     VERCEL_PROJECT_ID: ${{ secrets.VERCEL_PROJECT_ID }}
 on:
     push:
+        paths:
+            - 'highlight.io/**'
+            - 'docs-content/**'
         branches-ignore:
             - main
 jobs:
--- a/.github/workflows/landing-preview.yml
+++ b/.github/workflows/landing-preview.yml
@@ -1,0 +1,21 @@
+name: Vercel Preview Deployment
+env:
+    VERCEL_ORG_ID: ${{ secrets.VERCEL_ORG_ID }}
+    VERCEL_PROJECT_ID: ${{ secrets.VERCEL_PROJECT_ID }}
+on:
+    push:
+        branches-ignore:
+            - main
+jobs:
+    Deploy-Preview:
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v2
+              with:
+                  submodules: true
+            - name: Install Vercel CLI
+              run: npm install --global vercel@latest
+            - name: Pull Vercel Environment Information
+              run: vercel pull --yes --environment=preview --token=${{ secrets.VERCEL_TOKEN }}
+            - name: Deploy Project Artifacts to Vercel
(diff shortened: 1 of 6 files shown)


.github/workflows/landing-prod.yml
```diff
@@ -0,0 +1,24 @@
...
+            - main
+jobs:
+    Deploy-Preview:
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v2
+              with:
+                  submodules: true
+            - name: Install Vercel CLI
+              run: npm install --global vercel@latest
```
Vadman97, line 20: if you use yarn dlx vercel, no need for the install step here

.github/workflows/landing-prod.yml
```diff
@@ -0,0 +1,24 @@
+name: Vercel Production Deployment
+env:
+    VERCEL_ORG_ID: ${{ secrets.VERCEL_ORG_ID }}
+    VERCEL_PROJECT_ID: ${{ secrets.VERCEL_PROJECT_ID }}
+on:
+    push:
+        paths:
```
Vadman97, line 7: should run on top level yarn.lock changes

.github/workflows/landing-prod.yml
```diff
@@ -0,0 +1,24 @@
+name: Vercel Production Deployment
+env:
+    VERCEL_ORG_ID: ${{ secrets.VERCEL_ORG_ID }}
```
Vadman97, line 3: can we use turborepo remote caching like we do in turbo.yml

.github/workflows/landing-prod.yml
```diff
@@ -0,0 +1,24 @@
...
+    Deploy-Preview:
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v2
+              with:
+                  submodules: true
+            - name: Install Vercel CLI
+              run: npm install --global vercel@latest
+            - name: Pull Vercel Environment Information
+              run: vercel pull --yes --environment=preview --token=${{ secrets.VERCEL_TOKEN }}
```
Vadman97, line 22: should this be production?

.github/workflows/landing-preview.yml
```diff
@@ -0,0 +1,24 @@
+name: Vercel Preview Deployment
+env:
+    VERCEL_ORG_ID: ${{ secrets.VERCEL_ORG_ID }}
+    VERCEL_PROJECT_ID: ${{ secrets.VERCEL_PROJECT_ID }}
+on:
+    push:
```
Vadman97, line 6: same comments as other file

Vadman97 (changes requested):

.github/workflows/landing-prod.yml
jay-khatri: Are we looking to cache the `yarn install`? If so, that doesn't happen here; it happens in Vercel. I imagine Vercel already has knowledge of our turborepo settings?

jay-khatri: (note I'm doing `vercel`, not `vercel build`, so this is essentially just an API call to kick off the build)

jay-khatri: (i think)

.github/workflows/landing-prod.yml
jay-khatri: what does that do?

Vadman97: https://yarnpkg.com/cli/dlx/

to be consistent with our other usage of yarn, could simplify this so you don't need to install

.github/workflows/landing-prod.yml
Vadman97: this would cache the 'build' step so that it syncs with the vercel remote cache and speeds up subsequent builds in highlight.io package. this isn't a cache of the yarn install, that should be done by the yarn github action

.github/workflows/landing-prod.yml
jay-khatri: Doesn't it still have to install anyway? I can update for consistency.

I'll have to replace every instance of `vercel` with `yarn dlx vercel` though right?

Vadman97: yes that's correct.

Vadman97 (approved):

.github/workflows/landing-prod.yml
jay-khatri: Ok, but the build step is not happening on the github action runners. Will adding the turbo repo remote caching in this github action help w/ what you're talking about?

Vadman97: oh gotcha, i understand now. if this is a remote build then likely vercel already knows. just wanted to double check, we can follow up on this after this is merged.

jay-khatri: Yeah i think the `turbo.yml` is a flavor of this: https://turbo.build/repo/docs/ci/github-actions

But this situation is pretty different; basically just an API request to start the build.

PR #4748, opened 2023-03-29
remove alert setup modal

et:
## Summary

When setting up alerts for the first time on a new project, the user receives this prompt:

Screenshot 2023-03-29 at 9 35 03 AM

This is confusing to new users since it implies that Slack is a dependency to set up alerts - it's not, people can use discord or even just email.

This PR removes the modal.

Now when one visits the alerts page for the first time, they'll see this:

Screenshot 2023-03-29 at 9 37 58 AM

N/A


backend/private-graph/graph/schema.graphqls
```diff
@@ -1697,12 +1697,6 @@ type Mutation {
 		workspace_id: ID!
 	): Boolean!
 	syncSlackIntegration(project_id: ID!): SlackSyncResponse!
-	createDefaultAlerts(
```
et, line 1700: Removing this mutation since it's only used in the setup modal.

mayberryzane (approved):

PR #4758, opened 2023-03-29
add user setting for indexeddb cache

Vadman97:
## Summary

Allow disabling indexeddb caching in production. Needed for high-memory canvas projects that should not
preload any player data, as it is likely to OOM. Could be useful for customers with limited RAM as data
is loaded into memory before it can be stored in indexeddb.

https://www.loom.com/share/56d3a68e25d5411a815c3d81c0a302c0

No


frontend/src/pages/UserSettings/PlayerForm/PlayerForm.tsx
```diff
@@ -0,0 +1,44 @@
...
+	const [useIndexedDB, setUseIndexedDB] =
+		React.useState<boolean>(indexeddbEnabled)
+
+	return (
+		<>
+			<h3>App Settings</h3>
+			<form>
+				<p>
+					IndexedDB Browser Cache is used to preload all data in
+					highlight and speedup repeated loading, but can increase
```
et, line 20: * "highlight" -> "Highlight"
* "speedup" -> "speed up"
* "but can increase" -> "but it can increase"

frontend/src/pages/UserSettings/PlayerForm/PlayerForm.tsx
```diff
@@ -0,0 +1,44 @@
...
+					memory usage.
+				</p>
+				<div className={styles.fieldRow}>
+					<div className={styles.fieldKey}>
+						<Checkbox
+							checked={useIndexedDB}
+							onChange={(e) => {
+								const checked = e.target.checked
+								setUseIndexedDB(checked)
+								const env = import.meta.env.DEV ? 'dev' : 'prod'
```
et, line 30: Can we just use `mode`?
https://vitejs.dev/guide/env-and-mode.html

et (approved):

StephenHaney:
I will battle test once merged :) I can repro canvas recording OOM crashes very easily. (M1 MBP w/ 32GB memory)

frontend/src/pages/UserSettings/PlayerForm/PlayerForm.tsx
Vadman97: thank you!

frontend/src/pages/UserSettings/PlayerForm/PlayerForm.tsx
```diff
@@ -0,0 +1,44 @@
...
+export const PlayerForm = () => {
+	const [useIndexedDB, setUseIndexedDB] =
+		React.useState<boolean>(indexeddbEnabled)
+
+	return (
+		<>
+			<h3>App Settings</h3>
+			<form>
+				<p>
+					IndexedDB Browser Cache is used to preload all data in
```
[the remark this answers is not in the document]
Vadman97: For now, thinking that we just experiment with @StephenHaney 's project in particular to make sure that this fixes the issue, but I'll follow up to make this an automatic setting if that is the case. Using machine RAM would be reasonable, but figure out if sessions are using a canvas would be a bit more involved. If you'd like to follow along: https://github.com/highlight/highlight/issues/4769

PR #4762, opened 2023-03-29
add more context to logs datetime picker when it is closed

et:
## Summary

The relative date time picker doesn't include hours or minutes which can make it hard to understand what the time range is. For example, linking from the sessions, only [looks back the last 4 hours](https://github.com/highlight/highlight/pull/4653/files#diff-79ba7ec6cb62b65929c3c8a0aabc7b5857d3bd339dd50e40382ebf8f61f7194aR105).

This PR adjusts the label when the relative time picker is closed, namely:
* adds hours and minutes to when the relative date time picker is closed
* removes the day of week.

Before:
Screenshot 2023-03-29 at 2 59 29 PM

After:
Screenshot 2023-03-29 at 2 55 19 PM

N/A

### DIFF
--- a/packages/ui/src/components/DatePicker/PreviousDateRangePicker.tsx
+++ b/packages/ui/src/components/DatePicker/PreviousDateRangePicker.tsx
@@ -42,6 +42,16 @@ const isCustomSelected = ({
 	return !foundPreset
 }
 
+function toDateTimeString(date: Date) {
+	return date.toLocaleDateString('en-us', {
+		year: 'numeric',
+		month: 'short',
+		day: 'numeric',
+		hour: '2-digit',
+		minute: '2-digit',
+	})
+}
+
 export const getLabel = ({
 	selectedDates,
 	presets,
@@ -58,7 +68,9 @@ export const getLabel = ({
 	}
 
 	if (selectedDates.length == 2) {
-		return `${selectedDates[0].toDateString()} - ${selectedDates[1].toDateString()}`
+		return `${toDateTimeString(selectedDates[0])} - ${toDateTimeString(
+			selectedDates[1],
+		)}`
 	}
 
 	return ''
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/DatePicker/PreviousDateRangePicker.tsx
```diff
@@ -42,6 +42,16 @@ const isCustomSelected = ({
 	return !foundPreset
 }
 
+function toDateTimeString(date: Date) {
+	return date.toLocaleDateString('en-us', {
+		year: 'numeric',
```
et, line 47: We don't have moment in our ui library so we have to use the javascript built-ins.

packages/ui/src/components/DatePicker/PreviousDateRangePicker.tsx
```diff
@@ -42,6 +42,16 @@ const isCustomSelected = ({
 	return !foundPreset
 }
 
+function toDateTimeString(date: Date) {
+	return date.toLocaleDateString('en-us', {
```
et, line 46: Hardcoding to `en-us` simply because that's what something like moment would have done anyways. When the time comes, we handle localization, but for now, it's just easier to be consistent everywhere.

packages/ui/src/components/DatePicker/PreviousDateRangePicker.tsx
Vadman97: Should we add something to make this simpler? What about https://day.js.org/?

Vadman97 (approved):

packages/ui/src/components/DatePicker/PreviousDateRangePicker.tsx
et: day.js seems like a good alternative. Let's see if we add more date utilities in the future. I think it's wise to keep the component library as light as possible.

PR #4767, opened 2023-03-29
improve middleware to trace http request handling

Vadman97:
## Summary

Improve our highlight graphql `tracer` to allow logging all graphql operations. 
This is particularly handy until we have a better tracing product to allow folks to find
highlight frontend sessions where a particular network request (GraphQL operation) was
called by searching for it in the logs and then finding its session.

Fixes the go/fiber middleware to use the `c.Context()` which is more commonly 
available in customers' codebases rather than the `UserContext()`

Screenshot 2023-03-30 at 11 49 57 PM

Screenshot 2023-03-30 at 11 51 09 PM

Screenshot 2023-03-30 at 11 16 20 PM

Screenshot 2023-03-30 at 11 15 12 PM

Will deploy new version of the go package.

### DIFF
--- a/backend/clickhouse/log_row.go
+++ b/backend/clickhouse/log_row.go
@@ -15,7 +15,7 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-const LogAttributeValueLengthLimit = 2 << 10
+const LogAttributeValueLengthLimit = 2 << 12
 const LogRowSourceValueFrontend = "frontend"
 const LogRowSourceValueBackend = "backend"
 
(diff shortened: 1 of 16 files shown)


et (commented):
This is really cool! Super helpful.
Since docs are now in the same repo, can we include doc changes in this PR?

@jay-khatri - can you provide guidance on where we would document something like this?

Vadman97:
@et thanks! I'll add some docs to the `go.md` sdk docs as well as to all of the go error handling guides. Eventually this would belong in a 'tracing' product setup, with instructions per framework (99designs/GraphQL vs others)

et (approved):
Thanks for adding the docs!

PR #4775, opened 2023-03-30
improve logging and development documentation

Vadman97:
## Summary

The highlight.io build runs as part of the `turbo.yml` github action on all PRs.
For organization members, this uses turborepo remote caching so the build would only happen on relevant file changes.
For external members though, the remote caching secret is not accessible so all build steps run.
Before this change, building highlight.io required the `GRAPHCMS_TOKEN`.
This PR makes the build noop all hygraph operations when the token is not present.
Adds developer docs for developing in highlight.

cc @d29107d 

`yarn test:all` and local deploy of the docs

No


jay-khatri (approved):
lgtm. (maybe update the title of this PR?)

PR #4789, opened 2023-03-30
Import quickstart content directly into setup page

ccschmitz:
## Summary

Imports the quickstart content directly from `./highlight.io` instead of loading it as JSON from the website's endpoint. This allows us to remove the `quickstart` endpoint from the marketing site.

We also needed to make `highlight.io` a dependency of the `frontend` project in order to import the content. I initially tried to import it via a relative path (`../../../../../highlight.io/...`) which would work locally, but not in CI.

Local click test.

N/A


Vadman97 (approved):
awesome!

PR #4791, opened 2023-03-30
use materialized view for errors count

mayberryzane:
## Summary
- `daily_error_counts` was overcounting - switch this to use a materialized view similar to the sessions count instead
- cleanup some now-unused code
- increase timeout to 30 mins per refresh

- tested errors count in billing query locally
- tested `make refresh-materialized-views`

- no

### DIFF
--- a/backend/model/model.go
+++ b/backend/model/model.go
@@ -744,8 +744,6 @@ type DailySessionCount struct {
 
 const (
 	SESSIONS_TBL                    = "sessions"
-	DAILY_ERROR_COUNTS_TBL          = "daily_error_counts"
-	DAILY_ERROR_COUNTS_UNIQ         = "date_project_id_error_type_uniq"
 	METRIC_GROUPS_NAME_SESSION_UNIQ = "metric_groups_name_session_uniq"
 	DASHBOARD_METRIC_FILTERS_UNIQ   = "dashboard_metric_filters_uniq"
 )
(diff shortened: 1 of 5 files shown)


backend/model/model.go
```diff
@@ -1355,6 +1336,28 @@ func MigrateDB(ctx context.Context, DB *gorm.DB) (bool, error) {
 		return false, e.Wrap(err, "Error creating idx_daily_session_counts_view_project_id_date")
 	}
 
+	if err := DB.Exec(`
+		CREATE MATERIALIZED VIEW IF NOT EXISTS daily_error_counts_view AS
```
Vadman97, line 1340: should we only index error objects that are newer than 3 months old, similarly to how we optimized the sessions materialized view?

backend/worker/worker.go
```diff
@@ -1238,6 +1238,20 @@ func (w *Worker) RefreshMaterializedViews(ctx context.Context) {
 		log.WithContext(ctx).Fatal(e.Wrap(err, "Error refreshing daily_session_counts_view"))
 	}
 
+	if err := w.Resolver.DB.Transaction(func(tx *gorm.DB) error {
```
Vadman97, line 1241: should we run the two refreshes concurrently wrt each other?

Vadman97 (approved):

backend/worker/worker.go
mayberryzane: I think it's better to run them consecutively only because they each cause a large IOPS spike, more likely to cause resource contention if they're running concurrently

backend/model/model.go
mayberryzane: good point! updated (and updated the sessions one since that change wasn't in our code)

backend/worker/worker.go
Vadman97: sounds good!

backend/model/model.go
Vadman97: thanks!

PR #4792, opened 2023-03-31
fix STRIPE_INTEGRATION_ERROR for grandfathered-in plans

mayberryzane:
## Summary
- our stripe integration was failing to update details for older plans (e.g. ones without unlimited members) because `|SixMonths` was being appended to the lookup key
- default to the `3 month` prices instead (which doesn't add anything extra to the lookup key)
- don't update the workspace's retention period after the billing period ends to preserve the null  value

- ran `make report-stripe-usage` locally

- need to set `workspace.retention_period` to null for grandfathered 6 month retention customers


Vadman97 (approved):
nice job getting to the bottom of this!

PR #4793, opened 2023-03-31
use a materialized view to fetch log keys

et:


backend/clickhouse/logs.go
```diff
@@ -272,13 +273,30 @@ func (client *Client) ReadLogsHistogram(ctx context.Context, projectID int, para
...
+	if endDate.UTC().After(startOfCurrentHour) {
+		// If we're fetching keys within the last hour, we fallback to using the `logs` table
+		// because the materialized view hasn't populated yet.
+		sb.Select("arrayJoin(LogAttributes.keys) as Key, count() as Total").
+			From(LogsTable).
+			Where(sb.Equal("ProjectId", projectID)).
+			Where(sb.LessEqualThan("toUInt64(toDateTime(Timestamp))", uint64(endDate.Unix()))).
+			Where(sb.GreaterEqualThan("toUInt64(toDateTime(Timestamp))", uint64(startDate.Unix()))).
+			GroupBy("Key").
+			OrderBy("Total DESC")
```
Vadman97, line 287: could we share this common code between the `if` and the `else` blocks?

backend/clickhouse/logs.go
```diff
@@ -272,13 +273,30 @@ func (client *Client) ReadLogsHistogram(ctx context.Context, projectID int, para
...
-	sb.Select("arrayJoin(LogAttributes.keys) as key, count() as cnt").
-		From(LogsTable).
-		Where(sb.Equal("ProjectId", projectID)).
-		GroupBy("key").
-		OrderBy("cnt DESC").
-		Where(sb.LessEqualThan("toUInt64(toDateTime(Timestamp))", uint64(endDate.Unix()))).
-		Where(sb.GreaterEqualThan("toUInt64(toDateTime(Timestamp))", uint64(startDate.Unix())))
+
+	startOfCurrentHour := time.Now().UTC().Truncate(1 * time.Hour)
+	if endDate.UTC().After(startOfCurrentHour) {
```
Vadman97, line 278: why do we need a condition here at all if the materialized view is always up to date (there isn't a refresh interval for it, right?)

rounding to the start of the hour seems like would produce a skewed result (earlier times in the hour will have less of a chance to use the materialized view).

Issue #4744, opened 2023-03-29
Add opt-in filtering for chrome-extension errors


Vadman97:
Frontend errors with stacktraces that have chrome extensions in the path should be filtered out, with a simple toggle on the project settings page (can use #4631 as a reference on where to make this change). 

By default, all projects should have this off and we should also add a doc that documents this as a project setting (over here: https://www.highlight.io/docs/general/product-features/error-monitoring/overview)

d29107d:
@Vadman97 
I will take it.

Vadman97:
Hey @d29107d , answering your questions from discord: 

- how to migrate the db field?
  - Schema changes to [model.go](https://github.com/highlight/highlight/blob/83c412a0f3f28a13d6781593935dcee2f0e77403/backend/model/model.go#L1) will be automigrated. New tables should be added to [Models](https://github.com/highlight/highlight/blob/83c412a0f3f28a13d6781593935dcee2f0e77403/backend/model/model.go#L133-L187). [Migrations happen automatically](https://github.com/highlight/highlight/blob/83c412a0f3f28a13d6781593935dcee2f0e77403/backend/model/model.go#L1268) in dev and in github action as part of our production deploy. I would also recommend a postgres cli (or using `docker compose exec postgres`) to validate schema in your local database.

- how to generate the graph?
  -  per the [Makefile](https://github.com/highlight/highlight/blob/83c412a0f3f28a13d6781593935dcee2f0e77403/backend/Makefile), `cd backend; make private-gen` for changes to [private schema.graphqls](https://github.com/highlight/highlight/blob/83c412a0f3f28a13d6781593935dcee2f0e77403/backend/private-graph/graph/schema.graphqls) and `cd backend; make public-gen` for changes to [public schema.graphqls](https://github.com/highlight/highlight/blob/83c412a0f3f28a13d6781593935dcee2f0e77403/backend/public-graph/graph/schema.graphqls#L4). 

![image](https://user-images.githubusercontent.com/1351531/228711390-131671fe-91e4-49a3-890d-0872c8468fd5.png)

- Is there a way to execute the script to finish the above?
That should be all you need!

d29107d:
@Vadman97 
The highlight.io is public archive, what should I add a doc?
<img width="291" alt="image" src="https://user-images.githubusercontent.com/82889377/231104518-89643a9f-9784-4541-82c3-e9e92b1423c1.png">

PR #4794, opened 2023-03-31
[Feature 4744] Add opt-in filtering for chrome-extension errors

d29107d:
## Summary
Add opt-in filtering for chrome-extension errors
https://github.com/highlight/highlight/issues/4744

https://www.loom.com/share/8255701600d441339e92095f34c05ea6

### DIFF
--- a/backend/private-graph/graph/schema.resolvers.go
+++ b/backend/private-graph/graph/schema.resolvers.go
@@ -133,7 +133,13 @@ func (r *errorGroupResolver) StructuredStackTrace(ctx context.Context, obj *mode
 		stackTraceString = *obj.MappedStackTrace
 	}
 
-	return r.UnmarshalStackTrace(stackTraceString)
+	var project model.Project
+	filterChromeExtension := false
+	if err := r.DB.Where(&model.Project{Model: model.Model{ID: obj.ProjectID}}).First(&project).Error; err == nil {
+		filterChromeExtension = *project.FilterChromeExtension
+	}
+
+	return r.UnmarshalStackTrace(stackTraceString, filterChromeExtension)
 }
 
 // MetadataLog is the resolver for the metadata_log field.
@@ -231,7 +237,13 @@ func (r *errorObjectResolver) StructuredStackTrace(ctx context.Context, obj *mod
 		stackTraceString = *obj.MappedStackTrace
 	}
 
-	return r.UnmarshalStackTrace(stackTraceString)
+	var project model.Project
+	filterChromeExtension := false
+	if err := r.DB.Where(&model.Project{Model: model.Model{ID: obj.ProjectID}}).First(&project).Error; err == nil {
+		filterChromeExtension = *project.FilterChromeExtension
+	}
+
+	return r.UnmarshalStackTrace(stackTraceString, filterChromeExtension)
 }
 
 // Session is the resolver for the session field.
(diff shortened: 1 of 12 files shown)


backend/private-graph/graph/schema.resolvers.go
Vadman97, line 1: the changes to `StructuredStackTrace` mean that we are filtering out the stack trace frames on loading, right? I think what we want is to filter out the error entirely on error processing, so that the error is never saved (as similarly implemented in https://github.com/highlight/highlight/pull/4807). We'd want to update `public-graph` `resolver.go` to filter the errors out (see how we drop other errors in `HandleErrorAndGroup`)

d29107d: I think we should save all chrome extension errors so that we can check them anytime.
If missing the chrome extension errors maybe miss the way to debug something.

frontend/src/pages/ProjectSettings/FilterExtensionForm/FilterExtensionForm.tsx
```diff
@@ -0,0 +1,109 @@
...
+	}, [data?.project?.filter_chrome_extension, loading])
+
+	if (loading) {
+		return <LoadingBar />
+	}
+
+	const categories = [
+		{
+			label: 'Chrome extension',
+			info: 'Filter Chrome extension',
```
Vadman97, line 76: could this info text be a bit more detailed?
```suggestion
			info: 'Hide Chrome extension stack frames from errors.',
```

frontend/src/pages/ProjectSettings/FilterExtensionForm/FilterExtensionForm.tsx
```diff
@@ -0,0 +1,109 @@
...
+					OptInRow(c.label, c.info, c.checked, (isOptIn: boolean) => {
+						editProject({
+							variables: {
+								id: project_id!,
+								filter_chrome_extension: isOptIn,
+							},
+						})
+							.then(() => {
+								message.success(
+									`${c.label} turns ${
```
Vadman97, line 96:
```suggestion
									`${c.label} turned ${
```

frontend/src/pages/ProjectSettings/FilterExtensionForm/FilterExtensionForm.tsx
```diff
@@ -0,0 +1,109 @@
...
+		}
+	}, [data?.project?.filter_chrome_extension, loading])
+
+	if (loading) {
+		return <LoadingBar />
+	}
+
+	const categories = [
+		{
+			label: 'Chrome extension',
```
Vadman97, line 75:
```suggestion
			label: 'Chrome extension stackframes',
```

Vadman97 (approved):
this is a good starting point, thanks! we still may want to have the errors be excluded entirely (as with https://github.com/highlight/highlight/blob/5fbeb712285d24da355b6b4818140d08c9cfe7db/backend/public-graph/graph/resolver.go#L788-L801), but that can happen as a separate PR

frontend/src/pages/ProjectSettings/FilterExtensionForm/FilterExtensionForm.tsx
d29107d: :ok_hand:

d29107d:
@Vadman97 
Would you mind having a look why here failed?
https://github.com/highlight/highlight/actions/runs/4626461803/jobs/8183226472?pr=4794

Vadman97:
@d29107d the e2e test is broken in repo forks without some new commits in main. if you merge main, the test should pass. your changes shouldn't break the test though so i'll be merging this as is.

PR #4797, opened 2023-03-31
fix yaml merges

edde746:
## Summary

Currently fails to deploy with the docker compose file on versions greater than v2.17, see [this](https://github.com/docker/compose/issues/10407#issuecomment-1484787932)


Vadman97 (approved):
works great on 
```
Docker Version:           20.10.23
Docker Compose version v2.15.1
```

PR #4631, opened 2023-03-17
Ignore error groups by regex

deltaepsilon:
## To Do

- [x] Add a postgres project table `error_filter` regex column, which can be nil.
- [x] Add a private-graph operations to get and set the `error_filter` value.
- [x] Update `PublicResolver.ProcessBackendPayloadImpl` to use the `error_filter` to decide if errors should be ignored. Frontend errors go thru a different codepath in `PublicResolver.ProcessPayload`.
- [x] Build the frontend /<project>/settings/errors UI to edit / update the error pattern

## Summary

Addresses issue #4457 

Added `error_filters` to `projects` table and cloned front-end for `error_json_paths`.

Confirmed successful save.

## How did you test this change?

https://user-images.githubusercontent.com/878947/225923680-8fa075d7-b41f-4bcf-a3eb-943d7f3f4870.mp4

The back-end error filtering is harder to test. I added a `panic("test message")` into the front-end error handling code and got it to register against an `error_filter` in the back-end handling code.

## Are there any deployment considerations?

No need to backfill data. Migration should be automatic.

Issue #4457, opened 2023-03-02
Ignore error groups by regex


jay-khatri:
Currently, errors sent to highlight cannot be excluded.
For example, we may want to exclude all errors containing `admin not found` as expected.
<img width="1115" alt="Screenshot 2023-03-15 at 11 58 33 AM" src="https://user-images.githubusercontent.com/1351531/225414838-1c6b3d28-4505-4658-9242-85894a4c2cc9.png">

We have a [project error settings page](https://app.highlight.io/1/settings/errors) that allows configuring error changes, but we should allow a regex exclusion here to filter them out.
<img width="716" alt="Screenshot 2023-03-15 at 11 59 04 AM" src="https://user-images.githubusercontent.com/1351531/225414970-b265c51c-a35c-4e76-b10d-d449d1855724.png">

1. Add a postgres `project` table `error_filter` regex column, which can be nil.
2. Add a private-graph operations to get and set the `error_filter` value.
3. Update `PublicResolver.ProcessBackendPayloadImpl` to use the `error_filter` to decide if errors should be ignored. Frontend errors go thru a different codepath in `PublicResolver.ProcessPayload`.
5. Build the frontend `/<project>/settings/errors` UI to edit / update the error pattern

Vadman97:
cc @deltaepsilon

Vadman97:
fixed in #4389

PR #4807, opened 2023-03-31
Ignore error groups by regex

Vadman97:
Based on https://github.com/highlight/highlight/pull/4631

## To Do

- [x] Add a postgres project table `error_filter` regex column, which can be nil.
- [x] Add a private-graph operations to get and set the `error_filter` value.
- [x] Update `PublicResolver.ProcessBackendPayloadImpl` to use the `error_filter` to decide if errors should be ignored. Frontend errors go thru a different codepath in `PublicResolver.ProcessPayload`.
- [x] Build the frontend /<project>/settings/errors UI to edit / update the error pattern

## Summary

Addresses issue #4457 

Added `error_filters` to `projects` table and cloned front-end for `error_json_paths`.

Confirmed successful save.

https://user-images.githubusercontent.com/878947/225923680-8fa075d7-b41f-4bcf-a3eb-943d7f3f4870.mp4

The back-end error filtering is harder to test. I added a `panic("test message")` into the front-end error handling code and got it to register against an `error_filter` in the back-end handling code.

No need to backfill data. Migration should be automatic.

### DIFF
--- a/backend/private-graph/graph/schema.graphqls
+++ b/backend/private-graph/graph/schema.graphqls
@@ -315,7 +315,7 @@ type Project {
 	secret: String
 	workspace_id: ID!
 	excluded_users: StringArray
-	error_filters: StringArray
+	error_filters: StringArray!
 	error_json_paths: StringArray
 	rage_click_window_seconds: Int
 	rage_click_radius_pixels: Int
--- a/backend/private-graph/graph/schema.graphqls
+++ b/backend/private-graph/graph/schema.graphqls
@@ -315,6 +315,7 @@ type Project {
 	secret: String
 	workspace_id: ID!
 	excluded_users: StringArray
+	error_filters: StringArray
 	error_json_paths: StringArray
 	rage_click_window_seconds: Int
 	rage_click_radius_pixels: Int
@@ -1587,6 +1588,7 @@ type Mutation {
 		name: String
 		billing_email: String
 		excluded_users: StringArray
+		error_filters: StringArray
 		error_json_paths: StringArray
 		rage_click_window_seconds: Int
 		rage_click_radius_pixels: Int
--- a/backend/private-graph/graph/schema.graphqls
+++ b/backend/private-graph/graph/schema.graphqls
@@ -315,7 +315,7 @@ type Project {
 	secret: String
 	workspace_id: ID!
 	excluded_users: StringArray
-	error_filters: StringArray!
+	error_filters: StringArray
 	error_json_paths: StringArray
 	rage_click_window_seconds: Int
 	rage_click_radius_pixels: Int
--- a/backend/public-graph/graph/resolver.go
+++ b/backend/public-graph/graph/resolver.go
@@ -2344,7 +2344,7 @@ func (r *Resolver) ProcessBackendPayloadImpl(ctx context.Context, sessionSecureI
 	// Filter out empty errors
 	var filteredErrors []*publicModel.BackendErrorObjectInput
 	for _, errorObject := range errorObjects {
-		if r.isExcludedError(ctx, project.ErrorFilters, errorObject.Event) {
+		if r.isExcludedError(ctx, project.ErrorFilters, errorObject.Event, project.ID) {
 			continue
 		}
 		filteredErrors = append(filteredErrors, errorObject)
@@ -2829,26 +2829,19 @@ func (r *Resolver) ProcessPayload(ctx context.Context, sessionSecureID string, e
 		if hasBeacon {
 			r.DB.Where(&model.ErrorObject{SessionID: &sessionID, IsBeacon: true}).Delete(&model.ErrorObject{})
 		}
+
+		var project model.Project
+		if err := r.DB.Where(&model.Project{Model: model.Model{ID: projectID}}).First(&project).Error; err != nil {
+			return e.Wrap(err, "error querying project")
+		}
+
 		// filter out empty errors
 		seenEvents := map[string]*publicModel.ErrorObjectInput{}
 		for _, errorObject := range errors {
-			if errorObject.Event == "[{}]" {
-				var objString string
-				objBytes, err := json.Marshal(errorObject)
-				if err != nil {
-					log.WithContext(ctx).Error(e.Wrap(err, "error marshalling error object when filtering"))
-					objString = ""
-				} else {
-					objString = string(objBytes)
-				}
-				log.WithContext(ctx).WithFields(log.Fields{
-					"project_id":   projectID,
-					"session_id":   sessionID,
-					"error_object": objString,
-				}).Warn("caught empty error, continuing...")
-			} else {
-				seenEvents[errorObject.Event] = errorObject
+			if r.isExcludedError(ctx, project.ErrorFilters, errorObject.Event, project.ID) {
+				continue
 			}
+			seenEvents[errorObject.Event] = errorObject
 		}
 		errors = lo.Values(seenEvents)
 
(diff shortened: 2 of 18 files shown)


et (commented):
Test failure looks like legit. 

```
--- FAIL: TestProcessBackendPayloadImpl (0.00s)
    --- FAIL: TestProcessBackendPayloadImpl/trpc_test (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x11b7779]
```

Can you take a look and reassign to me when it's passing @Vadman97

Vadman97:
Passing checks now @et, thanks for catching that!

backend/public-graph/graph/resolver.go
```diff
@@ -2316,9 +2316,30 @@ func (r *Resolver) ProcessBackendPayloadImpl(ctx context.Context, sessionSecureI
...
 	var filteredErrors []*publicModel.BackendErrorObjectInput
 	for _, errorObject := range errorObjects {
+
+		// Filter out by project.ErrorFilters, aka regexp filters
+		var err error
+		matchedRegexp := false
+		for _, errorFilter := range project.ErrorFilters {
+			matchedRegexp, err = regexp.MatchString(errorFilter, errorObject.Event)
+			if err != nil {
+				log.WithContext(ctx).WithField("regex", errorFilter).WithError(err).Error("invalid regex: failed to parse backend error filter")
```
et, line 2334: Should we have any frontend validation on whether the regex is valid? This is going to silently fail for the end user and they'll have no idea.

frontend/src/pages/ProjectSettings/ErrorFiltersForm/ErrorFiltersForm.tsx
```diff
@@ -0,0 +1,99 @@
...
+			setErrorFilters(data?.project?.error_filters || [])
+		}
+	}, [data?.project?.error_filters, loading])
+
+	if (loading) {
+		return <LoadingBar />
+	}
+
+	return (
+		<FieldsBox id="errors">
```
et, line 61: I've never seen this component but looking the code...it looks like it scrolls that part of the page into view if the URL hash matches:

https://github.com/highlight/highlight/blob/fae0d6edd3f0590bbe72b38a4b6a8d024e8d64cd/frontend/src/components/FieldsBox/FieldsBox.tsx#L13-L19

Yet another part of the page uses the same `id`:
https://github.com/highlight/highlight/blob/fae0d6edd3f0590bbe72b38a4b6a8d024e8d64cd/frontend/src/pages/ProjectSettings/ErrorSettingsForm/ErrorSettingsForm.tsx#L61

So these two are going to conflict with each other.

Honestly, I can't see a reason for us to actually use this component. The page isn't even that long so it feels unnecessary. Can we just omit the usage?

frontend/src/pages/ProjectSettings/ErrorFiltersForm/ErrorFiltersForm.tsx
```diff
@@ -0,0 +1,99 @@
...
+			<h3>Error Filters</h3>
+
+			<form onSubmit={onSubmit} key={project_id}>
+				<p>Enter REGEXP patterns to filter out errors.</p>
```
et, line 65: Can we rename `REGEXP` -> `regular expression`

backend/private-graph/graph/schema.graphqls
```diff
@@ -315,6 +315,7 @@ type Project {
 	secret: String
 	workspace_id: ID!
 	excluded_users: StringArray
+	error_filters: StringArray
```
et, line 318: The db default is an empty array, can we make `StringArray` -> `StringArray!`

frontend/src/pages/ProjectSettings/ErrorFiltersForm/ErrorFiltersForm.tsx
```diff
@@ -0,0 +1,99 @@
...
+				<div className={styles.inputAndButtonRow}>
+					<Select
```
et, line 67: Can we tidy up the empty state?

Screenshot 2023-04-03 at 9 17 38 AM

et (commented):
We need some type of validation on the regex to ensure it's valid.

I was able to break this by doing the following:

* enter `a(b` into the input
* Visited https://localhost:3000/1/buttons and clicked "Throw an error"
* Saw an error when processing the payload

Screenshot 2023-04-03 at 9 22 38 AM

jay-khatri:
If you don't mind updating the docs here once this is ready: 
https://github.com/highlight/highlight/pull/4889/files#r1161038710

frontend/src/pages/ProjectSettings/ErrorFiltersForm/ErrorFiltersForm.tsx
Vadman97: updating the ids here for consistency, but the fields box sets styles and allows us to link directly to a section here (eg https://localhost:3000/2/settings/errors#sourcemaps)
Screenshot 2023-04-09 at 4 24 53 PM

frontend/src/pages/ProjectSettings/ErrorFiltersForm/ErrorFiltersForm.tsx
Vadman97: updated

frontend/src/pages/ProjectSettings/ErrorFiltersForm/ErrorFiltersForm.tsx
```diff
@@ -0,0 +1,99 @@
...
+					<Button
+						trackingId="UpdateErrorFilters"
+						htmlType="submit"
+						type="primary"
+						className={clsx(
+							commonStyles.submitButton,
+							styles.saveButton,
+						)}
+					>
+						{editProjectLoading ? (
```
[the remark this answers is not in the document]
Vadman97: cleaned this up, thanks!

Vadman97:
updated docs @jay-khatri

docs-content/general/6_product-features/2_error-monitoring/ignoring-errors.md
```diff
@@ -17,7 +17,7 @@ If your users are using chrome extensions, you may see errors that are not relev
 If you have alerts set up for your project, you can ignore specific error groups from triggering alerts. You can do this by clicking the "Ignore" button on the error group page.
 
 ## Ignore errors by regex on the error body
-Coming soon.
+If you'd like to ignore specific errors by regex pattern match against the error body, you can do so by adding error filters in your [project settings](https://app.highlight.io/settings/errors#filters).
```
jay-khatri, line 20:
```suggestion
If you'd like to ignore specific errors by a regex pattern match against the error body, you can do so by adding error filters in your [project settings] (https://app.highlight.io/settings/errors#filters) as well.
```

jay-khatri (approved):

et (approved):

backend/public-graph/graph/resolver.go
```diff
@@ -3024,6 +3027,28 @@ func (r *Resolver) ProcessPayload(ctx context.Context, sessionSecureID string, e
 	return nil
 }
 
+func (r *Resolver) isExcludedError(ctx context.Context, errorFilters []string, errorEvent string) bool {
+	if errorEvent == "[{}]" {
```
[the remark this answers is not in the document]
Vadman97: I believe this is something we've had from an old client version where error events might have been serialized like this. I'll readd the log here to see if this still happens and if not, we can drop this code.

PR #4809, opened 2023-03-31
disable reportConsoleErrors by default

Vadman97:
## Summary

Recently, we released a new flag to the frontend `H.init` to disable `console.error(...)` calls from being reported as Highlight Errors. This was done because frontend errors are still reported as error logs and are often noisy (not a real application-breaking error like a `window.onerror`, etc).
This PR switches `reportConsoleErrors` off by default to make this the default behavior for new SDK versions >= 5.5.0.
Adds a `disableSessionRecording` setting that allows using the javascript client without recording session data.

Disabling the flag locally and ensuring that frontend `console.error` calls are not reported as Highlight errors.
Screenshot 2023-03-31 at 4 21 42 PM

Error reporting by the sessionless client.
Screenshot 2023-04-11 at 12 06 31 AM

New `highlight.run` version 5.5.0

### DIFF
--- a/sdk/firstload/package.json
+++ b/sdk/firstload/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "highlight.run",
-	"version": "5.5.0",
+	"version": "6.0.0",
 	"scripts": {
 		"build": "rollup -c",
 		"dev": "rollup -c -w",
--- a/sdk/firstload/package.json
+++ b/sdk/firstload/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "highlight.run",
-	"version": "5.4.2",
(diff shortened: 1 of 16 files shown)


sdk/firstload/package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "highlight.run",
-	"version": "5.4.2",
+	"version": "5.5.0",
```
et, line 3: Shouldn't we bump this to `6.0.0` since it's sort of a breaking change?

Vadman97: Updated rrweb and bumped the major version.

et (approved):

PR #4813, opened 2023-04-02
Move from cjs to esm entry points for antd

lewisl9029:
## Summary

_This PR is part of a series of PRs being spun off from [my WIP branch](https://github.com/lewisl9029/highlight/pull/2) to get the Highlight web app ready for [Reflame](https://reflame.app/). Hopefully this makes things a bit easier to review, test, and merge. 🙂_ 

Previously, we were importing several antd components from `antd/lib/*`. 

If we look at their [package.json](https://github.com/ant-design/ant-design/blob/4.24.8/package.json#L44), we can see that the `/lib` directory is actually for their CommonJS modules. 

Vite can convert these to ES modules for us, but this can be a bit wasteful since Vite defaults to importing esm entry points where ever possible (in this case `/es/index.js`), so when we manually import CommonJS modules like this, we can end up including the same components twice in our bundles: once for the ESM version when we access it from the root (`import { blah } from 'antd'`) and once for the CJS version when we import from specific modules (`import blah from 'antd/lib/blah'`). It can in theory even cause bugs if we end up with duplicate context providers for instance.

The fix is pretty simple, I just did a global search & replace for `'antd/lib` -> `'antd/es`.

I also did a quick comparison of bundle sizes to validate my hypothesis:

Before:

```
@highlight-run/frontend:build: build/assets/index2.js                            1,451.09 kB │ gzip:   360.15 kB │ map:  4,608.04 kB
@highlight-run/frontend:build: build/assets/index.js                             7,632.86 kB │ gzip: 2,194.46 kB │ map: 25,321.52 kB
```

After:
```
@highlight-run/frontend:build: build/assets/index2.js                            1,463.51 kB │ gzip:   364.69 kB │ map:  4,645.33 kB
@highlight-run/frontend:build: build/assets/index.js                             7,541.09 kB │ gzip: 2,170.50 kB │ map: 25,073.95 kB
```

About 20KB saved. Not bad for such a tiny change. :)

I tried running both `yarn turbo run build --filter frontend...` and `yarn turbo run dev --filter frontend...`, and neither had any errors. And the app seemed to work fine in dev, but I could only check the sign in screen.

None that I can think of.


ccschmitz (approved):
This is great - thank you, @lewisl9029! Appreciate the contribution and detailed explanation in the PR description.

PR #4815, opened 2023-04-03
[hotfix] highlight.io build error

ccschmitz:
## Summary

We are hitting the following error in our docker builds on main:

```sh
#50 [public.ecr.aws/k9o6n7l8/highlight-frontend frontend-base 28/42] RUN yarn
#50 0.787 ➤ YN0000: ┌ Project validation
#50 0.787 ➤ YN0057: │ @highlight-run/frontend: Resolutions field will be ignored
#50 0.788 ➤ YN0057: │ @highlight-run/client: Resolutions field will be ignored
#50 0.788 ➤ YN0000: └ Completed
#50 0.788 ➤ YN0000: ┌ Resolution step
#50 1.439 ➤ YN0001: │ Error: highlight.io@workspace:*: Workspace not found (highlight.io@workspace:*)
```

The problem is that we don't copy over our `highlight.io` directory in our `frontend.Dockerfile`.

Set the workflow step to run and took it from [red](https://github.com/highlight/highlight/actions/runs/4598192564/jobs/8121903507?pr=4815) to [green (still failing because an ENV var wasn't set, but you can see we got past the failure)](https://github.com/highlight/highlight/actions/runs/4598207841/jobs/8121940258?pr=4815).

N/A

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -3,6 +3,9 @@ name: Docker
 on:
     push:
         branches: ['main']
+    # TODO: Remove
+    pull_request:
+        types: [opened, synchronize]
     workflow_dispatch:
     merge_group:
 
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -3,9 +3,6 @@ name: Docker
 on:
     push:
         branches: ['main']
-    # TODO: Remove
-    pull_request:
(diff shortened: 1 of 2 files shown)


Vadman97 (approved):

Vadman97:
nope, i'll open a pr for that. we could start running it even if it is not a 'required' job in case it takes a while to complete.

PR #4819, opened 2023-04-03
Finish setup docs for logs

ccschmitz:


docker/.env
```diff
@@ -4,7 +4,7 @@ COMPOSE_PROJECT_NAME=highlight
 # use for prod
 COMPOSE_FILE=compose.yml
 # use for dev
-# COMPOSE_FILE=compose.yml:compose.dev.yml
+COMPOSE_FILE=compose.yml:compose.dev.yml
```
ccschmitz, line 7: @Vadman97 do you think this change should be committed, or should I back it out?

Vadman97: fine to commit it as it has been a point of confusion now for a few folks and it's safer to default to the dev environment. i'll be changing this in #4820 to simplify this anyways.

Vadman97 (approved):
awesome work! and great to see the hubspot bug fixed with this as well

PR #4827, opened 2023-04-04
Log Alerts UI + Backend

mayberryzane:
## Summary
- builds off of #4554 
- added backend code for saving log alert metadata
- created a worker to poll clickhouse at alert-specific frequencies
- added a new key in redis to store the latest timestamp for logs for each project - this is used to determine what timestamps to poll for (assuming there could be a delay in log ingest)

- local click test for saving / editing / deleting log alerts from the alerts page
- ran the `log-alerts` worker locally, logged and monitored log alerts being fired

- will set up ECS service with a single task similar to metric monitors to run the `log-alerts` worker


backend/jobs/log-alerts/log-alerts.go
```diff
@@ -0,0 +1,194 @@
...
+				"🚨 *%s* fired!\nLog count %sis currently %s the threshold.\n"+
+					"_Count_: %d | _Threshold_: %d",
+				*alert.Name,
+				queryStr,
+				aboveStr,
+				count,
+				alert.CountThreshold,
+			)
+
+			fmt.Println(message)
```
Vadman97, line 149: should we logrus log instead with additional context?

backend/jobs/log-alerts/log-alerts.go
```diff
@@ -0,0 +1,194 @@
...
+		}
+	}()
+
+	for _, freq := range AlertFrequencies {
+		f := freq
+		go func() {
+			for range time.Tick(time.Duration(f) * time.Second) {
+				alerts := (*alertsByFrequency)[f]
+				if len(alerts) > 0 {
+					go func() {
```
Vadman97, line 60: nit - don't need the extra func closure, can just call the method

backend/private-graph/graph/schema.graphqls
```diff
@@ -1071,6 +1085,24 @@ type SessionAlert {
 	disabled: Boolean!
 }
 
+type LogAlert {
+	id: ID!
+	updated_at: Timestamp!
+	Name: String!
+	ChannelsToNotify: [SanitizedSlackChannel!]!
```
Vadman97, line 1092: could we rename this to SlackChannelsToNotify?

backend/private-graph/graph/schema.graphqls
```diff
@@ -1071,6 +1085,24 @@ type SessionAlert {
...
+	DiscordChannelsToNotify: [DiscordChannel!]!
+	EmailsToNotify: [String!]!
```
Vadman97, line 1094: could the log alerts also support a web hook endpoint?

backend/public-graph/graph/resolver.go
```diff
@@ -833,6 +833,12 @@ func (r *Resolver) HandleErrorAndGroup(ctx context.Context, errorObj *model.Erro
 			return nil, ErrNoisyError
 		}
 	}
+	if projectID == 3322 {
```
Vadman97, line 836: unrelated change, but feel free to keep this. just want to check if this is still relevant.

frontend/src/pages/Alerts/LogAlert/LogAlertRouter.tsx
```diff
@@ -0,0 +1,42 @@
...
+				alertsPayload: data,
+				loading,
+				slackUrl,
+			}}
+		>
+			<Helmet>
+				<title>Log Alerts</title>
+			</Helmet>
+			<Routes>
+				<Route path="new" element={<LogAlertPage />} />
```
Vadman97, line 35: LogAlertsPage will need to bring in we hook logic from other alert types

Vadman97 (approved):
🥳 awesome job getting this over the finish line

backend/jobs/log-alerts/log-alerts.go
mayberryzane: yep will do! copied this over from metric monitors

backend/jobs/log-alerts/log-alerts.go
mayberryzane: good point! my intention was to run each alert concurrently but this is operating on the list of alerts, will refactor

backend/private-graph/graph/schema.graphqls
mayberryzane: I wanted to keep it `ChannelsToNotify` just because that's the name on the base `Alert` model

PR #4830, opened 2023-04-04
add www to site images to ensure we do not redirect

et:
## Summary

Reusing images from highlight.io in app.highlight.io results in a 308 redirect error:
Screenshot 2023-04-04 at 11 21 07 AM

We observed that manually changing the URL to use www fixes that redirect error.

This is somewhat tricky to test. We're operating off of a hunch. If this does not work, we will just move the images over.

N/A


Vadman97 (approved):

et:
Confirmed that this did fix this image:

Screenshot 2023-04-04 at 12 08 21 PM

Issue #4816, opened 2023-04-03
Add support for multi-select menus


Vadman97:
session devtools network panel should have multi select for type and status codes. update styles per the figma

<img width="244" alt="image" src="https://user-images.githubusercontent.com/1351531/229589087-88ed7c22-608f-4b8a-ae71-e79dc88880b2.png">
<img width="218" alt="Screenshot 2023-04-03 at 10 56 21 AM" src="https://user-images.githubusercontent.com/1351531/229589119-ae9fc04b-790c-4a65-9626-5c2e480692be.png">

https://www.figma.com/file/rdBPNzn7Klk6RG1LHUCE5B/Screen-Designs?node-id=6701-621859&t=AkqGfHoXduYOKD7E-11

follow up on https://github.com/highlight/highlight/pull/4580

PR #4833, opened 2023-04-04
Fetch session logs via clickhouse

### LINKED ISSUE #4779: Render backend logs in session viewer
Now that we're piping the "source" of a log, a good next step is to actually render backend logs in the session viewer. I do think there's lots more cool things we can do to allow for fine-grained rendering of logs (put some ideas [here](https://github.com/highlight/highlight/issues/4692)), but a good first step is showing a new tab in the devtools.

- [x] #4786
- [ ] Rename the "Console" tab "Console Logs"
- [ ] Instead of "Related Logs" (see image below), call it "Show in Log Viewer", which should populate a search for the correct `source` and `session_secure_id`.

Image

### LINKED ISSUE #4871: Add a "goto" button to the right of each console log (and disable the goto action for clicking on the log text)

et:
## Summary

This PR's main purpose is to change the session logs source from S3 to Clickhouse (dedicated backend endpoint added in #4883). Additionally, we now have the ability to see Backend logs in the session logs viewer.

We are also making several UX change. See [figma](https://www.figma.com/file/rdBPNzn7Klk6RG1LHUCE5B/Screen-Designs?node-id=4451-205639&t=dnCWY9epGQJQ8INj-11) but also below for a detailed before/after.

Fixes #4779

**Note: we don't have a component for multi-select (#4816) so the filters use the existing menu button component which only permits single select**

### Adjust log level filter

Adjust the level filter to remove all colors and use the normalized log levels from clickhouse:
https://github.com/highlight/highlight/blob/83da5535aa0bca9239bd316a91c2f402e4f3f824/backend/private-graph/graph/schema.graphqls#L296-L303

Before

Screenshot 2023-04-10 at 2 58 50 PM

After:

Screenshot 2023-04-10 at 2 59 02 PM

### Add explicit level to the log line

Instead of a vertical bar to indiciate the log level, we have the explicit name of the level on the line. This uses the same colors as the log viewer.

Before:

Screenshot 2023-04-10 at 3 04 11 PM

After:
Screenshot 2023-04-10 at 3 02 04 PM

### Add source filter

Adds a source filter to distiguish between Frontend and Backend logs. This should default to Frontend

Screenshot 2023-04-10 at 3 00 01 PM

### Add Go to button

Instead of clicking the whole log to select the log (and move the session player), we now have an explicit button:

Screenshot 2023-04-10 at 3 00 51 PM

Fixes #4871

### Other things

* Rename the "Console" tab to "Console Logs"
* Rename "Related logs" button to "Show in Log Viewer"
  * Clicking this button should persist the `level`, `source`, and `secure_session_id` filters.
* Clicking "Related Session" in the main logs view will select the log line in the session logs and set the player to that time.
* Disabling preloading. This seems to load decently fast and we don't need to prefetch this data.

### DIFF
--- a/frontend/src/pages/Player/Toolbar/DevToolsWindowV2/ConsolePage/ConsolePage.tsx
+++ b/frontend/src/pages/Player/Toolbar/DevToolsWindowV2/ConsolePage/ConsolePage.tsx
@@ -1,189 +1,114 @@
 import LoadingBox from '@components/LoadingBox'
-import { useGetMessagesQuery } from '@graph/hooks'
-import { ConsoleMessage } from '@highlight-run/client'
+import { Log, LogLevel, LogSource } from '@graph/schemas'
 import { playerMetaData } from '@highlight-run/rrweb-types'
-import { Box, Text } from '@highlight-run/ui'
+import {
+	Box,
+	IconSolidArrowCircleRight,
+	Stack,
+	Tag,
+	Text,
+} from '@highlight-run/ui'
+import { useProjectId } from '@hooks/useProjectId'
+import { COLOR_MAPPING } from '@pages/LogsPage/constants'
+import { buildSessionParams } from '@pages/LogsPage/SearchForm/utils'
 import { ReplayerState } from '@pages/Player/ReplayerContext'
 import { EmptyDevToolsCallout } from '@pages/Player/Toolbar/DevToolsWindowV2/EmptyDevToolsCallout/EmptyDevToolsCallout'
-import { LogLevel, Tab } from '@pages/Player/Toolbar/DevToolsWindowV2/utils'
-import { indexedDBFetch } from '@util/db'
-import { useParams } from '@util/react-router/useParams'
+import { Tab } from '@pages/Player/Toolbar/DevToolsWindowV2/utils'
 import clsx from 'clsx'
-import { H } from 'highlight.run'
 import _ from 'lodash'
 import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react'
 import { Virtuoso, VirtuosoHandle } from 'react-virtuoso'
 
+import { useGetSessionLogsQuery } from '@/graph/generated/hooks'
 import { styledVerticalScrollbar } from '@/style/common.css'
 
 import { useReplayerContext } from '../../../ReplayerContext'
 import * as styles from './style.css'
 
-interface ParsedMessage extends ConsoleMessage {
-	selected?: boolean
-	id: number
-}
+type SessionLog = Pick<Log, 'timestamp' | 'message' | 'level'>
+type SessionLogEdge = { cursor: string; node: SessionLog }
 
 export const ConsolePage = ({
 	autoScroll,
 	filter,
-	logLevel,
+	sources,
+	levels,
 	time,
 }: {
 	autoScroll: boolean
 	filter: string
-	logLevel: LogLevel
+	sources: LogSource[]
+	levels: LogLevel[]
 	time: number
 }) => {
-	const [currentMessage, setCurrentMessage] = useState(-1)
+	const { projectId } = useProjectId()
+	const [selectedCursor, setSelectedCursor] = useState('')
 	const { session, setTime, sessionMetadata, isPlayerReady, state } =
 		useReplayerContext()
-	const [parsedMessages, setParsedMessages] = useState<
-		undefined | Array<ParsedMessage>
-	>([])
-	const { session_secure_id } = useParams<{ session_secure_id: string }>()
-	const [loading, setLoading] = useState(true)
-	const skipQuery = session === undefined || !!session?.messages_url
-	const { data, loading: queryLoading } = useGetMessagesQuery({
+
+	const { data, loading } = useGetSessionLogsQuery({
 		variables: {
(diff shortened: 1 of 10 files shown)


frontend/src/pages/Player/Toolbar/DevToolsWindowV2/ConsolePage/ConsolePage.tsx
```diff
@@ -1,199 +1,147 @@
...
-			}) ?? [],
-		)
-	}, [data?.messages])
+		return data?.sessionLogs
+	}, [filter, data?.sessionLogs])
 
 	// Logic for scrolling to current entry.
 	useEffect(() => {
-		if (state !== ReplayerState.Playing) {
+		if (state !== ReplayerState.Playing || !autoScroll) {
```
et, line 83: The previous code was missing `!autoScroll` which is a bug. The user should explicitly be enabling auto scrolling for the below code to run.

frontend/src/pages/Player/Toolbar/DevToolsWindowV2/ConsolePage/ConsolePage.tsx
```diff
@@ -1,199 +1,147 @@
...
-		}
-
-		return currentMessages.filter((message) => message?.value?.length)
-	}, [logLevel, filter, parsedMessages])
+	}, [state, time, messagesToRender, autoScroll])
 
 	const virtuoso = useRef<VirtuosoHandle>(null)
-	// eslint-disable-next-line react-hooks/exhaustive-deps
-	const scrollFunction = useCallback(
-		_.debounce((index: number) => {
```
et, line 180: Not sure why we were debouncing but it may have been trying to compensate for perf issues with smooth scrolling. I disabled smooth scrolling (see the link on the right for a comment)

packages/ui/src/css/theme.css.ts
```diff
@@ -28,7 +28,7 @@ export const themeVars = createGlobalTheme('.highlight-light-theme', {
 				informative: '#6346af',
 			},
 		},
-		divider: { strong: '#c8c7cb', default: '#dcdbdd', weak: '#e4e2e4' },
+		divider: { strong: '#c8c7cb', default: '#dcdbdd', weak: '#ebe9eb' },
```
et, line 31: Updated color to match what was in figma.

frontend/src/pages/Player/Toolbar/DevToolsWindowV2/DevToolsWindowV2.tsx
```diff
@@ -281,33 +296,59 @@ const DevToolsWindowV2: React.FC<
...
+														setLevels([
+															level as LogLevel,
+														])
+													}
+												}}
+											/>
+											<MenuButton
+												divider
+												size="medium"
+												selectedKey={LogSource.Frontend}
```
et, line 330: Kind of a hack here -- we need to set the default value of this dropdown to "Frontend". The state variable above is setting this correctly:

https://github.com/highlight/highlight/blob/2c0d451c3155f60a4cbfce46cf645f6c2df3d1c3/frontend/src/pages/Player/Toolbar/DevToolsWindowV2/DevToolsWindowV2.tsx#L80-L82

But it operates as being future proofed to handling multiple log sources (for when we have a proper multiselect component) whereas `<MenuButton />` can only handle one selected value. I could have done something like `selectedKey={sources[0]}` but that would have been kind of brittle. Hence, i'm opting for just duping the behavior and this will be cleaner when we have our multiselect component available.

frontend/src/pages/Player/Toolbar/DevToolsWindowV2/ConsolePage/ConsolePage.tsx
Vadman97: nice find! yup, that's exactly why we were trying to limit the scroll updates

frontend/src/pages/Player/Toolbar/DevToolsWindowV2/ConsolePage/ConsolePage.tsx
```diff
@@ -1,199 +1,147 @@
...
-			if (virtuoso.current) {
+	useEffect(() => {
+		if (
+			isPlayerReady && // ensure Virtuoso component is actually rendered
+			virtuoso.current &&
+			messagesToRender
+		) {
+			const index = messagesToRender.findIndex(
+				(logEdge) => logEdge.cursor === selectedCursor,
+			)
```
Vadman97, line 113: since this could be an expensive find, does it make sense to pull this into it's own `useMemo` value that just depends on `messagesToRender` and `selectedCursor` so that this doesn't change when the rest of the dependencies change?

Vadman97 (approved):
nice! 🏴‍☠️ 🚢

frontend/src/pages/Player/Toolbar/DevToolsWindowV2/ConsolePage/ConsolePage.tsx
et: Thanks for pairing. I've adjusted to use `useMemo` in aa48158ff9d44978ae6941535b7a8080decc913f

Will do another test on the render preview when it deploys.

Issue #4403, opened 2023-02-27
Websocket message recording


jay-khatri:
Consider supporting websocket messages in the highlight.io session viewer UI. Given this stackoverflow thread, it seems quite possible.

https://stackoverflow.com/questions/31181651/inspecting-websocket-frames-in-an-undetectable-way

That being said, some challenges that are worth considering:
- How do users want to filter websocket messages? By endpoint? Is there a concept of a "topic"
- Should this live in the network request UI?

Proposed Task list: 

- [x] #5544
- [x] #5541
- [x] #5542
- [x] #5543
- [x] #5780

PR #4835, opened 2023-04-05
Set up WebSocketListener to intercept WebSocket events

fwouts:
## Summary

This is a draft implementation of client-side recording of WebSocket events (#4403). It only implements the client-side interception and doesn't yet send events through.

Prior to this, it may be a good idea to improve type safety for the existing handling of `resources` so that we can integrate WebSocket events to that list in a safe manner. In particular, I suspect the current code may sometimes return objects that are missing `requestResponsePairs`, when a PerformanceResourceTiming isn't associated with a RequestResponsePair (e.g. for requests occurring before the Highlight client has finished loading). This would fail silently on the backend because [we just `return nil`](https://github.com/highlight/highlight/blob/fcb6dac185830b441f740733977395ccddf6fe5f/backend/public-graph/graph/resolver.go#L2810). Or perhaps Unmarshal defaults to an empty object when a field is missing?

Edit: re-reading the code, yes an empty object is expected. I wonder if a pointer would be good for RequestResponsePair so it would be null instead of empty when unmarshalled?

I haven't had time to write tests yet, but I manually tested by establishing a WebSocket connection to a local server and checking that events were intercepted correctly (including a failed connection).

No.

### DIFF
--- a/sdk/client/src/listeners/first-load-listeners.tsx
+++ b/sdk/client/src/listeners/first-load-listeners.tsx
@@ -270,10 +270,6 @@ export class FirstLoadListeners {
 				// Prior to this, it may be best to refactor the existing logic
 				// above to be less tied to PerformanceResourceTiming, since we
 				// only care about a subset of fields.
-				console.log(
-					'Intercepted WebSocket events',
-					sThis.wsNetworkContents,
-				)
 			}
 		}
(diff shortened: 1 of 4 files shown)


Vadman97:
implemented in #5541

PR #4839, opened 2023-04-05
use graphql enum for identifying log source

et:
## Summary

In preparation for #4779, we need a GraphQL enum to identify the log source (because Go types aren't usable on the frontend without a of work).

Additionally, this PR fixes a bug with the `IsBackend` function that wasn't correctly identifying backend logs correctly.

* Tests added
* Code should compile

N/A

### DIFF
--- a/backend/clickhouse/log_row.go
+++ b/backend/clickhouse/log_row.go
@@ -17,8 +17,6 @@ import (
 )
 
 const LogAttributeValueLengthLimit = 2 << 10
-const LogRowSourceValueFrontend = "frontend"
-const LogRowSourceValueBackend = "backend"
 
 func ProjectToInt(projectID string) (int, error) {
 	i, err := strconv.ParseInt(projectID, 10, 32)
@@ -77,7 +75,7 @@ func (l *LogRow) Cursor() string {
 }
 
 func (l *LogRow) IsBackend() bool {
-	return l.Source != highlight.SourceAttributeFrontend
+	return l.Source == modelInputs.LogSourceBackend.String()
 }
 
 type LogRowOption func(*LogRow)
@@ -91,8 +89,8 @@ func WithTimestamp(ts time.Time) LogRowOption {
 }
 
 func WithLogAttributes(ctx context.Context, resourceAttributes, spanAttributes, eventAttributes map[string]any, isFrontendLog bool) LogRowOption {
-	return func(h *LogRow) {
-		h.LogAttributes = GetAttributesMap(ctx, resourceAttributes, spanAttributes, eventAttributes, isFrontendLog)
+	return func(l *LogRow) {
+		l.LogAttributes = GetAttributesMap(ctx, resourceAttributes, spanAttributes, eventAttributes, isFrontendLog)
 	}
 }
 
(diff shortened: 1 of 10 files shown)


backend/clickhouse/log_row.go
```diff
@@ -104,39 +98,35 @@ func WithSeverityText(severityText string) LogRowOption {
...
 }
 
-func WithSource(source string) LogRowOption {
-	return func(h *LogRow) {
-		if source == highlight.SourceAttributeFrontend {
-			h.Source = LogRowSourceValueFrontend
-		} else {
-			h.Source = LogRowSourceValueBackend
-		}
+func WithSource(source modelInputs.LogSource) LogRowOption {
```
et, line 101: Note that `modelInputs.LogSource` is technically a `string`

https://github.com/highlight/highlight/blob/d8ee3158b7599a8e1fcecde73e3d701e4d84b274/backend/private-graph/graph/model/models_gen.go#L953

So any string can be passed in here. 

Not sure if it's possible for gqlgen to give Go a type that's more strict (see [discussion](https://github.com/99designs/gqlgen/discussions/1898)).

backend/clickhouse/logs_test.go
```diff
@@ -57,6 +57,33 @@ func assertCursorsOutput(t *testing.T, edges []*modelInputs.LogEdge, expectedCur
...
+	}
+
+	assert.NoError(t, client.BatchWriteLogRows(ctx, rows))
+
+	payload, err := client.ReadLogs(ctx, 1, modelInputs.LogsParamsInput{
+		DateRange: makeDateWithinRange(now),
+	}, Pagination{})
+	assert.NoError(t, err)
+	assert.Len(t, payload.Edges, 1)
+	assert.Equal(t, modelInputs.LogSourceFrontend.String(), *payload.Edges[0].Node.Source)
```
et, line 84: I just wanted to confirm that changing the type on `Source`:
https://github.com/highlight/highlight/blob/d8ee3158b7599a8e1fcecde73e3d701e4d84b274/backend/clickhouse/log_row.go#L47

still correctly writes (and reads) out the expected data.

backend/otel/otel.go
```diff
@@ -380,7 +395,7 @@ func (o *Handler) submitProjectLogs(ctx context.Context, projectLogs map[string]
 	for projectID, logRows := range projectLogs {
 		var hasBackendLogs bool
 		for _, logRow := range logRows {
-			if logRow.IsBackend() {
+			if logRow.Source == modelInputs.LogSourceBackend {
```
Vadman97, line 398: fine to do the check inline, but why not keep a helper method for the backend check?

Vadman97 (approved):

Vadman97:
with this PR, is it possible to also fix the error https://app.highlight.io/1/errors/BcMOkm6fhwU045BT6Kb0ZkS2319Z/instances/172080072

backend/otel/otel.go
et: Didn't really feel like it was needed since there's only two possible values for `Source` i.e. it feels like an unnecessary abstraction.

et:
> with this PR, is it possible to also fix the error https://app.highlight.io/1/errors/BcMOkm6fhwU045BT6Kb0ZkS2319Z/instances/172080072

Per slack, this is related to using the `useProjectId` on the integration page. @ccschmitz is going to handle that.

PR #4845, opened 2023-04-05
replace nestjs error filter with an interceptor

Vadman97:
## Summary

The nestjs `HighlightErrorFilter` would alter exceptions raised by applications.
This meant that exceptions that should return a body with a 400 exception would
be replaced by a 500 response coming from highlight.
The `NestInterceptor` interface allows us to observe exceptions raised by api handlers
without altering the response. This is a better fit for the highlight nestjs sdk to record errors. 

Screenshot 2023-04-05 at 12 59 52 PM

Screenshot 2023-04-05 at 1 01 23 PM

example usage
image

New nestjs sdk version published.


ccschmitz (approved):
Nice!

PR #4848, opened 2023-04-06
Import from react-router-dom instead of react-router

lewisl9029:
## Summary

_This is part of a [series](https://github.com/highlight/highlight/pull/4813) of PRs being spun off from [my WIP branch](https://github.com/lewisl9029/highlight/pull/2) to get the Highlight web app ready for [Reflame](https://reflame.app/). Hopefully this makes things a bit easier to review, test, and merge. 🙂_ 

There were several places in the frontend codebase where we were importing from `react-router`, a transitive dependency, instead of `react-router-dom`, a direct dependency.

This is often referred to as a phantom dependency, and can cause a number of issues, both in theory and practice:

- We have a rule against importing `useParams` from `react-router-dom`, but that does nothing to protect against importing `useParams` from `react-router`. In fact, this was something I [had to fix](https://github.com/highlight/highlight/commit/38a02f4ebcab29c1b8df368b864488b7bf736daf#diff-5fd6e69a538cbc878adc5b71eb5d9a6d68cd53d6588689284f4ecd9506343681L49) as part of this PR.

- Since the versions of transitive deps are not specified explicitly, they can easily change under our feet without us noticing and potentially cause us to import a different version than we were expecting, or can even inexplicably disappear altogether when seemingly unrelated deps change. The potential for spooky actions at a distance is greatly exacerbated in a large monorepo like we have here. The Rush.js folks did a great [writeup](https://rushjs.io/pages/advanced/phantom_deps/) on the perils of phathom dependencies, so I won't rehash all the details. 

- It's incompatible with stricter package management schemes like pnpm (and the one used by [Reflame](https://reflame.app/) itself, which was admittedly my primary motivation for this PR 😅) that don't expose non-direct dependencies to the resolution algorithm to begin with, specifically to combat the phantom dependency problem.

All I did to fix this was to search & replace all references of `'react-router'` to `'react-router-dom'`. And to prevent this specific issue from happening again, I added `react-router` to our existing list of restricted imports in eslint. For a more thorough defense against phantom deps, we'll need to switch to something like pnpm, npm's new [`linked` install strategy](https://github.com/npm/cli/pull/6078), or possibly yarn's [pnpm nodeLinker option](https://github.com/yarnpkg/berry/pull/3338) for a less disruptive migration (though I have no idea if it does what I think it does).

I ran `yarn turbo run dev --filter frontend...` locally and poked around the app.

None that I can think of.

### DIFF
--- a/frontend/.eslintrc
+++ b/frontend/.eslintrc
@@ -38,7 +38,10 @@
 		"@typescript-eslint/no-explicit-any": "off",
 		"@typescript-eslint/no-unused-vars": [
 			"error",
-			{ "argsIgnorePattern": "^_", "varsIgnorePattern": "^_" }
+			{
+				"argsIgnorePattern": "^_",
+				"varsIgnorePattern": "^_"
+			}
 		],
 		"unused-imports/no-unused-imports-ts": "error",
 		"@typescript-eslint/no-non-null-assertion": "off",
@@ -103,6 +106,10 @@
 						"name": "antd",
 						"importNames": ["InputNumber"],
 						"message": "Please use InputNumber from /src/components/InputNumber instead."
+					},
+					{
+						"name": "react-router",
+						"message": "Please use `react-router-dom` instead."
 					},
 					{
 						"name": "react-router-dom",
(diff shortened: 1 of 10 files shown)


frontend/.eslintrc
```diff
@@ -104,6 +107,10 @@
 						"importNames": ["InputNumber"],
 						"message": "Please use InputNumber from /src/components/InputNumber instead."
 					},
+					{
+						"name": "react-router",
+						"message": "Please use `react-router-dom` instead."
+					},
```
Vadman97, line 113: great to prevent this going forward with an eslint rule!

Vadman97 (approved):

Issue #4616, opened 2023-03-15
incremental snapshot before full snapshot should not break session processing


mayberryzane:
can discard these incremental snapshot events, should also see if there's a reason this error is happening frequently

PR #4850, opened 2023-04-06
#4616 ignore incremental snapshots before full snapshot

mayberryzane:
## Summary
- remove `FirstEventTimestamp` from the accumulator and use `FirstFullSnapshotTimestamp` instead
- skip incremental events if no full snapshot has occurred

- tested the happy path locally to make sure existing sessions weren't broken, will monitor prod for this error after deploying

- no, can revert if this is causing issues

### DIFF
--- a/backend/worker/worker_test.go
+++ b/backend/worker/worker_test.go
@@ -118,7 +118,7 @@ func TestCalculateSessionLength(t *testing.T) {
 }
 
 func TestGetActiveDuration(t *testing.T) {
-	zeroTime := time.Date(1, 1, 1, 0, 0, 0, 0, time.UTC)
+	zeroTime := time.Time{}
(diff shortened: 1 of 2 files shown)


backend/worker/worker_test.go
```diff
@@ -705,115 +709,3 @@ func TestGetActiveDuration(t *testing.T) {
 		})
 	}
 }
-
-func TestFullSnapshotValidation(t *testing.T) {
```
Vadman97, line 709: could we replace this with a test that has an incremental snapshot before a full snapshot, and see that we can process the session while just dropping the data before the full snapshot?

Vadman97 (approved):
thanks! this is great for helping us process sessions with extra incremental data that we can just ignore.

PR #4854, opened 2023-04-06
use log row constructor everywhere

et:
## Summary

We have a lot of data normalization when using the log row constructor (`NewLogRow`) and it's options (e.g. `WithSeverityText`). But some code calls the struct (`LogRow`) directly (specifically the tests) and that makes our code more brittle. 

Hence, this PR is doing a couple things:

* ensuring that all log rows are created with `NewLogRow`
* adjusted the constructor to require a timestamp and a project id (these are the bare minimum requirements for a log, everything else is optional and should use the `With*` functions)
* Got rid of `LogRowPrimaryAttrs` in favor of more `With*` functions (e.g. `WithTraceID`)
* Since timestamp is required, I got rid of `WithTimestamp` (the logic of truncating to the second is still present)
* Since project id is required, I got rid of `WithProjectIDString`. It wasn't doing error handling (see inline comment below).

Existing tests should pass and code should compile

N/A

### DIFF
--- a/backend/otel/otel.go
+++ b/backend/otel/otel.go
@@ -82,13 +82,13 @@ func setHighlightAttributes(attrs map[string]any, projectID, sessionID, requestI
 }
 
 func getLogRow(ctx context.Context, ts time.Time, lvl, projectID, sessionID, traceID, spanID string, excMessage string, resourceAttributes, spanAttributes, eventAttributes map[string]any, source modelInputs.LogSource) *clickhouse.LogRow {
+	projectIDInt, _ := clickhouse.ProjectToInt(projectID)
+
 	return clickhouse.NewLogRow(
-		clickhouse.LogRowPrimaryAttrs{
-			TraceId:         traceID,
-			SpanId:          spanID,
-			SecureSessionId: sessionID,
-		},
-		clickhouse.WithTimestamp(ts),
+		ts, uint32(projectIDInt),
+		clickhouse.WithTraceID(traceID),
+		clickhouse.WithSpanID(spanID),
+		clickhouse.WithSecureSessionID(sessionID),
 		clickhouse.WithBody(ctx, excMessage),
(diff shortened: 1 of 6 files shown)


backend/otel/otel.go
```diff
@@ -81,21 +81,24 @@ func setHighlightAttributes(attrs map[string]any, projectID, sessionID, requestI
...
-			SecureSessionId: sessionID,
-		},
-		clickhouse.WithTimestamp(ts),
+		ts, uint32(projectIDInt),
+		clickhouse.WithTraceID(traceID),
+		clickhouse.WithSpanID(spanID),
+		clickhouse.WithSecureSessionID(sessionID),
 		clickhouse.WithBody(ctx, excMessage),
 		clickhouse.WithLogAttributes(ctx, resourceAttributes, spanAttributes, eventAttributes, source == modelInputs.LogSourceFrontend),
-		clickhouse.WithProjectIDString(projectID),
```
et, line 94: Previously, when `WithProjectIDString` had an error, we should just not write the project id for a log. It makes no sense to store a log with a project id. Hence, that could have masked a bug.

See the code on the right ➡️ 
The code that could error was `ProjectToInt`. If there's an error in that code, we bubble it up, log an error message, and skip writing this log.

et: Looking into how `ProjectToInt` works, it appears it can error when someone passes in an invalid verbose project id:

https://github.com/highlight/highlight/blob/1cae712a59e123a3608d3701edaef72bb621aa3a/backend/clickhouse/log_row.go#L26-L30

Since the old code wouldn't actually set a project id for these logs, it defaults to 0 (clickhouse [defaults](https://clickhouse.com/docs/en/sql-reference/statements/create/table#default_values) to `0` for numeric columns).

I did a quick check on prod and it looks like there's a lot of logs with the bad data:

```sql
select count(*) from logs where ProjectId = 0;
=> 638693
```

Vadman97 (approved):
🏴‍☠️ 🚢

PR #4820, opened 2023-04-03
optimize docker deploy for development

Vadman97:
## Summary

Updates the dev workflow to only run the backend infrastructure in docker.
Adds new scripts in `docker` (`run.sh`, `start-infra.sh`, `run-backend.sh`, and `run-frontend.sh`) for spinning up the dev environment.
Updates documentation accordingly.

## How did you test this change?

CI `docker` step
```bash
cd docker; ./start-infra.sh
./run-backend.sh
# in another terminal
cd docker; ./run-frontend.sh
```

only infra running in docker with reduced memory footprint
Screenshot 2023-04-04 at 2 52 28 PM

new docs

image

opentelemetry now working
Screenshot 2023-04-05 at 12 42 57 PM

## Are there any deployment considerations?

No

PR #4855, opened 2023-04-06
fix ci check for opensearch migration

Vadman97:
## Summary

Make the check for migration success more strict so that it does not match irrelevant lines.
```bash
2023-04-06T04:19:40.4627731Z + grep -i 'error ' /tmp/highlightSetup.log
2023-04-06T04:19:40.4627833Z + grep -v lambda
2023-04-06T04:19:40.4628053Z go: downloading github.com/hashicorp/go-multierror v1.1.1
```
Improving the GitHub Action output as well to make it easier to triage the test.

Running in CI.
Testing the bad-path locally:
Screenshot 2023-04-05 at 9 50 08 PM

No

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -60,10 +60,13 @@ jobs:
 
                   yarn cy:run;
 
-            - name: Dump docker logs on failure
+            - name: Dump setup logs on failure
+              if: failure()
+              run: cat /tmp/highlight.log
+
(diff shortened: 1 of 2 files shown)


Vadman97:
merging to resolve test failure in `main`
for additional context - this didn't fail in #4820 CI because the `go mod download` output varies depending on the state of the go package cache. once in main, the cache was empty so the `go mod download` output is more verbose and started to match the grep pattern.

PR #4856, opened 2023-04-06
Logs polish 2.0

ccschmitz:
## Summary

* Raises matches log attributes to the top of results when not expanded
* Updates highlight of log body text to match the attribute matching styles
* Deselects the first item in the combobox when deleting the query
* Fixes an issue where applying a level filter from the histogram overwrote the entire query
* Updates CTAs for setting up backend logs/errors (not necessarily related to ticket)
* Updates the logs icon
* Add chevron on datepicker "Custom" option to indicate there is a sub menu
* Clean up menu heading styles
* Render quotes (`""`) for an empty attribute value

Local and PR preview click tests.

N/A - all client-side changes.


frontend/src/pages/LogsPage/LogsPage.tsx
```diff
@@ -91,7 +91,7 @@ const LogsPageInner = ({ timeMode, logCursor, startDateDefault }: Props) => {
 	}
 
 	const handleLevelChange = (level: LogLevel) => {
-		setQuery(`level:${String(level).toLowerCase()}`)
+		setQuery(`${query} level:${String(level).toLowerCase()}`)
```
et, line 94: Unrelated but `toLowerCase` is unnecessary:

https://github.com/highlight/highlight/blob/d4957a9e3bd0deb1dc45a00dfbc654463cdcff4f/backend/private-graph/graph/schema.graphqls#L296-L303

This code might have pre-dated when we finalized lowercasing everything.

et (approved):
Aweome work here! This really did clean up a lot of usability issues.

One thing I noticed that I believe is unrelated is that log attributes with capital letters are sorted above nested keys ([example](https://frontend-pr-4856.onrender.com/1/logs?query=ValueLength%3A8451))

Screenshot 2023-04-06 at 9 34 49 PM

Not sure if it's an easy fix to squeeze in. Happy to file a ticket as well.

ccschmitz:
> One thing I noticed that I believe is unrelated is that log attributes with capital letters are sorted above nested keys ([example](https://frontend-pr-4856.onrender.com/1/logs?query=ValueLength%3A8451))

@et thanks for calling that out. I don't think we're doing anything with sorting on the client, but we could. It looks like the attributes are coming back in that order from the server.

Screenshot 2023-04-07 at 6 56 33 AM

Do you think we should sort them alphabetically or something?

frontend/src/pages/LogsPage/LogsPage.tsx
ccschmitz: Just cleaned this up!

PR #4883, opened 2023-04-07
Create dedicated endpoint for loading session logs

et:
## Summary

This PR creates a dedicated endpoint for session logs. Unlike the log viewer that has to handle pagination, session logs are more well scoped. Specifically they:

* have a `secure_session_id` to filter on
* are [confined to a 4 hour window ](https://github.com/highlight/highlight/blob/et/session-logs-endpoint/frontend/src/pages/LogsPage/SearchForm/utils.ts#L126-L127)

Because it's more well scoped, it can avoid limitations of dealing with pagination (very unpredictable with virtualized lists).

Additionally, session logs do not need `LogAttributes` which is a very heavy column. A previous [observation](https://gist.github.com/et/bd25c44fd04b64bd76a0e35853e8cb3b) has noticed that omitting this column speeds up queries.

### Performance tests

Using a [30 minute session](https://app.highlight.io/1/sessions/6T5zO03EWSN4OBoblLfLBtTkxAQq), I generated this query on prod:

```sql
select Timestamp, SeverityText, Body from logs where SecureSessionId = '6T5zO03EWSN4OBoblLfLBtTkxAQq' and ProjectId = 1 and Timestamp >= '2023-04-06 20:28:57'  
```

where `2023-04-06 20:28:57` is the session start time (in UTC).

The above query returns 5,279 rows and takes 0.325s which is an acceptable duration to load all of this session's logs.

### Unit tests

I did add tests to confirm:

* we aren't respecting `LogLimit`
* we aren't loading `LogAttributes`

This is only the backend piece. A frontend PR will follow that will leverage this but we want this to be deployed first so we can test this endpoint with a render preview.

### DIFF
--- a/backend/clickhouse/logs.go
+++ b/backend/clickhouse/logs.go
@@ -166,6 +166,59 @@ func (client *Client) ReadLogs(ctx context.Context, projectID int, params modelI
 	return getLogsConnection(edges, pagination), rows.Err()
 }
 
+// This is a lighter weight version of the previous function for loading the minimal about of data for a session
+func (client *Client) ReadSessionLogs(ctx context.Context, projectID int, params modelInputs.LogsParamsInput) ([]*modelInputs.LogEdge, error) {
+	selectStr := "Timestamp, UUID, SeverityText, Body"
+	sb, err := makeSelectBuilder(selectStr, projectID, params, Pagination{}, OrderBackwardInverted, OrderForwardInverted)
+	if err != nil {
+		return nil, err
+	}
+
+	sql, args := sb.Build()
+
+	span, _ := tracer.StartSpanFromContext(ctx, "logs", tracer.ResourceName("ReadSessionLogs"))
+	query, err := sqlbuilder.ClickHouse.Interpolate(sql, args)
+	if err != nil {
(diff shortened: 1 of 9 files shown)


backend/clickhouse/logs.go
```diff
@@ -166,6 +166,59 @@ func (client *Client) ReadLogs(ctx context.Context, projectID int, params modelI
 	return getLogsConnection(edges, pagination), rows.Err()
 }
 
+// This is a lighter weight version of the previous function for loading the minimal about of data for a session
+func (client *Client) ReadSessionLogs(ctx context.Context, projectID int, params modelInputs.LogsParamsInput) ([]*modelInputs.LogEdge, error) {
```
et, line 170: The frontend is still responsible for constructing the query (e.g. `secure_session_id:foo`) ([example](https://github.com/highlight/highlight/pull/4833/files#diff-0f0e22bb410b4636bf449d61d3dc5d7faba340c3198f7baa2b284683774e7896R91)) because it'll need to do that anyways to construct the query for when we click the "Show in Log Viewer" button (see #4779 for requirements)

Vadman97 (approved):
nice!

PR #4900, opened 2023-04-09
Update InteractionOberserver Usage for iOS

jay-khatri:
## Summary

A few changes to our usage of interactionObserver. Before this PR, docs pages were broken on certain iOS devices.

Used https://saucelabs.com/resources/blog/how-to-inspect-element-on-iphone to inspect element on Safari for my iphone.

n/a

### DIFF
--- a/highlight.io/pages/docs/[[...doc]].tsx
+++ b/highlight.io/pages/docs/[[...doc]].tsx
@@ -118,7 +118,7 @@ const useIntersectionObserver = (setActiveId: (s: string) => void) => {
 
 		const observer = new IntersectionObserver(callback, {
 			root: document.body,
-			rootMargin: '0px',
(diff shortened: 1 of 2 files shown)


highlight.io/pages/docs/[[...doc]].tsx
```diff
@@ -117,8 +117,9 @@ const useIntersectionObserver = (setActiveId: (s: string) => void) => {
 		}
 
 		const observer = new IntersectionObserver(callback, {
-			rootMargin: ' -5% 0px -90% 0px',
-			threshold: 0.9,
+			root: document.body,
```
Comment (Crawron) [inline highlight.io/pages/docs/[[...doc]].tsx:120]: adding `document.body` as the root may break the sidebar on desktop. It can be completely omitted from the options instead

Does the right sidebar still work on desktop?

Crawron (approved):
yup, this looks all good!

PR #4901, opened 2023-04-09
Remove root from `root` from intersection observer calls

jay-khatri:
## Summary
Removes `root` from intersectionObserver calls because it is either:
- redundant
- problematic (causes issues)

clicktest

n/a


Crawron (approved):
Looks good!

PR #4918, opened 2023-04-11
Convert errorBody display to details and summary

shayneo:
## Summary

Converts the ErrorState component from a clickable span to `<details><summary/></details>` [elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details), which handles showing and hiding detailed information natively.

Tested with auth error. Validated in Chrome, Firefox, Safari.

### Before
Screenshot 2023-04-10 at 9 12 04 PM

Screenshot 2023-04-10 at 9 12 10 PM

### After
Screenshot 2023-04-10 at 9 12 50 PM

Screenshot 2023-04-10 at 9 12 58 PM

nope.

### DIFF
--- a/frontend/src/components/ErrorState/ErrorState.tsx
+++ b/frontend/src/components/ErrorState/ErrorState.tsx
@@ -48,7 +48,7 @@ export const ErrorState = ({
 				</p>
 				{errorString !== undefined && (
 					<details onToggle={() => setShowError((t) => !t)}>
-						<summary className="cursor-pointer">
+						<summary className="cursor-pointer text-gray-500">
 							{showError ? 'show less' : 'show more'}
 						</summary>
 						{showError && (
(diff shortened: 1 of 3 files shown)


frontend/src/components/ErrorState/ErrorState.tsx
```diff
@@ -45,18 +45,20 @@ export const ErrorState = ({
...
-							onClick={() => setShowError((t) => !t)}
-						>
-							{showError ? 'show less' : 'show more'}
-						</span>
-					)}
 				</p>
-				{showError && (
-					<code className={styles.errorBody}>{errorString}</code>
+				{errorString !== undefined && (
+					<details onToggle={() => setShowError((t) => !t)}>
```
Vadman97, line 50: could we also style the `>` icon and the `show more` text with the same gray font as the text above?

Vadman97 (approved):
thanks for catching this!

Vadman97:
e2e test is broken in forks right now - i have a fix for that in another pr but that shouldn't matter for this frontend-only change  (monorepo test is the one for that)

Issue #4735, opened 2023-03-28
For links without a project, they link to the first project by default


jay-khatri:
Something we noticed in James Quick stream.

For links without a project, (e.g. app.highlight.io/setup & app.highlight.io/alerts), we default to going to the first project that the user is in. Instead, can we default to the most recent project? We could do this by storing the most recent project in the Project Router, and reading from that variable when going to that link.

ccschmitz:
@jay-khatri do you happen to recall which links weren't working? Wondering if https://github.com/highlight/highlight/commit/420975ec7442aa87b7423ca5d35f79c35c080488 resolves this.

jay-khatri:
Yes. It was just the "Finish Setup" Cta in the center of the empty state for sessions and errors.

Issue #4873, opened 2023-04-07
Setup Page Polish


jay-khatri:
https://www.loom.com/share/33f1d70043724868b3595eefa0286b8c

- [ ] read the docs” link
- [ ] scrollbar overlaps with right section
- [ ] copy icon in the top right
- [ ] Rename “UX Monitoring” to Frontend Monitoring + Session Replay

jay-khatri:
Just added a last bullet here given that a new account is considered onboarded by default.

ccschmitz:
@highlight/jay I just incorporated your feedback from our 1-1 doc here.

ccschmitz:
@jay-khatri is there already an issue tracking the issue with shared sessions/errors not working?

ccschmitz:
@jay-khatri I broke that out to a separate issue: https://github.com/highlight/highlight/issues/4932

PR #4922, opened 2023-04-11
Setup page revisions and bug fixes

ccschmitz:
## Summary

This PR contains a number of small revisions and bug fixes for setup/onboarding.

* Changes the scrollable area of the setup docs page to not include the integration bar
* Renames "UX monitoring" to "Session replay / frontend monitoring"
* Updates some links from `/setup` to `/${projectId}/setup` (resolves https://github.com/highlight/highlight/issues/4735)
* Fixes the copy button being rendered in the top right of the page instead of inside code blocks
* Fixes an issue with the logic for retrieving the integration status for frontend, backend, and logging
* Changes the default behavior of the left panel to be open instead of closed on sessions and errors
* Only queries sessions and errors from the last 30 days in integration checks to avoid querying data outside the retention period

See more context in https://github.com/highlight/highlight/issues/4873.

Local click testing.

Will click test in production and watch some sessions of folks onboarding to ensure these issues ares resolved. I'll also be going through some SDK setup tomorrow and keeping an eye on things throughout that process.


backend/private-graph/graph/schema.resolvers.go
```diff
@@ -4619,9 +4619,12 @@ func (r *queryResolver) ClientIntegration(ctx context.Context, projectID int) (*
 	if err != nil {
 		return integration, e.Wrap(err, "error querying session for project")
 	}
-	integration.Integrated = true
-	integration.ResourceSecureID = &firstSession.SecureID
-	integration.CreatedAt = &firstSession.CreatedAt
+
+	if firstSession.ID != 0 {
```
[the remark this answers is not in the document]
Vadman97: i see, sorry for the hassle. so changing the `.First(` to a `.Find(` means that the `gorm.ErrRecordNotFound` is never returned, even if there are no results. that makes the zero-value check a safer one since you don't have to worry about the gorm internals.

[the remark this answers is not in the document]
et: Thanks for confirming. I do see in the [docs](https://gorm.io/docs/query.html) this callout:

> If you want to avoid the ErrRecordNotFound error, you could use Find like db.Limit(1).Find(&user), the Find method accepts both struct and slice data

Was hoping there was a way to support `ErrRecordNotFound` using this method but I guess this is the expected behavior.

I've opened up #4942 for our Q2 planning to track any GORM guardrails that can be put in place.

@ccschmitz - per this thread, we can remove the `if e.Is(err, gorm.ErrRecordNotFound) {` checks for clarity.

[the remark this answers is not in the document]
ccschmitz: Whoops, I removed it in some of the checks but not all - my bad. Will clean this up. Thanks for catching.

backend/private-graph/graph/schema.resolvers.go
```diff
@@ -4612,16 +4612,16 @@ func (r *queryResolver) ClientIntegration(ctx context.Context, projectID int) (*
 	}
 
 	firstSession := model.Session{}
-	err := r.DB.Model(&model.Session{}).Where("project_id = ?", projectID).Limit(1).Find(&firstSession).Error
-	if e.Is(err, gorm.ErrRecordNotFound) {
-		return integration, nil
-	}
+	err := r.DB.Model(&model.Session{}).Where("project_id = ?", projectID).Where("created_at > ?", time.Now().Add(time.Hour*24*-30)).Limit(1).Find(&firstSession).Error
```
ccschmitz, line 4615: @et ended up adding this additional `where` clause to ensure we don't query sessions or errors outside the retention period. That's the reason the session/error wasn't loading for you. I also updated the CTA text to say "View an error" and "View a session" since there's a chance it won't be the first.

et: Instead of hardcoding the time frame, can we change to actually use the retention date function:

https://github.com/highlight/highlight/blob/2614e6605eee217754bca2cb8ecf4e4f3ee47f1c/backend/private-graph/graph/schema.resolvers.go#L3773-L3782

ccschmitz: Yeah, I thought about doing that, but this query could potentially get hit so frequently I figured I'd avoid the extra query or join and just hard-code the 30 days.

et: @ccschmitz and I chatted about this over slack, particularly if this was going to be too brittle (e.g. 60 day retention but last session was 31 days ago). But in that scenario, we should consider setup as not integrated and advise the user to check their setup.

et (approved):
🚢

PR #4938, opened 2023-04-11
use rrweb events to record session properties

Vadman97:
## Summary

Updates the client-side logic for processing `track` events.
Addresses a customer issue where a session had no timeline indicators for track events despire `H.track` being called.
The reason was that we would pass these events as `addSessionProperties` calls which would update
the opensearch entries for the session and make it searchable, but would not add the events to the session timeline.

Local deploy using buttons page to call both `analytics.track` and `h.track`.

Before:

`analytics.track` events would be recorded, but only because they would pass thru the segment integration. pure `H.track` calls were not recorded even if the session was searchable by the event.

After:
 
Screenshot 2023-04-11 at 5 29 22 PM

Segment events still appear.

Screenshot 2023-04-11 at 5 29 43 PM

other instrumentation of frontend still appears.

Screenshot 2023-04-11 at 5 29 55 PM

Enhancing cypress test to check for `Track` events.

New highlight.run SDK released.


ccschmitz (approved):
Looks good, Vadim. Thanks for cleaning this up and updating the tests!

Issue #4932, opened 2023-04-11
Figure out why the share links and demo error/session links aren't working


ccschmitz:
Fairly confident this is because share session/error page links are not working in general. Need to figure out why this is happening.

PR #4943, opened 2023-04-12
public sessions fixup - remove auth check for retention period

mayberryzane:
## Summary
- this method is only called from other resolvers that are already doing their own auth checks
- this broke making a session public

- made a session public locally, validated I was able to view it incognito

- no


ccschmitz:
Resolves #4932

ccschmitz (approved):
LGTM - thank you, @mayberryzane!

PR #4950, opened 2023-04-12
Revert "Resolve @highlight-run/client without relative paths"

Vadman97:
Reverts highlight/highlight#4851

The changes to firstload -> client imports break types checking in packages using highlight.run after #4851
```
[1] 11:03:22 AM - Found 1 error. Watching for file changes.
[2] node_modules/highlight.run/dist/firstload/src/index.d.ts(1,65): error TS2307: Cannot find module '@highlight-run/client/src/types/types' or its corresponding type declarations.
[2] wait-on http://127.0.0.1:3000 && tsc -p electron && electron . exited with code 1
```
Looks like that's because our npm.js package packages the typescript definitions with their imports from the client package, which is not a public package.
```
 ~/w/react-typescript-electron-sample-with-create-react-app-and-electron-builder   master ±  cat node_modules/highlight.run/dist/firstload/src/index.d.ts
import type { HighlightOptions, HighlightPublicInterface } from '@highlight-run/client/src/types/types';
import HighlightSegmentMiddleware from './integrations/segment';
import configureElectronHighlight from './environments/electron';
export declare enum MetricCategory {
    Device = "Device",
    WebVital = "WebVital",
    Frontend = "Frontend",
    Backend = "Backend"
}
export declare const H: HighlightPublicInterface;
export { HighlightSegmentMiddleware, configureElectronHighlight };
export type { HighlightOptions };
```

To fix this, we need some way to tell firstload typescript to bundle the client types in firstload and reference the local imports.

cc @lewisl9029

### DIFF
--- a/sdk/firstload/CHANGELOG.md
+++ b/sdk/firstload/CHANGELOG.md
@@ -157,6 +157,12 @@ Ensures H.stop() stops recording and that visibility events do not restart recor
 
 ## 6.0.1
 
-### Minor Changes
+### Patch Changes
 
 - Fixes `H.track` reporting to ensure events are recorded as part of the session timeline indicators.
+
+## 6.0.2
+
+### Patch Changes
+
+- Fixes typescript definitions for `highlight.run` which referenced an internal unpublished package.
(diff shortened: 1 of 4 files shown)


Vadman97:
merged per https://highlightcorp.slack.com/archives/C04DVJJ0ZJ6/p1681323120285779

PR #4952, opened 2023-04-12
roadmap error checking

Abhishek-More:
## Summary

Added some checking for nonexistent data in the roadmap fetcher.

Also, made changes such that getStaticProps only calls roadmapFetcher for roadmap path, rather than all docs paths.


jay-khatri (approved):

Vadman97:
thanks @Abhishek-More !

PR #4958, opened 2023-04-12
Resolving highlight client from firstload - attempt 2

lewisl9029:
## Summary

This is another attempt at https://github.com/highlight/highlight/pull/4851, which had to be [reverted](https://github.com/highlight/highlight/pull/4950) because it broke type declarations. 😞 

TypeScript isn't planning to officially support import path rewriting in declarations (see https://github.com/microsoft/TypeScript/issues/31643#issuecomment-496995677), so to fully support arbitrary path remappings (like the ones to `@highlight-run/client`), the most promising solution I could find involved a combination of [ttypescript](https://github.com/cevek/ttypescript), [ts-transform-paths](https://github.com/zerkalica/zerollup/tree/master/packages/ts-transform-paths), and [rollup-plugin-typescript2](https://github.com/ezolenko/rollup-plugin-typescript2).

However in our case, we're only looking to remap a single reference, so a dumb script that looks for the reference and replaces them feels like a much simpler & lower risk solution. That's what I added in this PR in addition to the previously reverted changes.

Let me know if y'all would prefer looking into the other approach instead, which might be worthwhile if we wanted to start using arbitrary path remappings in libraries (say, to be able use absolute imports everywhere).

I ran `yarn build && yarn typegen` in both main and this branch, and ran `diff -rq` to compare both dist folders. There were no differences in any of the .d.ts files (indicating the import replacement script gave us the correct output), and the only differences were in `index.js`, `indexESM.js`, and their respective importmaps, due to the changes in the package.json file that ends up getting bundled.

Screenshot 2023-04-12 at 4 43 37 PM

This time I did make sure to include type declarations in the build output comparison, so I have pretty high confidence they'll behave identically this time.

Though one thing I couldn't confirm was whether the CI process that publishes this module runs the `yarn typegen` command or `yarn tsc` directly. The latter could produce the same problematic output as before since we depend on the script specified in `yarn typegen` to rewrite imports. Would appreciate it if someone could double check. 🙏

### DIFF
--- a/sdk/firstload/src/integrations/amplitude.ts
+++ b/sdk/firstload/src/integrations/amplitude.ts
@@ -1,6 +1,6 @@
 // @ts-nocheck
-import type { AmplitudeIntegrationOptions } from '../../../client/src/types/client'
-import type { Integration } from '../../../client/src/types/types'
+import type { AmplitudeIntegrationOptions } from '@highlight-run/client/src/types/client'
+import type { Integration } from '@highlight-run/client/src/types/types'
 
 interface Window {
 	amplitude?: AmplitudeAPI
--- a/sdk/firstload/src/integrations/mixpanel.ts
+++ b/sdk/firstload/src/integrations/mixpanel.ts
@@ -1,6 +1,6 @@
 // @ts-nocheck
-import type { MixpanelIntegrationOptions } from '../../../client/src/types/client'
-import type { Integration } from '../../../client/src/types/types'
+import type { MixpanelIntegrationOptions } from '@highlight-run/client/src/types/client'
+import type { Integration } from '@highlight-run/client/src/types/types'
 
 interface Window {
 	mixpanel?: MixpanelAPI
--- a/sdk/firstload/package.json
+++ b/sdk/firstload/package.json
@@ -6,7 +6,7 @@
 		"dev": "rollup -c -w",
 		"enforce-size": "size-limit",
 		"test": "vitest --run",
-		"typegen": "tsc"
+		"typegen": "tsc && node scripts/replace-client-imports.mjs"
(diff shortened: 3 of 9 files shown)


sdk/firstload/src/integrations/amplitude.ts
```diff
@@ -1,6 +1,6 @@
 // @ts-nocheck
```
Vadman97, line 1: not sure why we have `@ts-nocheck` here but does your change mean we can remove the override?

sdk/firstload/src/integrations/mixpanel.ts
```diff
@@ -1,6 +1,6 @@
 // @ts-nocheck
```
Vadman97, line 1: same

Vadman97 (approved):
> Though one thing I couldn't confirm was whether the CI process that publishes this module runs the yarn typegen command or yarn tsc directly. The latter could produce the same problematic output as before since we depend on the script specified in yarn typegen to rewrite imports. Would appreciate it if someone could double check.

since we use turborepo, we will first run `typegen` here and then `build` as part of the SDK build, never `tsc` directly, so this should be fine!

sdk/firstload/src/integrations/amplitude.ts
lewisl9029: Doesn't look like it. When I removed the `@ts-nocheck`, 49 errors end up showing up when running tsc.

It looks like these were mostly here to silence type errors from what looks like copy pasted init scripts?

Vadman97: ah gotcha, this is for the file rather than the import. thanks!

PR #4979, opened 2023-04-13
Ensure otel logs are processed with the correct source

Vadman97:
## Summary

Frontend logs passed through our otel ingest get a special attribute value set to indicate they are frontend.
Other SDKs may emit logs with no particular source, so we should use the default log source of `backend`.

New unit test for the otel protobuf parsing and otel processing to ensure correct sources are found.

New SDK version to be published.


backend/clickhouse/log_row.go
```diff
@@ -105,6 +105,9 @@ func WithSeverityText(severityText string) LogRowOption {
 func WithSource(source modelInputs.LogSource) LogRowOption {
 	return func(l *LogRow) {
 		l.Source = source
+		if l.Source == "" {
```
et, line 108: What do you think about making this:

```go
func WithSource(source modelInputs.LogSource) LogRowOption {
	return func(l *LogRow) {
		if l.Source == modelInputs.LogSourceFrontend {
      l.Source = modelInputs.LogSourceFrontend
		} else {
      l.Source = modelInputs.LogSourceBackend
		} 
	}
}
```

Per this [comment](https://github.com/highlight/highlight/pull/4839/files#r1158714695), we don't have a true union type but this type of logic would us basically there.

My thoughts are that if anyone passes "source = foo", we should just ignore it and default to "backend"

backend/otel/otel_test.go
```diff
@@ -25,8 +49,52 @@ func TestHandler_HandleLog(t *testing.T) {
 	h.HandleLog(w, r)
 }
 func TestHandler_HandleTrace(t *testing.T) {
+	inputBytes, err := os.ReadFile("./samples/traces.json")
```
et, line 52: Out of curiosity, how did you capture this data? Did you use something to sniff out network traffic?

backend/vercel/vercel.go
```diff
@@ -285,7 +285,9 @@ func HandleLog(w http.ResponseWriter, r *http.Request) {
 			http.Error(w, err.Error(), http.StatusBadRequest)
 			return
 		}
-		logs = append(logs, l)
+		if l.Message != "" {
```
et, line 288: Do we know why this was actually happening?

What I'm trying to figure out is whether a log without a message is actually valid (e.g. one only wanted to use fields).

I did confirm that logrus won't output anything without a message, e.g. this does nothing:

```go
import (
)
log.WithContext(ctx).WithField("foo", "bar")
```

Mostly wondering if you were able to replicate this and why vercel was sending us logs without a message in the first place.

et (approved):
Great work! Thanks for fixing this.

backend/vercel/vercel.go
et: Oh wait...I can do this:

```go
log.WithContext(ctx).WithField("foo", "bar").Info()
```

Screenshot 2023-04-14 at 9 24 40 AM

et: @Vadman97 - do you think we could bring this up with a team on whether or not we consider these types of logs to be valid?

backend/clickhouse/log_row.go
Vadman97: sure!

backend/otel/otel_test.go
Vadman97: i intercepted some incoming otel traces in `otel.go` like so

Screenshot 2023-04-14 at 1 14 46 PM

backend/vercel/vercel.go
Vadman97: Hmm, there might be some log drain data that has no message body. maybe other events like a deployment build workflow sends empty log messages populating other metadata fields? https://vercel.com/docs/concepts/observability/log-drains-overview/log-drains-reference#format-and-transport

I'll shoot them an email about this but in the meantime, I suggest that we merge this with a specific vercel warning about the empty messages we'd be dropping.

PR #4985, opened 2023-04-14
Many doc updates

et:
## Summary

I took a pass at various pages of the docs and cleaned up lots of grammar and inconsistent wording. I did find a bug when linking the ruby docs as well another bug when linking the Go SDK docs.

Visual test

N/A

### DIFF
--- a/highlight.io/components/QuickstartContent/self-host/shared-snippets.tsx
+++ b/highlight.io/components/QuickstartContent/self-host/shared-snippets.tsx
@@ -1,4 +1,3 @@
-import { siteUrl } from '../../../utils/urls'
 import { QuickStartStep } from '../QuickstartContent'
 
 export const dependencies: QuickStartStep = {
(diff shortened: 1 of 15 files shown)


highlight.io/components/QuickstartContent/self-host/shared-snippets.tsx
```diff
@@ -1,4 +1,3 @@
-import { siteUrl } from '../../../utils/urls'
```
et, line 1: This was an unused import that my editor clean up. Would be nice to add some eslint rules around that like we have in the frontend app:

https://github.com/highlight/highlight/blob/ebb0f1e751caa97b0a3399cfd2fc9615e196e916/frontend/.eslintrc#L46

Vadman97 (approved):
thank you! all great improvements :)

highlight.io/components/QuickstartContent/self-host/shared-snippets.tsx
Vadman97: could we share the `.eslintrc` among all packages in the monorepo?

jay-khatri (approved):

highlight.io/components/QuickstartContent/self-host/shared-snippets.tsx
et: Probably yes. Would likely require some clean up. I've added that as a hack week idea (#4862).

PR #4990, opened 2023-04-14
Enforce billing quotas for errors / logs on ingest

### LINKED ISSUE #4406: quota for errors on free tier

mayberryzane:
## Summary
- use clickhouse materialized view to aggregate log counts per project per day 
- add billing notification emails when exceeding usage thresholds for errors + logs
- genericize a few methods to check quotas for sessions or errors or logs
- add quota check in `HandleErrorAndGroup` to skip if quota exceeded
- add quota check in logs ingest to skip if quota exceeded (uses redis for performance)

- testing is a WIP

- no, can revert

Closes #4406

### DIFF
--- a/backend/clickhouse/migrations/000010_log_count_daily_mv.up.sql
+++ b/backend/clickhouse/migrations/000010_log_count_daily_mv.up.sql
@@ -1,0 +1,11 @@
+CREATE MATERIALIZED VIEW IF NOT EXISTS log_count_daily_mv
+ENGINE = SummingMergeTree
+PARTITION BY toDate(Day) ORDER BY (ProjectId, Day)
+POPULATE
+AS 
+SELECT
+    ProjectId,
+    toStartOfDay(Timestamp) AS Day,
+    count() as Count
+FROM logs
+GROUP BY ProjectId, Day;
--- a/backend/email/email.go
+++ b/backend/email/email.go
@@ -9,10 +9,14 @@ import (
 	"time"
 
 	e "github.com/pkg/errors"
+	"golang.org/x/text/language"
+	"golang.org/x/text/message"
 
 	"github.com/sendgrid/sendgrid-go"
 	"github.com/sendgrid/sendgrid-go/helpers/mail"
 	log "github.com/sirupsen/logrus"
+
+	modelInputs "github.com/highlight-run/highlight/backend/private-graph/graph/model"
 )
 
 var (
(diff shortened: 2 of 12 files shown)


backend/clickhouse/migrations/000010_log_count_daily_mv.up.sql
```diff
@@ -0,0 +1,11 @@
+CREATE MATERIALIZED VIEW IF NOT EXISTS log_count_daily_mv
+ENGINE = SummingMergeTree
```
et, line 2: Per our conversation earlier, why not AggregatingMergeTree?

backend/model/model.go
Vadman97, line 1: is `SendBillingNotifications` atomic? 
```
	// Skip sending email if sending was attempted within the cache TTL
	cacheKey := fmt.Sprintf("%s;%d", emailType, workspace.ID)
	_, exists := emailHistoryCache.Get(cacheKey)
	if exists {
		return nil
	}
	emailHistoryCache.Set(cacheKey, true)
```
i feel like there could be a race here where more than one kafka goroutine will trigger the notification since they can both read `false` simultaneously, then set the flag to `true`.

could we use redis as a lock? https://github.com/go-redsync/redsync

mayberryzane: it should be atomic, enforced by a unique constraint on `billing_email_histories` https://github.com/highlight/highlight/blob/main/backend/model/model.go#L1810

mayberryzane: cache is there just to limit having to check the db so frequently

Vadman97: oooh ok, makes sense!

backend/public-graph/graph/resolver.go
```diff
@@ -834,6 +835,25 @@ func (r *Resolver) HandleErrorAndGroup(ctx context.Context, errorObj *model.Erro
...
+				if err := model.SendBillingNotifications(ctx, r.DB, r.MailClient, email.BillingErrorsUsage100Percent, workspace); err != nil {
+					log.WithContext(ctx).Error(e.Wrap(err, "failed to send billing notifications"))
+				}
+			} else if quotaPercent >= .8 {
+				if err := model.SendBillingNotifications(ctx, r.DB, r.MailClient, email.BillingErrorsUsage80Percent, workspace); err != nil {
+					log.WithContext(ctx).Error(e.Wrap(err, "failed to send billing notifications"))
+				}
+			}
+		}
+	}()
```
Vadman97, line 852: would be nice to share this logic with the logs overage notifications logic (especially things like the `0.8` and `1` thresholds)

backend/email/email.go
```diff
@@ -148,16 +216,16 @@ func SendBillingNotificationEmail(ctx context.Context, mailClient *sendgrid.Clie
...
-	// 	estr := "error sending sendgrid email -> "
-	// 	estr += fmt.Sprintf("resp-code: %v; ", resp)
-	// 	if sendGridErr != nil {
-	// 		estr += fmt.Sprintf("err: %v", sendGridErr.Error())
-	// 	}
-	// 	return e.New(estr)
-	// }
+		Info("BILLING_NOTIFICATION email")
+
+	if resp, sendGridErr := mailClient.Send(m); sendGridErr != nil || resp.StatusCode >= 300 {
```
Vadman97, line 221: nice :)

Vadman97 (approved):
🚢 pending eric's point on the proper MV schema

backend/clickhouse/migrations/000010_log_count_daily_mv.up.sql
mayberryzane: Documentation was sparse here but from what I could tell, SummingMergeTree is a special case supporting only `sum` aggregations and may be more performant for this use case

Vadman97:
bump on this - is this waiting on anything else?

Issue #4933, opened 2023-04-11
next.js sdk should export `H` object from `highlight.run`


Vadman97:
highlight-run/next already depends on highlight.run
in our next.js docs, we tell folks to import `H` in their next.js frontend from `highlight.run`, meaning that they need to depend on two highlight SDKs.
https://github.com/highlight/highlight/blob/main/highlight.io/pages/_app.tsx

the highlight-run/next sdk already depends on highlight.run so we should just reexport the object

eightants:
After diving a little deeper with Vadim, it seems that exporting H from highlight/next is more difficult than expected. Since that means now we're importing highlight/next on the browser, the other dependencies (highlight/node) is throwing errors since OpenTelemetry uses node only functionality.

vasco3:
I just ran into this issue when I updated to ` "@highlight-run/next": "^7.2.1",` 

doesn't work with either 
`import { H } from "@highlight-run/next/client";`  
or `import { H } from "highlight-run";`

bloomfieldj:
> I just ran into this issue when I updated to ` "@highlight-run/next": "^7.2.1",` and "next": "^14.0.2"
> 
> doesn't work with either `import { H } from "@highlight-run/next/client";` or `import { H } from "highlight-run";`
> 
> Even removing the import causes the same error.

For now, I've found reverting to next@14.0.1 solves this for me.

PR #4996, opened 2023-04-16
Export H from next sdk

eightants:
## Summary

https://github.com/highlight/highlight/issues/4933

Testing by running e2e and checking session. 

It looks like the OpenTelemetry Tracer function from the dependency on highlight-node is causing an error as it's trying to import the highlight-next package on the client, but that's a server side function.

`Module not found: Can't resolve 'async_hooks'`

Need to find a way to not import that functionality on the client

NA


Vadman97 (approved):
nice, could you also bump the `sdk/next/package.json` version so that a new npm package is released? also, looks like build failures might be related?

sdk/highlight-node/src/index.ts
```diff
@@ -2,3 +2,4 @@ export { H, HIGHLIGHT_REQUEST_HEADER } from './sdk'
 export { Highlight } from './client'
 export { NodeOptions } from './types'
 export * as Handlers from './handlers'
+export { H as ClientH } from 'highlight.run'
```
Vadman97, line 5: i dont think we want node exporting the client, just `next`. reason for that is the `node` sdk is backend-only, whereas the `next` sdk is full-stack (and applies both to backend nodejs and the browser). sorry for missing that initially!

eightants:
After diving a little deeper with Vadim, it seems that exporting H from highlight/next is more difficult than expected. Since that means now we're importing highlight/next on the browser, the other dependencies (highlight/node) is throwing errors since OpenTelemetry uses node only functionality.

PR #4999, opened 2023-04-17
Suppoprt table markdown in docs

### LINKED ISSUE #4742: Support tables in our docs files
Right now, we don't support rendering tables in our docs. Lets do that.

In order to support it, we'll have to turn on the "GFM" plugin or GitHub Flavored Markdown, which supports tables. We already use a markdown library called `MDXRemote` which supports this plugin.

https://mdxjs.com/guides/gfm/

```[tasklist]
- [ ] Enable the plugin so that tables get compiled down to html table components
- [ ] Style the table according to @julian-highlight 's designs. 
```
cc @et

Crawron:
## Summary

Renders markdown tables in doc pages using `remark-gfm`. There is a test document at `/docs/general/test-table` (I should remove that before merging). fix #4742

image

Created a test page at `/docs/general/test-table`


Crawron:
> nice! want to remove the test table and merge?

yup! got it

et:
Thanks for adding this @Crawron. Using this already in #5080. Styling looks great!

PR #5002, opened 2023-04-17
Wrapping text in the Errors header

deltaepsilon:
This text could overflow in ugly ways. It's not common, but it shouldn't be breaking out of its containing box.

image

### DIFF
--- a/frontend/src/pages/ErrorsV2/ErrorBody/components/ErrorOccurenceDate/index.tsx
+++ b/frontend/src/pages/ErrorsV2/ErrorBody/components/ErrorOccurenceDate/index.tsx
@@ -13,17 +13,17 @@ const ErrorOccurenceDate = ({ errorGroup }: Props) => {
 			gap="4"
 			alignItems="center"
 			flexShrink={0}
-			flexWrap="nowrap"
+			flexWrap="wrap"
+			maxWidth="full"
 		>
-			<Text color="black" size="large" weight="bold">
+			<Text color="black" size="large" weight="bold" whiteSpace="nowrap">
 				{errorGroup?.last_occurrence
 					? moment(errorGroup.last_occurrence).fromNow(true)
-					: 'just now'}
+					: 'just now'}{' '}
+				/
 			</Text>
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/Text/Text.tsx
```diff
@@ -54,6 +56,7 @@ export const Text = React.forwardRef<unknown, Props>(
 				cssClass={clsx(styles.variants({ ...props }), cssClass)}
 				title={title}
 				overflowWrap={wrap}
+				whiteSpace={whiteSpace}
```
deltaepsilon, line 59: @ccschmitz Is this the right way to think about this styling??

Vadman97 (approved):
nice! wonder if there is a way to increase the gap between the rows once the line wraps

frontend/src/pages/ErrorsV2/ErrorBody/components/ErrorOccurenceDate/index.tsx
```diff
@@ -13,17 +13,18 @@ const ErrorOccurenceDate = ({ errorGroup }: Props) => {
 			gap="4"
 			alignItems="center"
 			flexShrink={0}
-			flexWrap="nowrap"
+			flexWrap="wrap"
+			maxWidth="full"
+			style={{ rowGap: 8 }}
```
deltaepsilon, line 18: @ccschmitz Is this the right way to approach this?

ccschmitz: We could also add this as a property in our sprinkles and allow it to accept our default space values. Would be similar to how we support `gap`:

https://github.com/highlight/highlight/blob/28138c5bc40addf98aec4162d1c9fd211141ffe7/packages/ui/src/css/sprinkles.css.ts/#L88

deltaepsilon: 💯 I need to do a deep dive into Sprinkles.

PR #5010, opened 2023-04-19
Competitor pages

Abhishek-More:
## Summary

PR for competitor pages. Added data + more customization for these pages. Also added to sitemap


jay-khatri (approved):

highlight.io/components/Competitors/hotjar.ts
```diff
@@ -1,37 +1,47 @@
...
 	sections: [
 		{
 			title: 'General',
 			rows: [
 				{
 					feature: 'Unlimited Team Members',
+					tooltip:
+						'LogRocket only allows 5 team members on their free plan.',
 					highlight: 1,
 					competitor: 0,
```
Vadman97, line 22: nit: numbers here could be booleans if that is their only purpose

Abhishek-More: For competitor, the only option is 0 or 1, but for highlight, the options are 0, 0.5, 1. 0.5 will represents "coming soon". For consistency, I opted to keep the numbers for the competitor.

Issue #4914, opened 2023-04-10
change caching strategy for 'slack channels to notify' 


Vadman97:
When we create an alert in highlight or comment on a session replay, we can tag slack users or channels via a `@...` reference.
The frontend uses an endpoint to lookup 'cached' slack channels that are stored in a specific postgres table.
This leads to outdated recommendations and requires a user to click a 'sync' button in the frontend, leading to a subpar UX.

Instead, remove the backend caching of the slack channels and rely on the indexeddb frontend caching.
Requests to 'get channel recommendations' will use a value from indexeddb if one exists, and update the frontend cache after the backend responds.

cc @mayberryzane

PR #5011, opened 2023-04-19
4914 slack caching strategy

shayneo:
## Summary
Addresses https://github.com/highlight/highlight/issues/4914

Removes the manual  "sync" button and instead makes a throttled request for the latest
and greatest slack data. **Note, the slack API rate limits at ~20 requests/min so we throttle to respect that**

1. navigate to a menu where you can select slack channels or `@` mentions users
2. type in the name of a channel that doesn't exist/shows no results
3. see the loading indicator in the search select menu
4. create the channel in your slack instance
5. search again
6. see it appear after the throttled request completes
7. rejoice 🎉  

Screenshot 2023-04-18 at 11 29 05 PM

This will add more server traffic for storing slack info in postgres. Probably nothing noticeable though given that users aren't doing this all the time.


frontend/src/pages/Alerts/AlertConfigurationCard/AlertConfigurationCard.tsx
```diff
@@ -782,6 +786,7 @@ export const AlertConfigurationCard = ({
 											mode="multiple"
 											onSearch={(value) => {
 												setSearchQuery(value)
+												syncSlack()
```
ccschmitz, line 789: As discussed in the demo, I think a better approach here could be to kick this off once, maybe in an `onFocus` or even when the component is rendered.

frontend/src/pages/Alerts/LogAlert/LogAlertPage.tsx
```diff
@@ -645,6 +640,10 @@ const LogAlertForm = ({
 						placeholder="Select Slack channels"
 						options={slackChannels}
 						optionFilterProp="label"
+						onSearch={(value) => {
+							setSlackSearchQuery(value)
+							syncSlack()
```
ccschmitz, line 645: Wondering if we can trigger this in an `onFocus` here as well 🤔

frontend/src/pages/Player/Toolbar/NewCommentForm/CommentTextBody/CommentTextBody.tsx
```diff
@@ -104,7 +115,17 @@ const CommentTextBody = ({
 			value={commentText}
 			className="mentions"
 			classNames={mentionsClassNames}
-			onChange={onChangeHandler}
+			onChange={(event, newValue, newPlainTextValue, mentions) => {
+				syncSlack()
```
ccschmitz, line 119: `onFocus` here as well?

frontend/src/pages/Player/Toolbar/NewCommentForm/CommentTextBody/CommentTextBody.tsx
```diff
@@ -58,6 +61,14 @@ const CommentTextBody = ({
 		}
 	}, [shouldAutoFocus])
 
+	useEffect(() => {
+		const parts = commentText.split('@')
+		const last = parts.at(-1)
+		if (last) {
+			setLatestAt(`@${last}`)
+		}
+	}, [commentText])
```
ccschmitz, line 70: Wondering if we need to be using state to calculate this or if it would be better to just derive `latestAt` on each render 🤔 - I think I'd go that route unless we noticed a performance issue or something that we're trying to work around.

frontend/src/pages/Alerts/MonitorConfiguration/MonitorConfiguration.tsx
```diff
@@ -440,6 +442,7 @@ const MonitorConfiguration = ({
 						mode="multiple"
 						onSearch={(value) => {
 							setSearchQuery(value)
+							syncSlack()
```
ccschmitz, line 445: `onFocus` here as well?

frontend/src/hooks/useSlackSync.ts
```diff
@@ -0,0 +1,42 @@
...
+		throttle(
+			async () => {
+				await syncSlackIntegration()
+				setSlackLoading(false)
+			},
+			3000,
+			{ leading: true },
+		),
+		[],
+	)
```
ccschmitz, line 31: If we end up calling this only in our initialization logic or when focusing an input, maybe we wouldn't need this throttling logic anymore either 🤔

ccschmitz (changes requested):
Looks great 💖  - the only change I think we should consider is changing the behavior for when we fetch from Slack. Thinking it would be best to handle this on render or maybe when an input is focused.

Great work overall!

ccschmitz:
Per discussion w/ @Vadman97, I'll pick this one up to incorporate changes from my feedback.

shayneo:
> Looks great 💖 - the only change I think we should consider is changing the behavior for when we fetch from Slack. Thinking it would be best to handle this on render or maybe when an input is focused.
> 
> Great work overall!

Good call. You beat me to updating! Made a few tiny tweaks.

ccschmitz:
@Vadman97 I pushed some changes to this to address PR comments - would you mind taking the final review?

frontend/src/pages/Alerts/AlertConfigurationCard/AlertConfigurationCard.tsx
```diff
@@ -784,9 +784,7 @@ export const AlertConfigurationCard = ({
 											className={styles.channelSelect}
 											options={channels}
 											mode="multiple"
-											onFocus={() => {
-												syncSlack()
-											}}
```
ccschmitz: @shayneo I prefer this as well, but followed the pattern I saw in `onSearch` - thanks for cleaning it up 😸

Vadman97:
build failure unrelated - https://github.com/highlight/highlight/commit/9cff1b597519b836e72ddb370fcecc2ed1053440#r111383332

but observing a small issue with the changes in aabc23f @ccschmitz 
https://www.loom.com/share/26aff145538a4984bde380825bf36bf9

et:
> build failure unrelated - https://github.com/highlight/highlight/commit/9cff1b597519b836e72ddb370fcecc2ed1053440#r111383332

Just a heads up, that this is fixed in #5206. Should be able to rebase main in this branch. Sorry about the disruption!

Vadman97:
> @Vadman97 I think this actually is working as expected, but you don't notice the loading state initially because it's already finished by the time you type your first @mention. It happens when the input is focused, not when the popup is shown. Let me know if you think that makes sense.
> 
> You can actually see in your demo at the end where it is loading when the input is focused: https://www.loom.com/share/26aff145538a4984bde380825bf36bf9?highlightComment=12120851&t=67

got it, thanks for clarifying @ccschmitz ! that makes sense then, just wanted to double check.

ccschmitz (approved):

PR #5014, opened 2023-04-19
add codeql.yml

Vadman97:
Adds automated codescanning for security issues.


mayberryzane (approved):
👍 just looking at the output, CodeQL looks pretty handy. Devskim looks a little noisy but it can't really hurt to run it as well, maybe we could skip certain files, e.g. it looks like it's finding strings like "md5" in image files

PR #5018, opened 2023-04-19
stop session feed query carryover on error feed

Vadman97:
## Summary

When reloading the sessions page and the URL is populated with a session query, 
the query params carry over to the error feed, breaking the query for errors.

This was caused by the differing error feed logic of triggering the `setSearchParams` update
on `segmentsLoading` change.

By making this consistent with the error feed, we get the right behavior on the errors feed.

Before: https://www.loom.com/share/af6cc98433214fdc9b73988da19f19c7

After: https://www.loom.com/share/5e9973ba3841426a97fe4a426216c686

No


ccschmitz (approved):
Nice - thank you for fixing this! Have been caught by this a few times when going between sessions and errors.

PR #5022, opened 2023-04-19
harden otel test

Vadman97:
## Summary

Improves stability of the `TestHandler_HandleTrace` test case which would
fail randomly because it assumed an order to opentelemetry messages.

Running the test in a benchmark to ensure it consistently passes

image

No

### DIFF
--- a/backend/otel/otel_test.go
+++ b/backend/otel/otel_test.go
@@ -7,6 +7,7 @@ import (
 	kafkaqueue "github.com/highlight-run/highlight/backend/kafka-queue"
 	"github.com/highlight-run/highlight/backend/private-graph/graph/model"
 	public "github.com/highlight-run/highlight/backend/public-graph/graph"
+	"github.com/samber/lo"
 	"github.com/stretchr/testify/assert"
 	"go.opentelemetry.io/collector/pdata/ptrace/ptraceotlp"
 	"net/http"
@@ -48,6 +49,7 @@ func TestHandler_HandleLog(t *testing.T) {
 	h := Handler{}
 	h.HandleLog(w, r)
 }
+
 func TestHandler_HandleTrace(t *testing.T) {
 	inputBytes, err := os.ReadFile("./samples/traces.json")
 	if err != nil {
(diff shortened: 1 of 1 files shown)


backend/otel/otel_test.go
```diff
@@ -84,19 +86,38 @@ func TestHandler_HandleTrace(t *testing.T) {
...
-		assert.Equal(t, 14, len(producer.messages[2].PushLogs.LogRows))
-		for _, log := range producer.messages[2].PushLogs.LogRows {
-			assert.Equal(t, model.LogSourceBackend, log.Source)
-		}
-	}
-	if assert.Equal(t, kafkaqueue.PushLogs, producer.messages[3].Type) {
-		assert.Equal(t, 1, len(producer.messages[3].PushLogs.LogRows))
-		for _, log := range producer.messages[3].PushLogs.LogRows {
-			assert.Equal(t, model.LogSourceFrontend, log.Source)
+	assert.GreaterOrEqual(t, 4, len(producer.messages))
```
et, line 89: Why did this need to change to `GreaterOrEqual`?

backend/otel/otel_test.go
```diff
@@ -84,19 +86,38 @@ func TestHandler_HandleTrace(t *testing.T) {
...
+		return message.Type == kafkaqueue.PushLogs
+	})
+	assert.Truef(t, ok, "did not find a PushLogs message")
+
+	allPushLogs := lo.Filter(producer.messages, func(message *kafkaqueue.Message, _ int) bool {
+		return message.Type == kafkaqueue.PushLogs
+	})
+
+	for _, pushLogs := range allPushLogs {
+		if len(pushLogs.PushLogs.LogRows) == 14 {
```
et, line 111: Is the issue for the flaking test that this test runs with different inputs? It feels a bit weird for the test runs to not be consistent. Is there a way to modify this test such that it runs with a consistent output?

Vadman97: the test runs with the same inputs but the data is submitted to the mock kafka object in an arbitrary order (no ordering guarantees), so the random order between frontend / backend logs and errors broke this previous logic. 
https://github.com/highlight/highlight/blob/1d0409e596a39955efded4c6c7b4403f15c44dba/backend/otel/otel.go#L182-L185
we use go maps to store the deserialized data before it is sent to kafka. since frontend and backend logs and errors live in the same maps, but maps have no guarantees about iteration order, the [order at kafka submission](https://github.com/highlight/highlight/blob/1d0409e596a39955efded4c6c7b4403f15c44dba/backend/otel/otel.go#L261) is arbitrary, so we end up with about 1/3 of the runs having `PushBackendPayload, PushLogs, MarkBackendSetup` message ordering and the rest having `PushBackendPayload, MarkBackendSetup, PushLogs` ordering.

et: Got it! Thanks for describing that.

et (approved):

PR #5025, opened 2023-04-19
add a GitHub integration for creating issues

Vadman97:
## Summary

Builds [a GitHub app](https://github.com/apps/highlight-io/) for Highlight for creating issues from comments and errors.
The GitHub app is installed on a Highlight workspace since the app is installed to a GitHub workspace.
Currently for creating issues, we show all repositories that the app has access to (within the org that it is installed to)
in the dropdown of the creation flow. In the future, the integration settings modal could be used to connect the
app to particular highlight projects (eg if we want to associate a highlight project with a repo for tracking deploys).

We use an app installation workflow to get access to the 'organization' into which
an app is installed. if we went with an oauth workflow, we would only have access
to the repositories the installing user owns. see [github docs](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/authenticating-as-a-github-app-installation) for more details on the auth flow.

We use the existing `model.IntegrationWorkspaceMapping` workflow for storing the integration state of 
the GitHub app with a workspace.

https://www.loom.com/share/29fe17a8d78f42d8bf2ca3ac1aaab982

No


frontend/src/pages/IntegrationsPage/components/GitHubIntegration/utils.ts
```diff
@@ -0,0 +1,72 @@
...
+		useGetGitHubIntegrationSettingsQuery,
+		useUpdateIntegrationProjectSettingsMutation,
+	)
+
+export const getGitHubOAuthUrl = (projectId: string, workspaceId: string) => {
+	let redirectPath = window.location.pathname
+	if (redirectPath.length > 3) {
+		// remove project_id and prepended slash
+		redirectPath = redirectPath.substring(redirectPath.indexOf('/', 1) + 1)
+	}
```
Vadman97, line 33: delete unused code

et (commented):
@Vadman97 and I chatted about this PR via slack/zoom but leaving comments for a papertrail.


* This does not link a github repo -> highlight project like other new integrations (height, clickup) do. As a result, the end user has an extra step when creating an issue (of selecting the rep). We decided to punt on someone might want multiple repos linked to a single project. If we really wanted to pursue this, we could link one project to a repo which is preselected when creating an issue but still give them an option to change it on the fly.
* This integration does not work with localhost. This seems to be a limitation with github apps that don't allow one to set multiple redirect urls when using the auth flow.
* There's an bug of linking multiple orgs seems to fetch the wrong repos ([loom](https://www.loom.com/share/b297c1e809a549eb82dd064a80bfd767)).

backend/integrations/github/github.go
```diff
@@ -0,0 +1,101 @@
...
+
+	return &config, nil
+}
+
+type Client struct {
+	client *github.Client
+	owner  string
+}
+
+func NewClient(ctx context.Context, installation string) *Client {
```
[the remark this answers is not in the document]
Vadman97: this is an installation ID which on its own doesn't give any access, but is used to look up what account we're looking at. adding a comment here accordingly.

Vadman97:
a fix for fetching repos from the correct org (where the app is installed) is added in ec305af 
https://www.loom.com/share/5058172b19dc49af91feb90c23769613

Vadman97:
bump on this @et review. regarding your other question:
> This does not link a github repo -> highlight project like other new integrations (height, clickup) do. As a result, the end user has an extra step when creating an issue (of selecting the rep). We decided to punt on someone might want multiple repos linked to a single project. If we really wanted to pursue this, we could link one project to a repo which is preselected when creating an issue but still give them an option to change it on the fly.

since the github integration is only used for filing issues and there are often cases issues should be created in different repositories, i suggest we keep the current workspace integration workflow and allow users to pick what repository they are creating a ticket in at the time of submission (as with the linear integration).

frontend/src/pages/IntegrationsPage/components/GitHubIntegration/GitHubIntegrationConfig.tsx
```diff
@@ -0,0 +1,100 @@
...
+					className={styles.modalBtn}
+					onClick={() => {
+						setModalOpen(false)
+						setIntegrationEnabled(false)
+					}}
+				>
+					Cancel
+				</Button>
+				<Button
+					trackingId="IntegrationConfigurationSave-Slack"
```
et, line 87: image

frontend/src/pages/IntegrationsPage/components/GitHubIntegration/GitHubIntegrationConfig.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+	return (
+		<>
+			<p className={styles.modalSubTitle}>
+				Connect GitHub to your Highlight workspace to create issues from
+				comments.
+			</p>
+			<footer>
+				<Button
+					trackingId="IntegrationConfigurationCancel-Slack"
```
et, line 77: Copy pasta error here.

frontend/src/pages/IntegrationsPage/components/GitHubIntegration/GitHubIntegrationConfig.tsx
```diff
@@ -0,0 +1,100 @@
...
+						className={styles.modalBtn}
+						onClick={() => {
+							setModalOpen(false)
+							setIntegrationEnabled(true)
+						}}
+					>
+						Cancel
+					</Button>
+					<Button
+						trackingId="IntegrationDisconnectSave-Slack"
```
et, line 51: Copy pasta error here.

et (approved):
Tested the following
* Confirmed it does not list repos of another org that has the highlight app configured
* Confirmed it works with granting access to a single public repo
* Confirmed it works with granting access to a single private repo
* Confirmed it works with granting access to all repos (will allow access to both public/private repos)
* Tested disconnection though Highlight.
* Confirmed disconnecting through GitHub shows the integration as disconnected in Highlight

Couple things I noticed but I don't think our blockers:
* Tags are carried over, I don't know if this is what we want or not:
Screenshot_2023-04-25_at_4_45_39_PM
* We need docs added to the [integrations](https://www.highlight.io/docs/general/integrations/overview) page.

frontend/src/pages/IntegrationsPage/components/GitHubIntegration/GitHubIntegrationConfig.tsx
Vadman97: my favorite pasta :yum:

Vadman97:
> Tags are carried over, I don't know if this is what we want or not:
carrying over tags, but also showing the github tags in the autocomplete of a issue creation in b03232dcdc3427559580993bbeea7659f8625fdd
> We need docs added to the [integrations](https://www.highlight.io/docs/general/integrations/overview) page
added in e80818f48dde3b2ad8b8740661d4a7a979492b6f

Issue #4764, opened 2023-03-29
Allow hours/seconds to be set in the datetime picker


et:
https://www.figma.com/file/rdBPNzn7Klk6RG1LHUCE5B/Screen-Designs?node-id=2513-150385

![Screenshot 2023-03-29 at 3 17 22 PM](https://user-images.githubusercontent.com/58678/228669802-62b7d6d6-ce79-4550-94e9-b7f2c7182076.png)

A relative datetime picker was added in #4245 and we need to add hours/minutes to the component. 

To run the ui library:
* cd packages/ui
* yarn dev
* Visit http://localhost:6006/
* Click DatePicker -> PreviousDateRangePicker

Requirements:
* Add two rows above the calendar -- one for month/day range and one for hour/minute range
* The hour/minute range should only show if the clock icon is selected

et:
> Hello @et i will love to work on this.

Sure! I've updated the issue description with a bit more detail. Let me know if you run into any issues.

betty-alagwu:
Alright thank you very much.

betty-alagwu:
Hello @et  can you please grant me access to the design for the date picker.

jay-khatri:
Just granted @betty-alagwu!

betty-alagwu:
Thank you very much @jay-khatri

betty-alagwu:
@julian-highlight thank you for the feedback, will work on it now.

Vadman97:
Thanks @betty-alagwu ! This is quite an important feature as the logs viewer is hard to use without the ability to pick a fine-grained time range. Thanks for picking this up.

betty-alagwu:
Thank you @Vadman97 . I'll be raising a first PR for feedback before EOD.

PR #5030, opened 2023-04-20
Feat/add hours mins to date picker

betty-alagwu:
Hello @Vadman97 @julian-highlight , thank you very much for your patience.  This PR allows the `hours` and `minute` to be set in the `DateTime picker`

Solves: #4764 

https://user-images.githubusercontent.com/94234459/233263271-fd73bad3-309d-406a-bc82-b30cf18856fe.mp4


julian-highlight:
Yo @betty-alagwu ! Already looking pretty good :) 

I have a few remaining questions though: is there a reason for the blue background inside the input? Also the hour/min picker seem to use a bolder font(?). Ideally the lower input should look identically to the upper one in terms of styles!

julian-highlight:
Btw there also was a little figma bug displaying the icon-switch unselected state with a grey background! Fixed this a few days ago. Sorry! @betty-alagwu 
image

betty-alagwu:
@julian-highlight Thank you for the feedback, I'll work on the changes right away

betty-alagwu:
> Yo @betty-alagwu ! Already looking pretty good :)
> 
> I have a few remaining questions though: is there a reason for the blue background inside the input? Also the hour/min picker seem to use a bolder font(?). Ideally the lower input should look identically to the upper one in terms of styles!

Hello @julian-highlight The blue background comes up when the use selects from the browser suggestion, but when the user types in the input, there is no background in the input.

julian-highlight:
> 

oh yea, thought it might have to be a browser side thing! Thanks for the headsup!

betty-alagwu:
Thank you @jay-khatri for the prompt, that's my priority for today.

packages/ui/src/components/Form/styles.css.ts
```diff
@@ -1,6 +1,7 @@
 import { recipe, RecipeVariants } from '@vanilla-extract/recipes'
 import { colors } from '../../css/colors'
 import { vars } from '../../css/vars'
+import { typographyStyles } from '@components/Text/styles.css'
```
Vadman97, line 4: looks like this is causing the [js build to fail](https://github.com/highlight/highlight/actions/runs/4772287183/jobs/8484704824?pr=5030)

```
Error: @highlight-run/frontend:typegen: ../packages/ui/src/components/Form/styles.css.ts(4,10): error TS2614: Module '"*.css"' has no exported member 'typographyStyles'. Did you mean to use 'import typographyStyles from "*.css"' instead?
```

i think the import needs to change in this case to
`import typographyStyles from "*.css"`

julian-highlight (approved):
Design-wise this looks good! Bump me as soon as I'm able to review it in prod!

packages/ui/src/components/Form/styles.css.ts
betty-alagwu: Thank you @Vadman97, i've resolved the issue.

betty-alagwu:
Hey @jay-khatri  thanks for your patience, the issue is resolved now, it's ready for review

packages/ui/src/components/DatePicker/PreviousDateRangePicker.tsx
```diff
@@ -232,15 +381,161 @@ const PreviousDateRangePickerImpl = ({
...
+								style={{
+									width: 28,
+									height: 28,
+									background: showingTime
+										? colors.p9
+										: colors.n4,
+								}}
+								onClick={handleShowingTimeToggle}
+							>
+								<IconSolidClock
```
Vadman97, line 519: icon is off-center

image

packages/ui/src/components/DatePicker/PreviousDateRangePicker.tsx
```diff
@@ -60,6 +66,74 @@ function toDateTimeString(date: Date, showYear: boolean) {
...
+	const hour24 =
+		isPM && hourInt !== 12
+			? hourInt + 12
+			: isAM && hourInt === 12
+			? 0
+			: hourInt
+	const timeOfDay = isAM ? 'AM' : isPM ? 'PM' : ''
+
+	return { hour: hourInt, minute: parseInt(minute, 10), timeOfDay, hour24 }
+}
```
Vadman97, line 90: could we use moment.js for this: https://momentjs.com/docs/#/parsing/
this already exists in moment, ie something like:
```js
let m = moment('12:34 pm', 'hh:mm a');
let data = { hour: m.format('HH'), minute: m.minute(), timeOfDay: m.hour() >= 12, hour24: m.hour() };
console.log({ data });
```

packages/ui/src/components/DatePicker/PreviousDateRangePicker.tsx
```diff
@@ -60,6 +66,74 @@ function toDateTimeString(date: Date, showYear: boolean) {
...
+ */
+const getTimeStringFromDate = (date: Date): string => {
+	if (!date) return '00:00 AM'
+
+	const hour = date.getHours() % 12 || 12
+	const minute = date.getMinutes().toString().padStart(2, '0')
+	const period = date.getHours() < 12 ? 'AM' : 'PM'
+
+	return `${hour}:${minute} ${period}`
+}
```
Vadman97, line 122: `moment().format()` would be a cleaner way to do that.
the reason we prefer to use a library for something like this is it is something well tested that covers all the edge-cases well. this would be a new implementation of something that already exists that we would need to test and maintain.

Vadman97 (commented):
looks great thus far, thanks for your work @betty-alagwu ! would you mind changing the time logic to moment.j to cover some of the edge cases and make the code more maintainable?

packages/ui/src/components/DatePicker/PreviousDateRangePicker.tsx
betty-alagwu: Thank you. Fixed the positioning and now it's centered.

packages/ui/src/components/DatePicker/PreviousDateRangePicker.tsx
betty-alagwu: Changed this to use moment for formatting.

betty-alagwu:
Thank you @Vadman97 for your review. I initially had reservations on using Moment because it was never used in the `ui` package before.

Thanks again for your time.

 It's ready for another pass now.

packages/ui/src/components/DatePicker/PreviousDateRangePicker.tsx
```diff
@@ -149,6 +245,49 @@ const PreviousDateRangePickerImpl = ({
 		}
 	}
 
+	const handleTimeChange = (
+		event: React.ChangeEvent<HTMLInputElement>,
+		input: 'start' | 'end',
+	) => {
+		const value = event.target.value
+		const isValid = value.match(VALID_FORMATTED_TIME_REGEX) !== null
```
Vadman97, line 253: can also use moment for this too? https://stackoverflow.com/a/40131561 - then we can get rid of the regex constant

Vadman97 (approved):
Much neater now, thank you Betty!

packages/ui/src/components/DatePicker/PreviousDateRangePicker.tsx
Vadman97: looks great now!

packages/ui/src/components/DatePicker/PreviousDateRangePicker.tsx
Vadman97: thanks!

packages/ui/src/components/DatePicker/PreviousDateRangePicker.tsx
```diff
@@ -60,6 +64,64 @@ function toDateTimeString(date: Date, showYear: boolean) {
...
+ */
+export const getTimeInfo = (timeString: string) => {
+	const date = moment(timeString, 'hh:mm a')
+	return {
+		hour: date.format('HH'),
+		minute: date.minute(),
+		timeOfDay: date.hour() >= 12,
+		hour24: date.hour(),
+	}
+}
```
Vadman97, line 80: image

just noticed this actually isn't quite working for AM/PM times (it works right for 24-hour times like 12:00 or 14:00). might be that the moment format needs to be different, could you please double check?

Vadman97: fyi just tried with 89ed966d297ea4ea607776271f58d871b87460a9 and this still isn't quite working.
we should also support inputs like `6pm` or `9 PM`

betty-alagwu:
Thank you @Vadman97 for bringing this to my attention,  I apologise for the delay and inconvenience. I have checked the moment format and modified to accommodate AM/PM times.

packages/ui/src/components/DatePicker/PreviousDateRangePicker.tsx
```diff
@@ -60,6 +72,64 @@ function toDateTimeString(date: Date, showYear: boolean) {
...
+ *
+ * @param {Date} date - The Date object to convert.
+ * @returns {string} A string in the format "HH:mm aa".
+ */
+const getTimeStringFromDate = (date: Date): string => {
+	if (!date) {
+		return '12:00 AM'
+	}
+
+	return moment(date).format(TIME_DISPLAY_FORMAT)
```
Vadman97, line 117: this placeholder text doesn't make sense now. 15:44 PM isn't a valid time - this should be 3:44 PM

image

packages/ui/src/components/DatePicker/PreviousDateRangePicker.tsx
```diff
@@ -149,6 +250,56 @@ const PreviousDateRangePickerImpl = ({
...
+		event: React.ChangeEvent<HTMLInputElement>,
+		input: 'start' | 'end',
+	) => {
+		const value = event.target.value
+		const isValid = [
+			TIME_INPUT_FORMAT,
+			TIME_INPUT_FORMAT_NO_SPACE,
+			TIME_INPUT_FORMAT_12_HOUR,
+			TIME_INPUT_FORMAT_12_HOUR_NO_SPACE,
+		]
```
Vadman97, line 263: could this also support `1pm` or `1 pm` without a minute input, assuming that minutes are 00?

Vadman97 (approved):
thanks again @betty-alagwu , really close now! just two more comments

packages/ui/src/components/DatePicker/PreviousDateRangePicker.tsx
betty-alagwu: Hey @Vadman97 it's looking good now, fixed it and updated the placeholder to display the time in 12 hours format.

packages/ui/src/components/DatePicker/PreviousDateRangePicker.tsx
betty-alagwu: I've just made an improvement to the time input format. Now, users can also input hours without minutes.

Vadman97: nice! this works well now

packages/ui/src/components/DatePicker/PreviousDateRangePicker.tsx
Vadman97: thanks, looks good!

packages/ui/src/components/DatePicker/PreviousDateRangePicker.tsx
```diff
@@ -232,15 +389,167 @@ const PreviousDateRangePickerImpl = ({
...
+												? 'secondary'
+												: 'error'
+										}
+										borderTop={
+											startTimeIsValid ? 'none' : 'error'
+										}
+										borderBottomLeftRadius={'6'}
+										borderBottomRightRadius={'6'}
+									>
+										<Form.Input
```
Vadman97, line 438: hey @betty-alagwu , apologies for all the back-and-forth. looks like there is a weird behavior now that bumps the input cursor to the end of the line whenever the time value is changed. here's an example of the behavior - https://www.loom.com/share/3c56e1726f784e26a4c0985d4914f53f

betty-alagwu: Thank you for the feedback @Vadman97 , I will implement the changes today and I am happy to making any additional modifications.

betty-alagwu: Hey @Vadman97 I've implemented some changes based on the bug you mentioned earlier.
Here's a video of how it works.

https://user-images.githubusercontent.com/94234459/234977177-ff92fc01-e2a3-4071-87fd-0a8902d9f987.mp4

packages/ui/src/components/DatePicker/TimeInput.tsx
```diff
@@ -0,0 +1,29 @@
...
+			color={'n8'}
+			value={value}
+			style={{
+				boxSizing: 'border-box',
+				border: 'none',
+				background: 'none',
+			}}
+			onChange={(event) => {
+				setValue(event.target.value)
+				onTimeChange?.(event.target.value)
```
Vadman97, line 25: i didn't know this was valid typescript syntax! awesome :)

packages/ui/src/components/DatePicker/TimeInput.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+export interface TimeInputProps extends InputProps {
+	onTimeChange?: (value: string) => void
+}
+
+export function TimeInput({ placeholder, name, onTimeChange }: TimeInputProps) {
+	const [value, setValue] = useState('')
+
+	return (
+		<Form.Input
```
Vadman97, line 12: kind of hacky to use the https://ariakit.org/ `Form` like this but it works! i think another way to have done something like the following (using a `<Form>` component with a ariakit form state)
https://github.com/highlight/highlight/blob/fb3b81e9988edbb3eda00f37b54ed9b900ed8602/frontend/src/pages/Auth/SignIn.tsx#L68-L120

Vadman97 (approved):
works much better now! just tested editing the time inputs locally

packages/ui/src/components/DatePicker/PreviousDateRangePicker.tsx
Vadman97: thanks @betty-alagwu ! this works great now

PR #5033, opened 2023-04-20
"Request a Demo" fixes

### LINKED ISSUE #5017: "Request a demo" fixes.
https://www.loom.com/share/25a47ed5a3934a8d89c8d3a73fe03b22

Figma "x" button (we can make it the same size as the current button, but hopefully this color + positioning it in the top righ makes it more obvious)
https://www.figma.com/file/lsi5qUTKAKxLv4VQ9K3cDJ/Highlight-%E2%80%93-Site-Build?node-id=5222-43844&t=aYmzVKmWHUJUZEOY-11

cc @Vadman97

Crawron:
## Summary

- Clicking out of the widget should close it consistently now.
  - Calendly includes an empty margin on its widget that may block some clicks
- Close button is now on the top right, and has an "X" icon instead of a left arrow
- Pressing Esc should close the widget if it's not focused.
  - When the widget is focused it seems to capture the key press and prevents headless from handling it
- Added button to feature pages

~~There's a big issue I have to figure out when it comes to using the widgets in other pages. When opening the widget, changing page, then opening it again, the widget will not appear. Calendly's inline widget assumes that after the div initially loads, it will *remain* on the page, which doesn't in our case. I think a different widget type may work better, but I have to test that.~~

I'm using a library `react-calendly` now, that seems to fix the loading issue

Closes #5017


jay-khatri (changes requested):
Fabio, this looks great. Styling is on point. Had a few comments:

Loom video: https://www.loom.com/share/5458fa0ea40d465a8940b01957ab5490
- mobile?
- how can we make sure it consistently renders the first time?
- making sure it works after switching pages?

https://github.com/tcampb/react-calendly

jay-khatri:
Fabio, we should use this link now: https://calendly.com/d/2gt-rw5-qg5/highlight-demo-call

This is probably what's causing the merge conflict.

jay-khatri (approved):
awesome!

Issue #3746, opened 2023-01-30
Initiative: Move away from Moment.js


jay-khatri:
Moment is an unnecessarily big library. [It is in maintenance mode, and its project authors are recommending to use something else.](https://momentjs.com/docs/#/-project-status/)

My personal favorite is [date-fns](https://npm.im/date-fns), but [Day.js](https://npm.im/dayjs) is also a good option, with a similar API to moment.

PR #5069, opened 2023-04-21
Optimize fetching for integration status

ccschmitz:
## Summary

Queries the `setup_events` table to determine if someone is integrated and performs an opensearch query to get a session/error to link to on the integration bar across the top of the setup page.

Also changes the behavior of the top-level app loading state so it does not wait for the integration status to load before rendering the page. This fixes an issue where shared sessions/errors can't be viewed when you are not logged in.

Finally, there is a little bit of cleanup of some janky styles on the header when you are logged in and viewing a shared session/error.

Local click test.

Should click test on some accounts that were taking a long time to load the integration status prior to this change.


backend/private-graph/graph/schema.resolvers.go
```diff
@@ -46,14 +46,14 @@ import (
...
-	"github.com/highlight/highlight/sdk/highlight-go"
+	highlight "github.com/highlight/highlight/sdk/highlight-go"
 	"github.com/leonelquinteros/hubspot"
 	"github.com/lib/pq"
 	"github.com/openlyinc/pointy"
 	e "github.com/pkg/errors"
 	"github.com/samber/lo"
 	log "github.com/sirupsen/logrus"
-	"github.com/stripe/stripe-go/v72"
+	stripe "github.com/stripe/stripe-go/v72"
```
ccschmitz, line 56: Not sure why these change in this PR, but was getting [a build error](https://github.com/highlight/highlight/actions/runs/4765793100/jobs/8471987110?pr=5069) without this.

frontend/src/pages/Setup/IntegrationBar.tsx
```diff
@@ -122,19 +154,20 @@ export const IntegrationBar: React.FC<Props> = ({ integrationData }) => {
...
+	const basePath = `/${projectId}/${CTA_PATH_MAP[area!]}`
+	let path
 
-	if (integrationData?.resourceSecureId) {
-		path = `${path}/${integrationData.resourceSecureId}`
+	if (resource?.secure_id) {
+		path = `${basePath}/${resource.secure_id}`
 	} else if (area === 'backend-logging') {
-		const logDate = moment(integrationData?.createdAt)
+		const logDate = moment(resource?.created_at)
```
deltaepsilon, line 165: Replacing `moment` with `date-fns` could be a fun project. It would do wonders for our bundle size.

ccschmitz: Definitely! I think we looked at this in the past, but not sure why we didn't go through with it. Maybe @jay-khatri knows 🤔 

https://github.com/highlight/highlight/issues/3746

jay-khatri: That wasn't my ticket; it was just created by me in the migration 😅 . Don't have opinions here tbh; are we worried about our bundle size atm?

ccschmitz: No, I don't think there are any major concerns right now. Never hurts to optimize, but if we start to see more mobile / slow networks trying to use Highlight we may want to consider it (I've never actually looked to check avg connection speed or browser/OS, but assume it's mostly desktop usage).

PR #5070, opened 2023-04-21
Fix overmatching for attribute and body filters

ccschmitz:
## Summary

Fixes issues with overmatching and highlighting attributes that aren't actually relevant to the logic for the logs being shown.

### Before

Screenshot 2023-04-21 at 9 25 21 AM
Screenshot 2023-04-21 at 9 25 36 AM

### After

Screenshot 2023-04-21 at 9 25 11 AM
Screenshot 2023-04-21 at 9 25 43 AM

Local click test.

N/A

### DIFF
--- a/frontend/src/pages/LogsPage/LogsTable/LogsTable.tsx
+++ b/frontend/src/pages/LogsPage/LogsTable/LogsTable.tsx
@@ -357,18 +357,18 @@ export const findMatchingLogAttributes = (
 		}
 
 		let matchingAttribute: string | undefined = undefined
-		if (bodyQueryValue && value.indexOf(bodyQueryValue) !== -1) {
+		if (
+			bodyQueryValue &&
+			key === 'message' &&
+			value.indexOf(bodyQueryValue) !== -1
+		) {
 			matchingAttribute = bodyQueryValue
 		} else {
 			queryTerms.some((term) => {
 				const queryKey = term.key
 				const queryValue = term.value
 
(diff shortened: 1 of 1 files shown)


frontend/src/pages/LogsPage/LogsTable/LogsTable.tsx
```diff
@@ -357,18 +357,18 @@ export const findMatchingLogAttributes = (
 		}
 
 		let matchingAttribute: string | undefined = undefined
-		if (bodyQueryValue && value.indexOf(bodyQueryValue) !== -1) {
+		if (
+			bodyQueryValue &&
+			key === 'message' &&
```
et, line 362: Use the graphql enum here:
https://github.com/highlight/highlight/blob/b3bb71fb7d62a108513e03518c05af0d345faffc/backend/private-graph/graph/schema.graphqls#L555

et: And is `bodyQueryValue` matching on the wrong thing? The backend remaps `body` -> `message` so `message` is what the frontend should be matching on.

ccschmitz: I think it's matching on the right thing, but we're using two different names for it, which is confusing. I'm calling it `body` but we have to match against the `message` attribute.

et (approved):
Thanks!

PR #5087, opened 2023-04-22
Fix unnested complex selector resulting in invalid css output

lewisl9029:
## Summary

According to [Vanilla Extract docs](https://vanilla-extract.style/documentation/styling/#complex-selectors), complex rules need to be nested under the `selectors` key. Looks like we had a spot that wasn't doing this, which was detected as a syntax error in the generated CSS on Reflame's side, and a warning in Vite's build output:

Screenshot 2023-04-21 at 8 29 54 PM

This was ending up in the final CSS output apparently (screenshot from prod):

Screenshot 2023-04-17 at 5 42 56 PM

Not valid syntax outside of browsers with draft CSS nesting support, though support is actually getting [surprisingly good](https://caniuse.com/css-nesting) for a draft, which might be why we haven't seen this in prod yet.

Here's the output in the generated CSS after the change:

Screenshot 2023-04-21 at 7 23 11 PM

### DIFF
--- a/packages/ui/src/components/Form/styles.css.ts
+++ b/packages/ui/src/components/Form/styles.css.ts
@@ -41,8 +41,10 @@ export const inputVariants = recipe({
 				width: 0,
 				padding: 0,
 				border: 'none',
-				'&:focus, &:active, &:placeholder-shown:hover': {
-					border: 'none',
+				selectors: {
+					'&:focus, &:active, &:placeholder-shown:hover': {
+						border: 'none',
+					},
 				},
 			},
 			false: {},


Vadman97 (approved):
very cool, thanks for finding that bug! does that mean that in the future, the reflame PR preview would catch this problem before it's merged because the reflame preview would fail?

lewisl9029:
@Vadman97 yep that's the behavior today, the transforms will throw on any parsing error for unrecognized syntax. Although I'll probably enable support for down-compiling the CSS nesting syntax at some point, so this particular issue will no longer show up as an error.

PR #5089, opened 2023-04-22
Various import updates

lewisl9029:
## Summary

A bunch of import updates split out from #4586, so they can be reviewed, tested, and shipped independently. I've structured these into separate commits for each set of related changes, with comments on each. Reviewing commit by commit is recommended.

Ran `yarn build:frontend`, was successful.
Ran `cd docker && ./run.sh`, everything seemed to be working as expected in localhost:3000.

Nothing I can think of.


frontend/src/components/Avatar/Avatar.tsx
```diff
@@ -3,12 +3,10 @@ import React, { useMemo } from 'react'
...
-import {
-	userAvatar,
-	userAvatarBorder,
-	userAvatarText,
-	userAvatarWrapper,
-} from './Avatar.module.scss'
+import styles from './Avatar.module.scss'
+
+const { userAvatar, userAvatarBorder, userAvatarText, userAvatarWrapper } =
+	styles
```
lewisl9029, line 9: The esbuild plugin I'm using for generating js modules for scss doesn't seem to support named exports. Luckily there's only a handful of named export usages across the codebase, so I've manually refactored them to use the default export.

frontend/src/components/SearchEmptyState/SearchEmptyState.tsx
```diff
@@ -3,14 +3,11 @@ import clsx from 'clsx'
...
-import {
-	emptyStateSection,
-	emptyStateWrapper,
-	emptySubTitle,
-	emptyTitle,
-} from './SearchEmptyState.module.scss'
 import styles from './SearchEmptyState.module.scss'
 
+const { emptyStateSection, emptyStateWrapper, emptySubTitle, emptyTitle } =
+	styles
```
lewisl9029, line 9: Same as https://github.com/highlight/highlight/pull/5089/commits/ad87492cfedd23531feb191b915b161ede267714#r1174279552

frontend/src/pages/Player/PlayerHook/utils/usePlayerConfiguration.tsx
```diff
@@ -6,7 +6,7 @@ import { Tab } from '@pages/Player/Toolbar/DevToolsWindowV2/utils'
 import useLocalStorage from '@rehooks/local-storage'
 import { useCallback, useMemo } from 'react'
 
-import { EventsForTimeline } from '.'
+import { EventsForTimeline } from './index'
```
lewisl9029, line 9: Had to change this because `'.'` isn't a valid relative URL specifier according to the spec (needs to start with `./` or `../`).

Reflame unlike most other tools tries to avoid module specifier rewriting as much as possible (instead preferring to rely on import maps for redirection at runtime), because specifier rewriting can be extremely wasteful in terms of write amplification when combined with content-addressed modules (i.e. changing 1 module can invalidate a boatload of dependent cached modules).

sdk/firstload/src/browserExtension/extensionListener.ts
```diff
@@ -1,4 +1,4 @@
-import { H } from '..'
+import { H } from '../index'
```
lewisl9029, line 1: Similar to https://github.com/highlight/highlight/pull/5089/commits/fd58dbaa5938a9ee77190946ddeafe8887fad2ce#r1174279883, `..` is also not a valid relative specifier.

frontend/src/util/attribution.ts
```diff
@@ -1,7 +1,6 @@
+import { GenerateSecureID } from '@highlight-run/client/src/utils/secure-id'
 import Cookies from 'js-cookie'
 
-import { GenerateSecureID } from '../../../sdk/client/src/utils/secure-id'
```
lewisl9029, line 3: Changed this because relative imports outside the package root is incompatible with Reflame's stricter package isolation, and as far as I could see, there's nothing that necessitates their use here in frontend.

packages/ui/src/components/DatePicker/Calendar/Calendar.tsx
```diff
@@ -1,5 +1,7 @@
 import React from 'react'
-import { Box, Stack, Text } from '@highlight-run/ui'
+import { Box } from '../../Box/Box'
+import { Stack } from '../../Stack/Stack'
+import { Text } from '../../Text/Text'
```
lewisl9029, line 4: Had to do a bunch of untangling of circular dependencies in this commit.

Reflame's development mode makes use of [es-module-shims](https://github.com/guybedford/es-module-shims)' shim mode to be able to hook into module loading and update import maps at runtime for its React Fast Refresh implementation. es-module-shims generally does a good job of supporting circular imports transparently, but it seems to be running into some trouble with this specific pattern where index.ts is full of reexports. I'll try to find a minimal repro and file an issue there.

In the longer term, it may be worth considering how we could avoid circular dependencies as much as possible regardless, since they are a blocker to effective code splitting in bundled workflows. Reflame can probably help with this by warning on any circular deps it encounters as part of its dependency analysis.

lewisl9029:
Screenshot 2023-04-21 at 9 49 40 PM
No clue why I'm getting [this error](https://github.com/highlight/highlight/actions/runs/4770785738/jobs/8482281991?pr=5089) on some PRs but not others. 🤔

frontend/src/util/attribution.ts
Vadman97: right, and unlike the `firstload` dependency on `client`, we don't need to worry about the frontend type defs referencing the private library

packages/ui/src/components/DatePicker/Calendar/Calendar.tsx
Vadman97: nice! yeah as i mentioned in the other comment, this will be much more manageable if someone introduces a PR that breaks this but the reflame preview fails (and hence notifies the author)

Vadman97 (approved):

Vadman97:
whoops, build failing because `turbo.yml` breaks for forks when there is a change affecting client. will fix that separately but force merge this since the build and tests are passing.

packages/ui/src/components/DatePicker/Calendar/Calendar.tsx
lewisl9029: This particular case is a bug on Reflame's side (or rather in es-module-shim's shim mode), since circular dependencies are supported just fine in standard modules (i.e. they work for production mode where we don't use es-module-shim's shim mode).

But yeah we'll definitely add warnings for circular dependencies sometime down the line since they cause plenty of other nasty problems (most notably for code-splitting since we can't split off anything within a cycle).

PR #5096, opened 2023-04-24
Sourcemap error style tweaks

deltaepsilon:
image

### DIFF
--- a/frontend/src/pages/ErrorsV2/SourcemapErrorDetails/SourcemapErrorDetails.tsx
+++ b/frontend/src/pages/ErrorsV2/SourcemapErrorDetails/SourcemapErrorDetails.tsx
@@ -1,11 +1,5 @@
-import { useAuthContext } from '@authentication/AuthContext'
-import { Button } from '@components/Button'
-import { LinkButton } from '@components/LinkButton'
-import {
-	Maybe,
-	SourceMappingError,
-	SourceMappingErrorCode,
-} from '@graph/schemas'
+import * as styles from './styles.css'
+
 import {
 	Box,
 	ButtonIcon,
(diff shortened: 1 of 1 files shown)


frontend/src/pages/ErrorsV2/SourcemapErrorDetails/SourcemapErrorDetails.tsx
```diff
@@ -263,7 +264,13 @@ const StackSectionError: React.FC<
 			cursor="default"
 			overflow="scroll"
 			boxShadow="medium"
-			style={{ maxWidth: '450px' }}
+			paddingTop="8"
+			style={{
+				maxWidth: '450px',
+				overflow: 'auto',
+				msOverflowStyle: 'none',
```
deltaepsilon, line 271: It's a windows thing.

MacOS doesn't show unnecessary scrollbars anymore, but Windows does. They look a little gross.

image

packages/ui/src/components/Box/Box.tsx
```diff
@@ -37,7 +40,11 @@ export const Box = React.forwardRef<unknown, BoxProps>(
 		}
 
 		return React.createElement(as, {
-			className: clsx([sprinkles(sprinklesProps), userClasses]),
+			className: clsx([
+				sprinkles(sprinklesProps),
+				userClasses,
+				props.hiddenScroll ? hiddenScroll : null,
```
deltaepsilon, line 46: @ccschmitz Is this the right way to implement `hiddenScroll`?

ccschmitz: My rule of thumb for adding something to the sprinkles classes or `Box` component has been to use something twice, and on the third time support it at this level. Maybe this is something we will need to deal with a number of places throughout the app, but I wonder if we should keep this isolated the sourcemap error details UI for now, or just make it a general utility class that can be `import`ed as needed rather than expanding the surface area of the `Box` props.

No strong opinions here, BTW. Cool with me if you want to merge this as is!

ccschmitz (approved):
One thought on the new `Box` prop, but looks great overall. Thanks for cleaning this up!

packages/ui/src/components/Box/Box.tsx
deltaepsilon: MacOS hides these scrollbars by default, but as the Windows Guy™, I use this snippet pretty frequently.

PR #5106, opened 2023-04-24
Next.js App Directory playground.

deltaepsilon:
## Summary

I'm working on `e2e/nextjs`. It a WIP, but I've got it in a better place and I'm sick of merging `main` over and over again.

Run `yarn e2e:nextjs` and navigate to `localhost:3005`.

Nope. Does not affect production code.


sdk/highlight-node/src/client.ts
```diff
@@ -57,7 +59,14 @@ export class Highlight {
 		this.processor = new BatchSpanProcessor(exporter, {})
 		this.otel = new NodeSDK({
 			autoDetectResources: true,
-			defaultAttributes: { 'highlight.project_id': this._projectID },
+			resource: {
+				attributes: { 'highlight.project_id': this._projectID },
+				merge: (resource) =>
```
deltaepsilon, line 64: This was required by the @opentelemetry upgrade.

sdk/highlight-node/package.json
```diff
@@ -15,18 +15,19 @@
...
 	"publishConfig": {
 		"access": "public"
 	},
 	"dependencies": {
-		"@opentelemetry/api": ">=1.0.0 <1.3.0",
-		"@opentelemetry/auto-instrumentations-node": "0.32.1",
-		"@opentelemetry/exporter-trace-otlp-http": "0.32.0",
-		"@opentelemetry/sdk-node": "0.32.0",
-		"@opentelemetry/sdk-trace-base": "1.6.0",
+		"@opentelemetry/api": "^1.4.1",
```
deltaepsilon, line 26: Upgrading @opentelemetry.

sdk/highlight-next/src/index.edge.ts
```diff
@@ -0,0 +1,9 @@
+import { ApiHandler, HasHeaders, HasStatus } from './util/withHighlight'
+
+import { NodeOptions } from '@highlight-run/node'
+
+export function Highlight(_: NodeOptions) {
```
deltaepsilon, line 5: We do not currently support the `edge` runtime for Next.js. This is an attempt to implement a noop for `edge` so it doesn't throw an error.

sdk/highlight-next/package.json
```diff
@@ -7,14 +7,20 @@
 	"types": "./dist/index.d.ts",
 	"exports": {
 		".": {
+			"edge": "./dist/index.edge.js",
```
deltaepsilon, line 10: Forking these runtimes to the `index.edge.js` noop.

See [vercel/otel](https://github.com/vercel/otel/blob/main/package.json) for the inspiration.

forks/ansi-color/ansi-color.js
```diff
@@ -0,0 +1,37 @@
+var ANSI_CODES = {
```
deltaepsilon, line 1: `ansi-color` threw a fit when I upgraded @opentelemetry. It was a single file, so I copy/pasted it, made a local patch, and resolved to the local file.

.vscode/settings.json
```diff
@@ -90,7 +99,9 @@
 	"[markdown]": {
 		"editor.defaultFormatter": "esbenp.prettier-vscode"
 	},
+	"editor.wordWrapColumn": 120,
+	"editor.wordWrap": "wordWrapColumn",
```
Vadman97, line 103: personally don't mind this and am not affected since i don't use vscode :) 
does this mean other folks who use vscode will start refactoring files, or is this just affecting new lines added?

sdk/highlight-next/package.json
```diff
@@ -7,14 +7,20 @@
...
 			"import": "./dist/index.mjs",
 			"types": "./dist/index.d.ts"
 		}
 	},
 	"scripts": {
-		"typegen": "tsup src/index.ts --dts-only",
-		"build": "tsup src/index.ts --format cjs,esm --dts",
+		"typegen": "tsup src/index.ts --dts-only && tsup src/index.edge.ts --dts-only",
+		"dev": "yarn build --watch",
+		"build": "tsup src/index.edge.ts --format cjs,esm --dts && tsup src/index.ts --format cjs,esm --dts",
```
Vadman97, line 23: did you mean to commit these changes? if so, we should bump the `version` references here and in all other affected sdk packages accordingly (and update changelog references). would you mind also updating the pr description about how this was tested with and without the app directory?

e2e/nextjs/src/app/utils/highlight.config.ts
Vadman97, line 1: should we consider updating the docs about where this file should go? right now, the docs recommend putting it at the top level of a customer's nextjs app

.vscode/settings.json
et: This is just going to affect the display in vscode:

Before:
Screenshot 2023-04-25 at 8 26 07 AM


After:
Screenshot 2023-04-25 at 8 25 49 AM

Notice the return values wrap at 120 columns.


Ideally, `gofmt` would handle wrapping the line for us but it looks like that's contentious (see https://github.com/golang/go/issues/11915). 

Googling around, I found [golines](https://github.com/segmentio/golines) ([blog post](https://yolken.net/blog/cleaner-go-code-golines)) which would extend `gofmt` to wrap the lines at a certain line length as well as formatting key/value pairs. 

I think there's a lot of merit to adding golines since we definitely have a lot of functions that have far too many params (that we should probably refactor into a struct) but may be difficult to change at the moment. Obviously adding that in now would be a lot of changes so adding it in as a vscode change is a nice middle ground in the short term.

PR #5107, opened 2023-04-25
Tweak doc tables style

Crawron:
## Summary

Adjust the table styles based on the design document.

image


jay-khatri (approved):
This looks great, except that the border radius doesn't match the border radius here: 
image

Can we change the table border radius to match that? 

Beyond that, looks great.

Vadman97 (approved):

Issue #4917, opened 2023-04-11
Redesign Settings Page(s)


Vadman97:
Consolidate the user, project, and workspace settings pages per new figma designs.
https://www.figma.com/file/rdBPNzn7Klk6RG1LHUCE5B/Screen-Designs?node-id=2737%3A99360&t=1O7dgx4sCuXQPrUv-1

things to address: 
- [ ] fix the white flash when we redirect to the settings page. 
- [ ] support multiple projects in the left side bar
- [ ] don't render the top right nav items for the settings page (github, discord, settings icon, etc..)

julian-highlight:
Updates on the Settings front @eightants 
https://www.loom.com/share/71dc85764d664a1bbc6cc9139961d7c5
https://www.figma.com/file/rdBPNzn7Klk6RG1LHUCE5B/Screen-Designs?node-id=2737-99360&t=50vTe5mzHuqEKfIX-11

PR #5097, opened 2023-04-24
Added sidebar to workspace & account settings

eightants:
## Summary

Initial work on implementing the new settings page from https://github.com/highlight/highlight/issues/4917. Since the new settings layout consolidates all 3 types of settings on one page, and also changes the content of some pages, this might be best done in a few stages for easy PR review. This PR adds the sidebar as shown in the Figma, and consolidates account and workspace settings on the same page. 

## How did you test this change?

Screenshot 2023-04-24 at 11 34 26 AM

## Are there any deployment considerations?

NA

PR #5110, opened 2023-04-25
Consolidated project settings into settings UI

eightants:
## Summary

Continuing work on new settings UI. 
- Consolidated project settings onto this page, multiple projects
- Removed top right buttons from settings ui
- Navbar only has the Back to Project button for all settings pages, including project settings. Initially the project settings page had Sessions, Errors, Logs, and all other settings on the navbar. Removed them to keep it consistent which I think makes sense?

Clicking around and changing settings to make sure they work. Also, checked that URLs are preserved.

Screenshot 2023-04-25 at 12 31 13 PM

NA


frontend/src/pages/SettingsRouter/SettingsRouter.tsx
```diff
@@ -180,6 +213,27 @@ export const SettingsRouter = () => {
...
+									[styles.menuItemActive]:
+										projectId === project.key,
+								})}
+							>
+								<Stack direction="row" align="center" gap="4">
+									<Text>{project.title}</Text>
+								</Stack>
+							</NavLink>
+						))}
+					</Stack>
```
Vadman97, line 236: in the [figma](https://www.figma.com/file/rdBPNzn7Klk6RG1LHUCE5B/Screen-Designs?type=design&node-id=7552%3A287169&t=CUu6aswgUI0QCnpi-1), this also has a color button on the side. could we add that?

frontend/src/pages/SettingsRouter/SettingsRouter.tsx
```diff
@@ -180,6 +213,27 @@ export const SettingsRouter = () => {
 							</NavLink>
 						))}
 					</Stack>
+					<Stack gap="0">
+						<Box mt="12" mb="10" ml="8">
```
Vadman97, line 217: @jay-khatri @julian-highlight does it make sense to have the left panel organized as `Account Settings`, `Workspace Settings`, `Project Settings` instead? more of a heirarchy of what belongs to what

Vadman97 (approved):

frontend/src/pages/SettingsRouter/SettingsRouter.tsx
```diff
@@ -31,7 +33,19 @@ export type WorkspaceSettingsTab =
 	| 'settings'
 	| 'current-plan'
 	| 'upgrade-plan'
-	| SettingGroups
+
+export const isSettingsPath = (path: string[]) => {
```
deltaepsilon, line 37: This always takes in a `path` value from `useLocation`, so it can easily be refactored into a custom hook to better encapsulate all of it's dependent logic.

frontend/src/pages/SettingsRouter/SettingsRouter.tsx
```diff
@@ -117,6 +143,13 @@ export const SettingsRouter = () => {
 		],
 	]
 
+	const projectSettingTabs = allProjects
```
deltaepsilon, line 146: **non-blocking**

These sorts of functions can be wrapped with a `useMemo`. The primary reason for `useMemo` is to maintain stable references. The secondary reason is to cut down on expensive operations.

This particular operation isn't particularly expensive, so it's up to you.

deltaepsilon (approved):

frontend/src/pages/SettingsRouter/SettingsRouter.tsx
eightants: What color should it be? Just some random color?

Vadman97: could we use the project name hashed to an int to look up from our color pallet? similar to the avatar code https://github.com/highlight/highlight/blob/09c50163505c36bcdeaaed265722c4cd0cc684a1/frontend/src/components/Avatar/Avatar.tsx#L78

eightants:
Addressed the comments and added the sidebar border from https://github.com/highlight/highlight/pull/5097

Will merge this Monday, since there wasn't any blocking changes when approved last week.

Issue #4890, opened 2023-04-07
improve chrome extension filter name


Vadman97:
Hey @d29107d  it's not super clear here what the name of the setting means.

Could you please update the setting title from `Filter Browser Extension` to  `Filter out errors thrown by chrome extensions` and completely remove the "Chrome extension stackframes" label and tooltip (i.e. only show the toggle)?

![image](https://user-images.githubusercontent.com/1351531/230687265-deb0aa08-5933-49d8-8b55-d0041165d332.png)

d29107d:
@Vadman97 ok

jay-khatri:
@d29107d I've just update the description slightly. fyi.

d29107d:
@Vadman97 @jay-khatri 
<img width="858" alt="image" src="https://user-images.githubusercontent.com/82889377/230713105-8b497697-78e9-4b39-ae7f-e36a8ae35b6f.png">
My idea is that there may be other filters(i.e. Firefox or Safari) here in the future.

jay-khatri:
That is true. Good point. For now, let's be more explicit, though.

d29107d:
> That is true. Good point. For now, let's be more explicit, though.

👌

d29107d:
@Vadman97 @jay-khatri 
<img width="872" alt="image" src="https://user-images.githubusercontent.com/82889377/231046212-f3dc289e-6ca3-415d-8038-3315249cd83d.png">
Does it like the above? If so, I will open a pr.

PR #5123, opened 2023-04-26
Filter sessions without an error setting

et:
## Summary

This PR adds the ability to filter sessions that do not have an error. 

The frontend for this PR adds a new project filter setting that works similar to how we filter Chrome extension errors (see #4890 and #4921).

Screenshot_2023-05-01_at_10_56_43_AM

The backend for this PR introduces a new table, `project_filter_settings` for storing various filter settings going forward (currently, they're just stuffed on the `project` model). 

The actual filtering works exactly how #5158 describes by simply checking to see if the session has an error.

Without setting checked, verified that all sessions are visible regardless if they have an error.

With setting checked
* session is excluded
* session is visible when an error is thrown

### DIFF
--- a/backend/model/model.go
+++ b/backend/model/model.go
@@ -191,6 +191,7 @@ var Models = []interface{}{
 	&SessionAdminsView{},
 	&ErrorGroupAdminsView{},
 	&LogAdminsView{},
+	&ProjectFilterSettings{},
 }
 
 func init() {
@@ -362,6 +363,12 @@ type SetupEvent struct {
 	CreatedAt time.Time            `json:"created_at" deep:"-"`
 	ProjectID int                  `gorm:"uniqueIndex:idx_project_id_type"`
 	Type      MarkBackendSetupType `gorm:"uniqueIndex:idx_project_id_type"`
+}
+
+type ProjectFilterSettings struct {
+	Model
+	ProjectID                  int  `gorm:"uniqueIndex"`
+	FilterSessionsWithoutError bool `gorm:"default:false"`
 }
 
 type HasSecret interface {
--- a/backend/model/model.go
+++ b/backend/model/model.go
@@ -364,10 +364,10 @@ type SetupEvent struct {
 	ProjectID int                  `gorm:"uniqueIndex:idx_project_id_type"`
 	Type      MarkBackendSetupType `gorm:"uniqueIndex:idx_project_id_type"`
 }
-
 type ProjectFilterSettings struct {
 	Model
-	ProjectID                  int  `gorm:"uniqueIndex"`
+	Project                    *Project
+	ProjectID                  int
 	FilterSessionsWithoutError bool `gorm:"default:false"`
 }
 
--- a/backend/public-graph/graph/resolver.go
+++ b/backend/public-graph/graph/resolver.go
@@ -2964,9 +2964,9 @@ func (r *Resolver) ProcessPayload(ctx context.Context, sessionSecureID string, e
 	updateSpan, _ := tracer.StartSpanFromContext(ctx, "public-graph.pushPayload", tracer.ResourceName("doSessionFieldsUpdate"))
 	defer updateSpan.Finish()
 
-	excluded, reason := r.isSessionExcluded(ctx, sessionObj)
+	excluded, reason := r.isSessionExcluded(ctx, sessionObj, sessionHasErrors)
 	if excluded {
-		log.WithContext(ctx).WithFields(log.Fields{"session_id": sessionObj.ID, "project_id": sessionObj.ProjectID, "reason": reason}).Infof("excluding session")
+		log.WithContext(ctx).WithFields(log.Fields{"session_id": sessionObj.ID, "project_id": sessionObj.ProjectID, "reason": *reason}).Infof("excluding session")
 	}
 
 	// Update only if any of these fields are changing
@@ -3031,7 +3031,7 @@ func (r *Resolver) ProcessPayload(ctx context.Context, sessionSecureID string, e
 	return nil
 }
 
-func (r *Resolver) isSessionExcluded(ctx context.Context, s *model.Session) (bool, *string) {
+func (r *Resolver) isSessionExcluded(ctx context.Context, s *model.Session, sessionHasErrors bool) (bool, *string) {
 	var project model.Project
 	if err := r.DB.Raw("SELECT * FROM projects WHERE id = ?;", s.ProjectID).Scan(&project).Error; err != nil {
 		log.WithContext(ctx).WithFields(log.Fields{"session_id": s.ID, "project_id": s.ProjectID, "identifier": s.Identifier}).Errorf("error fetching project for session: %v", err)
@@ -3042,7 +3042,7 @@ func (r *Resolver) isSessionExcluded(ctx context.Context, s *model.Session) (boo
 		return true, ptr.String("excluded user matches")
 	}
 
-	if r.isSessionExcludedForNoError(ctx, s, project) {
+	if r.isSessionExcludedForNoError(ctx, s, project, sessionHasErrors) {
 		return true, ptr.String("no associated error")
 	}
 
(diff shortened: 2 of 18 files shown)


backend/worker/worker.go
```diff
@@ -550,6 +566,19 @@ func (w *Worker) isSessionUserExcluded(ctx context.Context, s *model.Session) bo
 	return false
 }
 
+func (w *Worker) isSessionExcludedForNoError(ctx context.Context, s *model.Session, project model.Project) bool {
+	projectFilterSettings := model.ProjectFilterSettings{}
+
+	w.Resolver.DB.Where(model.ProjectFilterSettings{ProjectID: project.ID}).FirstOrCreate(&projectFilterSettings)
```
Vadman97, line 572: could we add a session `FilterSessionsWithoutError` column that is set in `InitializeSession` based on the `ProjectFilterSettings` value? that way we don't have to query the project filter settings here, and if the setting value changes between the initialization and the session processing (that window could be up to 4 hours, the length of the session), we'll have consistent behavior based on when the session was initialized.

Vadman97 (commented):
could we update `ProcessPayload` to determine whether a session should be un-excluded on events pushed based on the `FilterSessionsWithoutError` setting? that way, we won't show the live session in the feed until there is an error

https://github.com/highlight/highlight/blob/34a568681218b5929a8c55ba6a6e97f3dcf63652/backend/public-graph/graph/resolver.go#L2967-L3000

backend/worker/worker.go
et: Why does the db read matter if it's done here or upstream? And is the extra attribute on the session model really worth it for that edge case? Like we probably want to avoid polluting the exclusion logic to multiple places.
Also, I think we can make the argument that a customer would want the exclusion logic to take effect at the time of processing, not when the session started.

docs-content/general/6_product-features/1_session-replay/ignoring-sessions.md
```diff
@@ -10,19 +10,21 @@ updatedAt: 2022-06-27T03:34:47.000Z
 
 ## Ignore sessions by user identifier
 In some cases, you may want to ignore sessions from a specific user. You can do this by adding the user identifier to the "Excluded Sessions" input under the "Session Replay" tab in your [project settings](https://app.highlight.io/settings). Please note that we use the `identifier` (or first argument) sent in your `H.identify` method to ignore against (SDK docs [here](../../../sdk/client.md)).
+## Ignoring sessions without an error
```
et, line 14: Added to this [page](http://localhost:4000/docs/general/product-features/session-replay/ignoring-sessions):

Screenshot 2023-05-01 at 10 25 04 AM

backend/model/model.go
```diff
@@ -364,6 +365,12 @@ type SetupEvent struct {
 	Type      MarkBackendSetupType `gorm:"uniqueIndex:idx_project_id_type"`
 }
 
+type ProjectFilterSettings struct {
+	Model
+	ProjectID                  int  `gorm:"uniqueIndex"`
```
Vadman97, line 370: could remove the `Model` reference and make this could be the primary key to make it clear there is a 1-1 mapping with the `projects` table

backend/public-graph/graph/resolver.go
```diff
@@ -2963,9 +2964,9 @@ func (r *Resolver) ProcessPayload(ctx context.Context, sessionSecureID string, e
 	updateSpan, _ := tracer.StartSpanFromContext(ctx, "public-graph.pushPayload", tracer.ResourceName("doSessionFieldsUpdate"))
 	defer updateSpan.Finish()
 
-	excluded := r.isSessionUserExcluded(ctx, sessionObj)
+	excluded, reason := r.isSessionExcluded(ctx, sessionObj, sessionHasErrors)
 	if excluded {
-		log.WithContext(ctx).WithFields(log.Fields{"session_id": sessionObj.ID, "project_id": sessionObj.ProjectID, "identifier": sessionObj.Identifier}).Infof("excluding session due to excluded identifier")
+		log.WithContext(ctx).WithFields(log.Fields{"session_id": sessionObj.ID, "project_id": sessionObj.ProjectID, "reason": *reason}).Infof("excluding session")
```
Vadman97, line 2969: should be fine to log this. if the spam becomes a problem, we can clean this up.
guessing you removed the the `sessionObj.Identifier` logging because we also log it in `isSessionExcluded`?

backend/public-graph/graph/resolver.go
```diff
@@ -2963,9 +2964,9 @@ func (r *Resolver) ProcessPayload(ctx context.Context, sessionSecureID string, e
 	updateSpan, _ := tracer.StartSpanFromContext(ctx, "public-graph.pushPayload", tracer.ResourceName("doSessionFieldsUpdate"))
 	defer updateSpan.Finish()
 
-	excluded := r.isSessionUserExcluded(ctx, sessionObj)
+	excluded, reason := r.isSessionExcluded(ctx, sessionObj, sessionHasErrors)
```
Vadman97, line 2967: `sessionHasErrors` is only looking at the current batch of errors. 
```sessionHasErrors := len(errors) > 0```
with the logic below
```
        doUpdate := ... ||
		(sessionHasErrors && (sessionObj.HasErrors == nil || !*sessionObj.HasErrors))
	if doUpdate && !excluded {
		fieldsToUpdate := model.Session{
			PayloadUpdatedAt: &now, BeaconTime: beaconTime, HasUnloaded: hasSessionUnloaded, Processed: &model.F, ObjectStorageEnabled: &model.F, DirectDownloadEnabled: false, Chunked: &model.F, Excluded: false,
		}
```
that means that a session will not be updated by `PushPayload` until it gets an error (when the `FilterSessionsWithoutError` setting is enabled), right? i don't think that's right, as subsequent push payloads may not have an error.

ie, consider a project with `FilterSessionsWithoutError=true`
1. PushPayload with no errors -> does not update the session
2. PushPayload with an error -> updates the session
3. PushPayload with no errors -> does not update the session
4. PushPayload with no errors -> does not update the session
`PushPayloads` 3 and 4 will not update the payload time, so the session might be missing events in processing.

File: docs-content/general/6_product-features/2_error-monitoring/ignoring-errors.md [hunk 6]
```diff
@@ -9,8 +9,8 @@ updatedAt: 2022-06-27T03:34:47.000Z
 
 There are several options for ignoring errors, all of which can be found in the "Error Monitoring" tab of your [project settings](https://app.highlight.io/settings). Details on each option are below.
 
-## Ignore errors emmitted by chrome extensions
-If your users are using chrome extensions, you may see errors that are not relevant to your application. You can ignore these errors by checking the "Filter out errors thrown by chrome extensions" box in your [project settings](https://app.highlight.io/settings).
+## Ignore errors emitted by Chrome extensions
+If your users are using Chrome extensions, you may see errors that are not relevant to your application. You can ignore these errors by checking the "Filter errors thrown by Chrome extensions" box in your [project settings](https://app.highlight.io/settings).
```
Vadman97, line 13: thanks for the cleanup here!

backend/public-graph/graph/resolver.go
et: > sessionHasErrors is only looking at the current batch of errors.
sessionHasErrors := len(errors) > 0

Are you considering this block of code as well:

https://github.com/highlight/highlight/blob/cdc89e81676d125389a73cf0d5b17296ea9a6108/backend/public-graph/graph/resolver.go#L2954-L2961

et: We walked through the code together and the current implementation looks to be good.

backend/model/model.go
et: What's the syntax here? Per the [docs](https://gorm.io/docs/belongs_to.html), I tried changing to:

```go
type ProjectFilterSettings struct {
	Model
	ProjectID                  int
	Project                    Project
	FilterSessionsWithoutError bool `gorm:"default:false"`
}
```

and got this error:

```
2023/05/04 14:38:48 /Users/ericthomas/code/highlight/backend/model/model.go:1334
[error] invalid field found for struct github.com/highlight-run/highlight/backend/model.ProjectFilterSettings's field ProjectID, need to define a valid foreign key for relations or it need to implement the Valuer/Scanner interface
```

Vadman97: i think we'd want this to be 
```go
type ProjectFilterSettings struct {
	Project                     *Project
	ProjectID                  int `gorm:"foreignKey:ID;"`
	FilterSessionsWithoutError bool `gorm:"default:false"`
}
```

Vadman97 (approved):
looks good overall! :ship:

backend/model/model.go
et: Thanks. I guess we need a pointer here (c22dbd610c60cb5f6427222af3f8eac080de880e). 
I kept `Model` because it's nice to have the created/updated at columns.

PR #5127, opened 2023-04-26
fix incorrect usage of lazy query

et:
## Summary

There's an interesting bug whereby we can't scroll when changing the `query`. To reproduce:

* Visit the logs page and set the time frame to something small (Last 15 minutes is sufficient for local, you may need something smaller for prod)
* You should have more than one page of logs
* Set the query to: `level:trace`
* Keep scrolling until you hit the end of all trace logs
* Clear the query
* Observe that you cannot scroll past the first page of logs

The reason this was happening was because I initially configured the lazy query in #4535 to apply the variables at the instantiation of the lazy query, not when I actually call the query (`getLogs()`). This is a gotcha with apollo (see [#5912](https://github.com/apollographql/apollo-client/issues/5912)). 

The reason why we even use a lazy query in the first place is because changing the window info is dependent on what we're doing:
* first time the page loads, set window info from the results of invoking the lazy query
* variable (such as `query`, start time, end time) changes, set window info from the results of invoking the lazy query
* scrolling to the next page (or previous page), set window info based on the [`fetchMore`](https://www.apollographql.com/docs/react/pagination/core-api/#the-fetchmore-function) function

Per the test above, we should be able to scroll past that first page of logs after clearing the query.

N/A

### DIFF
--- a/frontend/src/pages/LogsPage/LogsTable/LogDetails.tsx
+++ b/frontend/src/pages/LogsPage/LogsTable/LogDetails.tsx
@@ -44,9 +44,10 @@ type Props = {
 	matchedAttributes: ReturnType<typeof findMatchingLogAttributes>
 }
 
-export const getLogURL = (row: Row<LogEdge>) => {
+export const getLogURL = (projectId: string, row: Row<LogEdge>) => {
 	const currentUrl = new URL(window.location.href)
-	const path = generatePath('/logs/:log_cursor', {
+	const path = generatePath('/:project_id/logs/:log_cursor', {
+		project_id: projectId,
 		log_cursor: row.original.cursor,
 	})
 	return currentUrl.origin + path
@@ -226,7 +227,7 @@ export const LogDetails: React.FC<Props> = ({
 						kind="secondary"
 						emphasis="low"
 						onClick={(e) => {
-							const url = getLogURL(row)
+							const url = getLogURL(projectId, row)
 							e.stopPropagation()
 							navigator.clipboard.writeText(url)
 							antdMessage.success('Copied link!')
--- a/frontend/src/pages/LogsPage/useGetLogs.ts
+++ b/frontend/src/pages/LogsPage/useGetLogs.ts
@@ -52,7 +52,7 @@ export const useGetLogs = ({
 
 	useEffect(() => {
 		setWindowInfo(initialWindowInfo)
-	}, [query])
+	}, [query, startDate, endDate])
 
 	const { data, loading, error, refetch, fetchMore } = useGetLogsQuery({
 		variables: {
--- a/frontend/src/pages/LogsPage/useGetLogs.ts
+++ b/frontend/src/pages/LogsPage/useGetLogs.ts
@@ -7,7 +7,7 @@ import {
 	parseLogsQuery,
 } from '@pages/LogsPage/SearchForm/utils'
 import moment from 'moment'
-import { useCallback, useEffect, useState } from 'react'
+import { useCallback, useState } from 'react'
 
 export type LogEdgeWithError = LogEdge & {
 	error_object?: Pick<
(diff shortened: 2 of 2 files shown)


Vadman97 (commented):
thanks for looking at this! 

the original issue is fixed but there are two new issues:
1. when querying a super-small time range, i see paginated results from some previous query (results that should definitely not be part of the result set)
2. i'm able to consistently repro the issue with out-of-order page results 

https://www.loom.com/share/71907c5e54b543218e4f3c390e0b6354

Vadman97:
hey @et, just wanted to bump this as I keep running into this in the logging product. thanks!

Vadman97:
friendly bump here, let me know if i can help with reproducing

et:
@Vadman97 - I rewrote this to:
* avoid the use of a lazy query
* avoid `windowInfo`. This was only needed to fetch the first and last cursor (in case we span multiple pages), but we can get this from whatever's in `data` since it includes results from previous requests.
* put a guard in place so subsequent requests to fetch more results will return early if there's an outgoing request. I think there's still a separate bug with relying on the scroll event (and calling too many times) that we could probably fix with an intersection observer but that can be tackled separately.

Please this this out when you have a chance. Thanks!

frontend/src/pages/LogsPage/LogsTable/LogDetails.tsx
```diff
@@ -226,7 +227,7 @@ export const LogDetails: React.FC<Props> = ({
 						kind="secondary"
 						emphasis="low"
 						onClick={(e) => {
-							const url = getLogURL(row)
+							const url = getLogURL(projectId, row)
```
et, line 230: Unrelated to this actual issue but we need to attach the project id to the permalink log. 
Permalinking is pretty darn slow (a distinct backend issue) but I figured I'd at least fix the frontend problem.

frontend/src/pages/LogsPage/useGetLogs.ts
```diff
@@ -29,111 +29,97 @@ export const useGetLogs = ({
...
-	// If the user scrolls forward to get the next 100 logs, the server will say that hasPreviousPage is `true` since we're on page 2.
-	// Hence, we track the initial information (where "window" is effectively multiple pages) to ensure we aren't making requests unnecessarily.
-	const [windowInfo, setWindowInfo] = useState<PageInfo>({
-		hasNextPage: false,
-		hasPreviousPage: false,
-		startCursor: '',
-		endCursor: '',
-	})
+	const [hasNextPage, setHasNextPage] = useState(true)
+	const [hasPreviousPage, setHasPreviousPage] = useState(false)
```
Vadman97, line 33: looks like the changes have fixed one bug where we would have incorrect results on follow-up queries, but we still have an issue where a query has no pages even when it should: https://www.loom.com/share/f5424be4731649b49fc640ee9752aadd

i think the culprit is that these two states need to be reset when the `serverQuery` or `startDate` and `endDate` change, as those values should retrigger the primary page query. currently, we set `hasNextPage` to a default `true` value on the intial load of this hook, but maybe these two values should be undefined and updated on `data` changing?

et:
@Vadman97 - I updated this to reset the page info whenever `query` changes. Relying on `data` being changed isn't reliable since it also updates for requests to `fetchMore`.
I also updated the default values for hasNextPage and hasPreviousPage to be `true`. Since we can't rely on that initial data request to populate the values, we have to just assume there's a next page. This means we may make an unnecessary request when fetching the next page, but the page info state will resolve itself when `fetchMore` is called.

I did some testing and this seems to solve the issue for me 🤞  Please give it another shot when you have a chance.

frontend/src/pages/LogsPage/useGetLogs.ts
```diff
@@ -37,103 +44,105 @@ export const useGetLogs = ({
...
-				direction: Types.LogDirection.Desc,
-				params: {
-					query: serverQuery,
-					date_range: {
-						start_date: moment(startDate).format(LOG_TIME_FORMAT),
-						end_date: moment(endDate).format(LOG_TIME_FORMAT),
-					},
+	useEffect(() => {
+		setWindowInfo(initialWindowInfo)
+	}, [query])
```
Vadman97, line 55: should this also reset when the `startDate` and `endDate` change, since those drive the initial `GetLogs` query?

https://www.loom.com/share/3b9c1aabc3ef497e8ef2937a257ed53c

Vadman97 (approved):

PR #5133, opened 2023-04-26
apply missing mui stylesheets

Vadman97:
## Summary

Inject MUI stylesheets for a customer.
This is a workaround for a problem in rrweb (which causes the stylesheets to be injected broken and repeated).

Pulling in a known broken session auH6iXPRSNSscJuiPwuJ2SFMeH8q
https://www.loom.com/share/13013fdbc8c04897889dd3feebad02f8

No, gated to customer's project.


et (approved):
Thanks for making the loom. Weird behavior but this is definitely good to unblock the customer.

Issue #4980, opened 2023-04-13
EM/SR: only record sessions that have errors


Vadman97:

Vadman97:
Discussed in Slack. Leaving my implementation thoughts here as well for posterity:

the simplest way to handle the exclusion:
* when the error-only-sessions mode is enabled, mark the session as `excluded=true` on every `PushPayload` while storing data as normal, but on `PushPayload` that contains an `error`, mark the session as `excluded=false, processed=false` so that the session becomes shown in the feed and is processed.
this is the most straightforward because we already have some logic that handles the state transitions (including logic on the billing end that will ignore these sessions from quotas) and you just need to toggle the flag. the downside - we would still be recording all the data and incurring all associated processing costs (though we won't incur storage costs in s3 since `excluded=true` sessions are not processed).

another way - along the lines you mentioned
* record the session as normal. when our session processing worker calls `ProcessSession` to store the session in s3, we can analyze whether it had any errors, and determine whether to keep or delete it.
the problem - we will still show such sessions in the feed until we determine whether they finish so we can determine whether they have errors or not. a customer might be confused since they wouldn't expect these sessions to be recorded, but they would still show in the sessions feed while live.

the cleanest way from the perspective of ingest cost
* record the session client-side-only (`client/index.tsx` pushing messages and storing them in the web worker or in indexeddb browser-side), and on client error, trigger an upload of the entire session and record as normal going forward
this requires the most browser-side changes. this also introduces a big problem of browser memory limits and upload speed constraints. ie. if the error happens near the end of a session, but the tab is closed shortly after, we might not have enough time to upload the entire session before the tab is closed. we might not even have enough browser memory to store the session client-side (unless we use indexeddb to store it).

> where would we actually do the deletion? i think i'm supposed to trigger off of `hasSessionUnloaded = true`...
i think `hasSessionUnloaded` has to do with beacon payloads marking whether they browser gracefully ended the session or not, and is likely not what you want here. we should go with the first option and just use the `excluded` flag to mark the session as one that is not going to be stored, so you don't actually have to worry about deleting it

PR #5158, opened 2023-04-27
Move excluded users up to ProcessPayload to avoid disappearing sessions

et:
## Summary

_This is some initial set up work to make #4980 easier._

This PR moves the excluded users logic from the worker process session up to ProcessPayload. 

The previous logic worked like this:
* `InitializeSession` (excluded = true)
* `ProcessPayload` (excluded = false)
* Session is visible in the feed
* Session finishes
* Process session kicks off
* `isSessionUserExcluded = true` and the session is marked as `excluded = false` and removed from the feed

Now the logic works like this:
The previous logic worked like this:
* `InitializeSession` (excluded = true)
* `ProcessPayload` (excluded = false **unless `isSessionUserExcluded = true`**)
* Session is **not** visible in the feed
* Session finishes
* Process session kicks off

* Confirmed that using the [excluded users setting](https://localhost:3000/1/settings) never shows the session in the feed (and that the db marked `excluded=true` on the session record).
* Confirmed that removing the [excluded users setting](https://localhost:3000/1/settings) does show in the feed (and that the db marked `excluded=false` on the session record).

N/A

### DIFF
--- a/backend/model/model.go
+++ b/backend/model/model.go
@@ -634,7 +634,7 @@ type Session struct {
 	VerboseID             string  `json:"verbose_id"`
 
 	// Excluded will be true when we would typically have deleted the session
-	Excluded *bool `gorm:"default:false"`
+	Excluded bool `gorm:"default:false"`
 
 	// Lock is the timestamp at which a session was locked
 	// - when selecting sessions, ignore Locks that are > 10 minutes old
--- a/backend/public-graph/graph/resolver.go
+++ b/backend/public-graph/graph/resolver.go
@@ -4,8 +4,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/highlight-run/highlight/backend/phonehome"
-	"go.opentelemetry.io/otel/attribute"
 	"hash/fnv"
 	"io"
 	"net/http"
@@ -16,6 +14,9 @@ import (
 	"strconv"
 	"strings"
 	"time"
+
+	"github.com/highlight-run/highlight/backend/phonehome"
+	"go.opentelemetry.io/otel/attribute"
 
 	"github.com/PaesslerAG/jsonpath"
 	"github.com/highlight-run/go-resthooks"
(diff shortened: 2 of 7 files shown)


backend/model/model.go
```diff
@@ -634,7 +634,7 @@ type Session struct {
 	VerboseID             string  `json:"verbose_id"`
 
 	// Excluded will be true when we would typically have deleted the session
-	Excluded *bool `gorm:"default:false"`
```
et, line 637: We don't need this to be a pointer since the default value is false (both at the database level and with the Go zero value). By not using a pointer, it makes the code way more readable, see below ⬇️

backend/public-graph/graph/resolver.go
```diff
@@ -3026,6 +3032,43 @@ func (r *Resolver) ProcessPayload(ctx context.Context, sessionSecureID string, e
 	return nil
 }
 
+func (r *Resolver) isSessionUserExcluded(ctx context.Context, s *model.Session) bool {
```
et, line 3035: This logic was simply cut/pasted from `worker.go`.

backend/public-graph/graph/resolver.go
```diff
@@ -2971,12 +2978,11 @@ func (r *Resolver) ProcessPayload(ctx context.Context, sessionSecureID string, e
 		(sessionObj.Processed != nil && *sessionObj.Processed) ||
 		(sessionObj.ObjectStorageEnabled != nil && *sessionObj.ObjectStorageEnabled) ||
 		(sessionObj.Chunked != nil && *sessionObj.Chunked) ||
-		(sessionObj.Excluded != nil && *sessionObj.Excluded) ||
 		(sessionHasErrors && (sessionObj.HasErrors == nil || !*sessionObj.HasErrors))
 
-	if doUpdate {
+	if doUpdate && !excluded {
```
et, line 2983: Because we don't update for excluded sessions, we'll cut down on unnecessary writes on the session object.

Per this query:

```sql
select id, excluded_users
from projects
where excluded_users is not null
and excluded_users != '{}'
```

we have 42 projects using this feature.

et:
@deltaepsilon - I rotated the reviewer since this is kind of a lot to review on just your second week 😉

backend/public-graph/graph/resolver.go
et: Not sure how much of a difference it'll make, but we can look at the p99 (or max) latency for this [trace](https://app.datadoghq.com/apm/traces?query=service%3Aworker-service%20operation_name%3AprocessPublicWorkerMessage&cols=core_service%2Ccore_resource_name%2Clog_duration%2Clog_http.method%2Clog_http.status_code&historicalData=true&messageDisplay=inline&sort=desc&spanType=all&start=1682026001766&end=1682630801766&paused=false) after this is rolled out.

ccschmitz (approved):

PR #5162, opened 2023-04-27
Use configurable project for demo account

ccschmitz:
## Summary

Revamps the setup for our demo project, making if configurable to a real Highlight project.

* Allows you to set a real project as the demo project and expose it to all logged in users by setting the `DEMO_PROJECT_ID`
* Remaps all the GraphQL variables we might pass to the backend with `project_id = 'demo'` to `project_id = DEMO_PROJECT_ID`
* Adds a redirect so any links going to `/${DEMO_PROJECT_ID}/*` are redirected to `/demo/*`
* Updates all admin checks on mutations to ensure only workspace admins can mutate data

* Created a new workspace
* Spun up a test app locally and sent some sessions to it
* Set `DEMO_PROJECT_ID` to the project ID in my new workspace (in the `docker/.env` and in Doppler) then restarted the app
* Removed my `workspace_admins` record for my user for the new workspace to ensure I wasn't an admin
* Loaded the demo project and ensured I saw the banner across the top and that I could view everything
* Added some early `return false`es in `isWhitelistedAccount` and `isDemoProject` to ensure I got the right UI
* Monitored network requests to ensure nothing was erroring (many requests were failing initially because we were sending `demo` as a project ID)

* Need to make sure `DEMO_PROJECT_ID` is set
* Should also set `DEMO_PROJECT_ID` in all the dev environments
* May want to figure out a way to test the demo account more easily locally 🤔


package.json
```diff
@@ -35,7 +35,7 @@
 		"dev": "run-p --print-label --race 'dev:*'",
 		"dev:backend": "cd backend && make start",
 		"dev:components": "yarn turbo run dev --filter @highlight-run/component-preview...",
-		"dev:frontend": "doppler run -- yarn turbo run dev --filter frontend... --filter=!@highlight-run/rrweb...",
+		"dev:frontend": "doppler run -- yarn turbo run dev --filter frontend... --filter=!@highlight-run/rrweb... --filter=!@highlight-run/ui...",
```
ccschmitz, line 38: The storybook build wasn't working for me so I decided to filter this out for now.

docker/.env
```diff
@@ -39,3 +37,6 @@ REACT_APP_PUBLIC_GRAPH_URI=https://localhost:8082/public
 REDIS_EVENTS_STAGING_ENDPOINT=redis:6379
 RENDER_PREVIEW=true
 SESSION_FILE_PATH_PREFIX=/tmp/
+
+# Uncomment and update ID to enable
```
Vadman97, line 41: a bit vague what enabling means in this case, mind clarifying in the comment?

package.json
Vadman97: thanks, i think one of the dependency upgrades a while back managed to break it  :(

backend/private-graph/graph/resolver.go
```diff
@@ -298,14 +300,23 @@ func (r *Resolver) isWhitelistedAccount(ctx context.Context) bool {
...
 
 func (r *Resolver) isDemoWorkspace(workspace_id int) bool {
 	return workspace_id == 0
 }
 
+func (r *Resolver) demoProjectID(ctx context.Context) int {
+	if demoProjectID, err := strconv.Atoi(DemoProjectID); err != nil {
+		log.WithContext(ctx).Error(err, "error converting DemoProjectID to int")
+		return 0
+	} else {
```
Vadman97, line 315: if `DEMO_PROJECT_ID` is not set, should we `return 0, e.New('')` here and have `isDemoProject` return false? otherwise, this logic implies that if it is not set, we fallback to demo project 0

docker/.env
```diff
@@ -39,3 +37,6 @@ REACT_APP_PUBLIC_GRAPH_URI=https://localhost:8082/public
...
+# DEMO_PROJECT_ID=0
```
Vadman97, line 42: this would effectively disable the demo for the hobby deploy, right? i think that is the behavior we want. should we make that clear in the comment?

Vadman97 (approved):
looks great locally! i had to `doppler secrets set DEMO_PROJECT_ID 1` locally to use my only project for the demo, so perhaps we should default to that value for the `dev` environments since folks wont have project 1344?

backend/private-graph/graph/resolver.go
ccschmitz: Yeah, I guess I figured `0` would be the same as disabled since there would be no project with that ID. I decided to log rather than erroring - do you think that's appropriate? Happy to change how this works if you think there is a better way to handle!

ccschmitz:
> i had to doppler secrets set DEMO_PROJECT_ID 1 locally to use my only project for the demo, so perhaps we should default to that value for the dev environments since folks wont have project 1344?

@Vadman97 good callout for the dev value. I just made it `2` in doppler and in the `.env` file. I didn't want to make the default project a demo project though since I thought that could be a little confusing/anoying.

docker/.env
ccschmitz: Yeah, without a valid project ID it would effectively be disabled. I'll note that in the comment above.

Vadman97: cool, thanks!

backend/private-graph/graph/resolver.go
Vadman97: given the comment now, i think this is fine! there isn't a way for a project to ever be 0 so it wouldn't ever match.

Vadman97 (approved):
awesome, thanks chris! :ship:

PR #5167, opened 2023-04-27
avoid duplicate errors from error boundary and show dialog by default

Vadman97:
## Summary

* Disables error reporting to highlight by the error boundary when highlight is present because highlight will be catching the same error via the `window.onerror` listener, resulting in duplicate errors. ie. 
image
* Changes the ErrorHandler props to default to `showDialog` and add an optional `noShowDialog` as we've heard from customers that the props are un-intuitive (the dialog is the whole point of the error boundary to begin with).

buttons page
image
only one error recorded
image

The `showDialog` is kept to keep this change non-breaking. Since we just released the major version 3.x today, thinking we can still sneak this in as 3.0.1 though the behavior of the dialog has changed.

### DIFF
--- a/sdk/highlight-react/src/components/ErrorBoundary.tsx
+++ b/sdk/highlight-react/src/components/ErrorBoundary.tsx
@@ -9,15 +9,13 @@ export type FallbackRender = (errorData: {
 
 export type ErrorBoundaryProps = {
 	children?: React.ReactNode
-	/** Deprecated: no-op for compatibility. Dialog is enabled by default, unless {@link noShowDialog} is true. */
+	/** If a Highlight report dialog should be rendered on error. Defaults to true. */
 	showDialog?: boolean
-	/** Set true to disable the Highlight report dialog being rendered on error  */
-	noShowDialog?: boolean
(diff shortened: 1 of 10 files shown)


sdk/highlight-react/src/components/ErrorBoundary.tsx
```diff
@@ -71,7 +73,7 @@ export class ErrorBoundary extends React.Component<
 		if (onError) {
 			onError(error, errorInfo.componentStack)
 		}
-		if (showDialog) {
+		if (!noShowDialog) {
```
et, line 76: Not a huge fan of prefixing a prop with "no".

What if we changed the default value of showDialog to true and if someone wanted to opt out, they could set it to false.

I get that would be a breaking change but I tend to think this is already a breaking change since the code is no longer referencing showDialog.

Vadman97: that's a great point, should have thought of that :man_facepalming: 
updated!

showDialog=true
image

showDialog=false
image

showDialog undefined
image

et (approved):
🎉

PR #5173, opened 2023-04-28
join on delete_sessions_tasks table in delete sessions / fields queries

mayberryzane:
## Summary
- the delete from `session_fields` query is timing out even when there are no records left to delete
- this new form of the query selecting from the `delete_sessions_tasks` table completes in `.093s` right now (for 0 rows to delete)

- ran query against prod

- no

### DIFF
--- a/backend/lambda-functions/deleteSessions/main.go
+++ b/backend/lambda-functions/deleteSessions/main.go
@@ -8,6 +8,7 @@ import (
 	"github.com/highlight-run/highlight/backend/lambda-functions/deleteSessions/handlers"
 	"github.com/highlight-run/highlight/backend/lambda-functions/deleteSessions/utils"
 	"github.com/highlight-run/highlight/backend/util"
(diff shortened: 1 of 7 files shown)


backend/lambda-functions/deleteSessions/main.go
```diff
@@ -16,6 +18,11 @@ func main() {
 		return
 	}
 
+	H.SetProjectID("1jdkoe52")
```
mayberryzane, line 21: @Vadman97 just added these lines here to start logging, anything else we need to do?

Vadman97: looks good! could we put this in the other lambdas too? you could also put this in `NewHandlers()` but then the `Stop()` won't be easy to add, so maybe just putting it in all the lambdas' `main` fns makes sense.

Vadman97 (approved):

PR #5175, opened 2023-04-28
React Code Guidlines

deltaepsilon:
Just some discussion points. Doesn't need to be merged.

### DIFF
--- a/frontend/CODE_GUIDELINES.md
+++ b/frontend/CODE_GUIDELINES.md
@@ -1,0 +1,105 @@
+# Code Guidelines for React
+
+Hat tip to Clean Code by Robert C. Martin (Uncle Bob)
+
+## Goals
+
+- Consistency
+- Predictability
+- Maintainability
+- Readability
+
+## Style
+
+- Take time to craft great variable and function names. It's ok to be verbose. Avoid abbreviations.
+- Replace magic numbers with named constants.
+- Remove dead code aggressively. We can always recover it from version control.
+- Use vertical whitespace to aid readability.
+  - Related lines should be vertically close.
+  - A block of variable declarations should be followed by a blank line.
+  - Use a blank line above return statements.
+  - Don't stack code too tight vertically, especially `if` blocks.
+- Use Prettier/ESLint to avoid bike-shedding.
+- Prefer copy/paste to premature abstraction.
+  - Abstractions tend to reveal themselves quickly on the second or third copy/paste.
+  - It's easier to abstract code than to remove an abstraction.
+- Break switch statements into their own functions with returns rather than `break;`.
+
+## File structure
+
+- Higher-order functions go at the top of the file when possible.
+- Named exports are preferred to default exports. It's easier to control naming.
+- Inline exports are preferred to the define-then-export pattern.
+
+## Testing
+
+Flakey tests are the worst. The following patterns ensure that tests don't interact.
+
+- Prefer many `describe` blocks.
+- Setup and teardown belong in `beforeEach`/`afterEach` blocks if possible. 
+- `beforeAll`/`afterAll` can be a reasonable fallback.
+- Some setup may be required in `it` blocks, but try to minimize it
+- Prefer `uuid` strings to `testing-testing-123`-style strings.
+- Don't over-test the DOM structure. 
+  - Focus on testing functionality.
+  - Simple components need tests too, mostly to make sure we don't break their imports.
+  - Testing is as much about easy refactors as anything. We need to be able to refactor confidently.
+- Don't be afraid to instrument your code for testing purposes.
+- Get good at mocking imports.
+  - Declare your mocks variables at the top of the file with the `let mock*` pattern.
+  - `jest.mock()` calls should follow the `let mock*` declarations.
+  - Jest automatically hoists your `jest.mock()` calls to the top of the file, so avoid confusion by keeping them near the top of the file. 
+  - Avoid `mockFn.mockImplementation` and `mockFn.mockImplementationOnce`. They're not **bad**, but they're more verbose and a bit harder to read than the `let mock*` pattern.
+
+## Javascript
+
+- Prefer `const` over `let`. Avoid `var`.
+- Prefer named functions, i.e., `function useSomething(){}` over anonymous functions.
+- Arrow functions are best for one-liners.
+- Write functional JS. `class` and `this` are unnecessary in most modern JavaScript.
+- Avoid mutating objects. This is bug-avoidance-rule-number-one in JavaScript.
+  - Object mutation is fine if it's visually close to the object declaration.
+  - There are times when object mutation simplifies the code, but BE CAREFUL AND EXPLICIT.
+- Be wary of using closure-scope for variable access. It's not functional and couples functions together, making refactors difficult.
+- Prefer "dot-syntax" over excessive destructuring.
+  - Well-crafted destructures can be very readable.
+  - Too much destructuring can kill your readability.
+  - For instance, always use `import constants from '~/constants';` This enables confident find/replace changes to the constants, because you can search for `constants.SOME_KEY`.
+  - Destructuring removes the context of the variable. The reader no longer knows where it came from. This is fine in vertically-adjacent code... but it can be hard to read otherwise.
+
+## Functions
+
+- Prefer a single return. There are definitely exceptions, but early returns can make debugging difficult.
+- Try to write small, single-purpose functions. Break functions down as much as is reasonable.
+- Prefer single-argument functions. 
+  - If multiple inputs are necessary, pass them in as a single object that can be destructured in the function signature.
+  - This is just a guideline, but fewer arguments is better.
+- Avoid nested conditionals when reasonable.
+
+## Comments
+
+- Comments lie and should be used sparingly.
+- If you're changing commented code, make sure to review and edit the comments carefully.
+- If you're about to commit a comment, see if you can avoid it with the following strategies:
+  - Replace comments with well-named, single-purpose functions where possible.
+  - Break multiple, ambiguous operations out into separate variable declarations.
+  - Prefer single variables in `if` and `switch` statements over inline operations.
+
+## React
+
+- Use hooks liberally.
+- Increase readability by extracting functionality from render functions out into private hooks.
+- Prop drilling is fine for one or two levels, but use a context before it gets out of hand.
+- Create "connected" functions explicitly in the same file as the main component.
+  - The connected function should be the default export
+  - Don't test the connected function.
+  - Always consume the context via a hook. Contexts are miserable to mock, but hooks are easy.
+- Try to keep render functions smallish.
+  - This is a balancing act, but large DOM trees in a single function can be hard to read.
+  - It's hard to break a render function into too many sub-components.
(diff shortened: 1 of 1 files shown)


frontend/CODE_GUIDELINES.md
```diff
@@ -0,0 +1,105 @@
...
+## Goals
+
+- Consistency
+- Predictability
+- Maintainability
+- Readability
+
+## Style
+
+- Take time to craft great variable and function names. It's ok to be verbose. Avoid abbreviations.
```
SpennyNDaJets, line 14: ++, hopefully this should also reduce the need for comments if everything is named carefully

frontend/CODE_GUIDELINES.md
```diff
@@ -0,0 +1,105 @@
...
+- Replace magic numbers with named constants.
```
SpennyNDaJets, line 15: 🔥

frontend/CODE_GUIDELINES.md
```diff
@@ -0,0 +1,105 @@
...
+- Use vertical whitespace to aid readability.
+  - Related lines should be vertically close.
+  - A block of variable declarations should be followed by a blank line.
+  - Use a blank line above return statements.
+  - Don't stack code too tight vertically, especially `if` blocks.
+- Use Prettier/ESLint to avoid bike-shedding.
+- Prefer copy/paste to premature abstraction.
+  - Abstractions tend to reveal themselves quickly on the second or third copy/paste.
+  - It's easier to abstract code than to remove an abstraction.
+- Break switch statements into their own functions with returns rather than `break;`.
```
SpennyNDaJets, line 26: I don't feel too strongly about this, but could be convinced

frontend/CODE_GUIDELINES.md
```diff
@@ -0,0 +1,105 @@
...
+
+## File structure
+
+- Higher-order functions go at the top of the file when possible.
+- Named exports are preferred to default exports. It's easier to control naming.
```
SpennyNDaJets, line 31: 🔥 - yes please

frontend/CODE_GUIDELINES.md
```diff
@@ -0,0 +1,105 @@
...
+- Inline exports are preferred to the define-then-export pattern.
```
SpennyNDaJets, line 32: I sort of like having all the exports at the bottom in one statement, but could see this making it easy to know what is private and public inline. I think explicitly making exports at the bottom could reduce the amount of methods that are exported based on copying/pasting another method.

```
const sum = (number1,  number2) => printNumber(number1 + number2)

const subtract = (minuend, subtrahend) => printNumber(minuend - subtrahend)

const printNumber = (number) => console.log(number)

export { sum, subtract }
```

frontend/CODE_GUIDELINES.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Testing
+
+Flakey tests are the worst. The following patterns ensure that tests don't interact.
+
+- Prefer many `describe` blocks.
+- Setup and teardown belong in `beforeEach`/`afterEach` blocks if possible. 
+- `beforeAll`/`afterAll` can be a reasonable fallback.
+- Some setup may be required in `it` blocks, but try to minimize it
+- Prefer `uuid` strings to `testing-testing-123`-style strings.
```
SpennyNDaJets, line 42: I'm guilty here, but like this

frontend/CODE_GUIDELINES.md
```diff
@@ -0,0 +1,105 @@
...
+
+## File structure
+
+- Higher-order functions go at the top of the file when possible.
+- Named exports are preferred to default exports. It's easier to control naming.
+- Inline exports are preferred to the define-then-export pattern.
+
+## Testing
+
+Flakey tests are the worst. The following patterns ensure that tests don't interact.
```
SpennyNDaJets, line 36: Wonder if there is anything we should call out related to tests involving times and timezones. Run into quite a few flakey specs that start failing because a certain date was past or daylight savings occurred

frontend/CODE_GUIDELINES.md
```diff
@@ -0,0 +1,105 @@
...
+- Don't be afraid to instrument your code for testing purposes.
+- Get good at mocking imports.
+  - Declare your mocks variables at the top of the file with the `let mock*` pattern.
+  - `jest.mock()` calls should follow the `let mock*` declarations.
+  - Jest automatically hoists your `jest.mock()` calls to the top of the file, so avoid confusion by keeping them near the top of the file. 
+  - Avoid `mockFn.mockImplementation` and `mockFn.mockImplementationOnce`. They're not **bad**, but they're more verbose and a bit harder to read than the `let mock*` pattern.
+
+## Javascript
+
+- Prefer `const` over `let`. Avoid `var`.
```
SpennyNDaJets, line 56: ++

frontend/CODE_GUIDELINES.md
```diff
@@ -0,0 +1,105 @@
...
+- Prefer named functions, i.e., `function useSomething(){}` over anonymous functions.
+- Arrow functions are best for one-liners.
```
SpennyNDaJets, line 58: I'm typically team arrow function in all cases I can

frontend/CODE_GUIDELINES.md
```diff
@@ -0,0 +1,105 @@
...
+  - Simple components need tests too, mostly to make sure we don't break their imports.
+  - Testing is as much about easy refactors as anything. We need to be able to refactor confidently.
+- Don't be afraid to instrument your code for testing purposes.
+- Get good at mocking imports.
+  - Declare your mocks variables at the top of the file with the `let mock*` pattern.
+  - `jest.mock()` calls should follow the `let mock*` declarations.
+  - Jest automatically hoists your `jest.mock()` calls to the top of the file, so avoid confusion by keeping them near the top of the file. 
+  - Avoid `mockFn.mockImplementation` and `mockFn.mockImplementationOnce`. They're not **bad**, but they're more verbose and a bit harder to read than the `let mock*` pattern.
+
+## Javascript
```
SpennyNDaJets, line 54: I'm also pro using the `?` operator over `&&` to safe navigate

frontend/CODE_GUIDELINES.md
```diff
@@ -0,0 +1,105 @@
...
+- Arrow functions are best for one-liners.
+- Write functional JS. `class` and `this` are unnecessary in most modern JavaScript.
+- Avoid mutating objects. This is bug-avoidance-rule-number-one in JavaScript.
+  - Object mutation is fine if it's visually close to the object declaration.
+  - There are times when object mutation simplifies the code, but BE CAREFUL AND EXPLICIT.
+- Be wary of using closure-scope for variable access. It's not functional and couples functions together, making refactors difficult.
+- Prefer "dot-syntax" over excessive destructuring.
+  - Well-crafted destructures can be very readable.
+  - Too much destructuring can kill your readability.
+  - For instance, always use `import constants from '~/constants';` This enables confident find/replace changes to the constants, because you can search for `constants.SOME_KEY`.
```
SpennyNDaJets, line 67: `import { constants } from '~/constants';` ;)

frontend/CODE_GUIDELINES.md
```diff
@@ -0,0 +1,105 @@
...
+  - Destructuring removes the context of the variable. The reader no longer knows where it came from. This is fine in vertically-adjacent code... but it can be hard to read otherwise.
+
+## Functions
+
+- Prefer a single return. There are definitely exceptions, but early returns can make debugging difficult.
```
SpennyNDaJets, line 72: Not sure I agree here, but would maybe a test case could convince me. Think early returns can make it easier on readability and debugging

frontend/CODE_GUIDELINES.md
```diff
@@ -0,0 +1,105 @@
...
+  - Replace comments with well-named, single-purpose functions where possible.
+  - Break multiple, ambiguous operations out into separate variable declarations.
+  - Prefer single variables in `if` and `switch` statements over inline operations.
+
+## React
+
+- Use hooks liberally.
+- Increase readability by extracting functionality from render functions out into private hooks.
+- Prop drilling is fine for one or two levels, but use a context before it gets out of hand.
+- Create "connected" functions explicitly in the same file as the main component.
```
SpennyNDaJets, line 93: ++

frontend/CODE_GUIDELINES.md
```diff
@@ -0,0 +1,105 @@
...
+## Comments
+
+- Comments lie and should be used sparingly.
+- If you're changing commented code, make sure to review and edit the comments carefully.
+- If you're about to commit a comment, see if you can avoid it with the following strategies:
+  - Replace comments with well-named, single-purpose functions where possible.
+  - Break multiple, ambiguous operations out into separate variable declarations.
+  - Prefer single variables in `if` and `switch` statements over inline operations.
+
+## React
```
SpennyNDaJets, line 88: I also like trying to limit the size of react files to help with readability. Maybe one react file should only be exporting one Component?

SpennyNDaJets (commented):
A lot of good ideas! I tried to leave some comments on ones I really like or may not agree with, but 90% are things I think I will keep in mind when coding

SpennyNDaJets (commented):
A lot of good ideas! I tried to leave some comments on ones I really like or may not agree with, but 90% are things I think I will keep in mind when coding

frontend/CODE_GUIDELINES.md
deltaepsilon: JavaScript STILL doesn't have `do` blocks. This is my pattern for hacking around the lack of `do` blocks. It's divisive. I've had some coworkers who've fallen in love with it, and other who think I'm ridiculous for using `switch` so much.

frontend/CODE_GUIDELINES.md
deltaepsilon: I used to use default exports all the time, but then I had to do some deep refactoring on a huge codebase and find/replace was basically broken, because the devs on our team were all naming things differently!

frontend/CODE_GUIDELINES.md
deltaepsilon: That's a point I hadn't considered.

I do a lot of `export function SomeComponent(){}`, so most files start out with a single top-level export and then might get more exports if necessary. I wonder if this is the `function` keyword pattern vs the `const` function assignment pattern.

frontend/CODE_GUIDELINES.md
deltaepsilon: TIME ZONES!!!!!

I've had similar experiences.

This section of the doc was a reaction to some sketchy use of Jest. I don't use Jest as much anymore, but I know all the ways to NOT use it.

frontend/CODE_GUIDELINES.md
deltaepsilon: It's a strict way to live, but I've found that it does surface errors a little better.

frontend/CODE_GUIDELINES.md
deltaepsilon: `?` has been nice.

frontend/CODE_GUIDELINES.md
deltaepsilon: We are on different sides of a holy war. I need to pick your brain on this.

PR #5180, opened 2023-04-28
fix go graphql sdk recording duplicate errrors

Vadman97:
## Summary

Fixes duplicate errors recorded by the go graphql SDK.
* Errors were recorded any time a `logrus.WithError(err)...` print statement was made. This is incorrect behavior - we should report the error as an attribute of the log, but never create error objects from the log statement.
* The graphql tracer would record two errors for the same request - once from `InterceptField` (the right place that has access to the stacktrace of the error) and once from `InterceptResponse` (an aggregate of the errors that do not have the source stacktraces).

* Checking the types of stacktraces present in go errors reported to highlight.
*Only two errors now reported in a clean local deploy
image
image
image

Will tag a new version of the Go SDK.

### DIFF
--- a/sdk/highlight-go/log/logrus.go
+++ b/sdk/highlight-go/log/logrus.go
@@ -3,7 +3,6 @@ package hlog
 import (
 	"context"
 	"fmt"
-	"runtime/debug"
 	"strings"
 
 	"github.com/highlight/highlight/sdk/highlight-go"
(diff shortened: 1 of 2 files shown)


sdk/highlight-go/log/logrus.go
```diff
@@ -78,22 +77,8 @@ func (hook *Hook) Fire(entry *logrus.Entry) error {
...
-			hasError = true
-			if err, ok := v.(highlight.ErrorWithStack); ok {
-				highlight.RecordSpanErrorWithStack(span, err)
-			} else if err, ok := v.(error); ok {
-				span.RecordError(err, trace.WithStackTrace(true))
-			}
-		} else {
-			attrs = append(attrs, attribute.String(k, fmt.Sprintf("%+v", v)))
-		}
-	}
```
Vadman97, line 93: this code meant that any `logrus.WithError(err)....` would report a highlight error object, even if the log was an `info`. Changing this since we don't want logs to ever be reported as an error (we just want that `WithError` to be a structured log attribute),

sdk/highlight-go/tracer.go
```diff
@@ -93,7 +92,6 @@ func (t Tracer) InterceptResponse(ctx context.Context, next graphql.ResponseHand
 	start := graphql.Now()
 	resp := next(ctx)
 	end := graphql.Now()
-	t.log(ctx, span, resp.Errors)
```
Vadman97, line 96: see description: `InterceptResponse` would be the aggregate error set for all fields in the request, meaning that it would duplicate errors (and not have full stacktraces for them).

et (approved):
Awesome. Thanks for fixing this.

PR #5206, opened 2023-05-02
Ensure that changes to schema.graphqls run typegen

### LINKED ISSUE #5204: Fix typegen failure
See https://github.com/highlight/highlight/commit/9cff1b597519b836e72ddb370fcecc2ed1053440#r111383332

Can reproduce with `yarn turbo typegen`

et:
## Summary

One of the goals of #5158 was to switch our backend booleans from avoiding a tribool state (true, false, nil) (see [comment](https://github.com/highlight/highlight/pull/5158/files#r1179684894)).

To make the type definition correct, this also updated the GQL schema:

Screenshot 2023-05-02 at 9 32 05 AM

However, our CI `typegen` check doesn't always run on changes to this file. As a result, any frontend PRs going forward that do run `typegen` are now broken (https://github.com/highlight/highlight/pull/5011#issuecomment-1530624736 and #5201).

Fixes #5204

I used another boolean that can safely avoid the tribool problem:
https://github.com/highlight/highlight/blob/48a8718b2247a91da3cf9770d1ac781e8b0ead30/backend/model/model.go#L623

and validated that this PR runs the `Monorepo / Build Yarn Turborepo (pull_request)` CI check.

See [failure](https://github.com/highlight/highlight/actions/runs/4864030064/jobs/8672513229) and that the build passes with the changes to `query.gql`.

N/A


backend/private-graph/graph/schema.resolvers.go
```diff
@@ -46,14 +46,14 @@ import (
 	"github.com/highlight-run/highlight/backend/util"
 	"github.com/highlight-run/highlight/backend/vercel"
 	"github.com/highlight-run/highlight/backend/zapier"
-	"github.com/highlight/highlight/sdk/highlight-go"
+	highlight "github.com/highlight/highlight/sdk/highlight-go"
```
et, line 49: #5183 should have failed `make check` (that runs `make private-gen`) but for some reason it was skipped. Not sure why but that's why this import change is here.

Vadman97 (approved):
thanks for fixing this!

PR #5217, opened 2023-05-02
upgrade to go 1.20

et:
# Need to upgrade?

If you're seeing this PR linked from somewhere else, you'll need to upgrade Go to [v1.20](https://go.dev/blog/go1.20).

### Homebrew

`brew install go@1.20`

### Other Go installations

https://go.dev/dl

## Summary

This PR upgrades us to Go 1.20 ([release notes](https://go.dev/blog/go1.20)). The reasoning behind the upgrade is obviously to be on the latest and greatest but also to help prove we don't need the [deprecated github.com/pkg/errors](https://github.com/pkg/errors/issues/245) library. Specifically, Go 1.20 has a built-in [`errors.Join`](https://pkg.go.dev/errors#example-Join) function that does what the [`errors.Wrap`](https://github.com/pkg/errors#adding-context-to-an-error) function was doing.

As we have discussed over slack, Go error wrapping is unnecessary assuming there's an associated stack trace. Long term, I still plan on removing all of the error wrapping in our code base but wanted to make that distinct from the Go upgrade.  

Updated our docs to reflect that Go 1.20 is now the minimum version. We don't [specify](https://www.highlight.io/docs/getting-started/self-host/dev-deployment-guide) how to install Go but for most of us, it's probably `brew install go@1.20` (other versions [here](https://go.dev/dl/)) and restart the backend.

I did a basic click test and didn't observe any issues. I didn't notice any breaking changes from the release notes either.

N/A

### DIFF
--- a/backend/alerts/integrations/discord/discord.go
+++ b/backend/alerts/integrations/discord/discord.go
@@ -1,10 +1,10 @@
 package discord
 
 import (
+	"errors"
 	"os"
 
 	"github.com/bwmarrin/discordgo"
-	"github.com/pkg/errors"
 )
 
 type Bot struct {
@@ -24,7 +24,7 @@ func NewDiscordBot(guildId string) (*Bot, error) {
 	session, err := discordgo.New("Bot " + DiscordBotSecret)
 
 	if err != nil {
-		return nil, errors.Wrap(err, "error creating Discord session")
+		return nil, errors.Join(err, errors.New("error creating Discord session"))
 	}
 
 	return &Bot{
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -28,7 +28,7 @@ jobs:
             - name: Setup Golang environment
               uses: actions/setup-go@v4
               with:
-                  go-version: '^1.19'
+                  go-version-file: 'backend/go.mod'
                   cache-dependency-path: '**/go.sum'
 
             - name: Login to Docker Hub
--- a/.github/workflows/go-sdk.yml
+++ b/.github/workflows/go-sdk.yml
@@ -52,6 +52,6 @@ jobs:
                   fetch-depth: 2
             - uses: actions/setup-go@v4
               with:
-                  go-version: '1.19'
+                  go-version-file: 'backend/go.mod'
             - name: Run tests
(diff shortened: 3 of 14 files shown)


backend/alerts/integrations/discord/discord.go
```diff
@@ -24,7 +24,7 @@ func NewDiscordBot(guildId string) (*Bot, error) {
 	session, err := discordgo.New("Bot " + DiscordBotSecret)
 
 	if err != nil {
-		return nil, errors.Wrap(err, "error creating Discord session")
+		return nil, errors.Join(err, errors.New("error creating Discord session"))
```
et, line 27: Just an example to prove that the Go upgrade worked (`errors.Join` is only available in 1.20).

In a follow up PR, this would simply be:

```go
if err != nil {
  return nil, err
}
```

Vadman97: ah, neat!

Vadman97 (commented):
could we also update the CI version to be consistent? mind checking the `.github/workflows` files

et:
> could we also update the CI version to be consistent? mind checking the .github/workflows files

Correct me if I'm wrong but isn't it using the same thing?

https://github.com/highlight/highlight/blob/4cac39963fca702a934900b38384d39281050e04/.github/workflows/backend.yml#L18-L19

ccschmitz:
> Correct me if I'm wrong but isn't it using the same thing?

@et I think there are still some CI files where we are hard-coded to `1.19`.

https://github.com/search?q=repo%3Ahighlight%2Fhighlight+%271.19%27&type=code

Looks good to me otherwise - thanks for upgrading us!

et:
> @et I think there are still some CI files where we are hard-coded to 1.19.

Nice find. I think I got the rest in bb84a5b2dd87586099ae40030937ec2106e55cbb and 86dab35656ce727a6d9c4e164861f3737b526dce

Vadman97 (approved):
cool, thanks eric! looks like updating the go version in the lint caught some important bugs ;)

et:
> cool, thanks eric! looks like updating the go version in the lint caught some important bugs ;)

Indeed. Just pushed some changes to fix some `rand` deprecation warnings.

Issue #4965, opened 2023-04-13
Live Mode: Improve entire live session experience (console, errors, network recording)


jay-khatri:

ccschmitz:
We discussed this in La Croix today and @et shared the idea of pushing all the event data into Clickhouse. Doesn't sound like this would bloat our costs much on the Clickhouse side, and it would provide some benefits like being able to search and filter across events more effectively.

@Vadman97 would like to chat through this with you when we meet later this week.

ccschmitz:
Here are some notes from chats I had with @Vadman97 and @mayberryzane today.

## Chat w/ Vadim

Below are some notes on potential new design for fetching data from the backend. We would try to consolidate to a single method of fetching events and resources.

```
# non-live, processed session

useGetSession -> {
    ... session metadata,
    events_url: 'https://d195z12t3tmt7c.cloudfront.net/v2/1/217804088/session-contents-compressed...',
    resources_url: 'https://d195z12t3tmt7c.cloudfront.net/v2/1/217804088/network-resources-compressed'
}

fetch(https://d195z12t3tmt7c.cloudfront.net/v2/1/217804088/session-contents-compressed...) -> [{e1}, {e2}]

# live session

useGetSession -> {
    ... session metadata,
    events_url: 'https://pri.highlight.io/sessions/217804088/events',
    resources_url: 'https://pri.highlight.io/sessions/217804088/resources'
}

backend http listener on `/sessions/:id/events` that returns events from redis
authenticated using project-token cookie based auth (see private middleware.go)

fetch(https://pri.highlight.io/sessions/217804088/events) -> [{e1}, {e2}]


# non-live, processed session
# alternative with s3 proxied by backend

useGetSession -> {
    ... session metadata,
}

useGetSessionEvents -> backend will proxy request from https://d195z12t3tmt7c.cloudfront.net/v2/1/217804088/session-contents-compressed... -> [{e1}, {e2}]
```

## Chat w/ @mayberryzane

Zane thought we could potentially return a redirect to the S3 file. Serving the whole file from S3 through our backend would be too much unless we could figure out a way to stream it to the client. Some of these files can be pretty big (>1GB) and we actually have had issues with this in the past, which was part of the motivation for offloading the fetching to the client.

Maybe we could store network request data in Clickhouse. Start with Highlight and roll out if it's scaling well. Would need to monitor costs.

Network resources are also stored in Redis. Check out `redis/utils.go` to see how the data is structured.

[...]

PR #5223, opened 2023-05-03
remove logic to fetch error instance that is not live

et:
## Summary

#3576 attempted to fetch the first error instance with a non-live session (`processed=true`). The reasoning behind this was the devtools in for livemode aren't great (#4965). Additionally, #4269 tried to avoid error instances tied to excluded sessions.

The problem with these approaches is that the error group's last instance timestamp doesn't match one these selected. Here's a visual:
236049033-4e4d768f-a4ad-441b-b453-f815aba8b9a7

This PR removes the backend logic and relies more on frontend UX to inform the user if there is an issue. 

If the session is excluded, we disable the "Show session" button and use a callout:
Screenshot 2023-05-04 at 3 04 10 PM

If the session is live, we provide a tooltip to warn them about dev tools:
Kapture 2023-05-04 at 10 11 41

Will confirm with the customer that reported this that this fixes their issue.

### DIFF
--- a/frontend/src/graph/operators/query.gql
+++ b/frontend/src/graph/operators/query.gql
@@ -1080,6 +1080,7 @@ fragment ErrorObject on ErrorObject {
 		country
 		user_properties
 		processed
+		excluded
 	}
 	error_group_id
 	error_group_secure_id
--- a/frontend/src/graph/operators/query.gql
+++ b/frontend/src/graph/operators/query.gql
@@ -1063,45 +1063,69 @@ query GetErrorObjectForLog($log_cursor: String!) {
 	}
 }
 
+fragment ErrorObject on ErrorObject {
+	id
+	created_at
+	project_id
+	session_id
+	trace_id
+	span_id
+	log_cursor
+	session {
+		identifier
+		fingerprint
+		secure_id
+		city
+		state
+		country
+		user_properties
+		processed
+	}
+	error_group_id
+	error_group_secure_id
+	event
+	type
+	url
+	source
+	lineNumber
+	columnNumber
+	stack_trace
+	structured_stack_trace {
+		fileName
+		lineNumber
+		functionName
+		columnNumber
+		lineContent
+		linesBefore
+		linesAfter
+		error
+		sourceMappingErrorMetadata {
+			errorCode
+			stackTraceFileURL
+			sourcemapFetchStrategy
+			sourceMapURL
+			minifiedFetchStrategy
+			actualMinifiedFetchedPath
+			minifiedLineNumber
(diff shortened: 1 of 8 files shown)


frontend/src/graph/operators/query.gql
```diff
@@ -1063,45 +1063,70 @@ query GetErrorObjectForLog($log_cursor: String!) {
...
-			country
-			user_properties
-		}
-		error_group_id
-		error_group_secure_id
-		event
-		type
-		url
-		source
+fragment ErrorObject on ErrorObject {
```
et, line 1066: Made a fragment for this such that we can fix our types later.

frontend/src/pages/ErrorsV2/ErrorInstance/ErrorInstance.tsx
```diff
@@ -278,9 +257,6 @@ const ErrorInstance: React.FC<Props> = ({ errorGroup }) => {
 		)
 	}
 
-	const errorObject =
-		errorInstance.error_object as ErrorInstanceType['error_object']
```
et, line 282: This casting wasn't actually correct. We don't have an `ErrorObject` but rather a `GetErrorObjectQuery['error_object']`. 

Using the graphql fragment makes the typing correct.

packages/ui/src/components/Tooltip/Tooltip.tsx
```diff
@@ -9,7 +9,7 @@ import { Box } from '../Box/Box'
 
 export type TooltipProps = Partial<TooltipState> &
 	React.PropsWithChildren<{
-		trigger: React.ReactElement
+		trigger: React.ReactNode
```
et, line 12: Had to change this since I'm using `React.PropsWithChildren` (see `WithDevToolsTooltip`) which returns a `React.ReactNode` instead of `React.ReactElement`

frontend/src/pages/ErrorsV2/ErrorInstance/ShowSessionButton.tsx
```diff
@@ -0,0 +1,74 @@
...
+					<IconSolidPlay />
+					Show session
+				</Box>
+			</LinkButton>
+		)
+
+		if (errorObject.session && errorObject.session.processed === false) {
+			return <WithDevToolsTooltip>{linkButton}</WithDevToolsTooltip>
+		} else {
+			return linkButton
```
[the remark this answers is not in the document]
et: I defined it as a const because I need it on line 57 as well. Let me know if that makes sense -- I might be misunderstanding your comment though.

PR #5224, opened 2023-05-03
fix misc issues

Vadman97:
## Summary

* fix frontend -> backend linking with tracingOrigins after pri domain migration
* fix kafka panic when no messages received
* fix fs session storage not resolving assets correctly
* reduce worker processing delay

Local deploy recording errors correctly.
image

Local deploy resolving in-lined assets correctly.
image

No

### DIFF
--- a/backend/main.go
+++ b/backend/main.go
@@ -293,7 +293,7 @@ func main() {
 		if os.Getenv("OBJECT_STORAGE_FS") != "" {
 			fsRoot = os.Getenv("OBJECT_STORAGE_FS")
 		}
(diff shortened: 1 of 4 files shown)


mayberryzane (approved):

ccschmitz:
@Vadman97 just curious, what was the reason for reducing the worker processing delay? I'm just digging through the resource fetching logic and realizing we are pulling this from file storage most of the time because sessions end up being processed so quickly locally because the delay is so short. Not an issue, just trying to understand the reasoning for the different behavior in dev.

PR #5235, opened 2023-05-05
clean up error grouping code

et:
## Summary

This PR cleans up some code I saw in the error grouping logic.

Build passes

N/A

### DIFF
--- a/backend/public-graph/graph/resolver.go
+++ b/backend/public-graph/graph/resolver.go
@@ -2311,7 +2311,7 @@ func (r *Resolver) ProcessBackendPayloadImpl(ctx context.Context, sessionSecureI
 	// Filter out empty errors
 	var filteredErrors []*publicModel.BackendErrorObjectInput
 	for _, errorObject := range errorObjects {
-		if r.isExcludedError(ctx, project.ErrorFilters, errorObject.Event, project.ID) {
+		if isExcludedError(ctx, project.ErrorFilters, errorObject.Event, project.ID) {
 			continue
 		}
 		filteredErrors = append(filteredErrors, errorObject)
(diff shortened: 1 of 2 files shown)


backend/public-graph/graph/resolver.go
```diff
@@ -464,7 +464,7 @@ func (r *Resolver) AppendFields(ctx context.Context, fields []*model.Field, sess
 	return nil
 }
 
-func (r *Resolver) getIncrementedEnvironmentCount(ctx context.Context, errorGroup *model.ErrorGroup, errorObj *model.ErrorObject) string {
```
et, line 467: Many of these functions do not need to be method functions on `*Resolver`

backend/public-graph/graph/resolver.go
```diff
@@ -575,47 +575,6 @@ func joinIntPtrs(ptrs ...*int) string {
 	return sb.String()
 }
 
-func (r *Resolver) GetOrCreateErrorGroupOld(errorObj *model.ErrorObject, stackTraceString string) (*model.ErrorGroup, error) {
```
et, line 578: Unused function `GetOrCreateErrorGroupOld`

backend/public-graph/graph/resolver.go
```diff
@@ -784,8 +743,8 @@ func (r *Resolver) HandleErrorAndGroup(ctx context.Context, errorObj *model.Erro
 	if errorObj == nil {
 		return nil, e.New("error object was nil")
 	}
-	if errorObj.Event == "" || errorObj.Event == "<nil>" {
```
et, line 787: Tracked this back to #1684. The original slack [message](https://highlightcorp.slack.com/archives/C01R04YNCEA/p1635803106004500) has no context as to why we are actually doing it. Would rather remove defensive guards unless we have a reason because this code prevents someone from seeing an error message of `"<nil>"` which (while silly) is totally a legitimate value.

ccschmitz (approved):

Issue #5231, opened 2023-05-04
Track why a session was excluded


et:
When a session is `excluded`, we simply mark `excluded=true` but we don't track why that is. Technically, we can figure it out through our logs but we can't surface it to the end user.

Examples of why a session is excluded:
* zero activity length (possibly from a bot crawler)
* no user interactions
* [ignored via a project setting filter](https://www.highlight.io/docs/general/product-features/session-replay/ignoring-sessions)
* no associated error (#https://github.com/highlight/highlight/pull/5123)

We should track this so we can present it to the UI if needed.

PR #5252, opened 2023-05-08
track why a session was excluded

et:
## Summary

This PR tracks why a session was excluded (see #5231 for reasons why a session could be excluded). We log this information but we don't persist it to the db. Hence, it's not possible for us to tell the end user why I session was excluded. With this change, we'll have the information available and can present it as such. 

Note: this is just the backend changes to record the excluded reason. A follow up PR will tidy up the [frontend code](https://github.com/highlight/highlight/pull/5223/files#diff-c740ecb3d6dc5071064893b4a639d723ff746a049d50e3dde47fe24b59ccd49aR531-R545). 

Tested [various exclusion reasons](https://github.com/highlight/highlight/issues/5231). Ensured that `excluded_reason` is set when `excluded=true`

Screenshot 2023-05-09 at 11 40 47 AM

N/A

### DIFF
--- a/backend/model/model.go
+++ b/backend/model/model.go
@@ -641,7 +641,8 @@ type Session struct {
 	VerboseID             string  `json:"verbose_id"`
 
 	// Excluded will be true when we would typically have deleted the session
-	Excluded bool `gorm:"default:false"`
+	Excluded       bool `gorm:"default:false"`
+	ExcludedReason modelInputs.SessionExcludedReason
 
 	// Lock is the timestamp at which a session was locked
 	// - when selecting sessions, ignore Locks that are > 10 minutes old
--- a/backend/model/model.go
+++ b/backend/model/model.go
@@ -642,7 +642,7 @@ type Session struct {
 
 	// Excluded will be true when we would typically have deleted the session
 	Excluded       bool `gorm:"default:false"`
-	ExcludedReason modelInputs.SessionExcludedReason
+	ExcludedReason *modelInputs.SessionExcludedReason
 
 	// Lock is the timestamp at which a session was locked
 	// - when selecting sessions, ignore Locks that are > 10 minutes old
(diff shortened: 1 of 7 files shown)


backend/public-graph/graph/resolver.go
```diff
@@ -2921,9 +2920,6 @@ func (r *Resolver) ProcessPayload(ctx context.Context, sessionSecureID string, e
 	defer updateSpan.Finish()
 
 	excluded, reason := r.isSessionExcluded(ctx, sessionObj, sessionHasErrors)
-	if excluded {
-		log.WithContext(ctx).WithFields(log.Fields{"session_id": sessionObj.ID, "project_id": sessionObj.ProjectID, "reason": *reason}).Infof("excluding session")
```
et, line 2925: I removed the log statements because they can be [quite noisy](https://github.com/highlight/highlight/pull/5123#discussion_r1181718382). Now that we have the reason recorded, we can refer to that for debugging if needed.

backend/public-graph/graph/resolver.go
```diff
@@ -2937,25 +2933,30 @@ func (r *Resolver) ProcessPayload(ctx context.Context, sessionSecureID string, e
...
+				Chunked:               &model.F,
+				Excluded:              false,
+				ExcludedReason:        "",
+				HasErrors:             &sessionHasErrors,
+			}).Error; err != nil {
+			log.WithContext(ctx).Error(e.Wrap(err, "error updating session"))
+			return err
 		}
-
-		// We only want to update the `HasErrors` field if the session has errors.
```
et, line 2944: Did some clean up here. See the comment on the right as to why we have to specify the column names.

backend/model/model.go
```diff
@@ -641,7 +641,8 @@ type Session struct {
 	VerboseID             string  `json:"verbose_id"`
 
 	// Excluded will be true when we would typically have deleted the session
-	Excluded bool `gorm:"default:false"`
+	Excluded       bool `gorm:"default:false"`
+	ExcludedReason modelInputs.SessionExcludedReason
```
Vadman97, line 645: if the session is not exclued, what is the value? should this be a pointer and be set to nil?

Vadman97 (approved):
nice!

Vadman97 (approved):

PR #5255, opened 2023-05-09
Repairing doc parser's UL handling.

deltaepsilon:
## Summary

The doc parser was only rendering complex unordered list items. Simple, text-only `<li>` elements under a `<ul>` were being skipped entirely.

I got it fixed 👇 
image

I worked on the Changelog 18 page: https://www.highlight.io/docs/general/changelog/changelog-18

Nope!

### DIFF
--- a/highlight.io/pages/docs/[[...doc]].tsx
+++ b/highlight.io/pages/docs/[[...doc]].tsx
@@ -1,45 +1,44 @@
-import { promises as fsp } from 'fs'
-import yaml from 'js-yaml'
-import { MDXRemote, MDXRemoteSerializeResult } from 'next-mdx-remote'
-import { serialize } from 'next-mdx-remote/serialize'
-import { GetStaticPaths, GetStaticProps } from 'next/types'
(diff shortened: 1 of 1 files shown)


highlight.io/pages/docs/[[...doc]].tsx
```diff
@@ -1111,12 +1106,19 @@ const DocPage = ({
...
-																						(
-																							e: any,
-																						) =>
-																							e,
-																					)}
+																					{c
+																						.props
+																						.children
+																						.map
+																						? c?.props?.children?.map(
```
Comment (ccschmitz) [inline highlight.io/pages/docs/[[...doc]].tsx:1113]: Nit, but should the safe navigation (`?.`) be swapped here so it's part of the condition? At this point I think we've already validated that `c.props.children.map` is present.

ccschmitz (approved):

highlight.io/pages/docs/[[...doc]].tsx
Comment (deltaepsilon) [reply highlight.io/pages/docs/[[...doc]].tsx:1113]: Good catch. I'll fix that on my next PR.

PR #5261, opened 2023-05-09
Fix logic for creating error setup events

ccschmitz:
## Summary

Noticed that setup events `where type = 'error'` haven't been created since April 5th, which led me to believe that https://github.com/highlight/highlight/pull/4839 might have changed some logic preventing these from being created. Turns out `err.Type` would never be equal to `modelInputs.LogSourceBackend.String()` (`"backend"`).

Thanks to @et for jumping on a call to help troubleshoot this one.

Local click test.

Will check prod DB to ensure setup events are being created as expected after this goes out.


Vadman97 (approved):
nice! :detective:

PR #5264, opened 2023-05-10
limit network payload size

Vadman97:
## Summary

Impose a limit on recorded network request bodies to avoid a session that has a large network payload that crashes on load.
Since a session can be long and can make network requests to binary content which can be arbitrarily large,
we should take care not to store such large bodies in our recorded payload to avoid issues at replay time.
For previous sessions recorded with older clients, this change will avoid crashing the replay by avoiding showing
network resources if the stored file is too large.
Updates the `fetch_prod_sessions.py` script to ensure session files are downloaded for local filesystem dev mode.

Local viewing of a very large session 1aZTcKkfurrOxCFFCXJSocM4Uhoc
image

On the [new client](https://frontend-pr-5264.onrender.com/1/sessions/4dEJtpsfnY5L7NUBsu7OcIyY8Q6X), network data for text data looks good
image

No


backend/storage/storage.go
```diff
@@ -380,21 +382,37 @@ func (f *FilesystemClient) writeFSBytes(ctx context.Context, fp string, data io.
 }
 
 func (f *FilesystemClient) SetupHTTPSListener(r chi.Router) {
-	r.Get("/direct/assets/{project-id}/{hash-val}", func(w http.ResponseWriter, r *http.Request) {
+	r.Use(cors.New(cors.Options{
+		AllowCredentials: true,
+		AllowedOrigins:   []string{"localhost:3000"},
```
mayberryzane, line 387: is hardcoding this an issue or is `FilesystemClient` only for local builds?

mayberryzane (approved):

backend/storage/storage.go
Vadman97: good catch! this would be only used in localhost but we already use a env var provided value here (`f.origin`) so we might as well use that here too. updated!

Vadman97: image
after 63c6887ab2a3dd370c9b3a9c603ec6bf1cb0248f

PR #5267, opened 2023-05-10
Organize stacktrace code

et:


backend/private-graph/graph/schema.resolvers.go
```diff
@@ -30,7 +30,7 @@ import (
 	"github.com/highlight-run/highlight/backend/clickhouse"
 	"github.com/highlight-run/highlight/backend/clickup"
 	Email "github.com/highlight-run/highlight/backend/email"
-	highlightErrors "github.com/highlight-run/highlight/backend/errors"
+	"github.com/highlight-run/highlight/backend/errorgroups"
```
Vadman97, line 33: nice, we'll no longer need an alias

Vadman97 (approved):

PR #5282, opened 2023-05-10
move firstload and client builds from rollup to vite

Vadman97:
## Summary

As [reported by a customer](https://discord.com/channels/1026884757667188757/1103576485777915945), our highlight.run build was broken
in certain applications, ie. in vue.js with vite-ssr. 
image

With a sample vue + vite.ssr app, I could easily reproduce this
image

Moving the bundling to vite simplifies our configuration and keeps our bundling process more consistent across packages.

Using yalc to publish the new `highlight.run` package locally, I was able to build the vue.js vite ssr app correctly.
image

Vite SSR app recording correctly
image

image

Local frontend session recording correctly
image

New highlight.run version.
@lewisl9029 , would this cause any issues for reflame?


lewisl9029:
> would this cause any issues for reflame?
Don't think so. We build everything in these packages from source for the reflame deploy (i.e. that's what the `libraries` config in .reflame.config.jsonc does), so changing the npm build/publishing process shouldn't make a difference. :)

deltaepsilon (approved):

Vadman97:
outstanding issue

Screenshot 2023-05-11 at 11 08 50 PM

GerryWilko:
Hello, this change appears to have caused an issue with using Highlight directly within the browser
image

This is using the "Other" workflow for the client SDK. Looks like we just need to add `type="module"` to the import but this is a breaking change for all existing "Other" workflow users like myself 😄

PR #5311, opened 2023-05-12
Add properties for tracking live mode usage

ccschmitz:
## Summary

We wanted to be able to see who was using live mode in Highlight but realized the tracking properties we have in place won't allow us to do this. We send an `is_live` property on `Session view` events, but that is always `false` because that's just whether the player is in live mode or not, which it never is when initialized.

This PR adds a new `Session live mode toggled` event and a `is_session_processed` property on `Session view` events so we can see how many people are viewing sessions that aren't fully processed and how many people are activating live mode.

Local click test.

N/A - only tracking


frontend/src/pages/Player/PlayerHook/PlayerState.ts
```diff
@@ -555,6 +555,9 @@ export const PlayerReducer = (
 		case PlayerActionType.setIsLiveMode:
 			s.isLiveMode = handleSetStateAction(s.isLiveMode, action.isLiveMode)
 			s = initReplayer(s, events, !!s.replayer?.config.mouseTail)
+			analytics.track('Session live mode toggled', {
```
Vadman97, line 558: just checking for my own understanding, this should get triggered from https://github.com/highlight/highlight/blob/18f8b27ff9094cb1feed0d9c48c5500f8b46cda7/frontend/src/pages/Player/Toolbar/ToolbarControlBar/ToolbarControlBar.tsx#L219-L232

so this should cover all the live mode cases

Vadman97 (approved):

frontend/src/pages/Player/PlayerHook/PlayerState.ts
ccschmitz: Yeah, exactly!

PR #5316, opened 2023-05-12
New Billing Page

mayberryzane:
## Summary
- builds on the changes for enforcing backend quotas here: https://github.com/highlight/highlight/pull/4990
- adds a new billing UI which shows current usage and lets customers add billing limits to each product
- uses billing limits to calculate quotas for each product during ingest and limit their recording

- clicktested locally

- will copy new products/prices from test to live mode in Stripe before deploying


Vadman97:
image

on the `add payment settings` screen, on a large monitor the `enter payment details` button seems hard to find. could we add one near the total amount, that's where i would expect to find it. cc @julian-highlight @jay-khatri

backend/worker/kafka_worker.go
```diff
@@ -121,30 +124,113 @@ func (k *KafkaBatchWorker) flush(ctx context.Context) {
...
+			if quotaPercent >= 1 {
+				if err := model.SendBillingNotifications(ctx, k.Worker.PublicResolver.DB, k.Worker.PublicResolver.MailClient, email.BillingLogsUsage100Percent, &workspace); err != nil {
+					log.WithContext(ctx).Error(e.Wrap(err, "failed to send billing notifications"))
+				}
+			} else if quotaPercent >= .8 {
+				if err := model.SendBillingNotifications(ctx, k.Worker.PublicResolver.DB, k.Worker.PublicResolver.MailClient, email.BillingLogsUsage80Percent, &workspace); err != nil {
+					log.WithContext(ctx).Error(e.Wrap(err, "failed to send billing notifications"))
+				}
+			}
+		}()
```
Vadman97, line 208: nit: these billing notification inlined fns could be their own package, esp since we have them in a few different places in the ingest path (sessions, errors, logs). `quotaPercent` could be a const?

backend/worker/kafka_worker.go
```diff
@@ -121,30 +124,113 @@ func (k *KafkaBatchWorker) flush(ctx context.Context) {
...
+		filteredRows = append(filteredRows, logRow)
+	}
+
+	wSpan, wCtx := tracer.StartSpanFromContext(ctx, "kafkaBatchWorker", tracer.ResourceName("worker.kafka.batched.process"))
+	wSpan.SetTag("BatchSize", len(k.BatchBuffer.messageQueue))
+
+	span, ctxT := tracer.StartSpanFromContext(wCtx, "kafkaBatchWorker", tracer.ResourceName("worker.kafka.batched.clickhouse"))
+	span.SetTag("NumLogRows", len(logRows))
+	span.SetTag("NumFilteredRows", len(filteredRows))
+	span.SetTag("PayloadSizeBytes", binary.Size(logRows))
```
Vadman97, line 228: nice to report the filtering as tags

Vadman97:
@mayberryzane does this PR stop ingesting logs when the quota is hit? i am on the 'free' tier and set $0 limits in my local workspace, but am still ingesting logs

jay-khatri:
> image
> 
> on the `add payment settings` screen, on a large monitor the `enter payment details` button seems hard to find. could we add one near the total amount, that's where i would expect to find it. cc @julian-highlight @jay-khatri

+1 here. @julian-highlight , I know this is a pattern we've used elsewhere, but I wonder if its worth revisiting. I think this is a weird place to generally put a button because if I'm filling out a form and the height of the form is larger than the viewport, I expect the submit button to be at the bottom.

mayberryzane:
> @mayberryzane does this PR stop ingesting logs when the quota is hit? i am on the 'free' tier and set $0 limits in my local workspace, but am still ingesting logs

Hmm I tested this previously and it did stop ingesting logs, though there might be a 5 min delay due to the Redis ttl here. Did you try after ~5 mins by chance?

backend/redis/utils.go
```diff
@@ -359,12 +372,12 @@ func (r *Client) SetIsPendingSession(ctx context.Context, sessionSecureId string
...
-func (r *Client) IsBillingQuotaExceeded(ctx context.Context, projectId int) (bool, error) {
-	return r.getFlag(ctx, BillingQuotaExceededKey(projectId))
+func (r *Client) IsBillingQuotaExceeded(ctx context.Context, projectId int, productType pricing.ProductType) (*bool, error) {
+	return r.getFlagOrNil(ctx, BillingQuotaExceededKey(projectId, productType))
 }
 
-func (r *Client) SetBillingQuotaExceeded(ctx context.Context, projectId int) error {
-	return r.setFlag(ctx, BillingQuotaExceededKey(projectId), true, 5*time.Minute)
+func (r *Client) SetBillingQuotaExceeded(ctx context.Context, projectId int, productType pricing.ProductType, exceeded bool) error {
+	return r.setFlag(ctx, BillingQuotaExceededKey(projectId, productType), exceeded, 5*time.Minute)
```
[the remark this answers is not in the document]
mayberryzane: yes, though I could also clear these keys in Redis on updating the billing settings to fix that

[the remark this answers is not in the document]
Vadman97: > > @mayberryzane does this PR stop ingesting logs when the quota is hit? i am on the 'free' tier and set $0 limits in my local workspace, but am still ingesting logs
> 
> Hmm I tested this previously and it did stop ingesting logs, though there might be a 5 min delay due to the Redis ttl here. Did you try after ~5 mins by chance?

will try that!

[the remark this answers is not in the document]
Vadman97: just tested, this looks great. after 5 minutes, got the 100% logs quota used email and no longer recorded any more logs
image

julian-highlight:
> image
> 
> on the `add payment settings` screen, on a large monitor the `enter payment details` button seems hard to find. could we add one near the total amount, that's where i would expect to find it. cc @julian-highlight @jay-khatri

Good call. Will add that asap!

backend/worker/kafka_worker.go
mayberryzane: :+1: will do this in a follow up

Vadman97: sgtm

julian-highlight:
> I know this is a pattern we've used elsewhere, but I wonder if its worth revisiting. I think this is a weird place to generally put a button because if I'm filling out a form and the height of the form is larger than the viewport, I expect the submit button to be at the bottom.

The reason we have that top panel, is mostly because of the pattern of moving back to a previous screen. I'd keep it for now and add the bottom button. Generally I still think CTA's on all settting screens could be improved!

PR #5320, opened 2023-05-12
<HighlightInit> transpilation

### LINKED ISSUE #5319: `<HighlightInit>` needs CJS
`<HighlightInit> is not CJS-compatible.

See https://discord.com/channels/1026884757667188757/1106642800474325042

Create a CJS-compatible version of the file.

deltaepsilon:
It turns out that we need to transpile `<HighlightInit>`.

closes #5319

### DIFF
--- a/sdk/highlight-next/bin/clean-highlight-init.sh
+++ b/sdk/highlight-next/bin/clean-highlight-init.sh
@@ -1,0 +1,5 @@
+sed -i '/use client/d' dist/HighlightInit.mjs
+sed -i '2s/^/"use client";\n/' dist/HighlightInit.mjs
+
+sed -i '/use client/d' dist/HighlightInit.js
+sed -i '2s/^/"use client";\n/' dist/HighlightInit.js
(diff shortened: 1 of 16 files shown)


sdk/highlight-next/bin/clean-highlight-init.sh
Vadman97, line 1: do we need the `use client` line at all in the `HighlightInit.tsx` file? isn't that only used by the next.js bundling process, and bundling with tsup isn't going to use it?

Vadman97 (approved):

Vadman97:
looks like the build is failing
```
nextjs:build: > 4 | import { HighlightInit } from '@***/next/HighlightInit'
nextjs:build:     |                               ^
```
i saw this issue too when digging into some of the conditional exports on friday. i think what's happening is that the `types` conditional export isn't actually being resolved by typescript, and typescript was resolving this before because it would look for the relative path of this file. now that it is under `src/HighlightInit`, that isn't the same as `/HighlightInit` so the types check fails. i wonder if this is a problem with our typescript configuration that could be fixed

sdk/highlight-next/bin/clean-highlight-init.sh
deltaepsilon: Basically, if someone finds `HighlightInit.tsx` and copies to code, you'd want them to copy `"use client";` as well.

Vadman97: gotcha - if you're certain we need `"use client";` in the final bundled files, then this looks good!

PR #5336, opened 2023-05-16
load timeline indicator events on frontend from s3

Vadman97:
## Summary

Timeline indicator events were loaded from the backend where the backend would download the file from S3.
Instead, pass the S3 signed url to the frontend as we do for resources and the events data
so that the frontend could asynchronously load the timeline without blocking the rest of the loading, and
to make sure the cloudfront download could happen faster and be cached by the browser.

Local fetch and render of large session VbEEeAuUMnOqTefUNruC7Dhd3Yb0
image

No


mayberryzane (approved):
sweet 🏎️

PR #5340, opened 2023-05-16
add highlight.run browser key for cdn distribution

Vadman97:
## Summary

As mentioned in https://github.com/highlight/highlight/pull/5282#issuecomment-1549450686, our highlight.run release
was not setting a `browser` key resulting in `unpkg` and `cdn.jsdelivr.net` using the `main` file, which is now an ESM build.

Adds the `browser` key to ship the `<script>` tag umd file.
https://www.jsdelivr.com/documentation#id-configuring-a-default-file-in-packagejson

Updates some other highlight.run package metadata.

Local deploy using `<script src="http://localhost:5173/dist/index.umd.cjs></script>`

https://www.loom.com/share/98a9863ada1143b197f3201bf6ad4400

New patch version of highlight.run released.

### DIFF
--- a/sdk/firstload/package.json
+++ b/sdk/firstload/package.json
@@ -1,6 +1,29 @@
 {
 	"name": "highlight.run",
-	"version": "6.4.0",
+	"version": "6.4.1",
+	"description": "Open source, fullstack monitoring. Capture frontend errors, record server side logs, and visualize what broke with session replay.",
+	"keywords": [
+		"highlight",
+		"session replay",
+		"error monitoring",
+		"logging",
+		"debugging",
+		"observability",
+		"browser",
+		"library"
+	],
+	"homepage": "https://github.com/highlight/highlight#readme",
+	"bugs": {
+		"url": "https://github.com/highlight/highlight/issues",
+		"email": "support@highlight.io"
+	},
+	"license": "Apache-2.0",
+	"repository": {
+		"repository": {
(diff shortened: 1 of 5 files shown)


et (approved):

sdk/firstload/package.json
```diff
@@ -12,6 +35,7 @@
 	},
 	"type": "module",
 	"main": "./dist/index.js",
+	"browser": "./dist/index.umd.cjs",
```
lewisl9029, line 38: @Vadman97 unfortunately this will also result in most esm-supporting dev tools and bundlers to use the umd build instead of the esm build, which is probably not what we want.

I think a better approach would be to add individual `unpkg` and `jsdelivr` keys for the umd build instead. 

Or possibly drop the umd build altogether and ask users to start importing these as modules from the cdns.

lewisl9029: Even with the exports field I think the existence of the browser field will force bundlers to switch to the umd entry point, I believe. That's the intended purpose of the browser field after all.

I think named exports getting broken is very likely a result of the umd module being used instead of esm as well. So removing the browser field should hopefully fix both.

lewisl9029: Here's the defacto spec for the browser field that most dev tools adhere to, for reference: https://github.com/defunctzombie/package-browser-field-spec

PR #5343, opened 2023-05-16
use consistent js extension for highlight.run

Vadman97:
## Summary

unpkg will use `application/node` when the file extension is `.umd.cjs`

image

new highlight.run patch


deltaepsilon (approved):
It is perfect. This will work, first try.

PR #5347, opened 2023-05-16
New UI for app settings

eightants:
## Summary

From the Figma for this issue:

Moved the ToggleRow to a separate reusable component

Screenshot 2023-05-16 at 3 24 17 PM


packages/ui/vite.config.ts.timestamp-1683841674780.mjs
```diff
@@ -0,0 +1,29 @@
+// vite.config.ts
```
et, line 1: How did this file make it here? Might be a reflame issue?

frontend/src/components/ToggleRow/ToggleRow.tsx
```diff
@@ -0,0 +1,39 @@
+import { Box, Stack, Text } from '@highlight-run/ui'
+
+import Switch from '@/components/Switch/Switch'
+
+export const ToggleRow = (
+	label: string,
+	info: string | undefined,
```
et, line 7: Realize you're just moving this component but can `info` be required? i.e. can we remove `| undefined`?

et (approved):
Looks good! There is a build failure. Let me know if you need any help diagnosing it.

packages/ui/vite.config.ts.timestamp-1683841674780.mjs
Vadman97, line 1: this is some weird vite actifact where if the build crashes, the file is left over. should delete it and commit the deletion

packages/ui/vite.config.ts.timestamp-1683841674780.mjs
Vadman97: this is some weird vite actifact where if the build crashes, the file is left over. should delete it and commit the deletion

Vadman97 (approved):

PR #5348, opened 2023-05-16
Use a single node & esbuild process for reflame scripts

lewisl9029:
## Summary

Some folks were trying to run the dev env with 8gb, which is probably not enough for the Vite setup, but might actually be ok  if we just run Reflame, except for the fact that our scripts were extremely wasteful when it comes to memory usage previously, spinning up a lot more node and esbuild processes than we needed.

This PR pulls all of them into the same script (sharing the same node and esbuild process), and adds a few more external resolution rules to stop esbuild from looking at any non-source files.

There are still a lot more low hanging fruits, but this is a super quick and easy change that cuts memory usage to a fraction of what it was before.

This is looking at just the memory used by esbuild processes alone (not counting the extra node processes):

Before: 
Screenshot 2023-05-16 at 2 27 43 PM

After:
Screenshot 2023-05-16 at 2 36 25 PM


Vadman97 (approved):
:tada:

lewisl9029:
A few more optimizations later, the steady state for the watcher looks like this:
Screenshot 2023-05-16 at 5 56 28 PM

However, when testing inside a 8gb codespace (with ~5gb free), there can be occasional spikes in memory usage that end up causing the esbuild process to crash (usually when rebuilding vanilla extract entry points).

I've gotten it down to a point where the watch script can at least survive startup, but afterwards ymmv...

A more reliable approach seems to be to run `yarn reflame-build` (the non-watcher version) when you need styles to update manually. Not a great experience, but at least it won't randomly fail on you.

@Vadman97 can probably merge in this state now since it's at least a very tangible improvement over the existing scripts.

frontend/scripts/build-vanilla-extract.mjs
```diff
@@ -1,103 +1,131 @@
...
-} else {
-	await build()
+		console.log(new Date(), 'built vanilla extract entry points')
+		return
+	}
+
+	const isEntryPoint = (path) => path.endsWith('.css.ts')
+
+	if (watch) {
+		let transactionIdLatest = randomString()
```
lewisl9029, line 80: This removes some race conditions and ensures at most 1 instance of build is running at a time, which helps put an upper bound on memory usage as well.

Vadman97: neat!

Vadman97 (approved):

PR #5349, opened 2023-05-16
fix issue where identify after payload isn't excluded

mayberryzane:
## Summary
- there are sessions for ignored users not getting flagged as excluded: 
`select * from sessions where excluded_reason = 'IgnoredUser' and not excluded limit 1`
- it seems like this is affecting sessions where a payload is ingested before the session is identified
- was able to repro by logging out, starting a new session, and logging in as an ignored user

- followed steps above before and after change

- no, but we can probably backfill the excluded flag for these sessions. looking into how many are affected


Vadman97 (approved):

Vadman97:
thank you for getting to the bottom of this :tada:

PR #5351, opened 2023-05-17
Lifted ApplicationContext and remove loading

eightants:
## Summary

An issue noticed from converging the settings is that projects are in a different router, so a loading screen comes up when switching between. This is because the data for projects in a workspace is loaded fresh each time. This PR removes that by removing some unnneeded logic that adds a loading state, and have the project list be fetched on the AppRouter level. 

https://www.loom.com/share/409dbcc354eb4bebbd716d5bf6fde872


mayberryzane (approved):
Looks good in preview env. One nit: I'm seeing this weird loading bar state, guessing that's the loading state we had on those project settings pages already? https://www.loom.com/share/2eeea1081cf74e58858dbc459f5dc745

eightants:
Yeah, that's the existing loading bar state. Will have a future PR to add our current loading state just in the right panel

Issue #5228, opened 2023-05-04
Auto-resolve errors if they haven't reoccurred


et:
Relying on users to actually click "Resolve" on an error when the error is fixed is a manual, tedious process.

We should support a project setting that auto resolves errors if they haven't been seen a certain amount of time.

et:
Fixed via https://github.com/highlight/highlight/pull/5354

PR #5354, opened 2023-05-17
Autoresolve errors + audit logs

et:
## Summary

This PR adds the backend for autoresolving errors (#5228). Additionally, audit logs are created whenever the error group state is updated.

The frontend is intentionally is not included as we need to do some deployment steps (see below) before exposing this to users.

This PR also attempts some code organization. For updating the error group state, there's a slew of actions:
* updating postgres
* updating opensearch
* creating an audit log entry

Having this consolidated into a packaged function (and out of the resolver file) makes it easier to test. I took some inspiration from the repository pattern (seen in another [opensource golang app](https://github.com/techschool/simplebank), see the [`sql`](https://github.com/techschool/simplebank/tree/bdfe8e7aa022d3129f3a2478e844bafab7cdf10f/db/sqlc) directory) as it lends itself to writing tests in isolation (particularly when trying to grok how gorm works). We don't have a strategy yet for testing opensearch through a test but it's something that should hopefully be easier now that the structure is in place.

See this loom for a walkthrough:
https://www.loom.com/share/6d1b5ab8ec7e464681ef403d4fc0d49c

If you want to test locally, set your project's `project_filter_settings.auto_resolve_stale_errors_day_interval` to a non-zero value:
Screenshot 2023-05-19 at 11 04 15 AM

* Setup a cron job on AWS to run `make auto-resolve-stale-errors` daily
* Set `project_filter_settings.auto_resolve_stale_errors_day_interval=30` for our project and ensure the job runs successfully
* Rolled out the frontend changes


backend/filtering/autoresolver.go
```diff
@@ -0,0 +1,83 @@
...
+		Model(model.ErrorObject{}).
+		Select("error_group_id").
+		Where("created_at >= ?", time.Now().AddDate(0, 0, -interval))
+
+	result := db.Debug().Where(model.ErrorGroup{
+		State:     privateModel.ErrorStateOpen,
+		ProjectID: project.ID,
+	}).
+		Where("id NOT IN (?)", subQuery).
+		FindInBatches(&errorGroups, 100, func(tx *gorm.DB, batch int) error {
```
et, line 61: Using [`FindInBatches`](https://gorm.io/docs/advanced_query.html#FindInBatches) such that for projects with lots of stale errors, we don't timeout the query.

For example, to get the last month of our project's stale errors:

```sql
SELECT
	*
FROM
	"error_groups"
WHERE
	"error_groups"."project_id" = 1
	AND "error_groups"."state" = 'OPEN'
	AND id NOT IN(
		SELECT
			"error_group_id" FROM "error_objects"
		WHERE
			created_at >= '2023-04-18 11:29:53.374')
ORDER BY
	"error_groups"."id"
```

takes 221.621s and returns 128,079 rows

Using `FindInBatches` will attach a `LIMIT 100` to the query and work through 100 error groups before fetching the next batch.

backend/worker/autoresolver.go
```diff
@@ -0,0 +1,86 @@
...
+		Model(model.ErrorObject{}).
+		Select("error_group_id").
+		Where("created_at >= ?", time.Now().AddDate(0, 0, -interval))
+
+	result := db.Debug().Where(model.ErrorGroup{
+		State:     privateModel.ErrorStateOpen,
+		ProjectID: project.ID,
+	}).
+		Where("id NOT IN (?)", subQuery).
+		FindInBatches(&errorGroups, 100, func(tx *gorm.DB, batch int) error {
```
et, line 64: Using [`FindInBatches`](https://gorm.io/docs/advanced_query.html#FindInBatches) such that for projects with lots of stale errors, we don't timeout the query.

For example, to get the last month of our project's stale errors:

```sql
SELECT
	*
FROM
	"error_groups"
WHERE
	"error_groups"."project_id" = 1
	AND "error_groups"."state" = 'OPEN'
	AND id NOT IN(
		SELECT
			"error_group_id" FROM "error_objects"
		WHERE
			created_at >= '2023-04-18 11:29:53.374')
ORDER BY
	"error_groups"."id"
```

takes 221.621s and returns 128,079 rows

Using `FindInBatches` will attach a `LIMIT 100` to the query and work through 100 error groups before fetching the next batch.

backend/model/model.go
```diff
@@ -929,6 +932,22 @@ type ErrorGroup struct {
...
+	ErrorGroupIgnoredEvent  ErrorGroupEventType = "ErrorGroupIgnored"
+	ErrorGroupOpenedEvent   ErrorGroupEventType = "ErrorGroupOpened"
+)
+
+type ErrorGroupActivityLog struct {
+	Model
+	ErrorGroupID int `gorm:"index"`
+	AdminID      int // when this is 0, it means the system generated the event
+	Admin        *Admin
+	EventType    ErrorGroupEventType
```
et, line 948: Not included here is the snoozed event. Will be interested to know if folks have thoughts on how generic we want this.

For example: I named this `EventType` such that it could in theory handle updating things besides the state. For snoozing, the action is state=OPEN and snoozedUntil=time. 

If we want to keep this generic, we should probably add a new column called `EventParams` that's a JSON blob. The downside of this is that it's not typed. 

Alternatively, we can take the less overengineered approach and just name these columns `UpdatedState` and `UpdatedSnoozedUntil` if we don't think this will get a lot more usage.

Vadman97: my instinct is to keep this less generic. i'd add another column like `UpdatedSnoozedUntil` that would be null when the event is not a snooze.

backend/model/model.go
```diff
@@ -356,11 +357,13 @@ type SetupEvent struct {
...
+
 type ProjectFilterSettings struct {
 	Model
-	Project                    *Project
-	ProjectID                  int
-	FilterSessionsWithoutError bool `gorm:"default:false"`
+	Project                           *Project
+	ProjectID                         int
+	FilterSessionsWithoutError        bool `gorm:"default:false"`
+	AutoResolveStaleErrorsDayInterval int  `gorm:"default:0"`
```
Vadman97, line 366: could this column be a `time.Duration` and stored as an `interval` in postgres? that way this can support finer time ranges than a day

backend/worker/autoresolver.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+func (autoResolver *AutoResolver) resolveStaleErrorsForProjectInBatches(ctx context.Context, project model.Project, interval int) error {
+	var errorGroups []model.ErrorGroup
+
+	db := autoResolver.db
+
+	subQuery := db.
+		Model(model.ErrorObject{}).
+		Select("error_group_id").
```
Vadman97, line 56: prefer to `select distinct error_group_id` to utilize the error_group_id index and not have to scan all error objects

backend/worker/autoresolver.go
```diff
@@ -0,0 +1,86 @@
...
+		Where("created_at >= ?", time.Now().AddDate(0, 0, -interval))
+
+	result := db.Debug().Where(model.ErrorGroup{
+		State:     privateModel.ErrorStateOpen,
+		ProjectID: project.ID,
+	}).
+		Where("id NOT IN (?)", subQuery).
```
Vadman97, line 63: i'm concerned this inverse subquery approach may break for projects with tons of new errors, since the subquery result is unbounded.

why do we need to use a subquery here? would something like this work:
```sql
select *
from error_groups eg
         inner join error_objects eo on eo.error_group_id = eg.id
where state = 'OPEN'
  and eg.project_id = 1
  and eg.created_at < NOW() - interval '30 day'
```


this will utilize some existing indexes and we can easily optimize further with a `state` and `created_at` index
Screenshot 2023-05-22 at 6 11 20 PM

backend/worker/autoresolver.go
Vadman97: i would avoid using `FindInBatches` here since postgres queries cannot time out but this adds some extra complexity. this will also mean that individual updates will not happen atomically since they will be part of different transactions.

backend/store/error_groups.go
```diff
@@ -0,0 +1,81 @@
...
+		Admin:        admin,
+		EventType:    eventType,
+		ErrorGroupID: errorGroup.ID,
+	}).Error
+
+	if err != nil {
+		return errorGroup, err
+	}
+
+	if err := store.opensearch.UpdateSynchronous(opensearch.IndexErrorsCombined, errorGroup.ID, errorGroup); err != nil {
```
Vadman97, line 60: i think we should change this to an asynchronous `.Update(` call, since `autoresolver.go` will call this method many times (for each error group that needs to be closed) resulting in a bunch of synchronous calls to opensearch

Vadman97:
thanks for tackling this @et! makes a lot of sense to do this in two parts (backend then frontend) - makes it easier to iterate on top of this change as well.

backend/model/model.go
et: I considered that but when @julian-highlight and I were scoping this out, we came to the conclusion that setting below a day is not a good user experience. I'm going to keep it as an `int` since it makes the code more readable.

backend/model/model.go
```diff
@@ -356,11 +357,13 @@ type SetupEvent struct {
 	ProjectID int                  `gorm:"uniqueIndex:idx_project_id_type"`
 	Type      MarkBackendSetupType `gorm:"uniqueIndex:idx_project_id_type"`
 }
+
 type ProjectFilterSettings struct {
```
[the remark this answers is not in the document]
et: That make sense although renaming will likely be a multi-step process. Filed a ticket for now and will tackle this later: #5428

backend/model/model.go
julian-highlight: Yea, I don't see a single use case to auto resolve errors < 1 day.

backend/model/model.go
et: What about error boundary feedback? That seem like we would want to store that in this table, yeah?

backend/worker/autoresolver.go
et: Adjusted in 87b02fa9f62a63b0cc1c4d7351dc80769769f0ab. Please check out the new query when you get a chance. Thanks!

backend/model/model.go
et: I've added a JSONB column to store arbitrary data:
8041d321901b1d485d4c58440b349e6b5e6aece7

backend/worker/autoresolver.go
```diff
@@ -0,0 +1,85 @@
...
+
+func (autoResolver *AutoResolver) resolveStaleErrorsForProject(ctx context.Context, project model.Project, interval int) error {
+	var errorGroups []model.ErrorGroup
+
+	db := autoResolver.db
+
+	err := db.Debug().
+		Table("error_groups").
+		Select("DISTINCT(error_groups.id), error_groups.project_id").
+		Joins("LEFT JOIN error_objects ON error_groups.id = error_objects.error_group_id").
```
Vadman97, line 57: this should be an `inner join`, since we search for `error_objects.created_at` which is `null` if there are no `error_objects` resulting from the `left join`

backend/model/model.go
Vadman97: that's fair, having an days in does keep the code simpler

Vadman97 (approved):

backend/worker/autoresolver.go
Vadman97: thanks! i believe the new query should use an `inner join` since we are filtering for `error_objects.created_at`. other than that, looks good.

backend/worker/autoresolver.go
et: Good catch. Also, looks like we were missing the project_id in the where clause. 
I fixed both in d6dadd54a

The generated query:

```sql
SELECT DISTINCT
	(error_groups.id),
	error_groups.project_id
FROM
	"error_groups"
	INNER JOIN error_objects ON error_groups.id = error_objects.error_group_id
WHERE (error_groups.state = 'OPEN')
	AND(error_groups.project_id = 1)
	AND(error_objects.created_at < '2023-04-25 09:48:15.667')
```

runs in 47.160s

Issue #5353, opened 2023-05-17
Bug w/ new date picker on session & error feeds


jay-khatri:
- date doesn't perist in URL
- adding a custom date range removes all filters

Loom link:
https://www.loom.com/share/5946ca66aa684d8298c3d3039d192d73

Vadman97:
@betty-alagwu - would you be able to address these bugs?

betty-alagwu:
Sure! Will start working on it

PR #5358, opened 2023-05-17
Fix/bug with new date picker

betty-alagwu:
hey @Vadman97 

This PR is the first pass for the bug fix on the session/error datePicker

#5353


betty-alagwu:
Closing in favour of #5369

PR #5366, opened 2023-05-17
Add generated files check for reflame to workflow

lewisl9029:
## Summary

We've seen a few instances where generated files in main are out-of-date and cause reflame deploys to fail. This PR adds a check to make sure generated files for reflame are up-to-date.

This does add ~30-40s to the `Build Yarn Turborepo` step, which is not ideal. We can run this as a separate workflow in parallel with `Build Yarn Turborepo`, but the step depends on `yarn install` so it will result in double the build minute usage for that step.

I enabled the action in my fork and verified that it fails on out-of-date generated files:

Screenshot 2023-05-17 at 1 54 46 PM

Adding the files got it passing again.


.github/workflows/turbo.yml
```diff
@@ -50,6 +50,9 @@ jobs:
             - name: Install js dependencies
               run: yarn
 
+            - name: Check generated files for Reflame
+              run: yarn reflame-check
```
Vadman97, line 54: nice!

Vadman97 (approved):

PR #5369, opened 2023-05-17
fix: set default selected date range from params

betty-alagwu:
hey @Vadman97 

This PR is the first pass for the bug fix on the session/error datePicker

#5353


jay-khatri (approved):
Looks great. @Vadman97, [this](https://www.notion.so/runhighlight/session-URL-46c0fde9fa1f41148bd9ab62786a758d?pvs=4) session URL doesn't work when I put it into the address bar in production; it defaults to "30 days" for some reason and also doesn't return any results. 

It seems like this change also fixes that issue (i did a click test), but just sharing in case there's anything you would change in the logic here.

Vadman97 (approved):
click test and code look good! thanks @betty-alagwu

PR #5371, opened 2023-05-18
Invite and account creationg flow fixes

ccschmitz:
## Summary

Ensures we don't ever delete workspace invite links from the database. By doing this we ensure that links won't be broken for a user when they use them, and if the link is expired we can show them that they are expired in the UI.

When someone is accessing Highlight from a workspace invite link we were dropping them on the `/sign_in` page, but most people clicking these links need to create an account, so this PR updates the redirect to the `/sign_up` path when a workspace invite code is present.

This PR also fixes an issue where the submit button for the about you form could be submitted more than once. I thought the `loading` prop also disabled the button, but it does not, so we set that explicitly and add an early `return` to prevent multiple submissions.

Local click test.

The only backend changes are removing some deletes for workspace invites. Should be a fairly safe change. Will monitor bug reports from users about trouble accepting invites or having duplicate workspaces created.


mayberryzane (approved):
clean!

PR #5382, opened 2023-05-19
Set hex dashboard link on Hubspot companies

ccschmitz:
## Summary

We are adding links to hex dashboards manually to Hubspot companies, but the link is in a consistent format and can be assigned automatically.

Will click test after this change goes outl.

Confirm property gets set as expected.

### DIFF
--- a/backend/hubspot/hubspot.go
+++ b/backend/hubspot/hubspot.go
@@ -4,10 +4,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	kafka_queue "github.com/highlight-run/highlight/backend/kafka-queue"
-	"github.com/highlight-run/highlight/backend/redis"
-	"github.com/openlyinc/pointy"
-	"github.com/samber/lo"
 	"io"
 	"math"
 	"net/http"
@@ -15,6 +11,11 @@ import (
 	"strconv"
 	"strings"
 	"time"
+
+	kafka_queue "github.com/highlight-run/highlight/backend/kafka-queue"
+	"github.com/highlight-run/highlight/backend/redis"
+	"github.com/openlyinc/pointy"
+	"github.com/samber/lo"
 
 	"github.com/aws/smithy-go/ptr"
 	"github.com/goware/emailproviders"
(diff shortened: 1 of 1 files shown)


Vadman97 (approved):
nice! thank you

backend/hubspot/hubspot.go
Vadman97, line 1: on line 349, if hubspot created the company already, we might need to update to add the custom properties

jay-khatri (approved):

backend/hubspot/hubspot.go
```diff
@@ -373,8 +366,41 @@ func (h *HubspotApi) CreateCompanyForWorkspaceImpl(ctx context.Context, workspac
...
+
+	if companyID, err = pollHubspot(func() (*int, error) {
+		id, err := h.getCompany(ctx, name)
+		if err != nil {
+			return nil, err
+		}
+
+		log.WithContext(ctx).
+			WithField("name", name).
+			Warnf("company already exists in Hubspot. updating")
```
Vadman97, line 385: this is the expected flow so this should be an `Infof`

backend/hubspot/hubspot.go
```diff
@@ -373,8 +366,41 @@ func (h *HubspotApi) CreateCompanyForWorkspaceImpl(ctx context.Context, workspac
...
+		}
+
+		return id, nil
+	}); companyID != nil {
+		return
+	}
+
+	log.WithContext(ctx).
+		WithField("name", name).
+		Warnf("failed to get client-side hubspot company. creating")
```
Vadman97, line 401: good idea to add this!

backend/hubspot/hubspot.go
```diff
@@ -373,8 +366,41 @@ func (h *HubspotApi) CreateCompanyForWorkspaceImpl(ctx context.Context, workspac
...
+		if id != nil {
+			_, err := h.hubspotClient.Companies().Update(*id, companyProperties)
+			if err != nil {
+				return nil, err
+			}
+		}
+
+		return id, nil
+	}); companyID != nil {
+		return
```
Vadman97, line 396: sorry, should have been a bit more precise here. i think the new logic should go out here to happen once, when the company is found. the inline function above is used to actually find the company in a loop until it appears

ccschmitz: Oh, right! My bad.

Vadman97 (approved):

jay-khatri:
Thanks Chris!

PR #5391, opened 2023-05-22
New project settings UI 

eightants:
## Summary

New project settings ui for sessions and error settings. Mostly the section cards and text. 

Currently not heavily modified to match the Figma: Input and Select components. 

The inputs seem to be reused throughout the app, so I kept them the same for now. Inputs can be changed in a future PR, possibly by making a SelectV2 or InputV2. 

https://www.figma.com/file/rdBPNzn7Klk6RG1LHUCE5B/Screen-Designs?type=design&node-id=7765-360301&t=Bxv5Eh5GdtXD0TwH-0

Screenshot 2023-05-22 at 11 30 42 AM
Screenshot 2023-05-22 at 11 30 52 AM


frontend/src/components/BorderBox/BorderBox.tsx
```diff
@@ -0,0 +1,12 @@
+import { Box } from '@highlight-run/ui'
+import React from 'react'
+
+const BorderBox: React.FC<React.PropsWithChildren> = ({ children }) => {
```
et, line 4: Nice!

I could see this going into `@highlight-run/ui`.
cc @ccschmitz - related to our convos before about how we use the box component with so many props defined.

et (approved):
Great cleanup here! 

One thing I noticed is that we don't have a loading state for these forms and the inputs jump when the data loads in. 
See this gif and watch the toggle for "Filter errors thrown by Chrome extensions"

 
Kapture 2023-05-22 at 12 31 02

I don't think we need to fix this immediately but it would be great to fix this at some point.

frontend/src/components/BorderBox/BorderBox.tsx
eightants: Yeah thought about that while building these components. The components are very barebones Box with a few props but there is value in making them reusable since we use them really often

eightants:
Loading state is currently being worked on. Will be getting a pr out for it later today!

PR #5392, opened 2023-05-22
Supporting Next.js middleware

deltaepsilon:
## Summary

DragonZSG discovered a problem with `middleware.ts` and `instrumentation.ts` on [this Discord thread](https://discord.com/channels/1026884757667188757/1109236631665385473/1109236631665385473)

I was able to reproduce and found this error:

```
An error occurred while loading instrumentation hook: (0 , _highlight_run_next__WEBPACK_IMPORTED_MODULE_1__.registerHighlight) is not a function
```

I was able to resolve the issue with a dynamic import in `instrumentation.ts`, because it appeared that `middleware.ts` was attempting to import Node.js-only code in a non-Node.js environment.

I've updated the docs.

image

I created a `middleware.ts` file at the root of `e2e/nextjs` and reproduced the error.

Subsequent changes to `instrumentation.ts` resolved the issue

Nope

### DIFF
--- a/e2e/nextjs/instrumentation.ts
+++ b/e2e/nextjs/instrumentation.ts
@@ -1,10 +1,20 @@
 // instrumentation.ts
 import CONSTANTS from '@/app/constants'
-import { registerHighlight } from '@highlight-run/next'
+// import { registerHighlight } from '@highlight-run/next'
 
 export async function register() {
-	registerHighlight({
-		projectID: CONSTANTS.NEXT_PUBLIC_HIGHLIGHT_PROJECT_ID,
-		otlpEndpoint: CONSTANTS.NEXT_PUBLIC_HIGHLIGHT_OTLP_ENDPOINT,
-	})
+	if (process.env.NEXT_RUNTIME === 'nodejs') {
+		/**
+		 * Conditional import required for use with Next middleware
+		 *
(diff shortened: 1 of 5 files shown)


e2e/nextjs/instrumentation.ts
```diff
@@ -1,10 +1,20 @@
 // instrumentation.ts
 import CONSTANTS from '@/app/constants'
-import { registerHighlight } from '@highlight-run/next'
+// import { registerHighlight } from '@highlight-run/next'
```
Vadman97, line 3: should we remove the import?

Vadman97 (approved):

MichaelCGrant:
I think having a middleware.ts file will also cause "__dirname" to not be defined since it is also a nodejs feature.
My solution was:
`// next.config.mjs

import { fileURLToPath } from "url";
import { dirname } from "path";

const __filename = fileURLToPath(import.meta.url);
const __dirname = dirname(__filename);

// ...rest of config`

e2e/nextjs/instrumentation.ts
deltaepsilon: Yeah. I was of a split mind on this, but you're pulling me over to removing it.

deltaepsilon:
image

PR #5403, opened 2023-05-22
Loading state in right panel

eightants:
## Summary

Added loading state to appear in right panel when navigating through setting pages and data is still being retrieved. It's not part of the App loading state since it's just individual setting pages, so I thought to replace many of the 
```
if (loading) {
return null
}
```

with this new flexible loading div. 

https://www.loom.com/share/85552ecf9fba4bcb9954927d30a509ab


eightants:
Adding you on this Eric since you have been reviewing the past few PRs that were part of this change. Also it is a follow up on https://github.com/highlight/highlight/pull/5391

et (approved):
Thanks.

PR #5405, opened 2023-05-22
Clean up unused dependencies

ccschmitz:
## Summary

Used [depcheck](https://github.com/depcheck/depcheck) to find and clean up some unused dependencies in our frontend code.

Local click test seemed to work as expected.

I think if click test and CI checks pass we are probably in good shape.


frontend/package.json
```diff
@@ -12,19 +12,16 @@
 		"@highlight-run/client": "workspace:*",
 		"@highlight-run/react": "workspace:*",
 		"@highlight-run/react-mentions": "4.4.1",
-		"@highlight-run/react-select": "^5.2.2",
```
ccschmitz, line 15: We also have the main `react-select` package installed and are importing from that everywhere.

Vadman97 (approved):
render.com preview looks good throughout our products

frontend/package.json
ccschmitz: Yeah, I was thinking we of doing that as a follow up. Wanted to make sure we weren't trying to install it in CI before cleaning it up.

ccschmitz: I did end up archiving this: https://github.com/highlight/react-select

Vadman97: thank you!

PR #5406, opened 2023-05-22
Changelog 19

deltaepsilon:
image


docs-content/general/changelog/changelog-19.md
```diff
@@ -0,0 +1,34 @@
+---
+title: Changelog 19 (05/22)
+slug: changelog-19
+---
+
+## Demo Project
+
+Check out [https://app.highlight.io/demo](https://app.highlight.io/demo) to check out our new demo project.
```
ccschmitz, line 8: Maybe intentional, but do we want to say "check out" twice in this sentence or should we come up with another way of saying this?

Also, do you think we should mention anything about what the project is (our marketing site)? Maybe we could even say something about people being able to find their own sessions on our marketing site if they look closely.

docs-content/general/changelog/changelog-19.md
```diff
@@ -0,0 +1,34 @@
...
+
+You can play around and see how the Highlight Dashboard works with a fully-featured project with plenty of live data.
```
ccschmitz, line 10: Nit, but "Dashboard" might be misleading here since we also have something called "Dashboards" in the product (https://app.highlight.io/1/dashboards).

Also, this technically isn't a "fully-featured project" since there are no backend logs or errors, but maybe that doesn't matter 🤷

docs-content/general/changelog/changelog-19.md
deltaepsilon: oooOOooOOO.

Thanks for catching this terrible copy. I'm fixing it now.

I'll try to come up with something fun around finding one's own sessions.

docs-content/general/changelog/changelog-19.md
deltaepsilon: Thanks for catching that. I'll tighten it up.

jay-khatri (approved):

PR #5413, opened 2023-05-22
remove feedbackWidget option from client

et:
## Summary

This PR removes the `feedbackWidget` option from the client. This is an unused feature from way back in the day that nobody really uses.
Note: this is different from the `<ErrorBoundary />` component but they both use the same endpoint of `addSessionFeedback`.

Confirmed `<ErrorBoundary />` feedback still works correctly.

Since this is a breaking change, I've bumped the major version.


et:
@jay-khatri - We talked about this in the past. Can you confirm you're still okay with this being removed?

jay-khatri (approved):

PR #5415, opened 2023-05-23
add cypress test that verifies the browser unpkg highlight.run bundle

Vadman97:
## Summary

Adds cypress tests for our umd browser installation workflow.
This change is motivated by a bug introduced in #5282 to our highlight.run published npm `package.json` 
which broke the bundle when used in html `<script/>` tags without a transpilation step.

The new testcase sets up the script tag highlight installation per our docs and ensures [public graph requests](https://www.highlight.io/docs/getting-started/client-sdk/other) are made

CI github action
image

No

### DIFF
--- a/cypress/e2e/umd.cy.js
+++ b/cypress/e2e/umd.cy.js
@@ -3,11 +3,6 @@ describe('web client recording spec', () => {
 		it('public graph requests are recorded', { baseUrl: null }, () => {
 			cy.intercept('POST', '/public', (req) => {
 				req.alias = req.body.operationName
-				req.continue(() => {
-					if (Array.isArray(req.body.variables.events?.events)) {
-						events.push(...req.body.variables.events.events)
-					}
-				})
 			})
 			cy.visit(`./cypress/pages/${source}.html`)
 			cy.window().then((win) => {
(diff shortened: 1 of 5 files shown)


cypress/e2e/umd.cy.js
```diff
@@ -0,0 +1,20 @@
+describe('web client recording spec', () => {
+	;['unpkg', 'jsdelivr'].map((source) => {
```
SpennyNDaJets, line 2: Is this `;` necessary?

cypress/e2e/umd.cy.js
```diff
@@ -0,0 +1,20 @@
...
+		it('public graph requests are recorded', { baseUrl: null }, () => {
```
SpennyNDaJets, line 3: Nit: could be nice to interpolate the `source` into this name incase one starts failing?

SpennyNDaJets (approved):
Looks good! Left a couple comments, but nothing blocking

cypress/e2e/umd.cy.js
Vadman97: eslint puts this here for some reason when i inline the array. i'll bring it out to clean this up.

cypress/e2e/umd.cy.js
Vadman97: good idea!

PR #5423, opened 2023-05-23
Spelling: GrqphQL to GraphQL

deltaepsilon:
☝️


deltaepsilon:
The diff on this is bad. It should be a one-line change, but it's getting confused.

For example, you can see that earlier misspellings were already corrected on `main`:

https://raw.githubusercontent.com/highlight/highlight/main/docs-content/general/6_product-features/1_session-replay/graphql.md

Issue #5190, opened 2023-05-01
Expand width of snooze window


et:
We can have undesirable wrapping when the snooze time has an hour window with two digits
![Screenshot 2023-05-01 at 9 30 51 AM](https://user-images.githubusercontent.com/58678/235478609-8d2ab9df-79ab-489a-a06b-2d0fc057d1ee.png)

et:
> Hello @et , i'd love to take this on.

Go for it!

et:
Fixed in #5427

PR #5427, opened 2023-05-23
chore: expand the width of the snooze time window

betty-alagwu:
Hello @et, this is the first pass for issue #5190.
I've been exploring different solution and i decided to go this. Please let me know if solution works

Here's how it looks now

Screenshot 2023-05-23 at 4 23 04 PM


frontend/src/pages/ErrorsV2/ErrorStateSelect/ErrorStateSelect.tsx
```diff
@@ -288,7 +288,11 @@ const ErrorStateSelectImpl: React.FC<Props> = ({
 										<Box color="n11" flex="stretch">
 											{option.title}
 										</Box>
-										<Box color="n9" flex="stretch">
+										<Box
+											color="n9"
+											flex="stretch"
+											style={{ width: 150 }}
```
et, line 294: @ccschmitz - can you provide guidance here? This feels like something the box component should just solve for us.

I tend to think the issue is that the menu component isn't granting enough horizontal space and that we should be making the changes at that level. Curious what you think. Thanks!

ccschmitz: Yeah, hard-coding `width` values in a `style` prop does feel janky right now. I know there are places where we need fixes values, and we can probably come up with a better API for that, but I'm wondering if that's the right way to solve this problem 🤔 - @betty-alagwu what do you think about setting `whiteSpace="nowrap"` here instead? I think the menu should automatically expand to the width of the contents.

betty-alagwu: Thank you for your suggestion @ccschmitz . I attempted to set whiteSpace="nowrap"  on the container holding the `option.title` and the `option.time.format(DATE_FORMAT)` but i wasn't happy with the way it solves the problem. Now I applied it in the container holding the `option.time.format(DATE_FORMAT)` and it works just fine.

Thank you.

frontend/src/pages/ErrorsV2/ErrorStateSelect/ErrorStateSelect.tsx
```diff
@@ -288,7 +288,13 @@ const ErrorStateSelectImpl: React.FC<Props> = ({
 										<Box color="n11" flex="stretch">
 											{option.title}
 										</Box>
-										<Box color="n9" flex="stretch">
+										<Box
+											color="n9"
+											flex="stretch"
+											style={{
+												whiteSpace: 'nowrap',
+											}}
```
ccschmitz, line 296: There should be a `whiteSpace` prop available on `Box`. Could we use that instead of the `style` prop?

ccschmitz, line 296: There should be a whiteSpace prop available on Box. Could we use that instead of the style prop?

betty-alagwu: Sure, just fixed it.

ccschmitz: Perfect - thank you!

frontend/src/pages/ErrorsV2/ErrorStateSelect/ErrorStateSelect.tsx
betty-alagwu: Thank you for your suggestion @ccschmitz . I attempted to set whiteSpace="nowrap"  on the container holding the `option.title` and the `option.time.format(DATE_FORMAT)` but i wasn't happy with the way it solves the problem. Now I applied it in the container holding the `option.time.format(DATE_FORMAT)` and it works just fine.

Thank you.

et:
Thanks @betty-alagwu!

PR #5433, opened 2023-05-23
fix billing details caching stripe url

Vadman97:
## Summary

Operation name had the wrong case in the indexeddb cache exclusion policy.
Also stops emitting github integration errors as they are not actual errors.

cache now bypassed for the payment settings page

Screenshot 2023-05-23 at 1 02 01 PM

different stripe urls loaded

Screenshot 2023-05-23 at 1 02 47 PM

No


deltaepsilon:
That was the bug!!!

PR #5434, opened 2023-05-23
Simplify logic for creating backend setup events

ccschmitz:
## Summary

Simplifies the logic for creating backend setup events by processing them inline before creating logs or errors rather than dispatching `MarkBackendSetup` messages in Kafka.

Local click test.

Will monitor the creation of backend setup events and performance of logs and errors processing.


frontend/src/components/Loading/Loading.tsx
```diff
@@ -5,7 +5,7 @@ import {
 } from '@context/AppLoadingContext'
 import SvgHighlightLogoWithNoBackground from '@icons/HighlightLogoWithNoBackground'
 import { Spin } from 'antd'
-import classNames from 'classnames'
+import clsx from 'clsx'
```
ccschmitz, line 8: This was removed in #5405

backend/kafka-queue/types.go
```diff
@@ -22,7 +22,6 @@ const (
 	AddSessionProperties             PayloadType = iota
 	PushBackendPayload               PayloadType = iota
 	PushMetrics                      PayloadType = iota
-	MarkBackendSetup                 PayloadType = iota
```
Vadman97, line 25: we'll want to keep this here and add a note about it being deprecated. otherwise messages currently in flight will be interpreted incorrectly, as these `iota` values will be offset by 1 from the values in the current kafka messages

```suggestion
	// Deprecated: setup events are written from other payload processing
	MarkBackendSetup                 PayloadType = iota
```

backend/kafka-queue/types.go
```diff
@@ -149,7 +141,6 @@ type Message struct {
 	AddSessionProperties             *AddSessionPropertiesArgs
 	PushBackendPayload               *PushBackendPayloadArgs
 	PushMetrics                      *PushMetricsArgs
-	MarkBackendSetup                 *MarkBackendSetupArgs
```
Vadman97, line 152: safe to remove this from the message though as the json deserialization will ignore keys not part of this struct

backend/public-graph/graph/resolver.go
```diff
@@ -2355,6 +2335,10 @@ func (r *Resolver) ProcessBackendPayloadImpl(ctx context.Context, sessionSecureI
 			continue
 		}
 
+		if v.Source == "" {
+			v.Source = "backend"
+		}
```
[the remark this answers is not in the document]
Vadman97: this will ensure that other backend errors reported without a source will use `backend` as the default value rather than an empty string

Vadman97 (approved):
thank you! just mentioning a small tweak to the kafka type constants

backend/kafka-queue/types.go
ccschmitz: Ahh, yes, I meant to ask about this because I saw another note about a deprecated message type. Thanks for flagging.

PR #5438, opened 2023-05-24
Request Headers

deltaepsilon:
## Summary

We've had consistent bug reports of missing headers.

Here's an example:
https://discord.com/channels/1026884757667188757/1109049760909885480

This likely closes a bunch of other issues in the repo.

I used [sindresorhus/ky](https://github.com/sindresorhus/ky) to reproduce. This library uses the native `Request` object in a way that we weren't supporting.

Nope! Just a patch version bump.

### DIFF
--- a/sdk/client/src/listeners/network-listener/utils/fetch-listener.ts
+++ b/sdk/client/src/listeners/network-listener/utils/fetch-listener.ts
@@ -1,17 +1,18 @@
-import { getBodyThatShouldBeRecorded } from './xhr-listener'
-import { NetworkListenerCallback } from '../network-listener'
 import {
-	RequestResponsePair,
-	Request as HighlightRequest,
-	Response as HighlightResponse,
-} from './models'
-import {
+	HIGHLIGHT_REQUEST_HEADER,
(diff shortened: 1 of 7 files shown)


sdk/client/src/listeners/network-listener/utils/fetch-listener.ts
```diff
@@ -43,10 +44,18 @@ export const FetchListener = (
 			init = init || {}
 			// Pre-existing headers could be one of three different formats; this reads all of them.
 			let headers = new Headers(init.headers)
+
+			if (input instanceof Request) {
```
deltaepsilon, line 48: This is the critical change.

sdk/firstload/package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "highlight.run",
-	"version": "6.5.2",
+	"version": "6.5.3",
```
deltaepsilon, line 3: @Vadman97 Can you confirm that this is the correct way to bump the version?

Vadman97 (approved):

PR #5448, opened 2023-05-25
Fix import missing from package.json

lewisl9029:
## Summary

Looks we removed the classnames package from package json without removing its one remaining usage. This worked fine for the vite build since it picked up classnames from a hoisted transitive dep, but reflame only exposes direct dependencies so previews ended up breaking.

This PR changes the last usage of classnames to use clsx instead, and also adds a lint rule to prevent situations like this from happening again.

Running `yarn lint` in frontend results in errors like this when importing from transitive deps:

Screenshot 2023-05-24 at 11 26 04 PM


Vadman97 (approved):

Vadman97:
thank you for the fix and the lint rule @lewisl9029 !

lewisl9029:
Hmm looks like the commit with the lint rule broke our esbuild script. Reverting it for now so we can get the fix in first. Will take another look in the morning.

lewisl9029:
Looks like it was caused by duplicated vanilla extract dependencies, just pushed a fix, should be good this time @Vadman97 🙏

Vadman97 (approved):

PR #5449, opened 2023-05-25
Refactor auth logic

ccschmitz:

### DIFF
--- a/frontend/src/authentication/AuthContext.tsx
+++ b/frontend/src/authentication/AuthContext.tsx
@@ -29,5 +29,7 @@ export const [useAuthContext, AuthContextProvider] = createContext<{
 	isAuthLoading: boolean
 	isLoggedIn: boolean
 	isHighlightAdmin: boolean
-	refetchAdmin: () => Promise<any>
+	signIn: () => void
+	signOut: () => void
+	fetchAdmin: () => void
 }>('AuthContext')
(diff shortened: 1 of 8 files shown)


frontend/src/components/Button/index.tsx
```diff
@@ -30,6 +30,7 @@ export const Button: React.FC<ButtonProps> = ({
 }) => {
 	const handleClick: React.MouseEventHandler<HTMLButtonElement> = (e) => {
 		if (loading) {
+			e.preventDefault()
```
ccschmitz, line 33: Unrelated, but a source of bugs where you could trigger the `onClick` handler multiple times even while the button was in the loading state.

ccschmitz:
@Vadman97 would you mind giving this one a look?

Vadman97 (approved):
code looks good, tested locally working with:
* firebase auth google new account, deleting admin record
* firebase auth regular account
* simple auth

however it does look like 2FA is broken:
image
i see a loading spinner after putting in the code and the network request to verify fires, but nothing happens after. trying to press next afterwards shows
image

ccschmitz:
@Vadman97 thanks for click testing! I just fixed this up in https://github.com/highlight/highlight/pull/5449/commits/7a55314515280ae7635f16e86dad2650e1ff9787.

PR #5450, opened 2023-05-25
Create pending invites tab

SpennyNDaJets:
## Summary
We want to allow the User Management page to show "Pending Invites" depending on the tab the user selects.
[Figma designs
](https://www.figma.com/file/rdBPNzn7Klk6RG1LHUCE5B/Screen-Designs?type=design&node-id=8287-324021&t=SPkMxFB7uP9tkc89-0)

This PR is a step towards these changes:
- Create nav bar that will allow user to select "Pending invites" tab"
  - "members" tab shown by default
  - "invites" tab is disabled, but navigatable through `/w/1/team/invites`
- Updates to the "Invite user" button
  - Update copy and add icon
  - Move out of the top level header down to the tab content
- Rename page from "User management" to "Members"

1. Select "Workspace settings" option in the settings menu
2. Switch the top level tabs on the left to ensure they still work
3. Select the "Members" tab to see the updated team page
- [ ] Confirm "Members" tab is active
- [ ] Confirm "Pending Invites" tab is disabled
- [ ] Can still update "Auto join" settings
- [ ] Can still invite users
- [ ] Can still manage members
4) Navigate to "Pending Invites" page (`/w/1/team/invites`)
- [ ] Confirm "Pending Invites" tab is selected
- [ ] "Coming soon" content is sufficient to display
- [ ] Can switch back to "Members" tab

https://github.com/highlight/highlight/assets/17744174/a2574ff3-e038-4850-a7b6-4c0d44d45d5d

N/A

## Considerations
- Is "Coming soon" content sufficient to display in production, or should we hide behind variable?
- Do we want to include hotkeys (`m` and `p`) for switch tabs?
  - There will be numbers in the places of hotkeys, so may need a different way to educate. So left them out for now.


frontend/src/pages/WorkspaceTeam/WorkspaceTeam.tsx
```diff
@@ -1,26 +1,62 @@
...
+				<Button
+					className={styles.inviteButton}
+					trackingId="WorkspaceTeamInviteMember"
+					type="primary"
+					onClick={toggleShowModal}
+					small
+				>
+					<IconSolidUserAdd />
+					<Text>Invite users</Text>
+				</Button>
```
ccschmitz, line 54: If you import `Button` from `components/Button` instead of `components/Button/Button/Button` you should be able to use the `iconLeft` prop to display the icon and text the way you want it here.

That button is also up to date with the latest mocks in Figma so the props will match up better too (e.g. `size="small"`, `kind="primary"`, and `emphasis="high"`).

image

frontend/src/pages/WorkspaceTeam/WorkspaceTeam.tsx
```diff
@@ -1,26 +1,62 @@
...
 	const { data, error, loading } = useGetWorkspaceAdminsQuery({
 		variables: { workspace_id: workspace_id! },
 		skip: !workspace_id,
 	})
+	const navigate = useNavigate()
 	const [showModal, toggleShowModal] = useToggle(false)
 
+	const TabContentContainer = ({ children }: { children: any }) => (
+		<Box mt="8">
+			<div className={styles.tabTitleContainer}>
```
ccschmitz, line 34: Nit, but I don't see anything being applied in the `tabTitleContainer` CSS properties that couldn't be applied directly through `Box` or `Stack`. e.g.

.tabTitleContainer {
	align-items: center;
	display: flex;
	justify-content: space-between;
	margin-bottom: var(--size-xSmall);
}


```tsx
<Stack
  mb="8"
  align="center"
  justify="space-between"
>
 ...
</Stack>
```

In general, we try to leverage the `Box` component for applying basic styles like this. It has access to all of our CSS utility classes (similar to something like Tailwind, but created with a tool called Sprinkles from [vanilla-extract](https://vanilla-extract.style/)) via its props. It's inevitable we will have to write some custom styles once in a while, but we're trying to move away from SCSS modules entirely + leverage the utility classes via `Box` and `Stack` whenever possible.

There is [a tech talk on some of this](https://www.notion.so/runhighlight/Building-A-Component-Library-e0b9e29f381342328532686a49c11780?pvs=4), but it's probably worth going over in more detail the next time we pair!

ccschmitz (approved):
Looks great, Spencer! ✨ 

Left some comments about leveraging our design system components + utility classes in general rather than using SCSS to provide custom styles, but feel free to handle some of that refactoring in a future PR.

ccschmitz (approved):
Looks great, Spencer! ✨ 

Left some comments about leveraging our design system components + utility classes in general rather than using SCSS to provide custom styles, but feel free to handle some of that refactoring in a future PR.

frontend/src/pages/WorkspaceTeam/WorkspaceTeam.tsx
SpennyNDaJets: Good to know! Thanks for sharing

PR #5465, opened 2023-05-26
Only create setup events for backend logs

ccschmitz:
## Summary

Only create `logs` setup events for backend logs.

Local click test.

Will monitor setup event creation in prod.


Vadman97 (approved):
makes sense! this makes the setup_event for `logs` backend only which was the original intention

PR #5473, opened 2023-05-30
Add delete invite functionality

SpennyNDaJets:


backend/private-graph/graph/schema.resolvers.go
```diff
@@ -926,6 +926,25 @@ func (r *mutationResolver) AddAdminToWorkspace(ctx context.Context, workspaceID
...
+	if err := r.validateAdminRole(ctx, workspaceID); err != nil {
+		return false, e.Wrap(err, "A non-Admin role Admin tried changing an admin role.")
+	}
+
+	result := r.DB.Where("id = ?", workspaceInviteLinkID).Where("workspace_id = ?", workspaceID).Delete(&model.WorkspaceInviteLink{})
+	if result.Error != nil {
+		return false, e.Wrap(err, "error deleting workspace invite link")
+	}
+
+	return int(result.RowsAffected) > 0, nil
```
SpennyNDaJets, line 945: Want to call out we are returning a boolean here to confirm that the row was deleted. It is possible to pass in an inviteId with an incorrect workspaceId in attempt to validate the admin has permissions, but should then be filtered out in the SQL query

backend/private-graph/graph/schema.resolvers.go
```diff
@@ -926,6 +926,25 @@ func (r *mutationResolver) AddAdminToWorkspace(ctx context.Context, workspaceID
...
 }
 
+// DeleteInviteLinkFromWorkspace is the resolver for the deleteInviteLinkFromWorkspace field.
+func (r *mutationResolver) DeleteInviteLinkFromWorkspace(ctx context.Context, workspaceID int, workspaceInviteLinkID int) (bool, error) {
+	_, err := r.isAdminInWorkspace(ctx, workspaceID)
+	if err != nil {
+		return false, e.Wrap(err, "current admin is not in workspace")
+	}
+
+	if err := r.validateAdminRole(ctx, workspaceID); err != nil {
```
SpennyNDaJets, line 936: Should only admins be able to delete invites? Seems like member users should not be able to delete other invites, but they can invite other members

backend/private-graph/graph/resolver_test.go
Vadman97, line 1: thanks for adding a test!

backend/private-graph/graph/schema.resolvers.go
Vadman97: maybe it doesn't make sense for a member to invite another admin, only another member?

frontend/src/pages/WorkspaceTeam/WorkspaceTeam.tsx
```diff
@@ -87,9 +87,7 @@ const WorkspaceTeam = () => {
 						{
 							key: 'invites',
 							// TODO(spenny): load count
```
Vadman97, line 89: is this still relevant?

backend/private-graph/graph/schema.resolvers.go
SpennyNDaJets: I'll add an issue for this

PR #5479, opened 2023-05-30
Fix autoresolve query

et:
## Summary

The previous query would overmatch since it was not considering whether there were error objects that were recently created.

Before:

```sql
SELECT DISTINCT (error_groups.id), error_groups.project_id
FROM
	"error_groups"
	INNER JOIN error_objects ON error_groups.id = error_objects.error_group_id
WHERE error_groups.state = 'OPEN'
	AND error_groups.project_id = 1
	AND error_objects.created_at < '2023-04-30 11:54:57.589'
```

After:

```sql
SELECT DISTINCT (error_groups.id), error_groups.project_id
FROM
	"error_groups"
WHERE
	"error_groups"."project_id" = 1
	AND "error_groups"."state" = 'OPEN'
	AND id NOT IN(
		SELECT
			"error_group_id" FROM "error_objects"
		WHERE
			created_at >= '2023-04-30 11:54:57.589')
```

Added a regression test.

N/A

### DIFF
--- a/backend/worker/autoresolver.go
+++ b/backend/worker/autoresolver.go
@@ -54,15 +54,18 @@ func (autoResolver *AutoResolver) resolveStaleErrorsForProject(ctx context.Conte
 
 	db := autoResolver.db
 
+	subQuery := db.
+		Model(model.ErrorObject{}).
+		Select("error_group_id").
+		Where("created_at >= ?", time.Now().AddDate(0, 0, -interval))
+
 	err := db.Debug().
-		Table("error_groups").
-		Select("DISTINCT(error_groups.id), error_groups.project_id").
-		Joins("INNER JOIN error_objects ON error_groups.id = error_objects.error_group_id").
-		Where("error_groups.state = ?", privateModel.ErrorStateOpen).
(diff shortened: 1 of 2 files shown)


backend/worker/autoresolver.go
```diff
@@ -54,15 +54,19 @@ func (autoResolver *AutoResolver) resolveStaleErrorsForProject(ctx context.Conte
 
 	db := autoResolver.db
 
+	subQuery := db.
+		Model(model.ErrorObject{}).
+		Select("error_group_id").
+		Where("created_at >= ?", time.Now().AddDate(0, 0, -interval))
+
 	err := db.Debug().
```
Vadman97, line 62: should we remove the `.Debug().` printing here for production?

Vadman97 (approved):
🙏

backend/worker/autoresolver.go
et: I don't think there's any harm in keeping it since there's nothing sensitive and that this will log every 30 minutes (current cron setup).

PR #5482, opened 2023-05-30
update next build clean script to be platform agnostic

Vadman97:
## Summary

Updates the next build pipeline to work on all platforms by switching the `sed` command to `printf`. 
`sed` on mac would error with `sed: 1: "dist/HighlightInit.js": extra characters at the end of d command` despite working on linux.
The build step is necessary because the bundle must be marked with `"use client"` to make sure next.js does not
try to use the highlight component in server-side rendered code.

image


SpennyNDaJets (approved):
LGTM

PR #5483, opened 2023-05-30
Improve onboarding experience from feedback

SpennyNDaJets:
Would love feedback for any of the solutions applied to address the customer feedback
**Recommend reviewing code hiding whitespace changes**

## Summary
We received feedback on the onboarding experience that certain sections are unclear
- copy button in setup is useless since it copies npm and yarn instructions
- `<YOUR_PROJECT_ID>` is confusing since it has already been replaced in the code snippet
- urlBlocklist should have an example

The following changes were made:
- Allow each section to pass in multiple code blocks as an array, so users can copy only one of the blocks
- Add `pnpm` command when appropriate
- Replace the use of `<YOUR_PROJECT_ID>` in the content section so that it is clear what to update
  -  Note: these are used when the user is signed in on the app and when an unauthenticated user views the setup docs, so we should make sure this is clear in both contexts.
- Add examples to each of the `urlBlocklist` with the urls that we already block by default

1. Start the Highlight app, and view a workspace that has not been initialized.
2. View each of the frontend frameworks
- [ ] Each initialization step of the frontend frameworks makes sense and is consistent
3. Start the Highlight website
4. View the Frontend setup guides (e.g. `/docs/getting-started/client-sdk/reactjs`)
- [ ] Each initialization step of the frontend frameworks makes sense and is consistent'

Screenshot 2023-06-05 at 1 11 34 PM

Screenshot 2023-06-05 at 1 12 40 PM

N/A


mayberryzane (approved):
looks good! spot checked the vercel and reflame previews

PR #5494, opened 2023-05-31
fix clickup integration

et:
## Summary

When checking if a project has clickup integrated, the `IsProjectIntegratedWith` endpoint was always returning `false`. This looks to be a regression that happened some time ago and we never realized the clickup integration was broken until a customer wrote in (clickup doesn't get a lot of usage). 

Added a regression unit test for verifying if a prorject is integrated with Clickup.

Verified manually that we can manually create an issue into a Clickup Task and can choose the Clickup List
Screenshot 2023-05-31 at 10 20 12 AM
Screenshot 2023-05-31 at 10 18 44 AM

N/A

### DIFF
--- a/backend/integrations/integrations.go
+++ b/backend/integrations/integrations.go
@@ -116,6 +116,22 @@ func (c *Client) GetWorkspaceAccessToken(ctx context.Context, workspace *model.W
 }
 
 func (c *Client) IsProjectIntegrated(ctx context.Context, project *model.Project, integrationType modelInputs.IntegrationType) (bool, error) {
+	if integrationType == modelInputs.IntegrationTypeClickUp {
+		var projectMapping *model.IntegrationProjectMapping
+		if err := c.db.Where(&model.IntegrationProjectMapping{
+			ProjectID:       project.ID,
+			IntegrationType: integrationType,
+		}).First(&projectMapping).Error; err != nil {
+			return false, nil
+		}
+
+		if projectMapping == nil {
+			return false, nil
+		}
+
+		return true, nil
+	}
+
 	var workspace model.Workspace
 	if err := c.db.Where(&model.Workspace{Model: model.Model{ID: project.WorkspaceID}}).First(&workspace).Error; err != nil {
 		return false, nil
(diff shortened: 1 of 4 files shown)


backend/private-graph/graph/schema.resolvers.go
```diff
@@ -6027,21 +6027,6 @@ func (r *queryResolver) IsProjectIntegratedWith(ctx context.Context, integration
...
-		if err := r.DB.Where(&model.IntegrationProjectMapping{
-			ProjectID:       project.ID,
-			IntegrationType: integrationType,
-		}).First(&projectMapping).Error; err != nil {
-			return false, nil
-		}
-
-		if projectMapping == nil {
-			return false, nil
-		}
```
et, line 6041: This was the bug, we should have had an `else` here with `return true, nil`

et: I moved the code into the `backend/integrations` directory to assist with writing a regression test.

backend/integrations/integrations.go
```diff
@@ -116,6 +116,22 @@ func (c *Client) GetWorkspaceAccessToken(ctx context.Context, workspace *model.W
 }
 
 func (c *Client) IsProjectIntegrated(ctx context.Context, project *model.Project, integrationType modelInputs.IntegrationType) (bool, error) {
+	if integrationType == modelInputs.IntegrationTypeClickUp {
+		var projectMapping *model.IntegrationProjectMapping
+		if err := c.db.Where(&model.IntegrationProjectMapping{
+			ProjectID:       project.ID,
+			IntegrationType: integrationType,
+		}).First(&projectMapping).Error; err != nil {
+			return false, nil
```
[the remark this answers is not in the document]
et: Good question. Gorm considers it an error if the [record](https://gorm.io/docs/query.html#Retrieving-a-single-object) isn't found.

Screenshot 2023-05-31 at 10 48 17 AM

Per the docs, we could switch to using `Limit(1).Find` if we find this behavior undesirable but I think it's okay to leave as is for now.

PR #5504, opened 2023-06-01
create a reusable select component

et:


ccschmitz (approved):
Nice! Looks great. Agreed that we can upgrade to use Ariakit's `Select` at some point, but I like the lightweight abstraction you have here. I created [a ticket](https://github.com/highlight/highlight/issues/5507) to track upgrading this later.

PR #5506, opened 2023-06-01
Better UX of switching between the inspect and comment modes

ccschmitz:
## Summary

Makes it more obvious what mode you are in when viewing a session by adding a switch in the top right that visually shows you what mode you're in and allows you to switch between modes.

Screenshot 2023-06-01 at 11 29 03 AM

Also tweaks some of the styles of other items in that toolbar: 

* Removes the icon on share button
* Converts dimensions + privacy to badges
* Moves dimensions + privacy left of the current URL

These changes came through some pairing with Julian.

Local click test.

N/A


Vadman97 (approved):
preview looks great. 🚢

PR #5508, opened 2023-06-01
Add tracking for auth pages

ccschmitz:
## Summary

Adds some missing tracking events for our pre-auth pages.

Local click test.

N/A - only tracking changes


ccschmitz:
cc @jay-khatri and @Vadman97 - I'm thinking these events could be useful in measuring the funnel in Mixpanel. We may have better ways of looking at this now in Hex, but thought it couldn't hurt.

SpennyNDaJets (approved):
image

PR #5512, opened 2023-06-01
Add config to test current file in debug mode

SpennyNDaJets:
## Summary
Its tough to run go tests in debug mode. Add a debug config to run a go test file in debug mode.

1. Open a go test file
2. Place a break point in the file
3. Open the "Run and Debug" tab in VSCode
4. Run "Test Current File"
- [ ] Only tests in the current directory are ran
- [ ] Catches the breakpoint

Screenshot 2023-06-01 at 4 10 13 PM

N/A

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,7 +1,7 @@
 {
 	// Use IntelliSense to learn about possible attributes.
 	// Hover to view descriptions of existing attributes.
-	// For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+	// For more information, visit: https://go.microsoft.com/fwlink/?linkid": " 830387
 	"version": "0.2.0",
 	"configurations": [
 		{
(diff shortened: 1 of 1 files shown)


.vscode/launch.json
```diff
@@ -11,6 +11,23 @@
...
 			"mode": "remote",
 			"port": 2345
-		}
+		},
+        {
+            "name": "Test Current File",
+            "type": "go",
+            "request": "launch",
+            "mode": "test",
+            "program": "${relativeFileDirname}",
```
SpennyNDaJets, line 20: Can't use file since it needs the package it is testing defined in its directory. As a result, if multiple test files are in the same directory, they will all get run, but "Test Current Directory" doesn't have the same ring to it

et:
Does "debug test" work for you?
Kapture 2023-06-01 at 14 17 00

Screenshot_2023-06-01_at_2_17_27_PM

SpennyNDaJets:
> Does "debug test" work for you?

@et I don't have the option looking at the file. What plugin are you using?

SpennyNDaJets:
Was able to get the test running in VSCode by installing the recommended
```
github.com/ramya-rao-a/go-outline@v0.0.0-20210608161538-9736a4bde949
golang.org/x/tools/gopls@latest
```
I think its most likely the latter but did them both at the same time

PR #5515, opened 2023-06-01
Network request multiselect

SpennyNDaJets:
## Summary
Allow the user to select multiple types of request types and statuses when viewing a session's network tab.

Click test: 
1. View a session from the sessions page (`/1/sessions/`)
2. Open the Network tab in the Dev tools tab
3. Click around with the filters for the Type and Status types
- [ ] Clicking a value when "All" is selected clears the "All" value
- [ ] Click the "All" value when other values are selected clears the other values
- [ ] Unselecting all the values selects the "All" value by default
- [ ] Logs respect multiple filters correctly 
- [ ] Counts respect multiple filters correctly

https://github.com/highlight/highlight/assets/17744174/446c877e-062c-4933-9f17-21bd862668a5

N/A


jay-khatri:
so cool! small comment:
- Could we give each dropdown a minimum width (that's maybe 20% larger than the current)? Its a bit weird that the width of each dropdown is different.

deltaepsilon (approved):

PR #5516, opened 2023-06-01
Create the multiselect component

SpennyNDaJets:
## Summary
Builds a reusable multiselect component in our UI package - use the accessible `ariakit` select components as a base

Implemented in https://github.com/highlight/highlight/pull/5515

https://github.com/highlight/highlight/assets/17744174/1dc22028-5ee9-4939-9aaf-2f4f393700d5

N/A


packages/ui/src/components/MultiSelectButton/MultiSelectButton.tsx
SpennyNDaJets, line 1: I originally just designed for the network page multi-select so would love feedback on what we be best to be abstracted out (especially regarding designs). happy to implement that as part of this PR or to build off as we go

et (approved):
Thank you! We've been needing this for awhile.
My only feedback is if we can change the cursor icon when hovering an option to a hand icon so it's more obvious that it's clickable.

Issue #5471, opened 2023-05-30
Improve how we handle unprocessed sessions that can't be replayed


ccschmitz:
Many live sessions seem to get stuck in the loading state when trying to view them on Highlight. Need to figure out why these sessions aren't loading and see if we can filter them out of the feed to avoid confusion about sessions not loading.

```[tasklist]
### Tasks
- [ ] Update exclusion logic in PushPayload when initializing a session. Should exclude a session by default unless it has real events.
- [ ] Update client logic to not create a new session when in an inactive tab.
```

## Sessions to Test

* A live session that does not have any events (click around in live sessions)
* A completed session that does not have any events ([example](https://app.highlight.io/6629/sessions/WEi6IB5Tc0BBF3RScCphQkyYlUGb))

ccschmitz:
Example of viewing a session that says it's live but couldn't be loaded. I _think_ this is because there are no real events to show for the session, but if that's the case, I don't understand the graph on the right side of the session card.

![Image](https://github.com/highlight/highlight/assets/308182/a3ea8c1f-1961-4462-82c3-35c62593fdb8)

Here's the payload from `GetSessionPayload`:

```json
{
  "data": {
    "events": [
      {
        "_sid": 55959,
        "data": {
          "payload": "{\"jsHeapSizeLimit\":2172649472,\"usedJSHeapSize\":246553566,\"relativeTimestamp\":14340.055,\"fps\":0}",
          "tag": "Performance"
        },
        "timestamp": 1685457527899,
        "type": 5
      },
      {
        "_sid": 55960,
        "data": {
          "payload": "{\"userIdentifier\":\"4871\"}",
          "tag": "Identify"
        },
        "timestamp": 1685457588141,
        "type": 5
      },
      {
        "_sid": 55961,
        "data": {
          "payload": "{\"userIdentifier\":\"4871\"}",
          "tag": "Identify"
        },
        "timestamp": 1685457588145,
        "type": 5
      },
      {
        "_sid": 55962,
        "data": {
          "payload": "https://app.highlight.io/1031/sessions/8e7OQDIzSGqnIvC5oTvODnWKCJqB",
          "tag": "Navigate"
        },
        "timestamp": 1685457589350,
        "type": 5
      },
      {
        "_sid": 55963,
        "data": {
          "payload": "{\"jsHeapSizeLimit\":2172649472,\"usedJSHeapSize\":246772086,\"relativeTimestamp\":0.265,\"fps\":0}",
[...]

PR #5518, opened 2023-06-02
add additional debugging information on when there are no timeline indicator events

et:
## Summary

Noticed this [comment](https://github.com/highlight/highlight/issues/5471#issuecomment-1568563563) and while looking at all the places where we exclude a session, I noticed this piece of code that excludes sessions when there are no timeline indicator events. I've added tracking to determine if this is a reason we exclude a session (see #5252).

I don't think this PR solves this comment linked but it may help with understanding why this query returns non-zero results:

```sql
SELECT
	created_at,
	excluded,
	excluded_reason
FROM
	sessions
WHERE
	excluded = TRUE
	AND excluded_reason IS NULL
ORDER BY
	created_at DESC
LIMIT 100
```

Screenshot 2023-06-01 at 8 03 27 PM

### DIFF
--- a/backend/worker/worker.go
+++ b/backend/worker/worker.go
@@ -526,7 +526,7 @@ func (w *Worker) excludeSession(ctx context.Context, s *model.Session, reason ba
 			"session_obj": s}).Warnf("error excluding session (session_id=%d, identifier=%s, is_in_obj_already=%v, processed=%v): %v", s.ID, s.Identifier, s.ObjectStorageEnabled, s.Processed, err)
 	}
 
-	if err := w.Resolver.OpenSearch.Update(opensearch.IndexSessions, s.ID, map[string]interface{}{
+	if err := w.Resolver.OpenSearch.UpdateSynchronous(opensearch.IndexSessions, s.ID, map[string]interface{}{
 		"Excluded":  true,
 		"processed": true,
 	}); err != nil {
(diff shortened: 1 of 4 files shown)


backend/worker/worker.go
```diff
@@ -526,7 +526,7 @@ func (w *Worker) excludeSession(ctx context.Context, s *model.Session, reason ba
 			"session_obj": s}).Warnf("error excluding session (session_id=%d, identifier=%s, is_in_obj_already=%v, processed=%v): %v", s.ID, s.Identifier, s.ObjectStorageEnabled, s.Processed, err)
 	}
 
-	if err := w.Resolver.OpenSearch.Update(opensearch.IndexSessions, s.ID, map[string]interface{}{
+	if err := w.Resolver.OpenSearch.UpdateSynchronous(opensearch.IndexSessions, s.ID, map[string]interface{}{
```
et, line 529: Updated to use `UpdateSynchronous` per this slack [convo](https://highlightcorp.slack.com/archives/C02CJANPHQS/p1685660764377189) as well as the observation that the non async version, `Update`, can have unpredictable behavior (see #5454).

et:
Re-rolling reviewers since chris is out. The build failure is unrelated (see #5519)

Vadman97 (approved):
awesome!

PR #5530, opened 2023-06-02
retry sync updates to opensearch in case of conflict

et:
## Summary

Since switching to sync updates to opensearch (#5524), we are getting some conflict updates:
Screenshot 2023-06-02 at 1 41 40 PM

See this stackoverflow: https://stackoverflow.com/questions/36188917/how-to-fix-elasticsearch-conflicts-on-the-same-key-when-two-process-writing-at-t

These are fine since kafka will retry, however it's a lot of noise.

Per looking at the source, it looks like the async update had RetryOnConflict set. This PR ports that behavior to the sync update.

We should monitor to confirm this cools off these logs:
https://app.highlight.io/1/logs?query=error%20updating%20session%20in%20opensearch%20level%3Aerror%20code.filepath%3A%2Fbuild%2Fbackend%2Fpublic-graph%2Fgraph%2Fresolver.go

### DIFF
--- a/backend/opensearch/opensearch.go
+++ b/backend/opensearch/opensearch.go
@@ -299,10 +299,11 @@ func (c *Client) UpdateSynchronous(index Index, id int, obj interface{}) error {
 	indexStr := GetIndex(index)
 
 	req := opensearchapi.UpdateRequest{
-		Index:      indexStr,
-		DocumentID: documentId,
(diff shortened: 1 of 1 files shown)


Vadman97 (approved):
lgtm, checked that this doesnt apply to `opensearchapi.IndexRequest`

PR #5531, opened 2023-06-02
make all updates to opensearch via private graph synchronous

et:
## Summary

Similar to #5524, this makes all private graph requests that update opensearch to be synchronous

Confirmed that an update to opensearch via private graph (e.g. marking an error group as viewed) updates as expected.

Will monitor the same as #5524. Private graph traffic is a lot more negligible than public graph so I don't expect this to have any impact.


mayberryzane (approved):
🥳

Issue #5511, opened 2023-06-01
Unresolving errors is getting out of sync with opensearch


et:
![Screenshot_2023-06-01_at_1_59_23_PM](https://github.com/highlight/highlight/assets/58678/26a5bd98-f9a3-45fe-8aff-77f397565ad0)

It seems like when an error state is updated, postgres and opensearch aren't in sync.

Tasks

- [x] Retry sync updates on conflict (#5530)
- [x] Skip retries on sync flow (#5528)
- [x] Make public graph requests sync (#5524)
- [x] Make private graph requests sync (#5531)
- [ ] Make worker requests sync
- [ ] Remove async update (#5539) (will solve #5457)
- [ ] Ensure there are no more [OPENSEARCH_ERROR](https://app.highlight.io/1/logs?query=OPENSEARCH_ERROR) logs
- [ ] Remove async index
- [ ] Reindex error groups

PR #5546, opened 2023-06-05
ensure ignored errors stay ignored

et:
## Summary

If an ignored error reoccurs, it is reopening instead of staying ignored. This PR fixes that and ensures the opensearch data is the same as postgres.

Relates to #5511

Tests added.

Manual test:
* Resolved errors reopen if they reoccur
* Ignored errors stay ignore if they reoccur

N/A

### DIFF
--- a/backend/public-graph/graph/resolver.go
+++ b/backend/public-graph/graph/resolver.go
@@ -393,6 +393,8 @@ func (r *Resolver) GetOrCreateErrorGroup(ctx context.Context, errorObj *model.Er
 
 		updatedState := errorGroup.State
 
+		// Reopen resolved errors
+		// Note that ignored errors do change state
 		if updatedState == privateModel.ErrorStateResolved {
 			updatedState = privateModel.ErrorStateOpen
 		}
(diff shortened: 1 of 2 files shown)


backend/public-graph/graph/resolver.go
```diff
@@ -391,31 +391,29 @@ func (r *Resolver) GetOrCreateErrorGroup(ctx context.Context, errorObj *model.Er
 
 		environmentsString := getIncrementedEnvironmentCount(ctx, errorGroup, errorObj)
 
+		updatedState := errorGroup.State
+
+		if updatedState == privateModel.ErrorStateResolved {
+			updatedState = privateModel.ErrorStateOpen
+		}
```
Vadman97, line 398: this is what toggles a resolved error getting a new instance to be open? could we add a comment to express that, as it's a bit confusing what this logic is for.

et: Added in b4946ada087149827901336432060a6cc82a0b40

I still am unable to track down if this is a regression or if this ever actually worked but at least there's a test now to validate it.

PR #5550, opened 2023-06-05
fix: resolve public/private graph endpoints at frontend runtime

Vadman97:
## Summary

Currently, the hobby deploy does not work when hosted outside of `localhost` because the
frontend assumes the backend will be hosted at `https://localhost:8082/{private,public}`

Ensure that public and private graph endpoints are resolved at frontend container run time
rather than during the vite build, so that `.env` file changes are reflected in the running frontend.

Only pushes docker image on new release publish.
Speeds up frontend image build locally by allowing layers to be cached (only copies relevant files).

Renames `prod` docker stack references to `hobby`.

Local docker build and deploy.

image

image

Will release new docker image.


SpennyNDaJets (approved):
Code and click test make sense to me!

PR #5552, opened 2023-06-06
improve experience of recording cross-origin iframes

Vadman97:
## Summary

Improves the experience of using highlight with a cross-origin iframe setup by removing the
`isCrossOriginIframe` option (now automatically inferred) and by making the `H.init` setup more robust to the order in which it is called.

Before this change, `H.init` had to be called first in the parent window and then in the cross-origin iframe
to make sure the iframe would be recorded. This change adds window message listeners to the parent
and iframe to allow them to communicate about the rrweb initialization state to allow the iframe to wait
for the parent window to set up rrweb.

https://www.loom.com/share/7727bf44b7e2418db9d8f9dc8238d220

New minor version 7.1.0
https://highlightcorp.slack.com/archives/C01JFQ77WTX/p1685063049376099

### DIFF
--- a/sdk/client/src/index.tsx
+++ b/sdk/client/src/index.tsx
@@ -71,6 +71,12 @@ import {
 	JankListener,
 	JankPayload,
 } from './listeners/jank-listener/jank-listener'
+import {
+	HighlightIframeMessage,
+	HighlightIframeReponse,
+	IFRAME_PARENT_READY,
+	IFRAME_PARENT_RESPONSE,
+} from './types/iframe'
 
 export const HighlightWarning = (context: string, msg: any) => {
 	console.warn(`Highlight Warning: (${context}): `, { output: msg })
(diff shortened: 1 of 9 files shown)


sdk/client/src/index.tsx
```diff
@@ -270,6 +276,14 @@ export class Highlight {
 		// Old firstLoad versions (Feb 2022) do not pass in FirstLoadListeners, so we have to fallback to creating it
 		this._firstLoadListeners =
 			firstLoadListeners || new FirstLoadListeners(this.options)
+		try {
+			// throws if parent is cross-origin
```
deltaepsilon, line 280: VERY helpful comment

deltaepsilon (approved):

sdk/client/src/index.tsx
Vadman97: yeah that's a sneaky one :)

PR #5553, opened 2023-06-06
add auth checks to error object endpoints

et:
## Summary

This PR adds some auth checks to error object endpoints

On the main branch:

Copy one of the queries as curl and paste it into Postman
Screenshot 2023-06-05 at 8 10 48 PM

In Postman, uncheck the `token` param and observe that the request goes through:
Screenshot 2023-06-05 at 8 11 28 PM

On this branch:
Perform the same request and observe the request is rejected
Screenshot 2023-06-05 at 8 12 33 PM

If you check the `token` param, the request should go through:

Screenshot 2023-06-05 at 8 13 04 PM

N/A

### DIFF
--- a/backend/model/model.go
+++ b/backend/model/model.go
@@ -886,9 +886,9 @@ type ErrorObject struct {
 	SessionID        *int
 	TraceID          *string
 	SpanID           *string
-	LogCursor        *string    `gorm:"index:idx_error_object_log_cursor,option:CONCURRENTLY"`
-	ErrorGroupID     int        `gorm:"index:idx_error_group_id_id,priority:1,option:CONCURRENTLY"`
-	ErrorGroup       ErrorGroup `gorm:"foreignkey:ErrorGroupID"`
+	LogCursor        *string `gorm:"index:idx_error_object_log_cursor,option:CONCURRENTLY"`
+	ErrorGroupID     int     `gorm:"index:idx_error_group_id_id,priority:1,option:CONCURRENTLY"`
+	ErrorGroup       ErrorGroup
 	Event            string
 	Type             string
 	URL              string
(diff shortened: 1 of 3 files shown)


backend/model/model.go
```diff
@@ -886,8 +886,9 @@ type ErrorObject struct {
 	SessionID        *int
 	TraceID          *string
 	SpanID           *string
-	LogCursor        *string `gorm:"index:idx_error_object_log_cursor,option:CONCURRENTLY"`
-	ErrorGroupID     int     `gorm:"index:idx_error_group_id_id,priority:1,option:CONCURRENTLY"`
+	LogCursor        *string    `gorm:"index:idx_error_object_log_cursor,option:CONCURRENTLY"`
+	ErrorGroupID     int        `gorm:"index:idx_error_group_id_id,priority:1,option:CONCURRENTLY"`
+	ErrorGroup       ErrorGroup `gorm:"foreignkey:ErrorGroupID"`
```
et, line 891: Defining FK so we can preload the error group later on.

backend/private-graph/graph/resolver.go
```diff
@@ -915,6 +915,28 @@ func (r *Resolver) doesAdminOwnErrorGroup(ctx context.Context, errorGroupSecureI
 	return eg, true, nil
 }
 
+func (r *Resolver) canAdminViewErrorObject(ctx context.Context, errorObjectID int) (*model.ErrorObject, error) {
+	authSpan, _ := tracer.StartSpanFromContext(ctx, "resolver.internal.auth", tracer.ResourceName("canAdminViewErrorObject"))
+	defer authSpan.Finish()
+
+	errorObject := model.ErrorObject{}
+	if err := r.DB.Where(&model.ErrorObject{ID: errorObjectID}).First(&errorObject).Error; err != nil {
```
mayberryzane, line 923: do we need this first db call if we have the other on line 927?

backend/model/model.go
mayberryzane: is there any db blocking while the foreign key constraint is added or chance it will fail (e.g. if any error objects reference a null or a non-existent error group id)? I think it would be good to add this FK manually before deploying this change

mayberryzane (approved):

backend/private-graph/graph/resolver.go
et: Good catch! This was leftover as I tried to switch to Preload

PR #5554, opened 2023-06-06
Stop minifying generated css again

lewisl9029:
## Summary

This PR reverts back to outputting non-minified generated css for Reflame, in hopes that it could minimize frequency of merge conflicts in the generated file. 

Previously we had to start minifying because vanilla-extract's esbuild plugin outputted comments that depend on absolute path, so would differ between CI and local machines. In this PR we work around that by manually stripping out those comments with a regex.

Also reverts #5551, since treating the file as binary actually does the opposite of what we want (minimize frequency of merge conflicts), since no automatic conflict resolution could ever occur with binary files.

Styles in latest Reflame deployment looks fine. Should be good to go if the Reflame generated files check passes in CI. Edit: looks like we're good to go!

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,9 +1,9 @@
 backend/private-graph/graph/schema.resolvers.go linguist-generated=false
 backend/public-graph/graph/schema.resolvers.go linguist-generated=false
 frontend/src/graph/generated/* linguist-generated=true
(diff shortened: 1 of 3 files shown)


et (approved):
Thanks!

PR #5555, opened 2023-06-06
Show join workspace page after signing up if the user has joinable workspaces

ccschmitz:
## Summary

We want to prompt users to join a workspace after they confirm their email if they have access to one based on their email domain. This will hopefully eliminate issues with users creating personal workspaces when they are meaning to create an account to join an existing workspace.

Screenshot 2023-06-06 at 9 24 52 AM

Local click test. Would be great if the reviewer could click test in a PR preview as well.

N/A


frontend/src/pages/Auth/JoinWorkspace.tsx
```diff
@@ -1,194 +1,147 @@
-import { ApolloError } from '@apollo/client'
```
ccschmitz, line 1: The logic in this file was moved to the `WorkspaceInvitation` component.

frontend/src/routers/AppRouter/AppRouter.tsx
```diff
@@ -72,10 +75,11 @@ export const AppRouter = () => {
 		useState<GetWorkspaceDropdownOptionsQuery>()
 	const [projectListData, setProjectListData] =
 		useState<GetProjectDropdownOptionsQuery>()
+	const isValidProjectId = Number.isInteger(Number(projectId))
 
 	const { data: projectDropdownData } = useGetProjectDropdownOptionsQuery({
 		variables: { project_id: projectId! },
-		skip: !isLoggedIn || !projectId,
+		skip: !isLoggedIn || !isValidProjectId,
```
ccschmitz, line 82: Fixes an issue where we would try to query on pages like `/about_you` where we didn't have a valid project ID.

Vadman97 (approved):
thanks for following up on this, chris!
just tested this with a new account and it worked correctly. firs time i did this i had some weird behavior showing me the workspace before i actually verified my email, but i think that's an unrelated frontend graphql cache issue
https://www.loom.com/share/2174e849c3464afe81ca49fbbd9f8dd6

Vadman97:
> > firs time i did this i had some weird behavior showing me the workspace before i actually verified my email, but i think that's an unrelated frontend graphql cache issue
> 
> @Vadman97 did you see the session viewer briefly? Are you thinking something like `GetAdmin` was cached and returned the cache value then was updated with the data it from the server which resulted in a redirect?

yup that's what i was thinking was happening @ccschmitz , though `GetAdmin` shouldn't be getting cached https://github.com/highlight/highlight/blob/c3534529c676ebbdb89fb5959932440ff03a0b1b/frontend/src/util/db.ts#L140

ccschmitz:
> yup that's what i was thinking was happening @ccschmitz , though `GetAdmin` shouldn't be getting cached, but maybe there is some other request that is that then causes a redirect after it updates

@Vadman97 I think I found the culprit of this and addressed it in https://github.com/highlight/highlight/pull/5555/commits/7cbb696ba2e989bc1617fb420c30e278e16b6373.

PR #5557, opened 2023-06-06
wire up backend to fetch error objects for an error group

et:


backend/private-graph/graph/schema.graphqls
```diff
@@ -577,20 +592,36 @@ type Log {
...
+type ErrorObjectNode {
+	id: ID!
+	createdAt: Timestamp!
+	event: String!
+	session: ErrorObjectNodeSession!
+}
+
+type ErrorObjectEdge implements Edge {
+	cursor: String!
+	node: ErrorObjectNode!
```
et, line 620: I would have liked to have used `ErrorObject` here but there seems to be something [broken](https://highlightcorp.slack.com/archives/C02CJANPHQS/p1686078353817089) with gqlgen.

backend/store/error_groups.go
```diff
@@ -3,13 +3,145 @@ package store
...
+
+	if err := query.Find(&errorObjects).Error; err != nil {
+		return privateModel.ErrorObjectConnection{}, err
+	}
+
+	if len(errorObjects) == 0 {
+		return privateModel.ErrorObjectConnection{}, nil
+	}
+
+	// Extract the non-null session IDs
```
et, line 53: Because some of the data is on the session (`session.user_properties` and `session.app_version`) (not the error_object), we load the session for each error_object.
Because the session_id is nullable, we have to do some manual work to support preloading.

et: To clarify why we need user_properties, the email in the listview is derived via this logic:

https://github.com/highlight/highlight/blob/1ef0610bc42906347fccb240100939455f3a9dbf/frontend/src/pages/ErrorsV2/ErrorInstance/ErrorInstance.tsx#L504

https://github.com/highlight/highlight/blob/1ef0610bc42906347fccb240100939455f3a9dbf/frontend/src/pages/Sessions/SessionsFeedV3/MinimalSessionCard/utils/utils.ts#L39-L55

meaning it won't always be email -- it's entirely dependent on how much we've identified on the session.

backend/store/error_groups_test.go
```diff
@@ -2,18 +2,179 @@ package store
...
+		connection, err = store.ListErrorObjects(errorGroup, ListErrorObjectsParams{})
+		assert.NoError(t, err)
+
+		assert.Len(t, connection.Edges, 1)
+
+		edge := connection.Edges[0]
+
+		assert.Equal(t, strconv.Itoa(errorObject.ID), edge.Cursor)
+		assert.Equal(t, errorObject.ID, edge.Node.ID)
+		assert.WithinDuration(t, errorObject.CreatedAt, edge.Node.CreatedAt, 10*time.Second)
```
et, line 57: Would have liked to have done a [direct equality check](https://github.com/highlight/highlight/pull/5557/commits/3cbc5cbc62d5057d1de5d394e0c09bbedd4f59d4) on the output but the time check results in a flakey test.

https://github.com/highlight/highlight/actions/runs/5203795531/jobs/9387189549?pr=5557

backend/store/error_groups.go
```diff
@@ -3,13 +3,139 @@ package store
...
+	var sessionIDs []int
+	for _, errorObject := range errorObjects {
+		if errorObject.SessionID != nil {
+			sessionIDs = append(sessionIDs, *errorObject.SessionID)
+		}
+	}
+
+	// Preload sessions for non-null session IDs
+	var sessions []model.Session
+	err := store.db.Where("id IN (?)", sessionIDs).Find(&sessions).Error
```
et, line 63: Loading all the sessions to avoid an N+1 query.

mayberryzane (approved):
cool! I like the more generic pagination concepts

PR #5558, opened 2023-06-06
fix slack channel noisy error

Vadman97:
## Summary

Fix nil pointer error from slack channel fetching from workspaces not integrated with slack.

Unit test.

No


et (approved):
Thanks!

PR #5575, opened 2023-06-07
Upgrade to Storybook 7 and explore new docs capabilities

ccschmitz:
## Summary

* Upgrades us to storybook 7
* Upgrades all storybook, vite, and vanilla-extract dependencies
* Fixes an issue we had in our build where we were getting an error about `TypeError: import_ws.WebSocketServer is not a constructor` (removed the `ws` dependency)
* Adds a script to only build the docs and not the stories (`build:docs`)
* Adds an example MDX documentation page for buttons
* Cleans up all the TS errors in the UI package - some were preexisting
* Updates out vitest `include` patterns because we started getting [an error](https://github.com/highlight/highlight/actions/runs/5216312898/jobs/9414908401?pr=5575) trying to run our setup files

Local click test.

N/A


packages/ui/src/components/Button/Button.stories.tsx
```diff
@@ -1,20 +1,18 @@
...
-} as ComponentMeta<typeof Button>
+} as Meta<typeof Button>
 
 export const ButtonVariants = () => {
-	const kind: Variants['kind'][] = ['primary', 'secondary']
-	const emphasis: Variants['emphasis'][] = ['high', 'medium', 'low']
-	const size: Variants['size'][] = ['xSmall', 'small', 'medium']
+	const kind: any[] = ['primary', 'secondary', 'asdf']
+	const emphasis: any[] = ['high', 'medium', 'low']
+	const size: any[] = ['medium', 'small', 'xSmall']
```
ccschmitz, line 15: I couldn't get these types to work anymore. Didn't seem worth spending too much time on, and felt like it's something that should be fixed in VE at some point.

cypress.config.js
```diff
@@ -4,7 +4,7 @@ module.exports = defineConfig({
 	e2e: {
 		baseUrl: 'https://localhost:3000',
 		pageLoadTimeout: 1200000,
-		video: true,
+		video: false,
```
ccschmitz, line 7: @Vadman97 have you ever used these? Didn't seem like they were used, so thought we could disable. Let me know what you think.

mayberryzane (approved):

PR #5587, opened 2023-06-08
move error body down to the instance level

et:
## Summary

This PR moves the error body (internally called `event`) down from the error group level to the error instance (internally called `error_object`) level. The reason for doing this is that the error group's body is set to whatever the latest error instance body is. However, error instances with different error bodies can be grouped together (e.g. same stacktrace but different error bodies due to using [variable interpolation](https://github.com/highlight/highlight/pull/5310)).

See [figma](https://www.figma.com/file/rdBPNzn7Klk6RG1LHUCE5B/Screen-Designs?type=design&node-id=8467-347211&t=iCRRrIXANlJHENtH-0).

* Find an error group with different error instance bodies.
* Traverse the error instances and observe the different error bodies.

Kapture 2023-06-08 at 09 57 35

Confirmed via [slack](https://highlightcorp.slack.com/archives/C0524MRA4NA/p1686239690884379?thread_ts=1685625555.000169&cid=C0524MRA4NA) that we don't want to change the main header when traversing error instances.

N/A

### DIFF
--- a/frontend/src/pages/ErrorsV2/ErrorBody/components/ErrorBodyText/index.tsx
+++ b/frontend/src/pages/ErrorsV2/ErrorBody/components/ErrorBodyText/index.tsx
@@ -1,5 +1,5 @@
 import { Button } from '@components/Button'
-import { ErrorGroup, Maybe } from '@graph/schemas'
+import { Maybe } from '@graph/schemas'
 import { Box, Text } from '@highlight-run/ui'
 import { getErrorBody } from '@util/errors/errorUtils'
 import { useEffect, useRef, useState } from 'react'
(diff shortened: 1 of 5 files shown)


frontend/src/pages/ErrorsV2/ErrorBody/components/ErrorBodyText/index.tsx
```diff
@@ -26,12 +26,7 @@ const ErrorBodyText = ({ errorGroup }: Props) => {
 
 	return (
 		<Box display="flex" flexDirection="column" gap="8">
-			<Box
-				py="8"
-				cssClass={style.errorBodyContainer}
-				overflowY="scroll"
```
et, line 32: Unrelated but there was a really ugly scrollbar showing:
Screenshot_2023-06-08_at_9_58_37_AM

This is unneeded since we truncate for long error bodies with a Show more/Show less button:

Kapture 2023-06-08 at 10 03 08

SpennyNDaJets (approved):

PR #5591, opened 2023-06-08
fix nil pointer references and remove noisy frontend errors

### LINKED ISSUE #5556: runtime error: invalid memory address or nil pointer dereference
workspace can be undefined

https://app.highlight.io/1/errors/MSQYy41pUCXpJREijV0zfWRmkAFv
```[tasklist]
### Tasks
- [ ] subscription details workspace
```

### LINKED ISSUE #5590: error getting linear teams: linear graphql API responded with error; status_code=400 Bad Request; body={"errors":[{"message":"Authentication required, not authenticated","extensions":{"code":"AUTHENTICATION_ERROR","type":"authentication error","userError":true,"userPresentableMessage":"You need to authenticate to access this operation.","meta":{}}}]}
api returns error when the integration is not configured

https://app.highlight.io/1/errors/ms7U1pwOsBoyV9SWmpnYIoHrtIQI

Vadman97:
## Summary

Fixes resolver logic that's causing nil pointers in our frontend.
Removes noisy frontend errors that are reported from the preload code.

Closes #5556
Closes #5590

New unit tests.
Testing the relevant graphql operations locally.

No

### DIFF
--- a/backend/private-graph/graph/resolver.go
+++ b/backend/private-graph/graph/resolver.go
@@ -474,9 +474,6 @@ func (r *Resolver) isAdminInProject(ctx context.Context, project_id int) (*model
 
 	span.SetTag("ProjectID", project_id)
 
-	if util.IsTestEnv() {
-		return nil, nil
-	}
 	if r.isWhitelistedAccount(ctx) {
 		project := &model.Project{}
 		if err := r.DB.Where(&model.Project{Model: model.Model{ID: project_id}}).First(&project).Error; err != nil {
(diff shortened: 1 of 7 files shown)


frontend/src/util/preload.ts
```diff
@@ -312,9 +312,7 @@ export const loadSession = async function (secureID: string) {
 			},
 		])
 	} catch (e: any) {
-		const msg = `failed to preload session ${secureID}`
```
et, line 315: Is there something else going on here? Like why are we trying to load a session that doesn't exist (or isn't available at the time)?

backend/private-graph/graph/schema.resolvers.go
```diff
@@ -6323,7 +6323,7 @@ func (r *queryResolver) GithubIssueLabels(ctx context.Context, workspaceID int,
 func (r *queryResolver) Project(ctx context.Context, id int) (*model.Project, error) {
 	project, err := r.isAdminInProjectOrDemoProject(ctx, id)
 	if err != nil {
-		return nil, nil
+		return nil, err
```
et, line 6326: Wonder if we can write a lint rule for this 🤔

et (approved):
Thanks for cleaning up the noise 🥇 
While looking through some [lint rule configs](https://golangci-lint.run/usage/linters/#errcheck), I did file #5593 for fixing another issue with our error handling.
If you get a moment, check out that page when you have a chance to see if there's anything else you think would help out here.

backend/private-graph/graph/resolver.go
```diff
@@ -474,9 +474,6 @@ func (r *Resolver) isAdminInProject(ctx context.Context, project_id int) (*model
 
 	span.SetTag("ProjectID", project_id)
 
-	if util.IsTestEnv() {
-		return nil, nil
-	}
```
Vadman97, line 479: this was to bypass the auth context, but i'm changing the tests to leverage the simple auth mode instead and not need this mock.

frontend/src/util/preload.ts
Vadman97: the session is available, but it might not have timeline indicators / network resources, so some of the other graphql calls here may fail.

PR #5592, opened 2023-06-08
ensure blank errors are checked

et:

### DIFF
--- a/backend/.golangci.yml
+++ b/backend/.golangci.yml
@@ -26,6 +26,9 @@ linters:
 
 # all available settings of specific linters
 linters-settings:
+    errcheck:
+        check-blank: true
+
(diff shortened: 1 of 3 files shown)


Vadman97:
added a `return nil, nil` linter in ac1d6e8

backend/.golangci.yml
```diff
@@ -26,6 +26,9 @@ linters:
 
 # all available settings of specific linters
 linters-settings:
+    errcheck:
+        check-blank: true
+
```
Vadman97, line 31: `check-type-assertions` is a useful one - i've bumped into that in the `otel.go` file (casting interfaces and not checking error)

Vadman97 (approved):

PR #5601, opened 2023-06-09
remove unneeded utils

et:
## Summary

We have some custom util code to get a pointer from a variable. While we use both [pointy](https://github.com/openlyinc/pointy) and [ptr](https://pkg.go.dev/github.com/aws/smithy-go@v1.13.5/ptr) in the codebase and probably need to pick one, we certainly don't need custom code.

Did a find and replace.
Build should pass.

N/A


SpennyNDaJets (approved):
LGTM

Issue #5606, opened 2023-06-09
Filtering chrome extension errors should apply to all browsers


et:
For example, it would be great to filter out firefox extensions which I believe look like this:

```
[{"functionName":null,"args":null,"fileName":"moz-extension://eaa42bd9-4641-4de4-b65e-784283b4fda1/js/inject.bundle.js","lineNumber":1,"columnNumber":2457,"isEval":null,"isNative":null,"source":"@moz-extension://eaa42bd9-4641-4de4-b65e-784283b4fda1/js/inject.bundle.js:1:2457"}]
```

error_group_secure_id: hiVVtcpIr8kAjSpXRugWUsCmdzoA

PR #5610, opened 2023-06-09
Filter chrome extension errors on the ingest side

et:
## Summary

This PR addresses the issue that filtering chrome extension errors had an incorrect implementation. Instead of filtering the frames from being visible (#4794), we need to drop the error objects on the ingest side. 

Unit test added pulling data from prod.

N/A


backend/errorgroups/filtering.go
```diff
@@ -0,0 +1,25 @@
...
+	privateModel "github.com/highlight-run/highlight/backend/private-graph/graph/model"
+	"github.com/samber/lo"
+)
+
+func IsErrorTraceFiltered(project model.Project, structuredStackTrace []*privateModel.ErrorTrace) bool {
+	if !*project.FilterChromeExtension {
+		return false
+	}
+
+	hasChromeExtensionFrame := lo.SomeBy(structuredStackTrace, func(frame *privateModel.ErrorTrace) bool {
```
et, line 16: Intentionally written like this to support additional filtering, e.g. #5606

backend/go.mod
```diff
@@ -73,7 +73,7 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/rs/cors v1.7.0
 	github.com/rs/xid v1.4.0
-	github.com/samber/lo v1.13.0
+	github.com/samber/lo v1.38.1
```
et, line 76: Updates `lo` so we can use [`SomeBy`](https://github.com/samber/lo#someby)

backend/util/environment.go
```diff
@@ -14,7 +14,7 @@ var (
 )
 
 func IsHubspotEnabled() bool {
-	return !IsDevEnv()
+	return !IsDevEnv() && !IsTestEnv()
```
et, line 17: Was getting a very tricky panic to debug: https://github.com/highlight/highlight/actions/runs/5225688025/jobs/9435696575

Turns out it was because Hubspot was getting called here:

https://github.com/highlight/highlight/blob/c7c5b2c63d51228b33c917c70b736b8d8bf7e3bd/backend/public-graph/graph/resolver.go#L1128-L1129

backend/public-graph/graph/resolver.go
```diff
@@ -553,7 +554,12 @@ func (r *Resolver) HandleErrorAndGroup(ctx context.Context, errorObj *model.Erro
 		return nil, errors.New("error object stacktrace was empty")
 	}
 
-	if projectID == 1 {
+	var project model.Project
+	if err := r.DB.Where("id = ?", projectID).First(&project).Error; err != nil {
```
et, line 558: We are now doing an extra lookup for the project which adds one query to this code path. However, the trade off does seem worth it as we are ensuring errors are only being processed for valid projects (i.e. garbage project ids can't be thrown in). See the tests that needed to be updated to support this.

backend/public-graph/graph/resolver.go
```diff
@@ -553,7 +554,12 @@ func (r *Resolver) HandleErrorAndGroup(ctx context.Context, errorObj *model.Erro
 		return nil, errors.New("error object stacktrace was empty")
 	}
 
-	if projectID == 1 {
+	project, err := r.Store.GetProject(projectID)
```
et, line 557: We are now doing an extra lookup for the project which adds one query to this code path. However, the trade off does seem worth it as we are ensuring errors are only being processed for valid projects (i.e. garbage project ids can't be thrown in). See the tests that needed to be updated to support this.

backend/store/projects.go
```diff
@@ -1,12 +1,10 @@
 package store
 
 import (
-	"context"
-
 	"github.com/highlight-run/highlight/backend/model"
 )
 
-func (store *Store) GetProject(ctx context.Context, id int) (model.Project, error) {
```
et, line 9: Didn't need context.

mayberryzane (approved):

PR #5627, opened 2023-06-13
Megalinter allowlist

et:


nvuillam:
Do you need any support ? :)

et:
> Do you need any support ? :)

Hey! Thanks for the help. This was for a hack week but I don't think we're going to pull this in. IIRC - I couldn't get the go linter to work with megalinter (it didn't detect the version of golangci-lint we were using) and we already have a linter working without it.

PR #5628, opened 2023-06-13
Start recording web socket events

SpennyNDaJets:
## Summary
Start recording web socket events and uploading them with the resources.

- Websocket recording will be on by default if `recordHeadersAndBody` is enabled
- It can be disabled by setting `disableWebSocketEventRecordings: true` in the `H.init()` options
- The WebSocket Proxy will be used even if recording is disabled since this needs to be set up before Highlight is initialized with the settings in order to catch earlier created WebSockets, but it will not record any events

Click test:
1. Find a live session, and view in Live Mode
- [ ] Confirm that a web socket was created in the Chrome Network Tab to the private graph
- [ ] Confirm that an `open`, `send`, and `received` event was sent as a resource in the `PushPayload` mutation
2. View the current session
- [ ] Confirm the WebSocket events are displayed in Highlight's Network Tab (may need to scroll to find)
- [ ] Confirm the WebSocket events are interspersed in the Network requests appropriately, sorted by time

https://www.loom.com/share/70ba511c01bc4850863070c70d0b78d2?sid=fbafcd12-c8fe-46db-9cd6-b74151414522

The npm package has not been updated, so this will not release a new build. The frontend should be updated to handle WebSocket events before the package is released


et:
@all-contributors please add @SpennyNDaJets for code

sdk/client/src/listeners/first-load-listeners.tsx
```diff
@@ -258,25 +276,31 @@ export class FirstLoadListeners {
...
 				)
 			}
 		}
-		return resources
+
+		if (!sThis.disableRecordingWebSocketContents) {
+			webSocketResources = sThis.webSocketNetworkContents
+		}
+
+		return [...httpResources, ...webSocketResources]
```
SpennyNDaJets, line 296: Assuming these do not need to be ordered by timestamp since we do this on the frontend, but should we do that here?

sdk/client/src/listeners/network-listener/utils/web-socket-listener.ts
```diff
@@ -0,0 +1,20 @@
...
+export type HighlightWebSocketWindow = Window & {
+	_highlightWebSocketCallback: WebSocketListenerCallback
+	WebSocket: any
+}
+
+declare var window: HighlightWebSocketWindow
+
+const WebSocketListener = (callback: WebSocketListenerCallback) => {
+	const initialHighlightWebSocketCallback = window._highlightWebSocketCallback
+	window._highlightWebSocketCallback = callback
```
SpennyNDaJets, line 13: Should we also block urls from the `urlBlocklist` here?

Follow up - what does blocking look like? Do we just not show the message, but that an event happen, or hide the event entirely?

sdk/client/src/listeners/network-listener/utils/models.ts
```diff
@@ -24,3 +24,13 @@ export interface RequestResponsePair {
 	/** Whether this URL matched a `urlToBlock` so the contents should not be recorded. */
 	urlBlocked: boolean
 }
+
+export interface WebSocketEvent {
+	socketId: string
+	initiatorType: 'WebSocket'
+	type: 'open' | 'error' | 'close' | 'received' | 'sent'
+	name: string
+	startTime: number
```
SpennyNDaJets, line 33: Called this `startTime` to avoid needing to patch the frontend to handle when sorting, but technically there is no endTime (i.e. this is a timeStamp)

backend/public-graph/graph/schema.graphqls
```diff
@@ -110,6 +110,7 @@ type Mutation {
 		events: ReplayEventsInput!
 		messages: String!
 		resources: String!
+		web_socket_events: String
```
SpennyNDaJets, line 113: Made this optional for backwards compatibility

Issue #5452, opened 2023-05-25
Make the date picker more usable


jay-khatri:
Make the date picker more usable:
https://www.notion.so/runhighlight/date-picker-audit-b4c5b85d347e46a593d7707825bf815a?pvs=4

```[tasklist]
### Tasks
- [ ] https://github.com/highlight/highlight/issues/5489
- [ ] https://github.com/highlight/highlight/issues/5490
- [ ] https://github.com/highlight/highlight/issues/5491
- [ ] https://github.com/highlight/highlight/issues/5492
```

jay-khatri:
@julian-highlight do you mind extracting these into issues?

julian-highlight:
Sure! Here's a tasklist. Adding a few comments into the issues! 

- [ ] Add disabled state for disabled days 
- [ ] Enable inputs
- [ ] Remove hover jiggle
- [ ] Use component styling

PR #5629, opened 2023-06-13
Fix/date picker input

betty-alagwu:
This PR is the first pass for issue #5452.

This PR allow users to type in the date in the `dateInput` and also select dates in the `datePicker`

Here's a video of how it works

https://github.com/highlight/highlight/assets/94234459/c92a1084-70c4-4a1d-97e2-77ef47a8411c


julian-highlight:
@betty-alagwu can we maybe try to make the individual dates use the respective sizes? It seems like the width is much higher than the height of the individual items. In the designs they just differ by 2px. This is a small detail, which isn't to dramatic, but I'd love to keep it close to the designs to ensure we don't have future frustration with sizes.

packages/ui/src/components/DatePicker/DateInput.tsx
```diff
@@ -0,0 +1,39 @@
...
+			name={name}
+			placeholder={placeholder}
+			type="input"
+			color={'n8'}
+			value={value}
+			style={{
+				boxSizing: 'border-box',
+				border: 'none',
+				background: 'none',
+				marginTop: '-5px',
```
Vadman97, line 34: is this required for the font changes?

Vadman97 (approved):
code for parsing date in multiple formats looks great. regarding @julian-highlight 's comment, could we update the heights of the date grid?

betty-alagwu:
Absolutely, will do that right away

packages/ui/src/components/DatePicker/DateInput.tsx
betty-alagwu: Not for the font change, It's required for the placement of the text and the borders of the container holding the text. I also made some changes to fix the fonts as well.
@Vadman97

PR #5639, opened 2023-06-13
Visual fixes to ErrorBody Page

jay-khatri:
### Remove bottom border on ErrorBody `Stat`
There's a weird bottom border on the error body `stat` section. It looks like this:
image

This change removes the bottom border on each `stat` component. It now looks like:
image

### Adds a border radius to the "Instance Error Body" section, see below:
image
previously, there was no radius.

## How did you test this change?

Visual test.

## Are there any deployment considerations?
no

PR #5641, opened 2023-06-14
Fix env script not updating config with trailing comma

lewisl9029:
## Summary

Fixes https://github.com/highlight/highlight/pull/5639#discussion_r1228794392, so you shouldn't need to manually run prettier-fix after the env script updates it. In fact it should no longer try to update it in the first place if it's identical (trailing comma included).

Ran the script locally and made sure trailing commas were added.


Vadman97 (approved):
nice! noticed that trailing comma issue today too, thanks for the quick fix

PR #5643, opened 2023-06-14
Using AI to provide insights for a session through rrweb events

eightants:
## Summary

AI experiment. Could we get interesting session insights by feeding in rrweb events into ChatGPT?

By asking ChatGPT to output insights and respective timestamps, you can skip to that area of the session by clicking on the insight. 

Function in packages/ai deployed onto AWS Lambda. For local development, run `yarn dev` in the /ai directory, the server is run on localhost:8765

https://www.loom.com/share/ddd0b0941bf541ae925f2e676950b91a

Just need to make sure that the API costs are considered


backend/lambda/lambda.go
```diff
@@ -60,3 +63,30 @@ func (s *Client) GetSessionScreenshot(ctx context.Context, projectID int, sessio
...
+
+	if util.IsDevEnv() {
+		b, _ = json.Marshal(&modelInputs.SessionQuery{
+			ID:        232563428,
+			ProjectID: 1,
+		})
+		req, _ = http.NewRequest(http.MethodPost, "https://ohw2ocqp0d.execute-api.us-east-2.amazonaws.com/default/ai-insights", bytes.NewBuffer(b))
+	} else {
+		req, _ = http.NewRequest(http.MethodPost, "https://ohw2ocqp0d.execute-api.us-east-2.amazonaws.com/default/ai-insights", bytes.NewBuffer(b))
+	}
```
Vadman97, line 83: pull the `http.NewRequest` out of the `if/else` since it is the same. just the `modelInputs.SessionQuery` needs to change in dev mode.

packages/ai/src/utils.ts
```diff
@@ -0,0 +1,53 @@
...
+	return parsedEvents
+}
+
+export const getInsightsForEvents = async (
+	openai: OpenAIApi,
+	events: any[],
+) => {
+	const parsedEvents = parseEventsForInput(events)
+	const completion = await openai.createChatCompletion({
+		model: 'gpt-3.5-turbo-16k',
```
Vadman97, line 40: cool that this is using the 16k-token model

packages/ai/README.md
```diff
@@ -0,0 +1 @@
+# ai
```
Vadman97, line 1: mind adding how one would test the lambda locally?

backend/lambda/lambda.go
eightants: Oh, good catch, it's meant to have the localhost url as the default in dev, but if that request fails it calls the lambda

Vadman97 (approved):
nice, thanks @eightants !

PR #5648, opened 2023-06-14
real time billing meters for sessions/errors

mayberryzane:
## Summary
- union the materialized view results with the count of sessions / errors after the materialized view was last updated (>= the latest date from the materialized view)
- added relevant indexes:
```
CREATE INDEX idx_error_objects_project_id_created_at ON public.error_objects USING btree (project_id, created_at)

CREATE INDEX idx_recent_sessions_billing ON public.sessions USING btree (project_id, created_at) WHERE ((excluded <> true) AND ((active_length >= 1000) OR ((active_length IS NULL) AND (length >= 1000))) AND (processed = true))
```
- slightly worse performance against the prod db (~.2s vs .06s for a larger project), difference will be more noticeable on ECS due to colocation so will monitor prod after deploying

- tested query in prod to make sure results were reasonable, click tested locally to make sure stuff wasn't breaking

- no

### DIFF
--- a/backend/pricing/pricing.go
+++ b/backend/pricing/pricing.go
@@ -2,6 +2,7 @@ package pricing
 
 import (
 	"context"
+	"database/sql"
 	"fmt"
 	"os"
 	"time"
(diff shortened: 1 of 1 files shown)


Vadman97 (approved):

Vadman97:
nice! with regards to degraded performance, this should be a constant degradation proportional to the interval from the last materialized view refresh, right? maybe we could add a trace to the sql queries to set up an alert in case the new query gets too slow?

mayberryzane:
> nice! with regards to degraded performance, this should be a constant degradation proportional to the interval from the last materialized view refresh, right? maybe we could add a trace to the sql queries to set up an alert in case the new query gets too slow?

yep, it's proportional to the number of sessions/errors in the project since the last refresh, will add tracing to both of these

PR #5651, opened 2023-06-15
Fix gem name usage in README.md

Syntaf:
Tiny update, just noticed this looking over the SDK implementation. `highlight_io` is the correct gem to mention here


Vadman97 (approved):

et:
@all-contributors please add @Syntaf for docs

Issue #5631, opened 2023-06-13
Wire up all-contributors


et:

et:
@all-contributors please add @jay-khatri for code and docs

et:
@all-contributors
please add @jay-khatri for code and docs
please add @Vadman97 for code and docs
please add @mayberryzane for code and docs
please add @et for code and docs
please add @ccschmitz for code and docs
please add @deltaepsilon for code and docs
please add @SpennyNDaJets for code and docs
please add @julian-highlight for design

et:
please add @Vadman97 for code and docs

et:
@all-contributors please add @Vadman97 for code and docs

PR #5652, opened 2023-06-15
Add more contributors

et:
## Summary

This updates our README with more contributors. Going forward, we should just use the bot to keep this in sync ([example](https://github.com/highlight/highlight/issues/5631#issuecomment-1590174525)) as the [CLI tool](https://allcontributors.org/docs/en/cli/overview) is kinda buggy.

Alternatively, if we find that this is too much of a hassle, we can use something like https://contrib.rocks. The downside of that approach is that it's reliant on git commits when we'd like to acknowledge other things (e.g. design).

[View Rendered](https://github.com/highlight/highlight/blob/3cc8e1e3a04a37ec5e0d2bbc2f6762a9cf412e94/README.md#contributors-)

N/A


SpennyNDaJets (approved):
LGTM

image

PR #5662, opened 2023-06-16
remove sessions outside billing quota from meter

mayberryzane:
## Summary
- billing meter was including sessions outside billing quota after the materialized view's end date, remove those
- add filter to the materialized view definition in code (was already included in prod)
- currently adding another index in prod to include `within_billing_quota`

- click tested locallly

- no

### DIFF
--- a/backend/model/model.go
+++ b/backend/model/model.go
@@ -1320,6 +1320,7 @@ func MigrateDB(ctx context.Context, DB *gorm.DB) (bool, error) {
 			SELECT project_id, DATE_TRUNC('day', created_at, 'UTC') as date, COUNT(*) as count
 			FROM sessions
 			WHERE excluded <> true
+			AND within_billing_quota
 			AND (active_length >= 1000 OR (active_length is null and length >= 1000))
 			AND processed = true
 			AND created_at > now() - interval '3 months'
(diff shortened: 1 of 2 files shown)


Vadman97 (approved):

backend/model/model.go
```diff
@@ -1320,6 +1320,7 @@ func MigrateDB(ctx context.Context, DB *gorm.DB) (bool, error) {
 			SELECT project_id, DATE_TRUNC('day', created_at, 'UTC') as date, COUNT(*) as count
 			FROM sessions
 			WHERE excluded <> true
+			AND within_billing_quota
```
SpennyNDaJets, line 1323: Just looking at line 1325, is this something we should explicitly check is `true`? Or is that basically what is happening by default?
(Assuming this is the boolean set on the `Session` table)

SpennyNDaJets (approved):
Makes sense to me! Just left one comment to double check

backend/model/model.go
mayberryzane: I don't think it makes a difference in the logic, the `excluded <> true` was probably written to include null results but I don't think it does, would probably be cleaner to write as `not excluded` everywhere

```
postgres> select * from (select true as excluded union select false union select null) a where a.excluded <> true
+----------+
| excluded |
|----------|
| False    |
+----------+
```

PR #5664, opened 2023-06-16
better error messaging for Vercel env var patch error

mayberryzane:
Screen Shot 2023-06-16 at 2 27 55 PM

## Summary
- patching a Vercel environment variable will fail if it's already been defined (since the Highlight integration does not own it)
-  show the user a message indicating which variable is causing the issue so they can fix it

- click tested locally

- no

### DIFF
--- a/backend/vercel/vercel.go
+++ b/backend/vercel/vercel.go
@@ -4,12 +4,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/go-chi/chi"
-	model2 "github.com/highlight-run/highlight/backend/model"
-	hlog "github.com/highlight/highlight/sdk/highlight-go/log"
-	highlightChi "github.com/highlight/highlight/sdk/highlight-go/middleware/chi"
-	"github.com/samber/lo"
-	log "github.com/sirupsen/logrus"
 	"io"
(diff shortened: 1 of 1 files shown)


backend/vercel/vercel.go
```diff
@@ -121,6 +122,9 @@ func SetEnvVariable(projectId string, apiKey string, accessToken string, teamId
 	b, err := io.ReadAll(res.Body)
 
 	if res.StatusCode != 200 {
+		if method == "PATCH" {
+			return errors.New(fmt.Sprintf("Could not patch environment variable %s - If this environment variable is already defined in Vercel, please remove it from your Vercel project and retry.", key))
```
Vadman97, line 126: is there a way our api could remove the env var for them?

Vadman97 (approved):

backend/vercel/vercel.go
mayberryzane: can't remove it or edit it if it was created manually, we only have permission to update env vars that we create. I think we added some instructions to our Next.js docs to add a `NEXT_PUBLIC_HIGHLIGHT_PROJECT_ID` env var, and if the users do this before installing the integration they'll get this error

PR #5673, opened 2023-06-16
add an RFC on testing

Vadman97:
## Summary

Starts to define testing strategies at highlight.io.
Adds code coverage reports for our Go SDK and Go backend.

CI

No

### DIFF
--- a/internal-docs/rfcs/003-testing.md
+++ b/internal-docs/rfcs/003-testing.md
@@ -1,0 +1,51 @@
+# RFC: Testing Strategy
+
+* Status: In Progress
+* Author: [@Vadman97](https://github.com/Vadman97)
+
+## Summary
+
+Testing software at highlight means validating that features 
+used by customers meet and exceed their expectations.
+These feature span our SDKs, the app.highlight.io frontend, 
+and the backend and associated serverless functions. 
+
+When we build features or make changes, we have a few ways to
+validate that our implementation matches desired behavior. 
+This RFC discusses what we do today and what we could do in the future 
+to continue to impress our customers.
+
+Because of the early nature of our company, **we prioritize shipping features quickly**
+and iterating with enhancements, rather than guaranteeing correctness of every code change. 
+However, while quick iteration works well for new product development, it is often a challenge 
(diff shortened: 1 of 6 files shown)


internal-docs/rfcs/003-testing.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Proposed Implementation
+
+We are introducing automated test coverage for our primary backend languages (Golang and others used by SDKs).
+Test coverage will be tracked incrementally on each Pull Request to help identify new features that may
+be introducing gaps in coverage.
+
+## Success Metrics
+
+We would like to achieve 50%+ test coverage of our backend code and 80%+ test coverage of our SDK code.
```
et, line 47: Any thoughts on a goal date of when we should hit this goal?

internal-docs/rfcs/003-testing.md
```diff
@@ -0,0 +1,62 @@
...
+product. At the same time, the code path is difficult to test manually because there are a number
+of nested conditions that are hard to hit during a short manual session recording. Moreover, the
+code is complex and is infrequently changed, so engineers do not have the full context on the
+associated code when jumping in to make changes.
+
+## Proposed Implementation
+
+We are introducing automated test coverage for our primary backend languages (Golang and others used by SDKs).
+Test coverage will be tracked incrementally on each Pull Request to help identify new features that may
+be introducing gaps in coverage.
```
et, line 43: How are gaps actually identified? For example: if I add something to public/private graph, how would I see if this lowered the coverage?

et (approved):
Nice work here. The PR description mentions Go SDK coverage existing. Can you point me to where that is?

Vadman97:
> Nice work here. The PR description mentions Go SDK coverage existing. Can you point me to where that is?

it's not that extensive, but here's an sdk test 
https://github.com/highlight/highlight/blob/main/sdk/highlight-go/highlight_test.go

et:
> is the coverage missing from the ci report? there might be a collision between the two go coverage actions overwriting each other

I was looking at the comment above that had coverage percentages for each of our backend packages and was hoping to see something similar for the go SDK.

PR #5680, opened 2023-06-20
WebSocket client side recording

SpennyNDaJets:
## Summary
Start recording web socket events and pushing them to the backend via PushPayload. Note that they are currently being accepted by the backend, but dropped (i.e. not stored in Redis or Postgres)

- Websocket recording will be on by default if `recordHeadersAndBody` is enabled
- It can be disabled by setting `disableWebSocketEventRecordings: true` in the `H.init()` options
- The WebSocket Proxy will be used even if recording is disabled since this needs to be set up before Highlight is initialized with the settings in order to catch earlier created WebSockets, but it will not record any events

Click test:
1. Find a live session, and view in Live Mode
- [ ] Confirm that a web socket was created in the Chrome Network Tab to the private graph
- [ ] Confirm that an `open`, `send`, and `received` event was sent as a resource in the `PushPayload` mutation
2. View the current session
- [ ] Confirm the WebSocket events are displayed in Highlight's Network Tab (may need to scroll to find)
- [ ] Confirm the WebSocket events are interspersed in the Network requests appropriately, sorted by time

https://www.loom.com/share/70ba511c01bc4850863070c70d0b78d2?sid=fbafcd12-c8fe-46db-9cd6-b74151414522

The npm package has not been updated, so this will not release a new build. The frontend should be updated to handle WebSocket events before the package is released

### DIFF
--- a/.github/workflows/turbo.yml
+++ b/.github/workflows/turbo.yml
@@ -56,20 +56,20 @@ jobs:
             - name: Check formatting
               run: yarn format-check
 
+            - name: Configure AWS credentials
+              if: github.event.pull_request.head.repo.full_name == 'highlight/highlight' || github.ref == 'refs/heads/main'
+              uses: aws-actions/configure-aws-credentials@v2
+              with:
+                  aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+                  aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+                  aws-region: us-east-2
+
             - name: Build & test
               run: yarn test:all
               env:
                   DOPPLER_TOKEN: ${{ secrets.DOPPLER_PROD_RENDER_SECRET }}
                   GRAPHCMS_TOKEN: ${{ secrets.GRAPHCMS_TOKEN }}
                   NEXT_PUBLIC_HIGHLIGHT_PROJECT_ID: 1jdkoe52
-
-            - name: Configure AWS credentials
-              if: steps.filter.outputs.npm-changed == 'true' && (github.event.pull_request.head.repo.full_name == 'highlight/highlight' || github.ref == 'refs/heads/main')
-              uses: aws-actions/configure-aws-credentials@v2
-              with:
-                  aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
-                  aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-                  aws-region: us-east-2
 
             - name: Validate project is ready to be published
               if: steps.filter.outputs.npm-changed == 'true' && (github.event.pull_request.head.repo.full_name == 'highlight/highlight' || github.ref == 'refs/heads/main')
--- a/e2e/nextjs/src/app/components/highlight-buttons.tsx
+++ b/e2e/nextjs/src/app/components/highlight-buttons.tsx
@@ -1,0 +1,32 @@
+'use client'
+
+import { Button } from '@/app/components/button'
+import { H } from 'highlight.run'
+
+export function HighlightButtons() {
+	return (
+		<div
+			style={{
+				display: 'grid',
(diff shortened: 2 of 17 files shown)


sdk/client/src/listeners/first-load-listeners.tsx
```diff
@@ -258,25 +282,47 @@ export class FirstLoadListeners {
...
 				)
 			}
 		}
-		return resources
+
+		if (!sThis.disableRecordingWebSocketContents) {
+			webSocketResources = sThis.webSocketNetworkContents
+		}
+
+		return [...httpResources, ...webSocketResources]
```
SpennyNDaJets, line 302: Assuming these do not need to be ordered by timestamp since we do this on the frontend, but should we do that here?

sdk/client/src/listeners/network-listener/utils/web-socket-listener.ts
```diff
@@ -0,0 +1,41 @@
...
+	/**
+	 * Callback for web socket message and error events that are displayed under the websocket requests
+	 */
+	_highlightWebSocketEventCallback: WebSocketEventListenerCallback
+	WebSocket: any
+}
+
+declare var window: HighlightWebSocketWindow
+
+const WebSocketListener = (
```
SpennyNDaJets, line 21: Should we also block urls from the urlBlocklist here?

Follow up - what does blocking look like? Do we just not show the message, but that an event happen, or hide the event entirely?

sdk/client/src/listeners/network-listener/utils/models.ts
```diff
@@ -24,3 +24,21 @@ export interface RequestResponsePair {
 	/** Whether this URL matched a `urlToBlock` so the contents should not be recorded. */
 	urlBlocked: boolean
 }
+
+export interface WebSocketRequest {
+	socketId: string
+	initiatorType: 'WebSocket'
+	type: 'open' | 'close'
+	name: string
+	startTime?: number
```
SpennyNDaJets, line 33: Called this startTime to avoid needing to patch the frontend to handle when sorting, but technically there is no endTime (i.e. this is a timeStamp)

sdk/firstload/src/listeners/web-socket/index.ts
```diff
@@ -0,0 +1,125 @@
+import { createNetworkRequestId } from '@highlight-run/client/src/listeners/network-listener/utils/utils'
+import type { HighlightWebSocketWindow } from '@highlight-run/client/src/listeners/network-listener/utils/web-socket-listener'
+
+declare var window: HighlightWebSocketWindow
+
+const placeholderCallback = () => null
+
+export const initializeWebSocketListener = () => {
```
Vadman97, line 8: as we learned with the next.js setup, we'll want to do something like the following to make sure the monkeypatch can't happen more than once

https://github.com/highlight/highlight/blob/9e898c677527fce49a0f4d49bbe18bc4383bf75c/sdk/firstload/src/listeners/fetch/index.ts#L12-L15

sdk/client/src/listeners/first-load-listeners.tsx
Vadman97, line 302: is it ok that the websocket resources are mixed with the http resources here? is that because this isn't for the messages themselves, but the overall websocket connection resources?

Vadman97 (approved):
looks good overall! great that we are shipping this incrementally.

sdk/client/src/listeners/network-listener/utils/models.ts
SpennyNDaJets: Yes, but this would be represented as `endTime` the `close` event that would be sent to the network resources

sdk/client/src/listeners/first-load-listeners.tsx
SpennyNDaJets: Exactly, so these are just the higher level connection events that would be displayed intertwined in the network resources.

PR #5685, opened 2023-06-20
avoid monkeypatching window.fetch more than once

### LINKED ISSUE #5619: using `HighlightInit` with `H.identify` will cause recursive network monkeypatch
setting up `<HighlightInit/>` in next.js and `highlight.run` for `H.identify` will set up double network listeners that will collide and cause a recursive error. in `highlight.run`, we should only set up network listeners if they have not been set up.

image

https://discord.com/channels/1026884757667188757/1116676770813464596

### LINKED ISSUE #5676: Page freezing with `H.init` in NextJS
**Describe the bug**
> Having an issue getting the H.identify() method to work on NextJS, despite following the directions in the user manual. Do you have any tips? The page just freezes when highlight makes the identify call.

**To Reproduce**
TBD

Vadman97:
## Summary

When the highlight firstload bundle is imported once from more than one place, we setup
multiple fetch monkeypatches which causes a recursive loop in `window.fetch`.
This happens from the next.js SDK because the firstload highlight.run package imported
from the `@highlight-run/next` SDK differs from the `highlight.run` package imported on the frontend.

Closes #5676, closes #5619.

While @deltaepsilon and I have not been able yet to reproduce the behavior in the issues mentioned above,
I've seen the same behavior when importing the highlight.run bundle more than once into a browser.

Testing that workflow which I'm suspecting is the root cause of the issue.

Before the change: https://www.loom.com/share/c2d48a1a18ad4786a0ba98bee7e502c6
After the change: https://www.loom.com/share/f198156569a840ebb0e076de458cbc26

Meanwhile, trying to reproduce what the customer may have seen with next.js.

New package versions published for firstload / next / node SDKs


deltaepsilon (approved):
Looks great!

SpennyNDaJets (approved):
Looks good to me!

I applied the changes here to https://github.com/highlight/highlight/pull/5680, but will hold off on merging the WebSocket PR so we can roll this out safely

PR #5707, opened 2023-06-22
Remove references to alert_events

SpennyNDaJets:
## Summary
We are not respecting the frequency setting set on alerts. This is due to a reference to a table that is no longer used, `AlertEvents`. Update the references to this table to use the respective table based on the type of event.

Click test:
1) Create a new error alert
  a) set it to send to #chris-errors-test
  b) Set the Alert Frequency to 5 minutes
2) Navigate to the buttons page (`/1/buttons`)
3) Clock on "Throw an Error" button
- [ ] Error is sent to slack channel
4) Click on "Throw an Error" again
- [ ] Error is not sent to slack channel
5) Click on "Throw a randomized Error"
- [ ] Error is sent to slack channel
6) Wait 5 minutes
7) Click on "Throw an Error" again
- [ ] Error is sent to slack channel

Screenshot 2023-06-22 at 1 18 41 PM

Screenshot 2023-06-22 at 5 01 12 PM

N/A

### DIFF
--- a/backend/model/model.go
+++ b/backend/model/model.go
@@ -2033,21 +2033,25 @@ func (obj *Alert) GetDailySessionEventFrequency(db *gorm.DB, id int) ([]*int64, 
 
 func (obj *Alert) GetDailyLogEventFrequency(db *gorm.DB, id int) ([]*int64, error) {
 	var dailyAlerts []*int64
-	// TODO(spenny): figure out where to filter by project_id
 	if err := db.Raw(`
 		SELECT COUNT(e.id)
 		FROM (
 			SELECT to_char(date_trunc('day', (current_date - offs)), 'YYYY-MM-DD') AS date
 			FROM generate_series(0, 6, 1)
 			AS offs
-		) d LEFT OUTER JOIN
-		log_alert_events e
-		ON d.date = to_char(date_trunc('day', e.created_at), 'YYYY-MM-DD')
-			AND e.log_alert_id=?
+		) d
+		LEFT OUTER JOIN (
+			log_alert_events e
+			INNER JOIN log_alerts la
+			ON la.id = e.log_alert_id
+				AND e.log_alert_id=?
+				AND la.project_id=?
+		)
+		ON d.date = to_char(date_trunc('day', e.sent_at), 'YYYY-MM-DD')
 		GROUP BY d.date
 		ORDER BY d.date;
-	`, id).Scan(&dailyAlerts).Error; err != nil {
-		return nil, e.Wrap(err, "error querying daily alert frequency")
+	`, id, obj.ProjectID).Scan(&dailyAlerts).Error; err != nil {
+		return nil, e.Wrap(err, "error querying daily log event frequency")
 	}
 
 	return dailyAlerts, nil
--- a/backend/model/model.go
+++ b/backend/model/model.go
@@ -1979,7 +1979,7 @@ func (obj *Alert) GetEmailsToNotify() ([]*string, error) {
 	return emailsToNotify, err
 }
 
-func (obj *Alert) GetDailyFrequency(db *gorm.DB, id int) ([]*int64, error) {
+func (obj *Alert) GetDailyErrorEventFrequency(db *gorm.DB, id int) ([]*int64, error) {
 	var dailyAlerts []*int64
 	if err := db.Raw(`
 		SELECT COUNT(e.id)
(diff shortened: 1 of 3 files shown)


backend/public-graph/graph/resolver.go
```diff
@@ -1628,15 +1628,16 @@ func (r *Resolver) sendErrorAlert(ctx context.Context, projectID int, sessionObj
...
+				ON obj.id = ev.error_object_id
 				WHERE
-					project_id=?
-					AND type=?
-					AND (error_group_id IS NOT NULL
-						AND error_group_id=?)
-					AND alert_id=?
-					AND created_at > NOW() - ? * (INTERVAL '1 SECOND')
-			`, projectID, model.AlertType.ERROR, group.ID, errorAlert.ID, errorAlert.Frequency).Scan(&numAlerts).Error; err != nil {
+					obj.project_id=?
```
SpennyNDaJets, line 1635: After repeating this some time, I'm curious if we can do one of the following:
1. Remove the need to filter by `project_id` in these queries - seems redundant, but let me know if there is a security concern
2. If not, add project_id to the different types of alert event tables like it was on the `AlertEvent` table

Either one would save a join for _LogAlertEvents_ and _SessionAlertEvents_ but still would still need to join for _ErrorAlertEvents_ in order to filter by `error_group_id`

backend/model/model.go
```diff
@@ -1979,24 +1979,79 @@ func (obj *Alert) GetEmailsToNotify() ([]*string, error) {
 	return emailsToNotify, err
 }
 
-func (obj *Alert) GetDailyFrequency(db *gorm.DB, id int) ([]*int64, error) {
+func (obj *Alert) GetDailyErrorEventFrequency(db *gorm.DB, id int) ([]*int64, error) {
```
SpennyNDaJets, line 1982: Want to double check there is no performance concerns with adding these joins

Vadman97: found a large error alert id 533 by running
```select error_alert_id, count(*)
from error_alert_events
group by error_alert_id having count(*) > 10
order by 2 desc limit 10;```

the query takes a while to run the first time (60 seconds)
but it is fast subsequently (less than 1 second) probably because this data isn't often queries but can be cached

```explain analyze
SELECT COUNT(e.id)
FROM (SELECT to_char(date_trunc('day', (current_date - offs)), 'YYYY-MM-DD') AS date
      FROM generate_series(0, 6, 1)
               AS offs) d
         LEFT OUTER JOIN (
    error_alert_events e
        INNER JOIN error_objects obj
    ON obj.id = e.error_object_id
        AND e.error_alert_id=533
        AND obj.project_id=649
    )
                         ON d.date = to_char(date_trunc('day', e.sent_at), 'YYYY-MM-DD')
GROUP BY d.date
ORDER BY d.date;
```

```
GroupAggregate  (cost=747170.27..747183.91 rows=7 width=40) (actual time=68002.612..68002.696 rows=7 loops=1)
"  Group Key: (to_char(date_trunc('day'::text, ((CURRENT_DATE - offs.offs))::timestamp with time zone), 'YYYY-MM-DD'::text))"
  ->  Sort  (cost=747170.27..747174.77 rows=1798 width=40) (actual time=68002.605..68002.686 rows=7 loops=1)
"        Sort Key: (to_char(date_trunc('day'::text, ((CURRENT_DATE - offs.offs))::timestamp with time zone), 'YYYY-MM-DD'::text))"
        Sort Method: quicksort  Memory: 25kB
        ->  Hash Right Join  (cost=1000.73..747073.07 rows=1798 width=40) (actual time=68002.583..68002.673 rows=7 loops=1)
"              Hash Cond: (to_char(date_trunc('day'::text, e.sent_at), 'YYYY-MM-DD'::text) = to_char(date_trunc('day'::text, ((CURRENT_DATE - offs.offs))::timestamp with time zone), 'YYYY-MM-DD'::text))"
              ->  Gather  (cost=1000.57..746390.47 rows=51359 width=16) (actual time=68002.545..68002.624 rows=0 loops=1)
                    Workers Planned: 2
                    Workers Launched: 2
                    ->  Nested Loop  (cost=0.57..740254.57 rows=21400 width=16) (actual time=65607.344..65607.345 rows=0 loops=3)
                          ->  Parallel Seq Scan on error_alert_events e  (cost=0.00..5999.23 rows=93178 width=24) (actual time=0.215..73.642 rows=75083 loops=3)
                                Filter: (error_alert_id = 533)
                                Rows Removed by Filter: 84154
                          ->  Index Scan using error_objects_pkey on error_objects obj  (cost=0.57..7.88 rows=1 width=4) (actual time=0.871..0.871 rows=0 loops=225249)
                                Index Cond: (id = e.error_object_id)
                                Filter: (project_id = 1)
                                Rows Removed by Filter: 1
              ->  Hash  (cost=0.07..0.07 rows=7 width=4) (actual time=0.023..0.024 rows=7 loops=1)
                    Buckets: 1024  Batches: 1  Memory Usage: 9kB
                    ->  Function Scan on generate_series offs  (cost=0.00..0.07 rows=7 width=4) (actual time=0.004..0.005 rows=7 loops=1)
Planning Time: 0.314 ms
Execution Time: 68002.756 ms
```

slow part is
```
                          ->  Parallel Seq Scan on error_alert_events e  (cost=0.00..5999.23 rows=93178 width=24) (actual time=0.215..73.642 rows=75083 loops=3)
                                Filter: (error_alert_id = 533)
                                Rows Removed by Filter: 84154
```

so just adding a new index on `error_alert_events` on `error_alert_id` should help keep this query fast

```
create index concurrently error_alert_events_error_alert_id_index
    on error_alert_events (error_alert_id);

-- probably on the other tables as well

create index concurrently session_alert_events_session_alert_id_index
    on session_alert_events (session_alert_id);

create index concurrently log_alert_events_log_alert_id_index
    on log_alert_events (log_alert_id);
```

@mayberryzane , do we have a best practice for adding indexes on existing table? should we just run this one-off on the prod db? or should we also add the necessary gorm tag to create the index in dev?

backend/public-graph/graph/resolver.go
Vadman97: don't think that the project_id is needed since the alert id should always be unique.

Vadman97 (approved):
nice, thanks for cleaning this up!

backend/model/model.go
SpennyNDaJets: I may be looking at it wrong, but looks like we have indices created for each of these?
https://github.com/highlight/highlight/blob/main/backend/model/model.go#L1707

I wonder if it something with the query that was causing the sql plan to not use the index. I also updated the query to avoid the joins tables, so maybe this is looking better?
```
explain analyze
SELECT COUNT(e.id)
FROM (
SELECT to_char(date_trunc('day', (current_date - offs)), 'YYYY-MM-DD') AS date
FROM generate_series(0, 6, 1) AS offs
) d
LEFT OUTER JOIN error_alert_events e 
ON d.date = to_char(date_trunc('day', e.sent_at), 'YYYY-MM-DD')
AND e.error_alert_id=533
GROUP BY d.date
ORDER BY d.date;
```

Side note: I tried to run `./db-prod-access.sh` but was receiving a doppler error, and may need additional access?
```
Doppler Error: Could not find requested config 'prod_aws'
```

mayberryzane: confirming we already have an index on `error_alert_events.error_alert_id` (because it's the first column in the list, the query planner can use it for this query)
```
select * from pg_indexes where indexname = 'idx_error_alert_event'
-[ RECORD 1 ]-------------------------
schemaname | public
tablename  | error_alert_events
indexname  | idx_error_alert_event
tablespace | <null>
indexdef   | CREATE INDEX idx_error_alert_event ON public.error_alert_events USING btree (error_alert_id, error_object_id)
```
Looks like the query planner behavior is different for different ids, e.g. 
`postgres> explain select count(*) from error_alert_events where error_alert_id = 1`
```
-[ RECORD 1 ]-------------------------
QUERY PLAN | Aggregate  (cost=17.57..17.58 rows=1 width=8)
-[ RECORD 2 ]-------------------------
QUERY PLAN |   ->  Index Only Scan using idx_error_alert_event on error_alert_events  (cost=0.42..17.42 rows=60 width=0)
-[ RECORD 3 ]-------------------------
QUERY PLAN |         Index Cond: (error_alert_id = 1)
```
`postgres> explain select count(*) from error_alert_events where error_alert_id = 533`
```
-[ RECORD 1 ]-------------------------
QUERY PLAN | Finalize Aggregate  (cost=7248.19..7248.20 rows=1 width=8)
-[ RECORD 2 ]-------------------------
QUERY PLAN |   ->  Gather  (cost=7247.98..7248.19 rows=2 width=8)
-[ RECORD 3 ]-------------------------
QUERY PLAN |         Workers Planned: 2
-[ RECORD 4 ]-------------------------
QUERY PLAN |         ->  Partial Aggregate  (cost=6247.98..6247.99 rows=1 width=8)
-[ RECORD 5 ]-------------------------
QUERY PLAN |               ->  Parallel Seq Scan on error_alert_events  (cost=0.00..6014.45 rows=93410 width=0)
-[ RECORD 6 ]-------------------------
QUERY PLAN |                     Filter: (error_alert_id = 533)
```
Presumably because postgres knows that error_alert_id 533 accounts for about half of the table. On manually adding an index on just `error_alert_id`
`CREATE INDEX concurrently idx_error_alert_events_error_alert_id ON public.error_alert_events USING btree (error_alert_id)` it uses the new index

SpennyNDaJets: Thanks for taking a look! Makes sense that the index may not be used for large percentage of rows

Issue #5164, opened 2023-04-27
Default filter selected on session page does not work


eightants:
**Describe the bug**
When loading in a saved filter (likely from the last selected filter on a previous session, the filter name is populated, but the filters in the actual filter field are not applied. (As you can see, only Status is Completed is selected. ). 

Clicking on the list of filters and selecting the same saved filter again, or another saved filter, applies the filters properly. 

<img width="903" alt="Screenshot 2023-04-27 at 4 51 00 PM" src="https://user-images.githubusercontent.com/29858539/234987945-90df7713-697b-475e-87b9-41dc3b3fe3ae.png">

When switching between workspaces or projects, the name of the last set filter is applied as a default. However, this filter might not exist in the project itself, as seen below. 

<img width="903" alt="Screenshot 2023-04-27 at 4 51 16 PM" src="https://user-images.githubusercontent.com/29858539/234987996-b3a62106-2d3b-4207-8e6f-3b23111b952a.png">

**To Reproduce**

1. Go to Sessions
2. Select a saved filter
3. Close the page
4. Open a new tab and visit Sessions again

For the second part of the bug
1. Go to Sessions
2. Select a saved filter
3. Switch to a different project or workspace
4. Filter is selected, but doesn't exist in project

**Expected behavior**
Filter is applied as expected on load if a filter was saved from last session, with all the filters from the saved filter. Cached filters in localstorage should be tied to a project, so that filters that don't exist aren't loaded.

jay-khatri:
Thanks @eightants. @mayberryzane to do you mind bucketing w/ your search filter changes?

PR #5712, opened 2023-06-22
refactor QueryBuilder to use a reducer pattern

### LINKED ISSUE #5505: Active segment requires multiple clicks
In some cases, when the user goes back to an active segment, it requires that they "re-select" that segment to get it rendered. Can we try and reproduce this?

https://www.loom.com/share/ade08f9d139b4c8a94606d864989a49f

mayberryzane:
## Summary
- refactoring this to use a reducer pattern to make it easier to reason about state transitions in the query builder
- UI is a bit less janky because multiple states are updated at the same time in the reducer rather than waiting on a chain of useEffects to run, each once per render
- fixes #5505 and #5164 where the selected segment was not reset on page change

- click tested locally

- no


frontend/src/components/CommandBar/utils.ts
```diff
@@ -79,32 +85,35 @@ export const useAttributeSearch = (form: FormState<CommandBarSearch>) => {
...
+		const basePath = `/${projectId}/${isError ? 'errors' : 'sessions'}`
 		const qbParams = buildQueryBuilderParams({
 			attribute,
 			query,
 			startDate: params?.withDate ? dates[0] : undefined,
 			endDate: params?.withDate ? dates[1] : undefined,
 		})
 
-		const searchQuery = buildQueryURLString(qbParams, {
-			reload: true,
```
mayberryzane, line 99: removed the `?reload` query parameter and corresponding `useEffect`. Setting the search query before navigating has the same effect

frontend/src/components/SearchPagination/SearchPagination.tsx
```diff
@@ -97,29 +96,6 @@ const SearchPagination = ({
 		return [START_PAGE, ExpandAction.Back, ...rightRange]
 	}, [$siblingCount, currentPage, sideItemCount, pageCount])
 
-	const [paginationToUrlParams, setPaginationToUrlParams] = useQueryParams({
```
mayberryzane, line 100: logic for setting the page state / url param moved to `SearchState.ts`

frontend/src/context/BaseSearchContext.tsx
```diff
@@ -19,33 +21,23 @@ export function normalizeParams<T>(params: object) {
 	)
 }
 
-export type BaseSearchContext<T> = {
```
mayberryzane, line 22: removed a few of the setters that were updating internal state - e.g. `backendSearchQuery` is now updated whenever `setSearchQuery` is called to avoid them being out of sync

frontend/src/components/QueryBuilder/QueryBuilder.tsx
```diff
@@ -1427,315 +1353,60 @@ function QueryBuilder<T extends SearchParamsInput | ErrorSearchParamsInput>(
 	const getDefaultOperator = (field: SelectOption | undefined) =>
 		((field && getCustomFieldOptions(field)?.operators) ?? OPERATORS)[0]
 
-	const parseInner = useCallback(
```
mayberryzane, line 1430: moved a lot of the logic for creating the `backendSearchQuery` to `SearchState.ts`

frontend/src/components/QueryBuilder/QueryBuilder.tsx
```diff
@@ -1355,52 +1306,27 @@ function QueryBuilder<T extends SearchParamsInput | ErrorSearchParamsInput>(
...
-				if (match !== undefined) {
-					const segmentParameters = normalizeParams(
-						gqlSanitize(match?.params),
+				if (match && match.params?.query) {
+					setSelectedSegment(
+						{ id: segment.id, name: segment.name },
+						match.params?.query,
 					)
-					// @ts-expect-error
-					setSearchParams(segmentParameters)
```
mayberryzane, line 1380: multiple states updated at once now handled in the reducer

frontend/src/components/QueryBuilder/QueryBuilder.tsx
```diff
@@ -1427,315 +1353,60 @@ function QueryBuilder<T extends SearchParamsInput | ErrorSearchParamsInput>(
...
+	if (startingDateRange) {
+		const [fromStr, toStr] = startingDateRange.split('_')
+		from = new Date(fromStr)
+		to = new Date(toStr)
+	}
 	const [dateRange, setDateRange] = useState<Date[]>([
-		defaultPreset.startDate, // Start at 30days
-		new Date(now.toISOString()),
+		from ?? defaultPreset.startDate, // Start at 30days
+		to ?? new Date(getNow().toISOString()),
```
mayberryzane, line 1384: `getNow` is a little hacky, but we were using a slightly different `now` date in a few places which broke logic comparing `now` with the relative dates that required them to be exactly the same

frontend/src/components/QueryBuilder/QueryBuilder.tsx
```diff
@@ -1427,315 +1353,60 @@ function QueryBuilder<T extends SearchParamsInput | ErrorSearchParamsInput>(
...
 
 	const [currentRule, setCurrentRule] = useState<RuleProps | undefined>()
+
+	const {
+		isAnd: serializedIsAnd,
+		rules: serializedRules,
+	}: { isAnd: boolean; rules: any } = JSON.parse(searchQuery)
+	const startingRules = deserializeRules(serializedRules)
+	const [isAnd, toggleIsAnd] = useToggle(serializedIsAnd)
+	const [rules, setRules] = useState<RuleProps[]>(startingRules)
```
mayberryzane, line 1369: set the default rules / date from the search context instead of waiting on a use effect / second render

frontend/src/components/QueryBuilder/QueryBuilder.tsx
```diff
@@ -1427,315 +1353,60 @@ function QueryBuilder<T extends SearchParamsInput | ErrorSearchParamsInput>(
...
+
+			if (readonly || !newRules.every(isComplete)) {
+				return
+			}
+
+			const newState = JSON.stringify({
+				isAnd,
+				rules: serializeRules(newRules),
+			})
+			setSearchQuery(newState)
```
mayberryzane, line 1405: `setSearchQuery` is called here to remove a useEffect

frontend/src/components/QueryBuilder/QueryBuilder.tsx
```diff
@@ -1746,35 +1417,34 @@ function QueryBuilder<T extends SearchParamsInput | ErrorSearchParamsInput>(
...
 
 	const timeRangeRule = useMemo<RuleProps>(() => {
 		const timeRange = rules.find(
 			(rule) => rule.field?.value === timeRangeField.value,
-		)
-		if (!timeRange) {
-			addRule(defaultTimeRangeRule)
-			return defaultTimeRangeRule
-		}
+		)! // ZANETODO: can we enforce this?
```
mayberryzane, line 1444: we should separate the timeRangeRule from the rules eventually since it is required

frontend/src/components/QueryBuilder/QueryBuilder.tsx
```diff
@@ -1919,184 +1565,38 @@ function QueryBuilder<T extends SearchParamsInput | ErrorSearchParamsInput>(
...
-		}
-	}, [
-		setSearchParamsToUrlParams,
-		searchParams,
-		selectedSegment,
-		activeSegmentUrlParam,
-		initialized,
-	])
-
-	useEffect(() => {
```
mayberryzane, line 1996: the sync button wasn't working previously because of some changes to the relative date logic

frontend/src/pages/Sessions/EmptySessionsSearchParams.tsx
```diff
@@ -1,24 +1 @@
...
-	track_properties: [],
-	excluded_track_properties: [],
-	visited_url: undefined,
-	first_time: false,
-	device_id: undefined,
-	environments: [],
-	app_versions: [],
-	show_live_sessions: false,
-	query: `{"isAnd":true,"rules":[["custom_processed","is","true"]]}`,
-}
```
ccschmitz, line 24: 🥳

frontend/src/pages/Sessions/SessionsFeedV3/SessionsFeedV3.tsx
```diff
@@ -58,94 +57,83 @@ import * as style from './SessionFeedV3.css'
...
+					start_date: roundFeedDate(
+						backendSearchQuery?.startDate.toISOString() ?? null,
+					).format(),
+					end_date: roundFeedDate(
+						backendSearchQuery?.endDate.toISOString() ?? null,
+					).format(),
 				},
 			},
-			skip: !backendSearchQuery || !project_id,
-			fetchPolicy: projectHasManySessions ? 'cache-first' : 'no-cache',
```
ccschmitz, line 94: Just curious, what's the reasoning for removing this?

frontend/src/pages/Sessions/SessionsFeedV3/SessionsFeedV3.tsx
```diff
@@ -331,25 +300,12 @@ export const SessionFeedV3 = React.memo(() => {
...
-										item="sessions"
-										customTitle="Your project doesn't have starred sessions."
-										customDescription={
-											'Starring a session is like bookmarking a website. ' +
-											'It gives you a way to tag a session that you want to look at again. ' +
-											'You can star a session by clicking the star icon next to the user details ' +
-											"in the session's right panel."
-										}
-									/>
-								) : (
```
ccschmitz, line 345: Do we not support starred sessions anymore? Wondering if we want to go another step in cleaning up starring logic in a future PR.

packages/ui/src/components/DatePicker/PreviousDateRangePicker.tsx
```diff
@@ -16,7 +16,7 @@ import { Box } from '../Box/Box'
 import { colors } from '../../css/colors'
 import { TimeInput } from './TimeInput'
 
-export { getDefaultPresets } from './utils'
+export { getDefaultPresets, getNow } from './utils'
```
ccschmitz, line 19: Don't see where this is being used 🤔 - could it be cleaned up?

ccschmitz (approved):
A couple minor questions/comments, but code + click test looks great! Love all the code that was cleaned up as part of this. Thanks for tackling!

frontend/src/pages/Sessions/SessionsFeedV3/SessionsFeedV3.tsx
mayberryzane: yeah, we don't expose this in the UI anymore, I agree that there's probably more starring logic to clean up

frontend/src/pages/Sessions/SessionsFeedV3/SessionsFeedV3.tsx
mayberryzane: this was causing the query to run twice if `projectHasManySessions` is true since it defaults to false before any results are loaded, and it just seemed odd to me to have the caching policy dependent on the number of results. I think this is from a time before we had a default date filter, so users in new projects might have been confused why new sessions would appear after modifying the query but not appear if they made a query that was already cached.

packages/ui/src/components/DatePicker/PreviousDateRangePicker.tsx
mayberryzane: this is an export - it's used in a few different places so they reference the same "now" time

frontend/src/pages/Sessions/SessionsFeedV3/SessionsFeedV3.tsx
ccschmitz: Ah, gotcha. Thanks for sharing, @mayberryzane!

packages/ui/src/components/DatePicker/PreviousDateRangePicker.tsx
ccschmitz: Oh, sorry, I wasn't looking closely and thinking this was an `import` 😅

PR #5716, opened 2023-06-23
Error Messages: fix order of sql arguments

SpennyNDaJets:
## Summary
The [last commit here](https://github.com/highlight/highlight/pull/5707/commits/d0cbf6e446bcda7dc5da859c3a07a3294e94b952) removed the  projecct_id from the sql query, but we are still passing it in as an argument to the incorrect place.

Click test:
1) Create a new error alert
  a) set it to send to #chris-errors-test
  b) Set the Alert Frequency to 5 minutes
2) Navigate to the buttons page (`/1/buttons`)
3) Clock on "Throw an Error" button
- [ ] Error is sent to slack channel
4) Click on "Throw an Error" again
- [ ] Error is not sent to slack channel
5) Click on "Throw a randomized Error"
- [ ] Error is sent to slack channel
6) Wait 5 minutes
7) Click on "Throw an Error" again
- [ ] Error is sent to slack channel

N/A


backend/model/model.go
```diff
@@ -2014,7 +2014,7 @@ func (obj *Alert) GetDailySessionEventFrequency(db *gorm.DB, id int) ([]*int64,
 			AND e.session_alert_id=?
 		GROUP BY d.date
 		ORDER BY d.date;
-	`, id, obj.ProjectID).Scan(&dailyAlerts).Error; err != nil {
```
SpennyNDaJets, line 2017: Also cleaned up these since they were not being used

backend/public-graph/graph/resolver.go
```diff
@@ -1636,7 +1636,7 @@ func (r *Resolver) sendErrorAlert(ctx context.Context, projectID int, sessionObj
 						AND obj.error_group_id=?)
 					AND ev.error_alert_id=?
 					AND ev.sent_at > NOW() - ? * (INTERVAL '1 SECOND')
-			`, projectID, group.ID, errorAlert.ID, errorAlert.Frequency).Scan(&numAlerts).Error; err != nil {
```
SpennyNDaJets, line 1639: We are passing in the projectID as the groupID, so this is always returning 0

Vadman97 (approved):
my bad for not catching this originally :(

PR #5736, opened 2023-06-26
Fix left panel closing on error page

SpennyNDaJets:

### DIFF
--- a/frontend/src/pages/ErrorsV2/utils/ErrorPageUIConfiguration.ts
+++ b/frontend/src/pages/ErrorsV2/utils/ErrorPageUIConfiguration.ts
@@ -17,7 +17,7 @@ const useErrorPageConfiguration = (): typeof DEFAULT_STATE => {
 	const { isLoggedIn } = useAuthContext()
 	const { projectId } = useNumericProjectId()
 	const [showLeftPanel, setShowLeftPanel] = useLocalStorage<boolean>(
-		'highlightErrorPageShowLeftPanel',
+		'highlightMenuShowLeftPanel',
 		true,
 	)
 


frontend/src/pages/ErrorsV2/utils/ErrorPageUIConfiguration.ts
```diff
@@ -16,10 +16,7 @@ const DEFAULT_STATE = {
 const useErrorPageConfiguration = (): typeof DEFAULT_STATE => {
 	const { isLoggedIn } = useAuthContext()
 	const { projectId } = useNumericProjectId()
-	const [showLeftPanel, setShowLeftPanel] = useLocalStorage<boolean>(
-		'highlightErrorPageShowLeftPanel',
-		true,
-	)
+	const { showLeftPanel, setShowLeftPanel } = usePlayerConfiguration()
```
ccschmitz, line 19: Do you think it makes sense to sync the left panel behavior across errors and sessions? I'd think we want these to be separate since they have pretty different use cases (I'm actually not sure why you'd ever minimize the left panel on errors).

frontend/src/pages/ErrorsV2/utils/ErrorPageUIConfiguration.ts
```diff
@@ -16,10 +16,7 @@ const DEFAULT_STATE = {
 const useErrorPageConfiguration = (): typeof DEFAULT_STATE => {
 	const { isLoggedIn } = useAuthContext()
 	const { projectId } = useNumericProjectId()
-	const [showLeftPanel, setShowLeftPanel] = useLocalStorage<boolean>(
-		'highlightErrorPageShowLeftPanel',
```
Hg6f9, line 20: - - [ ]

frontend/src/pages/ErrorsV2/utils/ErrorPageUIConfiguration.ts
SpennyNDaJets: I agree! From the tested behavior, the two different panels shouldn't be in sync. From the behavior I was seeing, the `showLeftPanel` value appears to be reset to `true` when the user changes pages, but let me know if you are experiencing something different

ccschmitz: Oh, sorry for the misunderstanding - I didn't click test! I thought it was meant to be persisted across page loads since it was being stored in local storage. Feels like we should either a) not persist to local storage, or b) have it not reset between pages and store it under different keys. I'd defer to @jay-khatri and @julian-highlight on the UX.

deltaepsilon (approved):

Issue #4142, opened 2023-02-03
Clean up SearchParams


et:
We have some old logic in search params that is no longer relevant:

https://github.com/highlight/highlight/blob/cfd279789c0185cc4ee1031a51f21f22cbe250ed/backend/model/model.go#L696-L715

The only thing we still use is `Query`. Everything else can be dropped.

We only support the other fields for some backwards compatibility with old segments. For those, we should be able to re-save those segments to shove everything into `Query`.

PR #5747, opened 2023-06-27
use strongly typed logs table

et:
## Summary

While working on a different table that uses react-table, I noticed that we can now make our react-table usages more typesafe (see #4142) and the utility in #4185

Confirmed keys work with typeahead:

Screenshot 2023-06-26 at 10 04 08 PM

Confirmed that keys are typesafe:

Screenshot 2023-06-26 at 10 03 46 PM

N/A

### DIFF
--- a/frontend/src/pages/LogsPage/LogsTable/LogsTable.tsx
+++ b/frontend/src/pages/LogsPage/LogsTable/LogsTable.tsx
@@ -2,7 +2,7 @@ import { ApolloError } from '@apollo/client'
 import { Button } from '@components/Button'
 import { Link } from '@components/Link'
 import LoadingBox from '@components/LoadingBox'
-import { LogLevel as LogLevelType, ReservedLogKey } from '@graph/schemas'
+import { ReservedLogKey } from '@graph/schemas'
 import { LogEdge } from '@graph/schemas'
 import {
 	Box,
@@ -24,7 +24,7 @@ import {
 } from '@pages/LogsPage/SearchForm/utils'
 import { LogEdgeWithError } from '@pages/LogsPage/useGetLogs'
 import {
-	ColumnDef,
+	createColumnHelper,
 	ExpandedState,
 	flexRender,
 	getCoreRowModel,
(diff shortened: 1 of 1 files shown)


frontend/src/pages/LogsPage/LogsTable/LogsTable.tsx
```diff
@@ -119,56 +119,48 @@ const LogsTableInner = ({
...
-								cssClass={styles.expandIcon}
-							>
-								{row.getIsExpanded() ? (
-									<IconExpanded />
-								) : (
-									<IconCollapsed />
-								)}
-							</Box>
-						)}
-						<LogTimestamp timestamp={getValue() as string} />
```
et, line 147: We no longer need `as string` since getValue() knows the type:
Screenshot 2023-06-26 at 10 06 12 PM

frontend/src/pages/LogsPage/LogsTable/LogsTable.tsx
```diff
@@ -119,56 +119,48 @@ const LogsTableInner = ({
...
-					</Box>
-				),
-			},
-			{
-				accessorKey: 'node.level',
-				cell: ({ getValue }) => (
-					<LogLevel level={getValue() as LogLevelType} />
```
et, line 154: Same here:
Screenshot 2023-06-26 at 10 06 47 PM

Note `LogLevelType` was an alias for `LogLevel`

frontend/src/pages/LogsPage/LogsTable/LogsTable.tsx
```diff
@@ -119,56 +119,48 @@ const LogsTableInner = ({
 	const queryTerms = parseLogsQuery(query)
 	const [expanded, setExpanded] = useState<ExpandedState>({})
 
-	const columns = React.useMemo<ColumnDef<LogEdge>[]>(
```
et, line 122: Didn't see a reason to keep this memoized. It seems to only make the typing more difficult to manage. I did a clicktest on the reflame preview and don't see any issues.

et: I added the `useMemo` originally here: https://github.com/highlight/highlight/commit/21771a06efa1279c79af2879bf4847c9fed2b8d8#diff-ee6f428905a95e722c24e36e545d285643f49fc7d84333333089c32ec2bbfaeeR44

But I believe it was just copy/pasted from some example.

I also tracked down the eslint ignore rule to here: https://github.com/highlight/highlight/commit/cedbdd740620cdcb3e6fb79e5bbf792dbe156bce#diff-ee6f428905a95e722c24e36e545d285643f49fc7d84333333089c32ec2bbfaeeR126-R127

since we needed the columns to rerender when the query was updated.

Since it's annoying to add eslint rules like this, it's best to avoid memoization until we can identify a reason to add it.

ccschmitz (approved):

PR #5748, opened 2023-06-27
Frontend rendering of websocket events in the network tab

SpennyNDaJets:
## Summary
WebSocket resources are being displayed in the Network tab, but without any formatting event though they differ from HTTP resources.

Update the UI to:
- display 1 request per open/closed connection
- display status as 101
- display waterfall and latency if connection was closed before exiting

This does not render the messages in the left panel when clicked.

https://www.loom.com/share/5e257b0df642434e82be7a0d37d15593?sid=0aea073b-6f1f-4814-b5e3-31f1da31ca3d

1. View a live session
2. View in Live mode
3. Close out your tab, and let your current session process
4. Find the session in the Session search page
5. Play with the filter, click the WebSocket event, etc

N/A


Vadman97 (approved):
🚢

Issue #5693, opened 2023-06-21
Text Overflow on DatePicker for large date ranges


jay-khatri:
For large custom date ranges, the text overflows beyond the height of the datepicker. Instead of doing this, we should truncate the text (so it stays on one line) and add an ellipsis at the end. 

<img width="361" alt="image" src="https://github.com/highlight/highlight/assets/20292680/8ddc8b23-b2d7-4946-a360-449970bfc303">

PR #5749, opened 2023-06-27
chore: fix the text overflow on DateInput

betty-alagwu:
## Summary

This PR is the first pass for issue #5693 

This PR enhances the control bar functionality by addressing the issue of text overflow when selecting date and time for large date ranges.

Here's a video of how it works now:

https://github.com/highlight/highlight/assets/94234459/3616af33-673d-47a4-b842-241bed4ebdf6

### DIFF
--- a/packages/ui/src/components/DatePicker/PreviousDateRangePicker.tsx
+++ b/packages/ui/src/components/DatePicker/PreviousDateRangePicker.tsx
@@ -328,6 +328,10 @@ const PreviousDateRangePickerImpl = ({
 			<Menu.Button
 				kind="secondary"
 				iconRight={<IconSolidCheveronDown />}
+				style={{
+					whiteSpace: 'nowrap',
+					overflow: 'hidden',
+				}}
 				{...(props as Omit<MenuButtonProps, 'ref'>)}
 			>
 				{buttonLabel}


betty-alagwu:
Hi, @jay-khatri! I wanted to mention that the `Menu.Button` component used in the dateInput is a package called `ariakit`, and the package's styles are currently overriding the ellipsis class that I'm passing to the button. 
I was wondering if there might be a more efficient approach for this. 

If so, I'd be more than happy to implement any additional changes.

ccschmitz:
@betty-alagwu yeah, I see how that's problematic when we try to pass that down from the datepicker comonent. I wonder if the problem here is something we should fix at the `Button` level, which is what `Menu.Button` uses under the hood. I can't think of an instance where we would want a `Button`'s content spanning multiple lines.

I'd be curious if we could wrap `ButtonContent` with a container that applies `white-space: nowrap` (`<Box whiteSpace="nowrap">`) to resolve this 🤔 

https://github.com/highlight/highlight/blob/8e4f4de32f637f90410eb7e61e9f0e2764c83f4c/packages/ui/src/components/Button/Button.tsx#L98C2-L98C2

betty-alagwu:
Thanks for the suggestion @ccschmitz , will try this out and revert back.

betty-alagwu:
> 

 @ccschmitz, I tested the solution you proposed. Adding the `white-space: nowrap` to the button yields the same results as my current solution. But I noticed it's affecting all the other buttons in the app. 
 
 I think it might be safe to just update the menu button at this moment.

Also, I tried  to get the ellipsis to show by modifying the `ButtonContent`, but that didn't work out too. 

I think the nowrap on the menu button (my current solution) might be the best way forward at this time.

Please let me know what you think. Thanks.

ccschmitz:
> I noticed it's affecting all the other buttons in the app.

@betty-alagwu did you notice any negative side effects from this in other areas of the app? I actually was thinking this might be a change we want to roll out across all buttons, but maybe not if it's having unintended consequences. Regardless, it looks like the fix you have here works great - thank you for tackling this!

Also, apologies if I misunderstood the issue you mentioned in https://github.com/highlight/highlight/pull/5749#issuecomment-1609583049. I thought you were saying the current changes weren't working because of an issue with Ariakit overriding the styles you wanted to apply, but I think I know what you mean about the `className` being overridden. Sorry for the confusion around that!

ccschmitz (approved):
I approved a [Reflame deploy](https://preview.highlight.io/?%7Er_preview=%7B%22action%22%3A%22start%22%2C%22variantId%22%3A%2201H417T1PMFHVG2W9AM5SYY1RH%22%2C%22region%22%3A%22sjc%22%2C%22variantData%22%3A%22%7B%5C%22type%5C%22%3A%5C%22branch%5C%22%2C%5C%22branch%5C%22%3A%5C%22fix%2Ftextoverflow-on-dateinput%5C%22%2C%5C%22forkSpecifier%5C%22%3A%5C%22betty-alagwu%2Fhighlight%5C%22%2C%5C%22githubOwnerName%5C%22%3A%5C%22highlight%5C%22%2C%5C%22githubRepositoryName%5C%22%3A%5C%22highlight%5C%22%7D%22%7D) for this and click tested - it looks great! Thanks for tackling this, @betty-alagwu!

Issue #5687, opened 2023-06-21
Add an explicit button for generating session summary


jay-khatri:
- For enhancing a session, add a button to "generate a summary" of a session

julian-highlight:
Created a simple design here: https://www.figma.com/file/rdBPNzn7Klk6RG1LHUCE5B/Screen-Designs?type=design&node-id=9187-575774&mode=design&t=wpyUSW9q6L8qWF4E-11

@eightants

PR #5750, opened 2023-06-27
Add explicit button for session summary

eightants:
## Summary

Added button to trigger session insights as specified in https://github.com/highlight/highlight/issues/5687

Also fixed the backend insights retry function when local server is unavailable. 

Screenshot 2023-06-27 at 12 43 11 PM

### DIFF
--- a/backend/lambda/lambda.go
+++ b/backend/lambda/lambda.go
@@ -77,11 +77,7 @@ func (s *Client) GetSessionInsight(ctx context.Context, projectID int, sessionID
 			ID:        232563428,
 			ProjectID: 1,
 		})
-		var localServerErr error
-		req, localServerErr = http.NewRequest(http.MethodPost, "http://localhost:8765/session/insight", bytes.NewBuffer(b))
-		if localServerErr != nil {
-			req, _ = http.NewRequest(http.MethodPost, "https://ohw2ocqp0d.execute-api.us-east-2.amazonaws.com/default/ai-insights", bytes.NewBuffer(b))
-		}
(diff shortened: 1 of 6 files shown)


Vadman97 (commented):
unrelated to this PR, but just noticed that insights can come out of order. can we update the prompt to fix that?

image

SpennyNDaJets (commented):
Wanted to just raise the following things I found from click testing:
1. The Button is not there after generating a response, and then changing sessions. Not sure if its automatically hitting our endpoint, or stuck in a loading loop though 🤔 
2. The times seem to be off, but not sure if this is a local/unrelated issue
Screenshot 2023-06-27 at 12 53 10 PM

eightants:
> Wanted to just raise the following things I found from click testing:
> 
> 1. The Button is not there after generating a response, and then changing sessions. Not sure if its automatically hitting our endpoint, or stuck in a loading loop though 🤔
> 2. The times seem to be off, but not sure if this is a local/unrelated issue
>    Screenshot 2023-06-27 at 12 53 10 PM

2. Yeah this is cause we're using a production session in s3 to get all session events, so the timestamps on that session doesn't match the local session

Vadman97 (changes requested):
small note on the loading button
also the insights seem to be broken in the lambda
https://www.loom.com/share/14373574d53942e390988be0efd2fffa

Vadman97 (approved):
realized that the lambda might be broken in the render.com preview since that doesn't deploy the backend changes, in which case disregard!

PR #5763, opened 2023-06-27
allow viewing logs for backend only errors

et:


frontend/src/pages/ErrorsV2/ErrorInstance/ErrorInstance.tsx
```diff
@@ -310,26 +269,7 @@ const ErrorInstance: React.FC<Props> = ({ errorGroup }) => {
...
-								display="flex"
-								alignItems="center"
-								flexDirection="row"
-								gap="4"
-							>
-								<IconSolidLogs />
-								Show logs
-							</Box>
-						</LinkButton>
+						<ShowLogsButton data={data} />
```
et, line 272: Moved the logic into a separate component because `ErrorInstance` is quite big and it's easier to wrangle this if it's in a separate file.

frontend/src/pages/ErrorsV2/ErrorInstance/ShowLogsButton.tsx
```diff
@@ -0,0 +1,79 @@
...
+export const ShowLogsButton = ({ data }: Props) => {
+	const { isLoggedIn } = useAuthContext()
+
+	const logsLink = getLogsLink(data)
+
+	return (
+		<LinkButton
+			kind="primary"
+			emphasis="high"
+			disabled={!isLoggedIn || logsLink === ''}
```
SpennyNDaJets, line 69: Not a blocker, but sort of makes me weary using the `===` comparison here. Is it possible to use `!logsLink` that would also catch null and undefined values?

SpennyNDaJets (approved):
Click test looks great! Left one question below

frontend/src/pages/ErrorsV2/ErrorInstance/ShowLogsButton.tsx
et: Good question. The return value of `getLogsLink` is a `string` so it can't be undefined or null.
https://github.com/highlight/highlight/blob/3363218b9b936c30d76fe083f6430b98caec7ea1/frontend/src/pages/ErrorsV2/ErrorInstance/ShowLogsButton.tsx#L15-L20

et: For additional context, I fallback to an empty string intentionally because the `to` prop is required for `<LinkButton />`. It's somewhat of a non-graceful way of handling dynamic links when they're disabled. Here's another example of us doing the same thing:

https://github.com/highlight/highlight/blob/3363218b9b936c30d76fe083f6430b98caec7ea1/frontend/src/pages/Setup/IntegrationBar.tsx#L131-L137

PR #5764, opened 2023-06-27
disable indexeddb when there is no way to read memory usage

Vadman97:
## Summary

On Safari, we don't have access to reading how much memory indexeddb is occupying.
This means we cannot clean up if we are getting close to memory usage limits.
Disable indexeddb altogether in this case to avoid https://app.highlight.io/1/errors/4E8fIhngcFgOOeWpkmG0A6Aq7bee

* Fixes backend env vars for the hobby deployment needed by the backend

Reflame preview

No


et (approved):
Awesome. This seems a lot more safe for the end user.

PR #5766, opened 2023-06-27
avoid overwriting native `window.Highlight` class

Vadman97:
## Summary

Our client would set `window.Highlight` so that firstload could reference the class.
However, this would replace the native `window.Highlight` [class ](https://developer.mozilla.org/en-US/docs/Web/API/Highlight)defined by browsers.
This change renamed `window.Highlight` to `window.HighlightIO` to avoid this problem.

Render / reflame preview.

No

### DIFF
--- a/experiments/chrome-extension/README.md
+++ b/experiments/chrome-extension/README.md
@@ -51,7 +51,7 @@ steps:
 
 -   `cd experiments/chrome-extension`
 -   `yarn build`
--   replace `new window.Highlight` with `Highlight.create` in `firstload/index.tsx`
+-   replace `new window.HighlightIO` with `Highlight.create` in `firstload/index.tsx`
 -   `yarn firstload:update`
 -   open chrome extensions page on chrome or sidekick
 -   enable developer mode on the top right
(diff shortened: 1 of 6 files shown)


mayberryzane (approved):

swanzeyb:
Is there a timeline for this fix being released? We're still getting this issue on our frontend resulting from this naming.

Screenshot 2023-06-30 at 9 52 56 AM

Vadman97:
Tests are failing in the firstload/client xversion test. This is because this client change is not compatible with old firstloads.
```
  web client recording spec
    1) public graph requests are recorded when highlight set up from unpkg
    ✓ public graph requests are recorded when highlight set up from unpkg-remote (4371ms)
    2) public graph requests are recorded when highlight set up from jsdelivr
    ✓ public graph requests are recorded when highlight set up from local (3669ms)


  2 passing (10s)
  2 failing
```

this is ok though because old firstloads are pinned to use their respective client. once this is merged, the unpkg/jsdelivr bundles will update to reference the new client, so this test will resume passing.

Vadman97:
alright @swanzeyb , this is released now https://www.npmjs.com/package/highlight.run/v/7.3.2

PR #5767, opened 2023-06-28
reset trial on billing update, check quota during processing

### LINKED ISSUE #5760: billing error where free tier customers who chose a paid plan are getting a warning

mayberryzane:
## Summary
- fixes #5760
- when subscribing to a plan, clear the `TrialEndDate` to prevent ingesting extra sessions
- check billing quota again during session processing (since live sessions don't count toward the limit)

- click tested locally

- no

### DIFF
--- a/backend/worker/worker.go
+++ b/backend/worker/worker.go
@@ -784,6 +784,12 @@ func (w *Worker) processSession(ctx context.Context, s *model.Session) error {
 
 	pagesVisited := len(visitFields)
 
+	workspace, err := w.Resolver.GetWorkspace(project.WorkspaceID)
+	if err != nil {
+		return err
+	}
+	withinBillingQuota, _ := w.PublicResolver.IsWithinQuota(ctx, pricing.ProductTypeSessions, workspace, time.Now())
+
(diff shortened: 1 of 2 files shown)


backend/worker/worker.go
```diff
@@ -784,6 +784,12 @@ func (w *Worker) processSession(ctx context.Context, s *model.Session) error {
 
 	pagesVisited := len(visitFields)
 
+	workspace, err := w.Resolver.GetWorkspace(project.WorkspaceID)
+	if err != nil {
+		return err
+	}
+	withinBillingQuota, _ := w.PublicResolver.IsWithinQuota(ctx, pricing.ProductTypeSessions, workspace, time.Now())
```
Vadman97, line 791: do we still need the check in `InitializeSessionImpl` ?

Vadman97 (approved):

backend/worker/worker.go
mayberryzane: I think it's better to have the check there too, just because otherwise users will be able to see all their live sessions even though they're not paying

PR #5768, opened 2023-06-28
ensure first frame of console.error is not dropped

Vadman97:
## Summary

Our console error listener would drop the first frame of errors even when they were caught with a full stack trace.
It seems like this was an error when sharing the logic with the other console listeners that need to drop
the first frame of the stack trace because it is produced from the console listener itself (keeping that top frame
would be redundant).
Issue originally reported by a customer.
https://highlightcorp.slack.com/archives/C04NQ25RV8B/p1687896259977009

Local deploy triggering a frontend error.

before
image

after
image

New patch version of the SDK released.

### DIFF
--- a/backend/stacktraces/enhancer_test.go
+++ b/backend/stacktraces/enhancer_test.go
@@ -2,6 +2,8 @@ package stacktraces
 
 import (
 	"context"
+	"github.com/openlyinc/pointy"
+	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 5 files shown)


et (approved):
Test failure looks related but otherwise looks good!

Vadman97:
e2e failure unrelated, caused by https://github.com/highlight/highlight/pull/5766

PR #5772, opened 2023-06-28
Revert "refactor QueryBuilder to use a reducer pattern"

mayberryzane:
Reverts highlight/highlight#5712


Vadman97 (approved):

mayberryzane:
opened because of an issue causing workspace settings not to load - that issue was fixed instead

PR #5773, opened 2023-06-28
move error/search context up so it's accessible from workspace / other pages

mayberryzane:
## Summary
- workspace level pages were breaking because of some code I added to the command bar to access the search context (the command bar is accessible outside of just the project-level pages)

- click tested locally, tested other non-project-router pages (workspace settings, login, etc) to make sure they weren't crashing

- need to turn on render autodeploys after this is merged


Vadman97 (approved):
reflame click test of workspace pages looks good!

PR #5781, opened 2023-06-28
constantize authorization errors

et:
## Summary

This PR makes all authorization errors (i.e. you're logged in but you don't have access to this resource) return a constant error message. Additionally, I removed all error wrapping for this scenario. Taking this approach will allow all these auth errors to be grouped together since we match if the error message is identical ([docs](https://www.highlight.io/docs/general/product-features/error-monitoring/grouping-errors)). 

If we didn't have error message matching, we could also fallback to [filtering](https://www.highlight.io/docs/general/product-features/error-monitoring/ignoring-errors#ignore-errors-by-regex-on-the-error-body) by error message.

Checked the graphql errors response when there is an auth issue:

Screenshot 2023-06-28 at 2 07 23 PM

Will monitor after rolling out. Assuming this groups as expected, we should ignore an error of "403 - AuthorizationError" and any subsequent auth errors should be filtered out.

### DIFF
--- a/backend/private-graph/graph/resolver.go
+++ b/backend/private-graph/graph/resolver.go
@@ -340,7 +340,7 @@ func (r *Resolver) isAdminInWorkspaceOrDemoWorkspace(ctx context.Context, worksp
 	} else {
 		workspace, err = r.isAdminInWorkspace(ctx, workspace_id)
 		if err != nil {
-			return nil, AuthorizationError
+			return nil, err
 		}
 	}
 	return workspace, nil
--- a/backend/private-graph/graph/resolver.go
+++ b/backend/private-graph/graph/resolver.go
@@ -316,7 +316,7 @@ func (r *Resolver) isAdminInProjectOrDemoProject(ctx context.Context, project_id
 	} else {
 		project, err = r.isAdminInProject(ctx, project_id)
 		if err != nil {
-			return nil, e.Wrap(err, "admin is not in project or demo project")
+			return nil, err
 		}
 	}
 	return project, nil
@@ -462,7 +462,7 @@ func (r *Resolver) isAdminInWorkspace(ctx context.Context, workspaceID int) (*mo
 	}
 
 	if workspaceID != workspace.ID {
-		return nil, e.New("workspace is not associated to the current admin")
+		return nil, errors.New(AuthorizationError)
 	}
 
 	return &workspace, nil
--- a/backend/private-graph/graph/resolver.go
+++ b/backend/private-graph/graph/resolver.go
@@ -76,6 +76,8 @@ const ErrorGroupLookbackDays = 7
 const SessionActiveMetricName = "sessionActiveLength"
 const SessionProcessedMetricName = "sessionProcessed"
 
+const AuthorizationError = "403 - AuthorizationError"
+
 var (
 	WhitelistedUID  = os.Getenv("WHITELISTED_FIREBASE_ACCOUNT")
 	JwtAccessSecret = os.Getenv("JWT_ACCESS_SECRET")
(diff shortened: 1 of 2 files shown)


backend/private-graph/graph/schema.resolvers.go
```diff
@@ -443,7 +443,7 @@ func (r *mutationResolver) CreateAdmin(ctx context.Context) (*model.Admin, error
 func (r *mutationResolver) CreateProject(ctx context.Context, name string, workspaceID int) (*model.Project, error) {
 	workspace, err := r.isAdminInWorkspace(ctx, workspaceID)
 	if err != nil {
-		return nil, nil
```
et, line 446: This was a bug. There's a potential lint rule here that could catch this:
https://github.com/highlight/highlight/pull/5592#issuecomment-1583574619

backend/private-graph/graph/resolver.go
```diff
@@ -76,6 +76,8 @@ const ErrorGroupLookbackDays = 7
 const SessionActiveMetricName = "sessionActiveLength"
```
et, line 76: When reviewing, please also expand `backend/private-graph/graph/schema.resolvers.go` as github closed that file

backend/private-graph/graph/resolver.go
```diff
@@ -76,6 +76,8 @@ const ErrorGroupLookbackDays = 7
...
 const SessionProcessedMetricName = "sessionProcessed"
 
+const AuthorizationError = "403 - AuthorizationError"
```
et, line 79: We can also add `401 - AuthenticationError` when someone isn't logged in:
https://github.com/highlight/highlight/blob/718d86dedddf1937c859959577e5b76cd005e06b/backend/private-graph/graph/schema.resolvers.go#L7217-L7220

backend/private-graph/graph/schema.resolvers.go
Vadman97: this change might have been made to omit this common 'expected' error. should be ok to return a real error here now though if we ignore them all, but might want to check that the frontend doesn't break if this returns an error

backend/private-graph/graph/resolver.go
Vadman97, line 79: nit: i think you could make this a `var AuthorizationError = errors.New("403 - AuthorizationError")` so that you could just `return nil, AuthorizationError` below

backend/private-graph/graph/resolver.go
```diff
@@ -338,7 +340,7 @@ func (r *Resolver) isAdminInWorkspaceOrDemoWorkspace(ctx context.Context, worksp
 	} else {
 		workspace, err = r.isAdminInWorkspace(ctx, workspace_id)
 		if err != nil {
-			return nil, e.Wrap(err, "admin is not in workspace or demo workspace")
+			return nil, errors.New(AuthorizationError)
```
Vadman97, line 343: could we keep this consistent with line 319? this should return err and we cleanup `isAdminInWorkspace` to only return an `AuthorizationError`, or else 319 should return `errors.New(AuthorizationError)`. just in case either of these are used elsewhere directly in a resolver.

Vadman97 (approved):

backend/private-graph/graph/resolver.go
et: Good idea. 
Fixed in d693f80befb5b3e33b1939617bafc2ae9c81583f

backend/private-graph/graph/resolver.go
et: Good catch.
f3e549e4a513bcb770a2c0d83f8ec9c917b098c9

backend/private-graph/graph/schema.resolvers.go
et: Ah I see. I'll revert for now since this was out of scope and we can take a look in #5592.
1c1b9dc96136ed95afdcc47e3c967adff230b360

PR #5783, opened 2023-06-28
use special pricing for non-usage based plans

mayberryzane:
## Summary
- customers committed to a monthly spend pay a different price - this fixes billing limit calculations for sessions

- unit test

- no

### DIFF
--- a/backend/public-graph/graph/resolver.go
+++ b/backend/public-graph/graph/resolver.go
@@ -1527,10 +1527,15 @@ func (r *Resolver) IsWithinQuota(ctx context.Context, productType pricing.Produc
 		return false, 1
 	}
 
+	basePrice := pricing.ProductToBasePriceCents(productType)
+	if stripePlan != privateModel.PlanTypeUsageBased {
+		basePrice = pricing.ProductToBasePriceCentsNonUsageBased(productType)
+	}
+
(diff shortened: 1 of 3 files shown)


backend/public-graph/graph/resolver.go
```diff
@@ -1527,10 +1527,15 @@ func (r *Resolver) IsWithinQuota(ctx context.Context, productType pricing.Produc
 		return false, 1
 	}
 
+	basePrice := pricing.ProductToBasePriceCents(productType)
+	if stripePlan != privateModel.PlanTypeUsageBased {
+		basePrice = pricing.ProductToBasePriceCentsNonUsageBased(productType)
+	}
```
Vadman97, line 1533: does this need to be a separate function call, or could `pricing.ProductToBasePriceCents` just have the logic for non-usage-based plans? or is this because we rely on that constant return value of `ProductToBasePriceCents` elsewhere?

Vadman97 (approved):

backend/public-graph/graph/resolver.go
mayberryzane: good point, will combine the two functions

Vadman97 (approved):

PR #5689, opened 2023-06-21
WebSockets: saving data to redis/postgres

SpennyNDaJets:
## Summary
WebSocket events are getting sent to the backend, but just dropped. This adds functionality to storage the WebSocket events in Redis and S3.

## TODOs
- [x] Handle large payloads -> leave for now, and come back to see if there is an issue when loading into JS
- [x] Handle empty payloads -> leave these in for now
- [x] What versions to bump

## How did you test this change?
**Click test:**
Before: Check out [this PR](https://github.com/highlight/highlight/pull/5727) that has the storage changes, along with some frontend changes to better test
1) Create a new session that watches a live session (i.e. creates web socket connections)
2) Exit the session and let it process
3) Find the `GetSession` graphql query, and find the `web_socket_events_url`
4) Click on the `web_socket_events_url` to view the events in your browser
- [x] Events are correct captured and stored

https://www.loom.com/share/66c5f3ed87a14d55b2b81838ebc151ce?sid=7fec7489-7a76-4fed-93cd-0f68aa871c96

## Are there any deployment considerations?
There should not be any events recorded to S3 since the new npm package has not been built yet - watch S3 storage and errors around payload being too large.

PR #5788, opened 2023-06-29
fix pushpayload processing of messages with nil websocket data

Vadman97:
## Summary

Fixup to #5689. PushPayloads from old clients were causing a panic due to nil pointer dereference.

N/A

Rewinding kafka offset.


mayberryzane (approved):

SpennyNDaJets (approved):
Thanks for catching this

PR #5789, opened 2023-06-29
constantize authentication errors

et:
## Summary

Similar to #5781 but for authentication errors (you're not logged in).

* Copied a GQL request into Postman and removed the `token` attribute
* Observed 401 error in Highlight

Screenshot 2023-06-29 at 2 21 59 PM

### DIFF
--- a/backend/private-graph/graph/resolver.go
+++ b/backend/private-graph/graph/resolver.go
@@ -493,6 +493,10 @@ func (r *Resolver) isAdminInProject(ctx context.Context, project_id int) (*model
 	if err != nil {
 		return nil, e.Wrap(err, "error querying projects")
 	}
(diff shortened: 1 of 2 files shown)


backend/private-graph/graph/resolver.go
```diff
@@ -487,6 +493,10 @@ func (r *Resolver) isAdminInProject(ctx context.Context, project_id int) (*model
 	if err != nil {
 		return nil, e.Wrap(err, "error querying projects")
 	}
+	if len(projects) == 0 {
```
et, line 496: If you have no projects then you're not logged in. We short circuit here because if we continued, we would return a `AuthorizationError` which wouldn't be semantically the same.

mayberryzane (approved):

PR #5792, opened 2023-06-29
fix panic in stacktrace parsing code

Vadman97:
## Summary

Some stacktraces included a sourcemap URL that would cause a panic.
Fixes the panic and adds logging to make sure we can pinpoint what invalid sourcemap URL is provided.

Ingesting valid errors correctly.
image

Ingesting a fake sourcemap error of this type into the frontend.
image

Will monitor new log.


et (approved):
👏

PR #5795, opened 2023-06-30
Normalness scoring during session processing

mayberryzane:
## Summary
- closes [#5720](https://github.com/highlight/highlight/issues/5720)
- records the user journey from the timeline indicator events
- calculates and saves a normalness score with the session
- future PR will allow sorting by this score in the sessions feed

- copied prod data to local, observed the normalness scores for processed sessions

- no


backend/lambda-functions/journeys/main.go
Vadman97, line 1: does this lambda need a deploy step in github actions `deploy.yml`?

backend/worker/worker.go
```diff
@@ -617,6 +619,43 @@ func (w *Worker) processSession(ctx context.Context, s *model.Session) error {
...
+			if err != nil {
+				return err
+			}
+			if err := w.Resolver.DB.Model(&model.UserJourneyStep{}).Save(&userJourneySteps).Error; err != nil {
+				return err
+			}
+
+			// Calculate the "normalness" score using the trailing 30 days
+			// of user journeys in the same project
+			if err := w.Resolver.DB.Raw(`
```
Vadman97, line 634: could we add a span around this SQL in case it gets slow?

backend/worker/worker.go
```diff
@@ -617,6 +619,43 @@ func (w *Worker) processSession(ctx context.Context, s *model.Session) error {
...
+					sum(case when f.url = u.url and f.next_url = u.next_url then count else 0 end) 
+					   * (sum(case when f.url = u.url then count else 0 end))
+						/ sum((case when f.url = u.url then count else 0 end)^2) as normalness
+				from frequencies f
+				inner join user_journey_steps u
+				on f.url = u.url or f.next_url = u.next_url
+				group by session_id, index
+				order by session_id, index) a
+				where a.session_id = ?
+				group by a.session_id
```
Vadman97, line 655: big beautiful query. could we indent it so it's a bit easier to read?
```sql
with frequencies as (select url, next_url, count(*)
                     from user_journey_steps ujs
                              inner join sessions s
                                         on s.id = ujs.session_id
                     where not s.excluded
                       and s.created_at > now() - interval '30 days'
                       and s.project_id = ?
                     group by 1, 2)
select exp(sum(ln(normalness))) as normalness
from (select session_id,
             index,
             sum(case when f.url = u.url and f.next_url = u.next_url then count else 0 end)
                 * (sum(case when f.url = u.url then count else 0 end))
                 / sum((case when f.url = u.url then count else 0 end) ^ 2) as normalness
      from frequencies f
               inner join user_journey_steps u
                          on f.url = u.url or f.next_url = u.next_url
      group by session_id, index
      order by session_id, index) a
where a.session_id = ?
group by a.session_id
```

backend/lambda-functions/journeys/handlers/handlers.go
```diff
@@ -0,0 +1,260 @@
...
+		return true
+	}
+
+	// Replace any non-alphanumeric character with a comma
+	input = nonAlphanumericRegex.ReplaceAllString(input, ",")
+	// Add a comma in front of any capital
+	input = capitalsRegex.ReplaceAllString(input, ",$1")
+	input = strings.ToLower(input)
+	// Split on comma boundaries
+	parts := strings.Split(input, ",")
```
Vadman97, line 86: clever!

backend/lambda-functions/journeys/handlers/handlers.go
```diff
@@ -0,0 +1,260 @@
...
+
+	for _, p := range parts {
+		// If there are no vowels in the word part, assume it's an id
+		if !strings.ContainsAny(p, "aeiouy") {
+			return true
+		}
+
+		// If there are 5 or more consonants in a row, assume it's an id
```
Vadman97, line 94: also very clever

Vadman97 (approved):
🎉

backend/lambda-functions/journeys/main.go
mayberryzane: It's set up so it could be deployed as a lambda, but I'm not using it as one at the moment - just calling the code in the session processing worker

PR #5809, opened 2023-07-03
Revamp panel for viewing an individual network request

ccschmitz:
## Summary

Breaks the network request details out into a new right panel that is rendered on top of the other content on the page.

Screenshot 2023-07-10 at 1 37 36 PM

This also changes how state is managed for the network request, pulling it out of context and putting it in the URL so that you can deep link to a request. Future PRs will add tabs for `Errors` and `Logs` - you'll see those stubbed out in the code.

Also removes the `ariakit` dependency in `frontend/package.json` and uses the `Ariakit` exported from the UI package. This way we can ensure versions of Ariakit don't get out of sync. Ultimately, we shouldn't need to import Ariakit directly in `frontend` once we have more building blocks built out in the UI package.

Local + Reflame preview click test.

N/A


frontend/src/pages/Player/styles.css.ts
```diff
@@ -68,6 +68,7 @@ export const playerCenterColumn = style({
 	flexGrow: 1,
 	position: 'relative',
 	minWidth: MIN_CENTER_COLUMN_WIDTH,
+	zIndex: 0,
```
ccschmitz, line 71: Was causing some layout issues with player elements appearing on top of the network request panel.

mayberryzane (approved):
Looks good in preview! 
I noticed that all fields are displayed by default now - for large network payloads, should we truncate or hide some of what is rendered in the UI with a "show more" button? Rendering gets really slow when I open a 4.7mb request/response: https://www.loom.com/share/f8236ac49f6e4c1caa42b178efb03b96
That's probably a rare case, but I'd imagine it's possible for large enough payloads to crash a user's browser

ccschmitz:
@mayberryzane good catch on the rendering performance issue! I'll look at collapsing those by default.

ccschmitz:
@mayberryzane I addressed the issue of large payloads locking up the page in https://github.com/highlight/highlight/pull/5809/commits/3761f076317daa3832ece8cd214fac020ce1ba49.

mayberryzane:
> @mayberryzane I addressed the issue of large payloads locking up the page in [3761f07](https://github.com/highlight/highlight/commit/3761f076317daa3832ece8cd214fac020ce1ba49) - let me know if you have any thoughts on that hack 😸

sweet lgtm!

PR #5823, opened 2023-07-05
fix OpenSearch indexing when appending both new and old fields in same request

mayberryzane:
## Summary
- a gorm `Create` with a `Returning` clause was not associating the returned ids with the correct records. e.g. if we attempted to create 10 records but only 5 were inserted, the new ids would be set on the first 5 records regardless of which ones were actually inserted
- instead of relying on the `Returning` clause, get the count of rows updated and sort the results of the second query by id to get the latest N rows

- set a watchpoint and confirmed via debugging

- will use the `init-opensearch` Make command to reindex the OpenSearch fields index


et (approved):
We went over this over slack. There's a guarantee that we're only indexing the non-conflicting records since they'll have a higher ID than the conflicting records.

PR #5825, opened 2023-07-05
Retry session insight lambda request on 500 errors

eightants:
## Summary
https://github.com/highlight/highlight/issues/5775

The lambda often returns 500 and 503s, not sure if it's something to do with OpenAI's API being unreliable/inconsistent or the API Gateway timing out. Either way, the backend will now retry up to 5 times on error

### DIFF
--- a/backend/lambda/lambda.go
+++ b/backend/lambda/lambda.go
@@ -49,14 +49,14 @@ func NewLambdaClient() (*Client, error) {
 		return nil, errors.Wrap(err, "error loading lambda credentials")
 	}
 
-	retryClient := retryablehttp.Client{}
+	retryClient := retryablehttp.NewClient()
 	retryClient.RetryMax = 5
 
 	return &Client{
 		Config:              &cfg,
 		Credentials:         &creds,
 		HTTPClient:          &http.Client{},
-		RetryableHTTPClient: &retryClient,
+		RetryableHTTPClient: retryClient,
 	}, nil
 }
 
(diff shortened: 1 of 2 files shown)


jay-khatri:
Do you see any errors in Highlight about this? That might be a good way to debug.

eightants:
> > Do you see any errors in Highlight about this? That might be a good way to debug.
> 
> The error is in the linked issue, #5775 but I don't think it's very helpful.
> 
> @eightants - how do we know retrying will work? Is there nothing in AWS to help us diagnose? We could try logging the body when the response is a 5xx here:
> 
> https://github.com/highlight/highlight/blob/000d64c114dad8bd5c82d71dbf910f4d7b7c8e01/backend/private-graph/graph/resolver.go#L1264-L1271

These are the two main errors we're getting from the endpoint: 
500: https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/http-500-lambda-execution-error.html
503: https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/http-503-service-unavailable.html

The errors are pretty opaque, and happen occasionally for the same exact request, even if it works 90% of the time. 

Through my testing, it works after a retry, and the backend logs show this.

mayberryzane (approved):

et:
> These are the two main errors we're getting from the endpoint:
500: https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/http-500-lambda-execution-error.html
503: https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/http-503-service-unavailable.html

The errors are pretty opaque, and happen occasionally for the same exact request, even if it works 90% of the time.

Through my testing, it works after a retry, and the backend logs show this.

Thanks for the write up @eightants. This definitely seems like a good approach to solving this 👍

PR #5826, opened 2023-07-05
Revert "Normalness scoring during session processing"

mayberryzane:
Reverts highlight/highlight#5795


mayberryzane:
reverting because of an increasing Kafka backlog, "normalness" query may be conflicting with session updates

PR #5833, opened 2023-07-06
Use matching version of antd

lewisl9029:
## Summary

Looks like we were overriding to a version of antd that was different than the one in our lockfile, which caused some subtle CSS differences between Reflame and production that affected tab layout.

Using the version in the lockfile fixes this.

Checked Errors page in Reflame previews.

Before:
image

After:
image


et (approved):
Thanks! Preview looks good.

PR #5837, opened 2023-07-07
add the ability to search by email for error instances

et:

### DIFF
--- a/backend/queryparser/queryparser.go
+++ b/backend/queryparser/queryparser.go
@@ -1,0 +1,19 @@
+package queryparser
+
+import "strings"
+
+func Parse(query string) map[string]string {
+	parsed := map[string]string{}
+
+	queries := strings.Split(query, " ")
+
+	for _, q := range queries {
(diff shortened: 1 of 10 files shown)


backend/queryparser/queryparser.go
```diff
@@ -0,0 +1,19 @@
+package queryparser
+
+import "strings"
+
+func Parse(query string) map[string]string {
```
et, line 5: I extracted this out into a utility so we can use it for various parts of the app (searching logs, error objects, traces?).

It's not as comprehensive as logs search so I didn't update that code path but I figured we can use this as a starting point on how our query language can be re-used across various products.

mayberryzane (approved):

backend/queryparser/queryparser.go
et: FYI @ccschmitz @Vadman97 - if we use the same query language for searching trace logs, it would be great to consolidate here.

PR #5838, opened 2023-07-07
addProperties - skip properties that cannot be serialized

mayberryzane:
## Summary
- in `addProperties`, remove any properties that throw an error when calling `structuredClone`
- these would previously cause the entire `addProperties` call to fail
- we could consider making this check stronger (e.g. only allow `string | boolean | number`) as the `H.track` method signature enforces this already
Screen Shot 2023-07-07 at 11 09 44 AM

- tested by calling `H.track` with invalid properties in prod vs locally
Screen Shot 2023-07-07 at 10 54 35 AM
Screen Shot 2023-07-07 at 10 51 59 AM

- no

### DIFF
--- a/sdk/client/src/index.tsx
+++ b/sdk/client/src/index.tsx
@@ -481,10 +481,20 @@ export class Highlight {
 	}
 
 	addProperties(properties_obj = {}, typeArg?: PropertyType) {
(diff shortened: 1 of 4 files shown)


et (approved):
Nice 👏

PR #5841, opened 2023-07-08
add echo middleware to Go SDK

chancesm:
## Summary

I was wanting something to work on so I felt like adding a middleware for echo for your Go SDK. I did follow the patterns established in the other middleware options while working on this one. This is also one of my first OSS PRs so if you need me to change anything at all, please let me know.

I have not tested this change yet

Nope :smile:


sdk/highlight-go/go.mod
```diff
@@ -6,6 +6,8 @@ require (
 	github.com/99designs/gqlgen v0.17.24
 	github.com/gin-gonic/gin v1.7.7
 	github.com/gofiber/fiber/v2 v2.42.0
+	github.com/highlight-run/highlight/backend v0.0.0-20230707192428-b66e47a33c89
```
Vadman97, line 9: we don't want the SDK to depend on the backend code. this might have been automatically added if you didn't set up go workspaces?

go.work.sum
Vadman97, line 1: try run `go work sync` first at the top level of highlight to make sure the workspace is set up. i think these removals are because of the new self-dependency in `sdk/highlight-go/go.mod`

sdk/highlight-go/middleware/echo/middleware.go
```diff
@@ -0,0 +1,48 @@
...
+			ids := strings.Split(string(highlightReqDetails), "/")
+			if len(ids) >= 2 {
+				ctx = context.WithValue(ctx, highlight.ContextKeys.SessionSecureID, ids[0])
+				ctx = context.WithValue(ctx, highlight.ContextKeys.RequestID, ids[1])
+			}
+
+			span, hCtx := highlight.StartTrace(ctx, "highlight/echo")
+			defer highlight.EndTrace(span)
+
+			c.SetRequest(c.Request().Clone(hCtx))
```
Vadman97, line 35: does this inject `hCtx` as the echo handler's context?

Vadman97 (commented):
sorry we haven't gotten back to you sooner here! code looks great other than a cleanup with the go dependencies. 

could you also add docs for the new middleware? you'll need to change
- https://github.com/highlight/highlight/tree/main/docs-content/getting-started/4_backend-sdk/go
- https://github.com/highlight/highlight/blob/6d643dc458fea223573e90cfb702e710068f8890/highlight.io/components/QuickstartContent/QuickstartContent.tsx#L4
- https://github.com/highlight/highlight/tree/6d643dc458fea223573e90cfb702e710068f8890/highlight.io/components/QuickstartContent/backend/go


it would also be nice if you could add an example echo app using the middleware in https://github.com/highlight/highlight/tree/main/e2e/go

Vadman97:
hey @chancesm , any chance you could take another look at the remaining questions?

chancesm:
Sorry. I've been busy with work and some personal stuff. I'll try to look
at it soon 🙂

sdk/highlight-go/go.mod
chancesm: I see that the main branch now has this dependency. As such, I would like to resolve this conversation point.

sdk/highlight-go/middleware/echo/middleware.go
chancesm: Yes, but there may be a cleaner way to do this. I'll check.

chancesm:
- [x] Add Documentation
- [x] Add Example
- [x] Answer other questions

If there is anything else you would like me to do as part of this PR, please let me know 🙂

Vadman97 (approved):
:ship: thank you for your contribution!

Vadman97:
@chancesm should be good to go, but mind updating the description about how this was tested? could you try running the example you are adding locally to ensure this works correctly as an echo middleware?

chancesm:
@Vadman97 I'll do that this weekend 👍🏻

PR #5844, opened 2023-07-10
Remove obsolete overrides, pin versions for problematic packages

lewisl9029:
## Summary

I just shipped a huge rework to the way we process npm packages (bumping versions should now take closer to 10s for this repo, compared to ~40s before 🥳 ), and it happened to also remove the need for many of the overrides we were using before.

We were also using overrides to pin to specific versions for dependencies that had incompatible versions, but I realized we should probably be pinning those in package.json instead of Reflame overrides to make sure they stay consistent across Vite and Reflame.

In Reflame previews.


et (approved):
Thanks!

PR #5857, opened 2023-07-10
support concurrent preloading of error groups

### LINKED ISSUE #5820: Data loading + loading states on the instance view
@et , in short, it would be nice if we served the data for the first instance of an error group in the high-level loading state. UX wold be much cleaner imo. cc @julian-highlight .

https://www.loom.com/share/41e83222610347e08c200519285b5dbd

Vadman97:
## Summary

Per [slack discussion](https://highlightcorp.slack.com/archives/C02CJANPHQS/p1688682516219369), we want to
increase the concurrent preloading of error groups as they are lightweight and can be preloaded quickly.
* audits the preloading requests made for error groups to make sure the right data is fetched and stored.
* removes the preloading of sessions for error instances as this blows up the amount of data prefetched.
* adds a default mutation refetch cache invalidation to make sure `refetchQueries` are reindexed in indexeddb
* fixes error with error state updates
image

https://www.loom.com/share/fde1fe65ccf24983a251c03181dc709a

Closes #5820


frontend/src/util/preload.ts
```diff
@@ -27,8 +27,11 @@ import { H } from 'highlight.run'
 import moment from 'moment'
 import { useEffect, useRef } from 'react'
 
-const CONCURRENT_PRELOADS = 1
-const PREVIOUS_ERROR_OBJECTS_TO_FETCH = 2
+const CONCURRENT_SESSION_PRELOADS = 1
+const CONCURRENT_ERROR_PRELOADS = 10
+// disable this for now as loading session data for each error instance requires a lot of preloading
+const SHOULD_PRELOAD_ERROR_INSTANCE_SESSION = false
```
et, line 33: Is there a reason we can't just remove the code that prefetches a session for an error instance instead?

frontend/src/util/preload.ts
```diff
@@ -27,8 +27,11 @@ import { H } from 'highlight.run'
 import moment from 'moment'
 import { useEffect, useRef } from 'react'
 
-const CONCURRENT_PRELOADS = 1
-const PREVIOUS_ERROR_OBJECTS_TO_FETCH = 2
+const CONCURRENT_SESSION_PRELOADS = 1
+const CONCURRENT_ERROR_PRELOADS = 10
```
et, line 31: Nice!

frontend/src/util/graph.tsx
```diff
@@ -160,6 +161,11 @@ export const client = new ApolloClient({
 			authLink.concat(splitLink || highlightGraph),
 		),
 	]),
+	defaultOptions: {
+		mutate: {
+			onQueryUpdated: invalidateRefetch,
```
et, line 166: Awesome. This is a lot more robust than what we were doing before.

et (approved):
Nice work!

frontend/src/util/preload.ts
```diff
@@ -333,6 +338,12 @@ const loadErrorGroup = async function (projectID: string, secureID: string) {
 				secure_id: secureID,
 			},
 		})
+		await client.query({
```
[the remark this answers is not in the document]
Vadman97: that's fair, just checked and the loading state for the issue button is very clean. removing this

frontend/src/util/preload.ts
```diff
@@ -374,22 +387,6 @@ const loadErrorGroup = async function (projectID: string, secureID: string) {
 				break
 			}
 		}
-		await client.query({
-			query: GetErrorDistributionDocument,
```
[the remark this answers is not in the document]
Vadman97: great catch! cleaning this up

frontend/src/util/preload.ts
Vadman97: was thinking to leave this here in case we want to enable it in the future, but it's in git anyways so we can always bring this up. thanks for the suggestion; cleaning this up

PR #5859, opened 2023-07-10
support private channel slack integration

Vadman97:
## Summary

See #3497. Now [requested by another customer](https://discord.com/channels/1026884757667188757/1126738180553593026).
Allow Alerts / Comments to message private / DM conversations that the Highlight bot is added to in Slack.
Since the Bot has to be explicitly added in Slack by the members of the private channel, it wouldn't bypass slack's access controls. 
A channel name could only be shown to all highlight users if the bot is erroneously added to a slack channel.

Local deploy shows private channel that bot was added to.
image
image

No.


et (approved):
🥳

PR #5862, opened 2023-07-11
update docs for reportConsoleErrors

et:
## Summary

* Adds `reportConsoleErrors` to the SDK docs
* Updates the default value comment (the behavior was flipped [here](https://github.com/highlight/highlight/pull/4809/files#diff-a9b14a05444fa154939000009533ef701b98ad1a561962e9d9905953bd96b2b8L44) but missed a doc update).

`cd highlight.io && yarn run dev && open http://localhost:4000`


ccschmitz (approved):
Thanks for cleaning this up!

PR #5864, opened 2023-07-11
fix exclusion reason updated too frequently

mayberryzane:
## Summary
- `sessionObj.ExcludedReason != reason` was doing a pointer comparison and always returning true if both were non nil

- debugged locally

- no


Vadman97 (approved):
nice

PR #5866, opened 2023-07-11
normalize more auth errors

et:
## Summary

See [error](https://app.highlight.io/1/errors/0DEF02alRmqVT5dUvuQF2nm76L18?page=1&query=and%7C%7Cerror_state%2Cis%2COPEN%7C%7Cerror-field_timestamp%2Cbetween_date%2C2023-06-11T18%3A34%3A57.039Z_2023-07-11T18%3A34%3A57.039Z) in Highlight.

Similar to #5854, this PR normalizes more auth errors so they can be grouped better.

Auth error should be grouped.

N/A


et:
> to remove the accidental %v?

Sorry, I don't understand -- what is wrong with this?

et:
Will merge. I think the other error is distinct enough that we can do a follow up on this.

Issue #5828, opened 2023-07-06
Hobby deploy script is broken 


Plorenzo:
**Describe the bug**
I tried to deploy Highlight with the `./run-hobby.sh` scrip in both arm and intel machines, it fails to work in both cases.

First, `go` is required to be installed in the host machine but not mentioned on the [docs](https://www.highlight.io/docs/getting-started/self-host/self-hosted-hobby-guide). 
https://github.com/highlight/highlight/blob/d0e197e8fd3326cfc6d5e7b6c6155160cb400772/docker/env.sh#L37

```
+++ go env GOPATH
env.sh: line 37: go: command not found
```
Same for `yarn`
https://github.com/highlight/highlight/blob/d0e197e8fd3326cfc6d5e7b6c6155160cb400772/docker/run-hobby.sh#L10

```
+ yarn dlx wait-on -l -s 3 https://localhost:3000/index.html http://localhost:8080/dist/index.js https://localhost:8082/health
./run-hobby.sh: line 10: yarn: command not found
```

Even after installing both, a few of the containers never become healthy:
```
CONTAINER ID   IMAGE                                         COMMAND                  CREATED              STATUS                        PORTS                                                                                                                                     NAMES
981d73b86db9   ghcr.io/highlight/highlight-backend:latest    "/build/backend -run…"   About a minute ago   Up About a minute (healthy)   0.0.0.0:8082->8082/tcp, :::8082->8082/tcp                                                                                                 backend
9018d8c20ebc   ghcr.io/highlight/highlight-frontend:latest   "/docker-entrypoint.…"   About a minute ago   Up About a minute (healthy)   0.0.0.0:3000->3000/tcp, :::3000->3000/tcp, 0.0.0.0:6006->6006/tcp, :::6006->6006/tcp, 80/tcp, 0.0.0.0:8080->8080/tcp, :::8080->8080/tcp   frontend
afc11caa18b1   confluentinc/cp-kafka:7.3.0                   "/etc/confluent/dock…"   About a minute ago   Up About a minute             0.0.0.0:9092->9092/tcp, :::9092->9092/tcp                                                                                                 kafka
4dc28d531923   confluentinc/cp-zookeeper:7.3.0               "/etc/confluent/dock…"   About a minute ago   Up About a minute             2181/tcp, 2888/tcp, 3888/tcp                                                                                                              zookeeper
[...]

PR #5874, opened 2023-07-11
update docs for hobby networking

Vadman97:
## Summary

As mentioned in #5828, our hobby deploy would not work for EC2 / remote-server deploys because
the frontend did not route correctly to the remote backend.
Confirmed that setting these env. vars solved the issue for this customer.

[Docs preview](https://highlight-landing-k9ikag77h-highlight-run.vercel.app/docs/getting-started/self-host/self-hosted-hobby-guide).

No

### DIFF
--- a/highlight.io/components/QuickstartContent/self-host/self-host.tsx
+++ b/highlight.io/components/QuickstartContent/self-host/self-host.tsx
@@ -9,14 +9,21 @@ export const SelfHostContent: QuickStartContent = {
 		{
 			title: 'Prerequisites',
 			content:
-				'Before we get started, you should have the latest version of [Docker](https://docs.docker.com/engine/install/) (19.03.0+) ' +
+				'Before we get started, you should install [Go](https://go.dev/) (1.20), [Node.js](https://nodejs.org/en) (18), and [yarn](https://yarnpkg.com/getting-started/install) (v3+).' +
+				'You should have the latest version of [Docker](https://docs.docker.com/engine/install/) (19.03.0+) ' +
 				'and [Git](https://git-scm.com/downloads) (2.13+) installed. ' +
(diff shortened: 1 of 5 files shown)


highlight.io/components/QuickstartContent/self-host/self-host.tsx
```diff
@@ -9,21 +9,46 @@ export const SelfHostContent: QuickStartContent = {
...
+				'Before we get started, you should install [Go](https://go.dev/) (1.20), [Node.js](https://nodejs.org/en) (18), and [yarn](https://yarnpkg.com/getting-started/install) (v3+).' +
+				'You should have the latest version of [Docker](https://docs.docker.com/engine/install/) (19.03.0+) ' +
 				'and [Git](https://git-scm.com/downloads) (2.13+) installed. ' +
 				'For a local hobby deploy, we suggest [configuring docker](https://docs.docker.com/desktop/settings/mac/#resources) ' +
 				'to use at least 8GB of RAM, 4 CPUs, and 64 GB of disk space.',
 			code: [
 				{
 					language: 'bash',
-					text: `$ docker --version
+					text: `$ go version
```
SpennyNDaJets, line 20: One thing I wanted to raise about this code snippet is I don't think the copy/paste button functionality will work appropriately for the terminal. But looks like this was the existing case, so not blocking

SpennyNDaJets (approved):
LGTM

highlight.io/components/QuickstartContent/self-host/self-host.tsx
Vadman97: good point, makes sense to break this up into separate code blocks and make the copy icon only copy the bash command. making that change

PR #5877, opened 2023-07-11
add lambdas for session insights digests

### LINKED ISSUE #5721: Build a digest for "session insights" (with and w/o AI insights)
Session Insights email w/o "AI enhanced" session insights
Sessions insights email w/ "AI enhanced" session insights

mayberryzane:
## Summary
- closes #5721 
- creates 3 new lambdas:
  - `getSessionInsightsData` is a Go lambda, retrieves metadata about the top 3 "interesting" sessions for the project
  - `session-insights-email` is a Node.js lambda, renders an HTML email using react-email
  - `sendSessionInsightsEmails` is a Go lambda, sends the HTML email to all non-unsubscribed users in the project

- deployed lambdas + tested in a step function: https://us-east-2.console.aws.amazon.com/states/home?region=us-east-2#/statemachines/view/arn:aws:states:us-east-2:173971919437:stateMachine:TestSessionInsights

- will create a step function to chain these lambda executions


packages/react-email-templates/package.json
```diff
@@ -0,0 +1,23 @@
+{
+	"name": "emails",
+	"version": "0.0.18",
+	"private": true,
+	"workspaces": [
+		".react-email"
```
Vadman97, line 6: what's this for?

packages/react-email-templates/package.json
```diff
@@ -0,0 +1,23 @@
...
+	],
+	"scripts": {
+		"build": "tsc --build ./",
```
Vadman97, line 9: for consistency with the `render` lambda, could we use `tsup` here too?
also do the dependencies get zipped correctly? i found that deps get hoisted up to the top level node_modules unless you add the following
```
	"installConfig": {
		"hoistingLimits": "dependencies"
	},
```

Vadman97 (approved):

packages/react-email-templates/package.json
mayberryzane: this was how it was after the automatic setup https://react.email/docs/getting-started/automatic-setup - there's a .react-email subfolder but everything seems to work ok even if I delete this :)

packages/react-email-templates/package.json
mayberryzane: Earlier I was testing by just building within the `react-email-templates` folder, added this config!

Issue #5780, opened 2023-06-28
Display related websocket events in the right panel


SpennyNDaJets:
Mocks: https://www.figma.com/file/rdBPNzn7Klk6RG1LHUCE5B/Screen-Designs?type=design&node-id=8989-214445&mode=design&t=BcgK5MW0Dh0nQ9eq-0

The websocket data is being stored in Redis/S3, but is not currently being used on the frontend. The data should be retrieved from the web_socket_events_url (if in s3) or redis and displayed in the side panel when clicked.

More info: https://www.notion.so/runhighlight/RFC-Web-Socket-Recording-70b320060816488d954d2d55b5bf9c6a

PR #5881, opened 2023-07-12
Display websocket messages in right panel

eightants:
## Summary

https://github.com/highlight/highlight/issues/5780

Adapted the new right panel to display websocket messages and details

https://www.loom.com/share/12017f2a0fa142c082b9487f7272e6f4

The message panel is using the Virtuoso, but it's weirdly rendering the 1st element for 1 second then renders the rest as shown in the Loom. All the data is present, so loading state is not helping. Not sure why this is happening

Tested out multiple sessions with websockets. Clicked around for the messages. Verified with Chrome dev tools


frontend/src/pages/Player/RightPlayerPanel/components/WebSocketMessages/WebSocketMessages.tsx
```diff
@@ -0,0 +1,167 @@
...
+		? 'Websocket connection has been opened'
+		: event.type === 'close'
+		? 'Websocket connection has been closed'
+		: event.type === 'error'
+		? 'Error'
+		: event.message
+}
+
+const getWebSocketEventSize = (event: any) => {
+	return event.type === 'open' ? '' : event.type === 'close' ? '' : event.size
```
SpennyNDaJets, line 124: Nit: `event.type === 'open' || event.type === 'close' ? "" : event.size`

SpennyNDaJets (changes requested):
Looking good! I think the only case I see here is if there are multiple WebSocket connects, we show all the events. I think we should be able to sort by `socketId` to display the relevant events

Screenshot 2023-07-13 at 11 40 52 AM

SpennyNDaJets (approved):
image

PR #5889, opened 2023-07-12
return a proper err instead of getting a graphql error

### LINKED ISSUE #5818: the requested element is null which the schema does not allow
Enhance the apollo error message with more details. The above error isn't verbose enough.

https://app.highlight.io/1/errors/yk2zrzYteBjhuQQDaizBlL9KiVX5

et:
## Summary

These endpoints should be returning proper errors otherwise there will be a graphql schema issue:

Incorrect:
Screenshot 2023-07-12 at 12 53 01 PM

Correct:
Screenshot 2023-07-12 at 12 53 11 PM

Related to #5818

There may be other graphql schema errors but this should get some of them.

N/A


SpennyNDaJets (approved):
Makes sense to me, but curious if someone may have more context here

et:
> Makes sense to me, but curious if someone may have more context here

There's really no reason to return `nil, nil` in Go. Either we have the data and a nil error or we have nil data and a non-nil error. We actually looked into a lint rule [here](https://github.com/highlight/highlight/pull/5592#issuecomment-1583574619) but we need to spend a little more effort on it to ensure it doesn't produce false positives.

PR #5894, opened 2023-07-12
cleanup some errors

Vadman97:
## Summary

* fixes undefined access to chunk_index when a chunk does not exist
* ensure error instance query returns nil instead of 0 for next/previous ids when they do not exist

image
image

No

### DIFF
--- a/frontend/src/pages/Player/PlayerHook/PlayerHook.tsx
+++ b/frontend/src/pages/Player/PlayerHook/PlayerHook.tsx
@@ -346,7 +346,7 @@ export const usePlayer = (): ReplayerContextInterface => {
 			const lastChunkIdx =
 				eventChunksData?.event_chunks[
 					eventChunksData?.event_chunks.length - 1
-				].chunk_index
+				]?.chunk_index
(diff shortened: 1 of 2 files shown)


frontend/src/pages/Player/PlayerHook/PlayerHook.tsx
```diff
@@ -346,7 +346,7 @@ export const usePlayer = (): ReplayerContextInterface => {
 			const lastChunkIdx =
 				eventChunksData?.event_chunks[
 					eventChunksData?.event_chunks.length - 1
-				].chunk_index
+				]?.chunk_index
```
et, line 349: Kind of surprised typescript didn't complain here.

et (approved):
Thanks!

frontend/src/pages/Player/PlayerHook/PlayerHook.tsx
Vadman97: yeah, don't think typescript handles array out of bounds, though i think this only is out of bounds if `eventChunksData?.event_chunks.length` is 0

PR #5898, opened 2023-07-12
replace first with take

Vadman97:
## Summary

GORM's `Take` is essentially a `Limit(1).Find()` for one item, where if no item is found, an error is returned.
In most places, this makes more sense than our use of `First()` because we do not actually care about the ordering.
`First()` always adds a `ORDER BY` clause on the primary key, causing us to miss indexes in some situations
(we've had to manually bypass this for querying sessions by secure id). This cleans up the code and
makes it consistent for faster loading of error groups. 

Testing locally. https://www.loom.com/share/faaf23aafa5049fdb63aa08f06ef50e4

before
```sql
SELECT *
FROM "error_groups"
WHERE "error_groups"."secure_id" = 'O5am2YlVI3A6Cj8V9sUIgzDxTjkl'
ORDER BY "error_groups"."id"
LIMIT 1;
```
after
```sql
SELECT * 
FROM "error_groups" 
WHERE "error_groups"."secure_id" = 'D8Bon2X2pbnA0TCSTUO0sr9zaPeM' 
LIMIT 1
```

some new instrumentation on the influx part of the loading
image

Will monitor deployment for errors. The risky part is that changing from `First()` to `Take()` means the `.Error` will return something now if no results are found, while `First()` would return `nil`. I've tried to vet the changes to make sure that's ok (and make according adjustments where necessary to check the error type), but let me know if I missed any.

### DIFF
--- a/backend/hubspot/hubspot.go
+++ b/backend/hubspot/hubspot.go
@@ -393,11 +393,11 @@ func (h *HubspotApi) CreateContactCompanyAssociationImpl(ctx context.Context, ad
 	}
 
 	admin := &model.Admin{}
-	if err := h.db.Model(&model.Admin{}).Where("id = ?", adminID).First(&admin).Error; err != nil {
+	if err := h.db.Model(&model.Admin{}).Where("id = ?", adminID).Take(&admin).Error; err != nil {
 		return err
 	}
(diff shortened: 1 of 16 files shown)


mayberryzane (approved):
neat!

PR #5901, opened 2023-07-12
update excluded / excluded reason in opensearch w/ sql update

mayberryzane:
## Summary
- if a session was being excluded for an ignored user, its excluded flag / reason was updated in postgres but not updated in opensearch
- these sessions would keep showing as live since excluded sessions are skipped by the session processing worker

- local click test

- no


Vadman97 (approved):
nice catch! should we run a script to update excluded live sessions accordingly to exclude them in opensearch?

PR #5903, opened 2023-07-13
set a "normalness" score during session processing

mayberryzane:
## Summary
- this fixes an issue in #5795 that was causing locking on the sessions table (see https://github.com/highlight/highlight/commit/05416d0b3d44bc9ce05244fc2dfb695d4cc7c9d3 for changes since that PR)
- adding `project_id` and `created_at` columns to the `user_journey_steps` so we can query without joining on the sessions table

- tested locally

- no, will monitor kafka backlog and session processing to make sure this change is performant


Vadman97 (approved):
lgtm per #5795

PR #5905, opened 2023-07-13
Setting up Edge Function test with Cloudflare worker.

deltaepsilon:
The Cloudflare SDK does not work in the Next/Vercel Edge runtime. We're out of luck there, but we can still use the test harness that I built.

The only path forward at this point is to try to reverse-engineer the [RichiCoder1](https://github.com/RichiCoder1/opentelemetry-sdk-workers) lib that we're using for Cloudflare and try to strip it down for the Edge runtime.

image

## How to test

Run `yarn e2e:nextjs` to spin up the Next.js testbed at `http://localhost:3005/`. Click the "Edge Function: *" buttons to create successful or errorful results.

*******************
@Vadman97 and I figured out that the Node SDK was broken due to some dependency hoisting limits, so we removed those limits.

I also discovered that we need a `src/instrumentation.ts` file in addition to `instrumentation.ts`.

image


sdk/highlight-cloudflare/src/sdk.ts
```diff
@@ -63,8 +63,17 @@ export const H: HighlightInterface = {
 				['highlight.trace_id']: requestID,
 			}),
 		})
+
 		for (const m of RECORDED_CONSOLE_METHODS) {
-			console[m] = sdk.logger[m]
+			const originalConsoleMethod = console[m]
+
+			console[m] = (message: string, ...args: unknown[]) => {
```
deltaepsilon, line 70: I noticed that once I'd called `H.init`, I never saw anything in my console, ever again.

This logic preserves the original console methods and calls them alongside the `sdk.logger`.

e2e/nextjs/src/app/api/edge-test/route.ts
```diff
@@ -0,0 +1,45 @@
...
+		throw error
+	}
+}
+
+export const POST = async function POST(
+	request: NextRequest,
+	context: NextFetchEvent,
+) {
+	console.log('initializing cloudflare POST')
+	H.init(request, env, context)
```
Vadman97, line 37: per our pairing, the `context` needs to patch `waitUntil`

Vadman97 (approved):
mind also making that `sdk/highlight-node/package.json` change here to remove
```
	"installConfig": {
		"hoistingLimits": "dependencies"
	},```

**Comment by deltaepsilon** [comment]
> thanks again for tackling this! mind also making that `sdk/highlight-node/package.json` change here to remove
> 
> ```
> 	"installConfig": {
> 		"hoistingLimits": "dependencies"
> 	},```
> ```

Will do. I'll also version bump.

File: sdk/highlight-node/CHANGELOG.md [hunk 3]
```diff
@@ -55,3 +55,9 @@
 
 - Adds a `Handlers.serverlessFunction` for use as a error wrapper in AWS Lambda.
 - Adds a `H.stop()` method for shutting down the SDK and flushing unsent data.
+
+## 3.1.1
```
Vadman97, line 59: should be 3.1.2

Vadman97 (approved):
as you mentioned, apollo and nest (things that depend on node) might need a bump

sdk/highlight-node/CHANGELOG.md
deltaepsilon: Does this look right?

https://github.com/highlight/highlight/pull/5905/commits/b2d95ed381f0f58b59653ebc281847e5b7f643ba

I figured I'd bump everthing to v3.1.2 to match `@highlight-run/node`.

PR #5928, opened 2023-07-13
build frontend to be more similar to render

Vadman97:
## Summary

Based on [slack thread](https://highlightcorp.slack.com/archives/C02N2AZS8BV/p1689260173138899), we've seen 
a number of errors for our project in highlight having an invalid stack trace when enhanced. The root cause
seems to be that our sourcemap uploader uploads the `.js` and `.map` files from the github actions build,
but the app is hosted from a different build done by render.com. Since the code differs in the inlined environment,
the sourcemap lines/columns do not match up resulting in incorrect stack traces.

Diffing the s3 sourcemap-uploaded index.js file against the prod app.highlight.run index.js
and comparing that to a local build with these changes.

No, will monitor future sourcemaps on errors in our project.

### DIFF
--- a/.github/workflows/turbo.yml
+++ b/.github/workflows/turbo.yml
@@ -95,7 +95,7 @@ jobs:
               run: yarn publish:client
 
             - name: Upload frontend sourcemaps
-              if: github.ref == 'refs/heads/main'
+              #              if: github.ref == 'refs/heads/main'
               run: yarn && yarn sourcemaps:frontend
               env:
                   HIGHLIGHT_API_KEY: ${{ secrets.HIGHLIGHT_SOURCEMAP_API_KEY }}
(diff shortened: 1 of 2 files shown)


SpennyNDaJets (approved):
LGTM

.github/workflows/turbo.yml
```diff
@@ -64,12 +61,16 @@ jobs:
                   aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
                   aws-region: us-east-2
 
+            - name: Install Doppler CLI
+              uses: dopplerhq/cli-action@v2
+
             - name: Build & test
-              run: yarn test:all
+              run: doppler run -- bash -c 'RENDER_PREVIEW=false yarn test:all'
```
Vadman97, line 68: despite the render preview end, override `RENDER_PREVIEW=false` so that sourcemaps could be generated

PR #5930, opened 2023-07-13
improve old commandbar performance

Vadman97:
## Summary

@jay-khatri recently noticed a stutter when loading app.highlight.io
Turns out, this was due to some logic in the old command bar for processing suggested workspaces.
image
The change removes the workspace grouping logic since we now auto-populate project names in the about you form.

[Reflame preview](https://preview.highlight.io/1/sessions?page=1&query=and%7C%7Ccustom_processed%2Cis%2Ctrue%7C%7Ccustom_created_at%2Cbetween_date%2C2023-06-13T22%3A49%3A05.448Z_2023-07-13T22%3A49%3A05.448Z&%7Er_preview=%7B%22action%22%3A%22start%22%2C%22mode%22%3A%22production%22%2C%22region%22%3A%22sjc%22%2C%22variantId%22%3A%2201H58Q09N1S5MAPGSB47P8328N%22%2C%22variantData%22%3A%22%7B%5C%22type%5C%22%3A%5C%22branch%5C%22%2C%5C%22branch%5C%22%3A%5C%22vadim%2Ffix-load-perf%5C%22%2C%5C%22githubOwnerName%5C%22%3A%5C%22highlight%5C%22%2C%5C%22githubRepositoryName%5C%22%3A%5C%22highlight%5C%22%7D%22%7D)
no commandbar slowness in preview
image

No

### DIFF
--- a/backend/private-graph/graph/generated/generated.go
+++ b/backend/private-graph/graph/generated/generated.go
@@ -920,7 +920,6 @@ type ComplexityRoot struct {
 		WorkspaceInviteLinks         func(childComplexity int, workspaceID int) int
 		WorkspacePendingInvites      func(childComplexity int, workspaceID int) int
 		WorkspaceSettings            func(childComplexity int, workspaceID int) int
-		WorkspaceSuggestion          func(childComplexity int, query string) int
 		Workspaces                   func(childComplexity int) int
 		WorkspacesCount              func(childComplexity int) int
 	}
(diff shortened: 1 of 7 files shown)


et (approved):
Nice find. Looks like we can also get rid of the workspaceSuggestion field since it's no longer being used.

https://github.com/highlight/highlight/blob/53a7e6e8543ca5eee65ad0366fb448b42313a0b5/backend/private-graph/graph/schema.resolvers.go#L5812-L5821

PR #5932, opened 2023-07-14
Add logs to network request view

ccschmitz:
## Summary

Adds a `Logs` tab to the network request details panel.

Screenshot 2023-07-14 at 11 21 28 AM

Also refactors some of the tab components to their own files.

Local and PR preview click testing. An easy way to find a network request with some associated logs is to check out the [PR preview](https://frontend-pr-5932.onrender.com/1/sessions/41afLzfRoLuTkjCcI0KqiyvDxhEX?network-resource-id=40&page=1&query=and%7C%7Ccustom_processed%2Cis%2Ctrue%7C%7Ccustom_created_at%2Cbetween_date%2C2023-06-14T15%3A51%3A23.588Z_2023-07-14T15%3A51%3A23.588Z) and filter to requests for `MarkErrorGroupAsViewed`, which is what I did in the screenshot above ☝️ 

This is a UI-only change so there isn't much to monitor, but will keep an eye on sessions where people are interacting with the network resource panel.


jay-khatri:
image

> To match backend logs to network request 

should be

> To match backend logs to network requests,

ccschmitz:
@jay-khatri thanks for catching that - fixed in https://github.com/highlight/highlight/pull/5932/commits/84970c441aef48694fb2c54f617b24d84ed6f010!

frontend/src/pages/Player/RightPlayerPanel/components/NetworkResourcePanel/NetworkResourcePanel.tsx
SpennyNDaJets, line 1: 🙏 for breaking this down

SpennyNDaJets (commented):
Code looks good to me. Left a screenshot with a couple comments/questions from click testing, and wanted to raise before approving

Screenshot 2023-07-14 at 11 31 57 AM

SpennyNDaJets:
> @SpennyNDaJets yeah, good callout... there is no request ID for that request so it's not applying a trace ID filter. Maybe a better treatment would be to just say that there are no logs for the request. What do you think?

I think that sounds good! Seems a little confusing to be showing all those logs

ccschmitz:
@SpennyNDaJets pushed an update to address that feedback in 08eff9a7e508ecb309cbf804d203dd65c8ee34ba. Thanks for the review!

SpennyNDaJets (approved):
LGTM

PR #5941, opened 2023-07-14
prevent jumping in viewport in the listview

et:
## Summary

The error instances listview jumps when we are loading the page. This PR sets a min height on the listview to prevent the jumping. I tried figuring out a different solution that doesn't require a fixed value like this but nothing really seemed to do the job.

There is now a stable height on the loading and table view.

Screenshot 2023-07-14 at 3 25 04 PM

Screenshot 2023-07-14 at 3 25 26 PM

N/A


jay-khatri:
Just to confirm, the fixed height we decided on here is the size as if the list were full?

et:
Correct

PR #5950, opened 2023-07-16
Revert workaround for minified umd package

lewisl9029:
## Summary

Just shipped an update to our exports detection for several common minified UMD module patterns. This means we should see errors like `module does not have named export` and `minified react error #130`(often caused by the entire module namespace being dumped into the default export, causing us to use an object as a jsx component instead of an actual component) much less often.

It also means we can remove the workaround we had in place for `react-collapsible` which uses a minified umd module format that will now get properly transformed into proper esm with named exports.

Click into error details page using react-collapsible, no longer crashes.


et (approved):
Confirmed in the preview that the errors page works now. Thanks.

PR #5963, opened 2023-07-17
wildcard searching for email in error instances view

et:
## Summary

Adjusts the logic for searching an error instance's email to use a case insensitive, wildcard search. 
This also highlights what part of the email matched.

_Note: this makes the query slower since it won't utilize an index but we opted for usability over performance ([slack](https://highlightcorp.slack.com/archives/C02N2AZS8BV/p1689273119805889?thread_ts=1689185062.220409&cid=C02N2AZS8BV))._

https://www.loom.com/share/3a22bac77fc6480c8ac5c3dc163d28b3

N/A

### DIFF
--- a/frontend/src/pages/ErrorsV2/ErrorInstances/ErrorInstances.tsx
+++ b/frontend/src/pages/ErrorsV2/ErrorInstances/ErrorInstances.tsx
@@ -35,7 +35,7 @@ export interface SearchFormState {
 }
 
 export const ErrorInstances = ({ errorGroup }: Props) => {
-	const [currentSearchedEmail, setCurrentSearchEmail] = React.useState('')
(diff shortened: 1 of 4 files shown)


frontend/src/pages/ErrorsV2/ErrorInstances/ErrorInstances.tsx
```diff
@@ -35,6 +35,7 @@ export interface SearchFormState {
 }
 
 export const ErrorInstances = ({ errorGroup }: Props) => {
+	const [currentSearchEmail, setCurrentSearchEmail] = React.useState('')
```
et, line 38: Making a separate state (instead of relying on form) because we only want the highlighted text to change after the form is submitted (instead of as the user is typing).

deltaepsilon (approved):

PR #5974, opened 2023-07-17
set up services data model

et:
## Summary

Sets up the services table model per https://www.notion.so/runhighlight/GitHub-enriched-stack-traces-Product-Spec-ac7d8fda396848b3a207a4ac5e61b8a5?pvs=4#15b7a08d9d7f46f9a5ee2cf651ddb52b

We'll add some additional columns over time but this is the bare minimum of required fields.

Currently unused but tests added for fetching a service.


SpennyNDaJets (approved):
LGTM

PR #5978, opened 2023-07-17
simplify loading state for the error instance page

et:
## Summary

This PR simplifies the loading state for loading the error instance data. I also added an error state if we fail to load the data (previously we just returned `null`).

Before:
Screenshot 2023-07-17 at 3 23 56 PM

After:
Screenshot 2023-07-17 at 3 22 02 PM

Error state
Screenshot 2023-07-17 at 3 20 08 PM

N/A


ccschmitz (approved):
✨

PR #5983, opened 2023-07-17
improve performance of session loading

Vadman97:
## Summary

Improve performance of loading sessions.
On the frontend:
* parallelizes graphql operations responsible for loading parts of the session.
On the backend:
* improves performance of the backend queries loading parts of the session by avoiding concurrent session queries.
    * ie. `GetSessionPayload` queries `errors` `rage_clicks` and `session_comments`, all of which make a db query to the `session`.
    * this PR introduces a redis + LRU in-memory caching layer to the query so that multiple follow-up queries can reuse the database query result. 
Fixes a bug with hubspot that failed to create contacts because the redis locking logic would error on timeouts.

Local deploy and reflame preview for frontend.

Inspecting load times for `GetSessionPayload` (among other graphql operations responsible for loading a session)

Before:
image

After:
image

No

### DIFF
--- a/backend/hubspot/hubspot.go
+++ b/backend/hubspot/hubspot.go
@@ -382,7 +382,9 @@ func (h *HubspotApi) CreateContactCompanyAssociation(ctx context.Context, adminI
 
 func (h *HubspotApi) CreateContactCompanyAssociationImpl(ctx context.Context, adminID int, workspaceID int) error {
 	key := fmt.Sprintf("hubspot-association-%d-%d", adminID, workspaceID)
-	if acquired := h.redisClient.AcquireLock(ctx, key, ClientSideCreationPollInterval); acquired {
+	// wait for up to 5 seconds in case another worker is creating the same association
+	// we don't expect the action to take longer than 5 seconds
+	if acquired := h.redisClient.AcquireLock(ctx, key, 5*time.Second); acquired {
 		defer func() {
 			if err := h.redisClient.ReleaseLock(ctx, key); err != nil {
 				log.WithContext(ctx).WithError(err).WithField("url", key).Error("failed to release hubspot association lock")
--- a/backend/hubspot/hubspot.go
+++ b/backend/hubspot/hubspot.go
@@ -382,9 +382,7 @@ func (h *HubspotApi) CreateContactCompanyAssociation(ctx context.Context, adminI
 
 func (h *HubspotApi) CreateContactCompanyAssociationImpl(ctx context.Context, adminID int, workspaceID int) error {
 	key := fmt.Sprintf("hubspot-association-%d-%d", adminID, workspaceID)
-	if err := h.redisClient.AcquireLock(ctx, key, ClientSideCreationPollInterval); err != nil {
-		return e.New("failed to acquire hubspot association lock")
-	} else {
+	if acquired := h.redisClient.AcquireLock(ctx, key, ClientSideCreationPollInterval); acquired {
 		defer func() {
 			if err := h.redisClient.ReleaseLock(ctx, key); err != nil {
 				log.WithContext(ctx).WithError(err).WithField("url", key).Error("failed to release hubspot association lock")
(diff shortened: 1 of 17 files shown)


backend/redis/utils.go
```diff
@@ -480,24 +480,23 @@ func (r *Client) GetHubspotCompanies(ctx context.Context, companies interface{})
...
 			TTL:  timeout,
 			Mode: "NX",
 			Get:  true,
 		})
 		// error means value is not set
 		if cmd.Err() != nil {
-			return nil
+			return true
+		}
+		if time.Since(start) > timeout {
```
SpennyNDaJets, line 495: 🔥

backend/store/error_groups.go
```diff
@@ -27,8 +27,7 @@ func (store *Store) ListErrorObjects(errorGroup model.ErrorGroup, params ListErr
 
 	var errorObjects []model.ErrorObject
 
-	query := store.db.Debug().
```
SpennyNDaJets, line 30: Thanks for catching this

backend/hubspot/hubspot.go
```diff
@@ -382,9 +382,7 @@ func (h *HubspotApi) CreateContactCompanyAssociation(ctx context.Context, adminI
 
 func (h *HubspotApi) CreateContactCompanyAssociationImpl(ctx context.Context, adminID int, workspaceID int) error {
 	key := fmt.Sprintf("hubspot-association-%d-%d", adminID, workspaceID)
-	if err := h.redisClient.AcquireLock(ctx, key, ClientSideCreationPollInterval); err != nil {
-		return e.New("failed to acquire hubspot association lock")
-	} else {
+	if acquired := h.redisClient.AcquireLock(ctx, key, ClientSideCreationPollInterval); acquired {
```
SpennyNDaJets, line 385: Not blocking since pre-existing - I was a little confused by the the term `PollInterval` part of the variable here since it seems to be used as a timeout value instead. Just want to make sure we want to be used here in sync with its other usages.

i.e. if we change the interval in `pollHubspot` that we are okay with it changing here as well

backend/store/sessions.go
```diff
@@ -1,15 +1,58 @@
...
+			return
+		}
+	}
 
-	return session, err
+	return
+}
+
+func (store *Store) GetSessionFromSecureID(ctx context.Context, secureID string) (*model.Session, error) {
+	return cachedEval(ctx, store.redis, fmt.Sprintf("session-secure-%s", secureID), 150*time.Millisecond, time.Second, func() (*model.Session, error) {
```
SpennyNDaJets, line 41: I like this, and just want to make sure my understanding here is correct

We try the cache, if we get a hit return that value. If miss or timeout trying to get lock, we execute the code below?

frontend/src/pages/Player/PlayerHook/PlayerHook.tsx
```diff
@@ -315,6 +314,43 @@ export const usePlayer = (): ReplayerContextInterface => {
 		[getChunkIdx, showPlayerMouseTail, state.sessionMetadata.startTime],
 	)
 
+	const loadEventChunk = useCallback(
+		async (_i: number) => {
+			try {
+				const url = await fetchEventChunkURL({
```
SpennyNDaJets, line 320: Not blocking since pre-existing- noticed above that `fetchEventChunkURL` uses `rawFetchEventChunkURL` which hardcodes `skip: true` - just wanted to make sure this was expected

Saw that we cleaned up a lot of calls to `fetchEventChunkURL`

SpennyNDaJets (approved):
Code and click test look good to me! Left a couple non-blocking comments

Screenshot 2023-07-18 at 10 30 51 AM

backend/hubspot/hubspot.go
Vadman97: great point... this should just be a different short time constant and has nothing to do with the poll interval. updating!

backend/store/sessions.go
Vadman97: yup exactly! adding docs to the `cachedEval` function.

frontend/src/pages/Player/PlayerHook/PlayerHook.tsx
Vadman97: this is expected - that `skip` means we never call the query when we actually instantiate the hook `useGetEventChunkUrlQuery`. that hook just exposes a `{ refetch: rawFetchEventChunkURL }` refetch fn that we later use, and calling the refetch method makes the query.

Issue #5947, opened 2023-07-15
'Refresh suggestion' button on AI Error Suggestion loading state doesn't represent data fetching status. 


eightants:
**Describe the bug**
The 'refresh suggestion' button on AI Error Suggestions only loads for a split second, then data is displayed. But this data is the same as the data before refetch. The actual refetched data only updates after 5 seconds. This doesn't match the loading state. 

First discussed here: https://github.com/highlight/highlight/pull/5902#discussion_r1262894335

Checked the loading, networkStatus states and they match, it only loads for like 1 second. I'm assuming this is cause of caching though, since there's no way the API is returning so quick. It looks like it returns data from cache (triggering onComplete, so we can't set loading state through that either) and then that cached data is updated like 5 seconds later.

Tried updating fetchPolicy to no-cache or network-only, but doesn't seem to be doing anything. Could be something to investigate in another PR.

**To Reproduce**

1. Go to 'Errors'
2. Click on 'Get suggestion'
3. Scroll down to 'Refetch suggestion'
4. See that the loading state is instant, but data isn't updated, the data changes 5-10 seconds after `loading == false`

**Expected behavior**

On 'Refetch suggestion' the loading state persists until the actual refetch data is returned. Loading should take 5-10 seconds considering OpenAI API speeds. 

**Screenshots**

**Additional context**

PR #5984, opened 2023-07-18
Handle state switching for AI Suggestions

eightants:
## Summary

Fixed issues mentioned here:
 https://www.loom.com/share/c9c25313814548ee9f0cd312c478c8d8

Main issues I ran into involve LazyQuery, where after running LazyQuery once it behaves like useQuery. When the variables change it automatically refreshes. Added a state to check if the current error is the same error as the one AI Suggestions are requested for. 

Fixes https://github.com/highlight/highlight/issues/5947

### DIFF
--- a/frontend/src/pages/ErrorsV2/ErrorInstance/AiErrorSuggestion.tsx
+++ b/frontend/src/pages/ErrorsV2/ErrorInstance/AiErrorSuggestion.tsx
@@ -13,7 +13,7 @@ import {
 } from '@highlight-run/ui'
 import { message } from 'antd'
 import clsx from 'clsx'
-import { useState } from 'react'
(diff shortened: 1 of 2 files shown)


et (commented):
Hey @eightants - I experienced a similar issue with apollo's lazy query (see #5127) and found this issue ([#5912](https://github.com/apollographql/apollo-client/issues/5912)) that it's expected behavior.

Instead of having to track state like this, what do you think about just calling the client directly (https://github.com/apollographql/apollo-client/issues/5912#issuecomment-797060422)

eightants:
Yeah that makes sense. I can switch to doing that instead.

et (approved):
Awesome. Thank you!

PR #5989, opened 2023-07-18
activity graph lambda

mayberryzane:
## Summary
- given an events count array (e.g. "[1, 3, 6, 3, 2, 2, ...]"), return a base64 png for the activity graph, to be included in the email
- uses puppeteer to screenshot the SVG and save as a PNG, so I've included it in the `render` package and created a new lambda in AWS - the entry point for the lambda is `activity.handler`

- tested locally and viewed the output png:
test

- will add this to the session insights digest email, there's a placeholder rn

### DIFF
--- a/render/package.json
+++ b/render/package.json
@@ -7,8 +7,9 @@
 	"module": "./dist/index.js",
 	"scripts": {
 		"build": "tsup",
-		"test": "DEV=true node dist/index.js",
+		"test:screenshots": "DEV=true node dist/index.js",
 		"test:activity": "DEV=true node dist/activity.js",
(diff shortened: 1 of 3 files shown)


render/package.json
```diff
@@ -8,9 +8,12 @@
 	"scripts": {
 		"build": "tsup",
 		"test": "DEV=true node dist/index.js",
+		"test:activity": "DEV=true node dist/activity.js",
```
Vadman97, line 11: not confident this will be run by the `yarn test:render` from the top `package.json`

Vadman97 (approved):
nice!

PR #5992, opened 2023-07-18
Remix Support

deltaepsilon:
First draft of Remix support.

Test Highlight dashboard: https://app.highlight.io/8481

### Notes

- Remix requires importing `firstload` as `index.cjs` instead of `index.cjs.js`
- Server instrumentation appears to be trivial.
- Client instrumentation requires a `useEffect` just like Next.js, but that's it.

image

image

image

image

image


sdk/firstload/package.json
```diff
@@ -41,8 +41,8 @@
 	"exports": {
 		"types": "./dist/firstload/src/index.d.ts",
 		"import": "./dist/index.es.js",
-		"require": "./dist/index.cjs.js",
-		"node": "./dist/index.cjs.js",
+		"require": "./dist/index.cjs",
```
deltaepsilon, line 44: Remix did NOT like importing `index.cjs.js`. It demanded `index.cjs` instead.

This change does not appear to break our Next.js implementation. I'm inclined to believe that switching to `index.cjs` is the right move.

deltaepsilon:
- [x] Client-side session replay + error monitoring
- [ ] Server-side wrappers

This is looking more complicated than I'd hoped. I'm investigating the Remix server-side runtime now to determine a path forward.

deltaepsilon:
We made great progress today. Waiting on a fix from @opentelemetry to clean up our imports.

https://github.com/open-telemetry/opentelemetry-js/issues/3954

deltaepsilon:
I split it into three exports:

1. `@highlight-run/remix/client`
2. `@highlight-run/remix/server`
3. `@highlight-run/remix/report-dialog`

I've updated the docs and the screenshots at the top of this page ☝️

sdk/firstload/src/index.tsx
```diff
@@ -2,6 +2,7 @@ import {
 	AmplitudeAPI,
 	setupAmplitudeIntegration,
 } from './integrations/amplitude'
+import { SESSION_STORAGE_KEYS } from '@highlight-run/client/src/utils/sessionStorage/sessionStorageKeys'
```
Vadman97, line 5: i think you need to update https://github.com/highlight/highlight/blob/32faa2a2ad782369c8c11ebeef367e574974a4ed/.reflame.config.jsonc#L245-L252
to fix reflame, cc @lewisl9029

sdk/firstload/src/index.tsx
```diff
@@ -75,9 +79,26 @@ const H: HighlightPublicInterface = {
...
 			}
 
+			let previousSession = getPreviousSessionData()
+			let sessionSecureID = GenerateSecureID()
+			if (previousSession?.sessionSecureID) {
+				sessionSecureID = previousSession.sessionSecureID
+			} else {
+				const sessionData: SessionData = {
+					...previousSession,
+					projectID: +projectID,
```
Vadman97, line 89: i'm going to use this trick more :)

sdk/highlight-remix/package.json
```diff
@@ -0,0 +1,50 @@
...
+			"require": "./dist/report-dialog.js"
+		}
+	},
+	"scripts": {
+		"typegen": "tsup --dts-only",
+		"dev": "tsup --watch",
+		"build": "tsup"
+	},
+	"peerDependencies": {
+		"highlight.run": ">=3.0.0",
```
Vadman97, line 37: doesn't this need to be `>=7.4.0` to get the new `init` return behavior?

sdk/highlight-remix/package.json
```diff
@@ -0,0 +1,50 @@
+{
+	"name": "@highlight-run/remix",
+	"version": "0.0.0",
```
Vadman97, line 3: not sure what the convention is here, but i've only seen `0.0.0` for unpublishable packages. should this be `0.1.0`?

sdk/highlight-remix/package.json
```diff
@@ -0,0 +1,50 @@
...
+	"description": "Client for interfacing with Highlight in Remix",
+	"packageManager": "yarn@3.5.0",
+	"author": "",
+	"license": "ISC",
```
Vadman97, line 7:
```suggestion
	"license": "Apache-2.0",
```

sdk/highlight-remix/src/client.tsx
```diff
@@ -0,0 +1,30 @@
...
+			}
+
+			if (sessionSecureID) {
+				Cookies.set(
+					SESSION_STORAGE_KEYS.SESSION_SECURE_ID,
+					sessionSecureID,
+				)
+			}
+		}
+	}, []) // eslint-disable-line react-hooks/exhaustive-deps
```
Vadman97, line 27: should this rerun if `projectID` or `highlightOptions` change?

Vadman97 (commented):
gargantuan effort! leaving some thoughts but looks good overall!

.github/workflows/turbo.yml
Vadman97, line 1: when we are ready to deploy this, we might need to do a step to bootstrap the new SDK. can pair on this

sdk/firstload/src/index.tsx
lewisl9029: Ah yep, I added automatic entry point detection for npm packages, but not for libraries yet. The entry point mapping for libraries is a bit trickier since we have cases like this where entry point names and target locations don't quite match up 1:1.

I've been thinking about how best to address this and I think making libraries act more like workspace packages might be the best way forward in the long term (for instance, making paths relative to the root package directory instead of source (/src) subdirectory), both for compatibility with existing code that makes use of workspaces and to eventually make it possible to auto-configure libraries based on existing workspace configs.

jay-khatri (changes requested):
Can we add remix to this page as well? 
https://www.highlight.io/frameworks

sdk/firstload/package.json
deltaepsilon: It worked!

highlight.io/components/QuickstartContent/frontend/remix.tsx
```diff
@@ -0,0 +1,202 @@
...
+# with pnpm
+pnpm add @highlight-run/remix
+				`,
+					language: 'bash',
+				},
+			],
+		},
+		{
+			title: 'Initialize the client SDK.',
+			content: `Grab your project ID from [app.highlight.io/setup](https://app.highlight.io/setup), inject it into your client application using the \`loader\` export from your \`root.tsx\` file, and set the \`projectID\` in the \`<HighlightInit/>\` component.
```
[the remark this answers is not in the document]
deltaepsilon: Good consideration.

The env var methodology is the most idiomatic, but it's also arbitrary. Some apps will be fine directly setting the prop as a string.

I'm inclined to leave it.

highlight.io/components/QuickstartContent/frontend/remix.tsx
```diff
@@ -0,0 +1,202 @@
...
+		return (
+			<div>
+				<script src="https://unpkg.com/highlight.run"></script>
+				<script
+					dangerouslySetInnerHTML={{
+						__html: \`
+							H.init('\${process.env.HIGHLIGHT_PROJECT_ID}');
+						\`,
+					}}
+				/>
```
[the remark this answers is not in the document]
deltaepsilon: This drove me crazy.

The `ErrorBoundary` component is 100% server rendered. There's no client-side React involved. The only way to call `H.init` is to load it with the script tag and inject the `H.init` call manually.

The `dangerouslySetInnerHTML` trick with the script tag is, surprisingly enough, copied directly off of the Remix docs. `dangerouslySetInnerHTML` is used often enough in Reactland and is understood to be safe when NOT injecting user-generated data.

sdk/firstload/src/index.tsx
deltaepsilon: Thanks for the tip! Fixing now.

sdk/firstload/src/index.tsx
deltaepsilon: There are so many of these stupid tricks. For example, you can use `!!~someArrayIndex` to test if `someArrayIndex` is `-1`. The `~` sign is the bitwise NOT operator that just so happens to shift `-1` to `0`. It's typically used to test if `Array.findIndex` was successful.

image

sdk/highlight-remix/package.json
deltaepsilon: Yep! Fixing.

sdk/highlight-remix/package.json
deltaepsilon: Fixing.

sdk/highlight-remix/package.json
deltaepsilon: Done!

sdk/highlight-remix/src/client.tsx
deltaepsilon: Hmmm. I'd typically not do this, but I'm coming up with some edge cases where our clients may try to mutate these values, in which case we'd need to retry the `H.init`. I'll make the change.

deltaepsilon:
I've been manually editing the `node_modules` file to repair this.

We need to either fork `@opentelemetry/sdk-node` or write a `/bin` file to replace `* as ImportInTheMiddle` with `ImportInTheMiddle`.

image

deltaepsilon:
> Can we add remix to this page as well? https://www.highlight.io/frameworks

Done!

image

sdk/highlight-remix/package.json
```diff
@@ -0,0 +1,52 @@
...
+	},
+	"peerDependencies": {
+		"highlight.run": ">=7.4.0",
+		"react": ">=18"
+	},
+	"dependencies": {
+		"@highlight-run/node": "workspace:^",
+		"@highlight-run/react": "workspace:^",
+		"@opentelemetry/instrumentation": "0.41.1",
+		"highlight.run": "workspace:^",
```
Vadman97, line 44: these should be `workspace:*` versions to make sure we are referencing the latest published.

Vadman97 (approved):
🚢

sdk/highlight-remix/package.json
deltaepsilon: Done!

PR #5993, opened 2023-07-18
add activity graphs to session insights digest

mayberryzane:
## Summary
- call activity graph lambda from session insights digest lambda

- deployed lambdas and ran a test step function

- no


SpennyNDaJets (approved):
Code makes sense to me!

PR #5994, opened 2023-07-18
Add a way to track the referrer of error alerts

et:
## Summary

This PR attaches a `referrer` query param onto error alert URLs. When the URL is visited, we detect for the presence of this and attach it to the mixpanel `Viewed Error` [event](https://mixpanel.com/project/2504464/view/3045720/app/events#VsdjWdksXZzA). 

Screenshot_2023-07-18_at_2_46_52_PM

_Note: I only applied this logic to error alerts since it's what we're focused most on right now. Doing the other alerts would require some additional digging into how to integrate into the existing tracking events we have (or if we have to create brand new ones)._

To test:
* Set up an error alert with slack, discord, email, webhook
* The link should have `referrer=slack|discord|email|webhook`
* The tracking property should show up in mixpanel

Confirmed slack
See above

Confirmed discord:
Screenshot 2023-07-18 at 2 52 39 PM

Confirmed email:
Screenshot 2023-07-18 at 2 50 05 PM

Confirmed webhook:
Using this [ngrok guide](https://ngrok.com/docs/integrations/hostedhooks/webhooks/) for webhooks, confirmed the passed URL has the `referrer` property:
Screenshot 2023-07-18 at 3 10 20 PM

I copied/pasted the URL and confirmed it shows up in mixpanel

Screenshot 2023-07-18 at 3 14 35 PM

N/A

### DIFF
--- a/backend/model/model.go
+++ b/backend/model/model.go
@@ -2673,7 +2673,7 @@ func (obj *Alert) sendSlackAlert(ctx context.Context, db *gorm.DB, alertID int, 
 		errorLink = routing.AttachReferrer(ctx, errorLink, routing.Slack)
 		// construct Slack message
 		previewText = fmt.Sprintf("Highlight: Error Alert: %s", shortEvent)
-		textBlock = slack.NewTextBlockObject(slack.MarkdownType, fmt.Sprintf("*Highlight Error Alert: %d Recent Occurrences*\n\n%s\n<%s/|View>", *input.ErrorsCount, shortEvent, errorLink), false, false)
+		textBlock = slack.NewTextBlockObject(slack.MarkdownType, fmt.Sprintf("*Highlight Error Alert: %d Recent Occurrences*\n\n%s\n<%s|View>", *input.ErrorsCount, shortEvent, errorLink), false, false)
 		messageBlock = append(messageBlock, slack.NewTextBlockObject(slack.MarkdownType, "*User:*\n"+identifier, false, false))
 		if input.URL != nil && *input.URL != "" {
 			messageBlock = append(messageBlock, slack.NewTextBlockObject(slack.MarkdownType, "*Visited Url:*\n"+*input.URL, false, false))
--- a/backend/model/model.go
+++ b/backend/model/model.go
@@ -19,6 +19,7 @@ import (
 
 	"github.com/aws/smithy-go/ptr"
 	Email "github.com/highlight-run/highlight/backend/email"
+	"github.com/highlight-run/highlight/backend/routing"
 	"github.com/jackc/pgconn"
 	"github.com/jackc/pgerrcode"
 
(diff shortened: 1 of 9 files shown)


frontend/src/pages/Alerts/AlertConfigurationCard/AlertConfigurationCard.tsx
```diff
@@ -909,7 +909,7 @@ export const AlertConfigurationCard = ({
 							<p>
 								Add webhook destinations for this alert, sent as
 								JSON over HTTP. See the{' '}
-								<Link to="https://www.highlight.io/docs/general/product-features/general-features/alerts/webhooks">
+								<Link to="https://www.highlight.io/docs/general/product-features/general-features/webhooks">
```
et, line 912: Drive by bug fix to update to this URL
https://www.highlight.io/docs/general/product-features/general-features/webhooks

The previous URL is broken.

backend/model/model.go
```diff
@@ -2668,9 +2670,10 @@ func (obj *Alert) sendSlackAlert(ctx context.Context, db *gorm.DB, alertID int,
 			shortEvent = input.Group.Event[:50] + "..."
 		}
 		errorLink := fmt.Sprintf("%s/%d/errors/%s/instances/%d", frontendURL, obj.ProjectID, input.Group.SecureID, input.ErrorObject.ID)
+		errorLink = routing.AttachReferrer(ctx, errorLink, routing.Slack)
 		// construct Slack message
 		previewText = fmt.Sprintf("Highlight: Error Alert: %s", shortEvent)
-		textBlock = slack.NewTextBlockObject(slack.MarkdownType, fmt.Sprintf("*Highlight Error Alert: %d Recent Occurrences*\n\n%s\n<%s/|View>", *input.ErrorsCount, shortEvent, errorLink), false, false)
```
et, line 2673: There was an extra trailing slash

backend/model/model.go
```diff
@@ -2697,7 +2700,7 @@ func (obj *Alert) sendSlackAlert(ctx context.Context, db *gorm.DB, alertID int,
 					false,
 				),
 			)
-			button.URL = fmt.Sprintf("%s?action=%s", errorLink, strings.ToLower(string(action)))
+			button.URL = routing.AttachQueryParam(ctx, errorLink, "action", strings.ToLower(string(action)))
```
et, line 2703: Have to use a proper query builder since we already have the `?` set from the `referrer` property above.

backend/routing/routing.go
```diff
@@ -0,0 +1,36 @@
+package routing
```
et, line 1: There's also a slew of route helpers in `backend/alerts/routing.go`
Ideally those functions would go here. However, because the slack logic is in `model.go` (and not `backend/alerts`), we get a circular dependency (because those routes reference model code).

cc @SpennyNDaJets - when we take on refactoring slack alerts, we should move it to `backend/alerts/slack` and we can make the logic in `backend/alerts/routing.go` more useful to other parts of the codebase.

mayberryzane (approved):

PR #6001, opened 2023-07-19
fix monkeypatch of window.promise

Vadman97:
## Summary

The `window.Promise` monkeypatch would break in Next.js because, ..., I have no idea.
Something about how Next.js `<Link>` components use `new Promise`s under the hood would break
with the existing monkeypatch, even though the existing monkeypatch should have properly copied the full class.
This change monkeypatches only the constructor (adding a new method as well) which seems to fix the problem.

Promise errors still get recorded with the right stack trace
image
https://www.loom.com/share/dbfb50baac484911b032af02e36447d1
image

New highlight.run and @highlight-run/next versions.


deltaepsilon (approved):
Great find.

PR #6011, opened 2023-07-19
fix session insights formatting for outlook

mayberryzane:
## Summary
- remove the outer borders from each session in the template and use horizontal dividers instead
- add some properties to prevent outlook from expanding the table past its 400px width

- deployed and verified on outlook and other browsers

- no

### DIFF
--- a/packages/react-email-templates/emails/session-insights.tsx
+++ b/packages/react-email-templates/emails/session-insights.tsx
@@ -1,6 +1,5 @@
 import {
 	Body,
-	Button,
 	Column,
 	Container,
 	Head,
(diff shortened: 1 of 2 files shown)


packages/react-email-templates/emails/session-insights.tsx
```diff
@@ -64,114 +63,122 @@ export const SessionInsightsEmail = ({
...
-										alt="activity"
-									/>
+								<Column
+									width={360}
+									style={{ wordBreak: 'break-all' }}
+								>
+									<Text style={insightText}>{i}</Text>
+									{/* <Link style={hoverAnchor} href={s.url}>
+										<Text style={insightText}>{i}</Text>
+									</Link> */}
```
Vadman97, line 149: comment

Vadman97 (approved):

PR #6013, opened 2023-07-19
Remove additional focusing which causes an additional slack sync

et:
## Summary

We are issuing a duplicate request to `SyncSlackIntegration` when opening up the new comment modal on the session viewer. This can cause us to get rate limited.

#5011 added some nifty logic to re-sync slack when a component is focused. The problem is that we had some existing clunky logic for autofocusing in one of these components.
This PR removes some clunky logic that tries to autofocus and instead relies on the user to focus the window.

Turn off [strict mode](https://react.dev/reference/react/StrictMode) (to ensure your local dev acts like production) and doesn't issue an additional re-render:
```
diff --git a/frontend/src/index.tsx b/frontend/src/index.tsx
index 30b830e9f..2d96222f5 100644
--- a/frontend/src/index.tsx
+++ b/frontend/src/index.tsx
@@ -407,8 +407,4 @@ const AuthenticationRoleRouter = () => {
 
 const container = document.getElementById('root')!
 const root = createRoot(container)
-root.render(
-       <React.StrictMode>
-               <App />
-       </React.StrictMode>,
-)
+root.render(<App />)
```

Before - observe 2 requests to `SyncSlackIntegration` after opening up the new comment modal
Screenshot 2023-07-19 at 3 35 40 PM

After: observe 1 requests to `SyncSlackIntegration` after opening up the new comment modal
Screenshot 2023-07-19 at 3 34 14 PM


frontend/src/pages/Player/Toolbar/NewCommentForm/CommentTextBody/CommentTextBody.tsx
```diff
@@ -11,13 +11,12 @@ import { useSlackSync } from '@hooks/useSlackSync'
...
 import { splitTaggedUsers } from '@util/string'
 import clsx from 'clsx'
-import React, { useEffect, useState } from 'react'
+import React from 'react'
 import Linkify from 'react-linkify'
 
 import { AdminSuggestion } from '@/components/Comment/utils/utils'
 import SlackLoadOrConnect from '@/pages/Alerts/AlertConfigurationCard/SlackLoadOrConnect'
 
-import newCommentFormStyles from '../NewCommentForm.module.css'
```
et, line 20: Doesn't look these styles are used anymore.

SpennyNDaJets (approved):
This is better and removes the duplicate call. Looks like it still focuses on the text box when opened which is nice, so the user doesn't have to double click.

Overall, I think we may want to remove the `onFocus={syncSlack}` and put it in a better place, because any time the user clicks the `@` or `+` it is firing off another request. For now this should improve though and reduce the burst

PR #6014, opened 2023-07-19
Extract query parsing logic out of logs

et:
## Summary

This PR extracts out the query parsing logic from logs. The intention here is that the syntax can be re-used easily without being tied to a specific table or database.

Furthermore, it becomes incredibly easier to write additional unit tests without a clickhouse dependency (see `queryparser_test.go`).

This was already heavily tested in `logs_test.go`. I added additional tests for sanity.

Confirmed searching for logs and error instances works as expected

N/A

### DIFF
--- a/backend/clickhouse/logs.go
+++ b/backend/clickhouse/logs.go
@@ -6,12 +6,12 @@ import (
 	"math"
 	"strings"
 	"time"
-	"unicode"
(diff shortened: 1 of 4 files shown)


SpennyNDaJets (approved):
Click test and code look good

PR #6028, opened 2023-07-20
remove normalness query in session processing for now

mayberryzane:
## Summary
- as `user_journey_steps` grows, this is using an increasing amount of CPU in RDS
- will be replacing this with a batched query, but wanted to disable this for now as CPU is very high in prod at the moment
Screen Shot 2023-07-20 at 10 19 36 AM

- will monitor prod after deployment

- no


Vadman97 (approved):

Vadman97 (approved):
for these sessions that don't get `normalness` set, could we backfill it later with the batched query?

mayberryzane:
> for these sessions that don't get `normalness` set, could we backfill it later with the batched query?

yep, will backfill with a batched query, can select all sessions `where created_at > '2023-07-20' and normalness is null`

PR #6032, opened 2023-07-20
add a dynamic maintenance banner

Vadman97:
## Summary

Show the maintenance banner based on a window stored in Postgres, so that
we can dynamically set the maintenance window without having to push code.

Screenshot 2023-07-20 at 2 34 10 PM

No

### DIFF
--- a/frontend/src/components/Header/Header.tsx
+++ b/frontend/src/components/Header/Header.tsx
@@ -10,6 +10,7 @@ import {
 	useGetBillingDetailsForProjectQuery,
 	useGetProjectQuery,
 	useGetSubscriptionDetailsQuery,
+	useGetSystemConfigurationQuery,
 } from '@graph/hooks'
 import { Maybe, PlanType, ProductType, Project } from '@graph/schemas'
 import {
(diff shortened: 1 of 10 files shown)


frontend/src/components/Header/Header.tsx
```diff
@@ -721,8 +727,8 @@ const BillingBanner: React.FC = () => {
 	])
 
 	const isMaintenance = moment().isBetween(
-		'2023-03-05T23:30:00Z',
-		'2023-03-06T01:45:00Z',
+		systemData?.system_configuration?.maintenance_start,
+		systemData?.system_configuration?.maintenance_end,
 	)
```
Vadman97, line 732: checked that `moment().isBetween(undefined, undefined)` returns `false` so we should be ok if no data is loaded

mayberryzane (approved):

et:
Hey @Vadman97 - no rush but I think this is causing a `record not found` [error](https://app.highlight.io/1/errors/07qW0iRfsSAcwv9GoK0SDCD07Oas/instances/202249581?page=1&query=and%7C%7Cerror_state%2Cis%2COPEN%7C%7Cerror-field_timestamp%2Cbetween_date%2C2023-07-03T19%3A18%3A15.267Z_2023-08-02T19%3A18%3A15.267Z). 

I get this on local dev because I don't have a `SystemConfiguration` record. 

We could ignore the error but it feels like we should be handling this more gracefully. Particularly, the boolean check of "active" doesn't even seem accurate 

Screenshot 2023-08-02 at 1 25 09 PM

PR #6034, opened 2023-07-20
batch job for normalness scoring

mayberryzane:
## Summary
- adds a lambda for updating sessions' normalness scores in batches (runs for all processed and not excluded sessions without a score)
- after the last change to remove normalness scoring in the worker, re-processed sessions have a score of 0, so they will be re-scored by this job too
- originally wanted to do this in the worker outside of `processSession`, but looking at the logs, most processing batches have <10 sessions so the speedup won't be huge
- it's reasonably fast (depending on how large the input size is, the first was for about 2 hours worth of unscored sessions)
```
UPDATE 28202
Time: 101.153s (1 minute 41 seconds), executed in: 101.153s (1 minute 41 seconds)
UPDATE 274
Time: 10.094s (10 seconds), executed in: 10.094s (10 seconds)
```

- ran script in prod

- will create a step function for it and schedule it to run every 5 mins

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -224,6 +224,44 @@ jobs:
                   function_name: sendSessionInsightsEmails
                   zip_file: backend/sendSessionInsightsEmails.zip
 
+    deploy-user-journey-lambdas:
+        needs: migrate-database
+        runs-on: ubuntu-latest
+        steps:
+            - name: Checkout
+              uses: actions/checkout@v3
+            - name: Setup Go
+              uses: actions/setup-go@v4
+              with:
+                  go-version-file: 'backend/go.mod'
+            - name: Install Doppler CLI
+              uses: dopplerhq/cli-action@v2
+            - name: Build and zip
+              run: |
+                  cd backend/
+                  CGO_ENABLED=0 go build ./lambda-functions/journeys/updateNormalnessScores
+                  zip updateNormalnessScores.zip updateNormalnessScores
+            - name: Configure AWS credentials
+              uses: aws-actions/configure-aws-credentials@v2
+              with:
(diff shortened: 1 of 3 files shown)


backend/lambda-functions/journeys/handlers/handlers.go
```diff
@@ -260,3 +262,50 @@ func (h *handlers) GetJourney(ctx context.Context, input utils.JourneyInput) (*u
...
+		if err != nil {
+			return err
+		}
+
+		return tx.Exec(`
+			with frequencies as (
+				select project_id, url, next_url, count(*)
+				from user_journey_steps
+				where created_at > now() - interval '30 days'
+				group by 1, 2, 3),
```
Vadman97, line 278: do we need a new index to group on project_id, url, and next_url?

.github/workflows/deploy.yml
```diff
@@ -224,6 +224,44 @@ jobs:
                   function_name: sendSessionInsightsEmails
                   zip_file: backend/sendSessionInsightsEmails.zip
 
+    deploy-user-journey-lambdas:
```
Vadman97, line 227: i'll file a ticket for the next tech debt week to see if we can clean up this file to avoid having to copy-paste for each lambda

Vadman97 (approved):

backend/lambda-functions/journeys/handlers/handlers.go
mayberryzane: good point, already ran
```
CREATE INDEX concurrently idx_user_journey_steps_project_id_c
 reated_at_v3 ON public.user_journey_steps USING btree (project_id, cre
 ated_at, url, next_url)
 ```
 in prod, will add to the gorm model

PR #6051, opened 2023-07-21
Standardize Reflame config formatting

lewisl9029:
## Summary

I've been working on a new feature to programmatically update the Reflame config based on the NPM package entry points that are actually used (sneak peak [here](https://github.com/lewisl9029/highlight/pull/15/commits/05c4b7f817aa60f73a580e3dd560de886b354af5)). Realized that we need to take over formatting duties for the config from individual projects' prettier configurations, or otherwise there will always be discrepancies between our programmatic updates and prettier output, leading to issues in CI. This PR formats the config according to our new standard and updates prettier configs to ignore it to prevent CI failures.

Prettier's handling of jsonc files was a bit of a hack anyways (it's actually treating it as json5, which can result in subtle bugs and inconsistencies), so this should also make the formatting a bit more robust going forward.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -18,3 +18,4 @@ scripts/venv
 **/.eslintrc.json
 .yarnrc.yml
 __generated
+.reflame.config.jsonc
--- a/.prettierrc
+++ b/.prettierrc
@@ -10,16 +10,5 @@
 	"arrowParens": "always",
 	"requirePragma": false,
 	"insertPragma": false,
-	"proseWrap": "preserve",
-	"overrides": [
-		{
-			"files": [".reflame.config.jsonc"],
-			"options": {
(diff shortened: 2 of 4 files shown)


Vadman97 (approved):

Vadman97:
looks like e2e test requires doppler now 🤔 
@ccschmitz do you think we should just skip it in forks?

i'll override and merge this in the meantime

ccschmitz:
@Vadman97 ah, I didn't think of the implications for forks. I think we could also manually assign the necessary env vars, similar to what we do in env.sh. I haven't looked through everything we need from Doppler, but I'm guessing it's a relatively small list of vars we need to ensure are set for the build to be in the right state for tests. I can look into this next week.

PR #6053, opened 2023-07-21
Fix deduping service name

### LINKED ISSUE #6045: Logs do not dedupe mapped attributes
When we map data directly to a column and not `LogAttributes` for performance reasons, we do not handle removing that data from `LogAttributes`

For example, we map `service.name` to the column `ServiceName`. We do this for performance reasons (it's more scalable to index a column than an arbitrary unbounded map) Yet it's still showing up in `LogAttributes`.

Screenshot_2023-07-21_at_9_39_25_AM

Screenshot 2023-07-21 at 9 39 59 AM

To reproduce this problem, send a [curl](https://www.highlight.io/docs/getting-started/backend-logging/http) with the following payload:

```
{
    "resourceLogs": [
        {
            "resource": {
                "attributes": [
                    {
                        "key": "service.name",
                        "value": {
                            "stringValue": "my service"
                        }
                    }
                ]
            },
            "scopeLogs": [
                {
                    "scope": {},
                    "logRecords": [
                        {
                            "timeUnixNano": "1689889514000000000",
                            "severityNumber": 9,
                            "severityText": "Info",
                            "name": "logA",
                            "body": {
                                "stringValue": "Hello, world! This is sent from a curl command."
                            },
                            "attributes": [
                                {
                                    "key": "highlight.project_id",
                                    "value": {
                                        "stringValue": "1"
                                    }
                                },
                                {
                                    "key": "foo",
                                    "value": {
                                        "stringValue": "bar"
                                    }
                                }
                            ],
                            "traceId": "08040201000000000000000000000000",
                            "spanId": "0102040800000000"
                        }
                    ]
                }
            ]
        }
    ]
}
```

et:
## Summary

This PR fixes service name being duped in the log

image

Additionally, this PR also refactors the OTEL ingest fairly significantly. Namely, the following has been changed:

* Handle all data extraction and data massaging in one place (previously this was scattered across many functions in the OTEL code and in the clickhouse code).
* Use structs for the above (previously we had a number of args and it was difficult to track what mapped to what).
* Tests!

Fixes #6045

```
curl -X POST http://localhost:4318/v1/logs \
-H 'Content-Type: application/json' \
-d '{
      "resourceLogs": [
        {
          "resource": {
            "attributes": [
                {
                    "key": "service.name",
                    "value": {
                        "stringValue": "my service"
                    }
                }
            ]
        },
          "scopeLogs": [
            {
              "scope": {},
              "logRecords": [
                {
                  "timeUnixNano": "'$(date +%s000000000)'",
                  "severityNumber": 9,
                  "severityText": "Info",
                  "name": "logA",
                  "body": {
                    "stringValue": "Hello, world! This is sent from a curl command."
                  },
                  "attributes": [
                    {
                      "key": "highlight.project_id",
                      "value": {
                        "stringValue": "1"
                      }
                    },
                    {
                      "key": "foo",
                      "value": {
                        "stringValue": "bar"
                      }
                    }
                  ],
                  "traceId": "08040201000000000000000000000000",
                  "spanId": "0102040800000000"
                }
              ]
            }
          ]
        }
      ]
    }'
```

Find the log and confirm that service name is not duped:
Screenshot 2023-07-25 at 11 01 19 AM


backend/clickhouse/log_row.go
```diff
@@ -46,6 +29,12 @@ type LogRow struct {
 }
 
 func NewLogRow(timestamp time.Time, projectID uint32, opts ...LogRowOption) *LogRow {
+	// if the timestamp is zero, set time
+	// TODO(et) - should we move this up to extractFields?
```
Vadman97, line 33: should this be a new ticket or can this be handled in this PR?

backend/otel/extract.go
```diff
@@ -0,0 +1,201 @@
...
+	source       modelInputs.LogSource
+	serviceName  string
+
+	// The original map data. Should only be used for debugging.
+	// TODO(et) - if we only use this for debugging purposes, should we just omit?
+	resourceAttributes map[string]any
+	spanAttributes     map[string]any
+	eventAttributes    map[string]any
+	scopeAttributes    map[string]any
+	logAttributes      map[string]any
```
Vadman97, line 38: could we just make these `*map[string]any` so that the struct keeps pointers here rather than copying the whole map? then i'd say it's safe to keep these. otherwise if they are not useful for debugging, we should drop

Vadman97 (approved):
nice! really helps clean up `otel.go`

backend/otel/otel.go
```diff
@@ -236,79 +155,90 @@ func (o *Handler) HandleTrace(w http.ResponseWriter, r *http.Request) {
...
 				span := spans.At(k)
-				spanAttributes := span.Attributes().AsRaw()
-				setHighlightAttributes(spanAttributes, &projectID, &sessionID, &requestID, &source, &service, &host)
 				events := span.Events()
 				for l := 0; l < events.Len(); l++ {
 					event := events.At(l)
 					eventAttributes := event.Attributes().AsRaw()
-					setHighlightAttributes(eventAttributes, &projectID, &sessionID, &requestID, &source, &service, &host)
+
+					fields, err := extractFields(ctx, extractFieldsParams{
```
SpennyNDaJets, line 169: Just to confirm - this method is completing the logic we used to complete down the line in `setHighlightAttributes`?

SpennyNDaJets (commented):
Makes sense to me! No other comments besides those Vadim raised

backend/clickhouse/log_row.go
et: Tracked here for now #6116.

We can dig into this in our next pairing.

PR #6080, opened 2023-07-24
Update Service model with Github repo

SpennyNDaJets:
## Summary
Update the Service model to have:
- github repo link so we can fetch when appropriate
- status incase the repo link causes errors

Adds a query to fetch services by project id
Adds a mutation to update the service object

Postman - send GraphQL requests to the new Service query and EditService mutation

https://www.loom.com/share/6d46202e8fc0494ca798a571b87754c9?sid=69238973-4045-411f-b0df-4db3873ce5ee

N/A

### DIFF
--- a/backend/private-graph/graph/schema.graphqls
+++ b/backend/private-graph/graph/schema.graphqls
@@ -647,6 +647,14 @@ type ErrorObjectConnection implements Connection {
 	pageInfo: PageInfo!
 }
 
+type Service {
+	id: ID!
+	projectID: ID!
+	name: String!
+	status: String!
+	githubRepoPath: String
+}
+
 enum ReservedLogKey {
 	"""
 	Keep this in alpha order
@@ -1709,6 +1717,7 @@ type Query {
 	error_resolution_suggestion(error_object_id: ID!): String!
 	session_insight(secure_id: String!): SessionInsight
 	system_configuration: SystemConfiguration!
+	services(project_id: ID!): [Service!]!
 }
 
 type Mutation {
@@ -2073,6 +2082,7 @@ type Mutation {
 		is_opt_out: Boolean!
 		project_id: Int
 	): Boolean!
+	editService(id: ID!, project_id: ID!, github_repo_path: String): Service
 }
 
 type Subscription {
--- a/backend/private-graph/graph/schema.graphqls
+++ b/backend/private-graph/graph/schema.graphqls
@@ -647,11 +647,17 @@ type ErrorObjectConnection implements Connection {
 	pageInfo: PageInfo!
 }
 
+enum ServiceStatus {
+	healthy
+	error
+	created
+}
+
 type Service {
 	id: ID!
 	projectID: ID!
 	name: String!
-	status: String!
+	status: ServiceStatus!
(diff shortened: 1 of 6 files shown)


backend/private-graph/graph/schema.graphqls
```diff
@@ -647,6 +647,14 @@ type ErrorObjectConnection implements Connection {
 	pageInfo: PageInfo!
 }
 
+type Service {
+	id: ID!
+	projectID: ID!
+	name: String!
+	status: String!
```
et, line 654: Not blocking but it would be great if this was an enum so we can map the data more cleanly later on.

backend/private-graph/graph/schema.resolvers.go
```diff
@@ -7491,6 +7515,21 @@ func (r *queryResolver) SystemConfiguration(ctx context.Context) (*model.SystemC
...
+// Services is the resolver for the services field.
+func (r *queryResolver) Services(ctx context.Context, projectID int) ([]*model.Service, error) {
+	project, err := r.isAdminInProjectOrDemoProject(ctx, projectID)
+	if err != nil {
+		return nil, err
+	}
+
+	services := []*model.Service{}
+	if err := r.DB.Order("name ASC").Model(&model.Service{}).Where(&model.Service{ProjectID: project.ID}).Scan(&services).Error; err != nil {
+		return nil, e.Wrap(err, "error getting associated services")
```
et, line 7527: Don't really need error wrapping here.

backend/private-graph/graph/schema.resolvers.go
```diff
@@ -3632,6 +3632,30 @@ func (r *mutationResolver) UpdateEmailOptOut(ctx context.Context, token *string,
...
+func (r *mutationResolver) EditService(ctx context.Context, id int, projectID int, githubRepoPath *string) (*model.Service, error) {
+	project, err := r.isAdminInProject(ctx, projectID)
+	if err != nil {
+		return nil, err
+	}
+
+	serviceUpdates := map[string]interface{}{}
+	if githubRepoPath != nil {
+		serviceUpdates["GithubRepoPath"] = *githubRepoPath
+		serviceUpdates["Status"] = "healthy"
```
et, line 3645: This feels kind of misleading to me. Wouldn't `healthy` be the situation where we can actually verify data coming in from that service? In other words, just the process of linking a github repo doesn't really feel like it should be any different from the `created` status.

et (approved):
Looks good! Left some comments but nothing blocking.

backend/private-graph/graph/schema.resolvers.go
SpennyNDaJets: Yeah, I don't feel too strongly here - I was thinking a connection would be deemed "healthy" until its not. Not sure if "healthy" is the right word, but for complexity not sure if we want to add logic that would come back and update the service when it starts processing to start

backend/private-graph/graph/schema.graphqls
SpennyNDaJets: I agree! I tried this and hit a snag with something in the private-graph but will give it a look

PR #6101, opened 2023-07-25
cleanup sdk unused dependencies

Vadman97:
## Summary

As reported on [discord](https://discord.com/channels/1026884757667188757/1133363145885372517), our sdks have some unneeded dependencies. ie: https://npmgraph.js.org/?q=%40highlight-run%2Fnest

E2E test.

New package versions released. Bumping all node-dependent versions to be the same.

### DIFF
--- a/sdk/highlight-apollo/CHANGELOG.md
+++ b/sdk/highlight-apollo/CHANGELOG.md
@@ -9,3 +9,7 @@
 ### 3.1.2
 
 - Bumping to match `@highlight-run/node`
+
+### 3.1.6
+
+- Bumping to match `@highlight-run/node`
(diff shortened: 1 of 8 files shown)


Vadman97:
another mysterious failure @lewisl9029 - did the lock file change in a way that breaks the frontend?

lewisl9029:
@Vadman97 sorry about that, this likely didn't have anything to do with your changes here. I misconfigured my redis cache with `volatile-lru` when it should have been using `allkeys-lru`, causing it to not evict all old keys properly, causing OOM errors.

Just shipped an update, so you should be good to retry anytime. 🙏

deltaepsilon (approved):

PR #6104, opened 2023-07-25
Stop minifying rrweb generated modules

lewisl9029:
## Summary

Quick update to minimize opportunity for merge conflicts in rrweb generated modules.

Longer term, I'll be looking into how we could handle cross-repo projects natively to remove the need for these.

### DIFF
--- a/frontend/scripts/build-rrweb.mjs
+++ b/frontend/scripts/build-rrweb.mjs
@@ -41,7 +41,7 @@ export const run = async ({ rootDirectory }) => {
 		format: 'esm',
 		platform: 'browser',
 		outdir: path.join(rootDirectory, 'src/__generated/rr'),
-		minify: true,
+		minify: false,
 		target: 'esnext',
 		// external: Object.keys(packageJson.dependencies).concat(
 		//   Object.keys(packageJson.peerDependencies),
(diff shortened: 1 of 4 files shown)


Vadman97 (commented):
would this mean that our future production hosting of app.highlight.io would not be using a minified build of rrweb?

lewisl9029:
@Vadman97 great question! The generated modules here are actually treated as source code by Reflame, so they'll go through the same process as any other piece of source code (transpilation, minification, bundling) before making it into the production bundle, so everything will be minified and optimized in the prod deployment. You can verify this in the [preview](https://preview.highlight.io/?%7Er_preview=%7B%22action%22%3A%22start%22%2C%22variantId%22%3A%2201H677PKTER4WHA5RQZ8A2H3Y0%22%2C%22region%22%3A%22sjc%22%2C%22variantData%22%3A%22%7B%5C%22type%5C%22%3A%5C%22branch%5C%22%2C%5C%22branch%5C%22%3A%5C%22stop-minifying-generated-rrweb%5C%22%2C%5C%22forkSpecifier%5C%22%3A%5C%22lewisl9029%2Fhighlight%5C%22%2C%5C%22githubOwnerName%5C%22%3A%5C%22highlight%5C%22%2C%5C%22githubRepositoryName%5C%22%3A%5C%22highlight%5C%22%7D%22%7D), which should be byte-for-byte identical to what customers will end up getting served once it's merged.

Vadman97 (approved):
ah got it, so the committed `rr.js` file is just for reflame to know the 'state' of the rrweb repo. makes sense!

lewisl9029:
Yep, it's a workaround for the fact that our GitHub app can't access the rrweb submodule. Will eventually put together a more elegant solution for this.

PR #6109, opened 2023-07-26
fix e2e test on main

Vadman97:
## Summary

After #6015, the e2e test was broken in main as it would be trying to reference a dev client version.
This PR makes the e2e test default to `undefined` (script version set by the firstload package.json) in main.

CI

No


ccschmitz (approved):
Nice. Thanks for fixing!

PR #6115, opened 2023-07-26
Create services page using table

SpennyNDaJets:
## Summary
Create a table to allow users to view and link Github repos to their services.

https://www.loom.com/share/d42ee41190b4444d91189e180c99c4b0?sid=0af1d61b-de06-40fb-a997-098beb6e9eaa

1) Create multiple services in Postgres
```
INSERT INTO services (project_id, name, github_repo_path, status) VALUES (1, 'public_graph', 'highlight/highlight', 'error');
INSERT INTO services (project_id, name, status) VALUES (1, 'react_frontend', 'created');
INSERT INTO services (project_id, name, github_repo_path, status) VALUES (1, 'private-graph', 'highlight/highlight', 'healthy');
```
2) Created a dev Github app on Github to avoid setting/reseting the integration on the prod account 
- can use Spenny's Highlight Dev by updating [this line](https://github.com/highlight/highlight/blob/main/frontend/src/pages/IntegrationsPage/components/GitHubIntegration/utils.ts#L54) to `https://github.com/apps/spenny-s-highlight-dev/installations/new`
3) Copy over the following env variables from [dev-spencer on Doppler](https://dashboard.doppler.com/workplace/98837/projects/highlight-main/configs/dev_spencer)
```
GITHUB_APP_ID
GITHUB_CLIENT_ID
GITHUB_CLIENT_SECRET
GITHUB_PRIVATE_KEY
```
4) View the services on the services setting page (`/1/settings/services`)
- [ ] When clicking on the Github repo, user is prompted to integrate Github
4) Create a Github integration on the Integrations page (`/1/integrations`)
5) View the services on the services setting page (`/1/settings/services`)
- [ ] User can clear repo
- [ ] User can add repo
- [ ] User can switch repo
- [ ] Link to Github repo works
- [ ] User can search for service

No - only available for Highlight admins


frontend/src/pages/ProjectSettings/ServicesTable/ServicesTable.tsx
```diff
@@ -0,0 +1,288 @@
...
+		data: githubData,
+	} = useGitHubIntegration()
+
+	const [editServiceGithubSettings] = useEditServiceGithubSettingsMutation()
+
+	useEffect(() => {
+		loadServices({
+			variables: {
+				project_id: project_id!,
+				query: query!,
```
et, line 54: I don't think you need the `!` for query, before, and after since they're optional

frontend/src/pages/ProjectSettings/ServicesTable/ServicesTable.tsx
```diff
@@ -0,0 +1,288 @@
...
+import { GitHubSettingsModal } from '../GitHubSettingsModal/GitHubSettingsModal'
+import * as styles from './ServicesTable.css'
+
+type Pagination = {
+	after?: string
+	before?: string
+}
+
+export const ServicesTable: React.FC = () => {
+	const { project_id } = useParams<{ project_id: string }>()
```
et, line 37: Should we use `useProjectId` instead since it works with the demo project.

frontend/src/pages/ProjectSettings/ServicesTable/ServicesTable.tsx
```diff
@@ -0,0 +1,288 @@
...
+	const [loadServices, { error, data, loading }] = useGetServicesLazyQuery()
```
et, line 38: Do we need a lazy query here? 
I think you can get by with a normal query. There's some gotchas with lazy queries that I tend to avoid them if possible: https://github.com/highlight/highlight/pull/5984#pullrequestreview-1535274804

frontend/src/pages/ProjectSettings/ServicesTable/ServicesTable.tsx
```diff
@@ -0,0 +1,288 @@
...
+			debounce((e) => {
+				setQuery(e.target.value)
+				setPagination({})
+			}, 300),
+		[],
+	)
+
+	const handlePreviousPage = () => {
+		setPagination({
+			before: data?.services?.pageInfo.startCursor,
```
et, line 73: Should we null out the `after` in this case? If we don't, the after cursor is pointing to the wrong value.

frontend/src/pages/ProjectSettings/ProjectSettings.tsx
```diff
@@ -91,128 +94,133 @@ const ProjectSettings = () => {
...
-													display="flex"
-													gap="8"
-													justifyContent="space-between"
-													alignItems="center"
+							noHeaderPadding
+							noPadding
+							id="settingsTabs"
+							tabs={[
+								{
+									key: 'general',
```
et, line 119: Was there a reason for moving this to the first tab?

frontend/src/pages/ProjectSettings/ProjectSettings.tsx
```diff
@@ -91,128 +94,133 @@ const ProjectSettings = () => {
...
-						<ProjectSettingsContextProvider
-							value={{
-								allProjectSettings,
-								setAllProjectSettings,
-								loading,
+			<Box style={{ maxWidth: 560 }} my="40" mx="auto">
+				<Heading mt="16" level="h4">
+					Project Settings
+				</Heading>
+				<div className={styles.tabsContainer}>
```
et, line 101: I'm not sure what is going on but the header looks bunch up as a result of these changes:

Screenshot_2023-08-11_at_3_17_30_PM

Compared to what is on prod:


Screenshot 2023-08-11 at 3 19 59 PM

frontend/src/pages/ProjectSettings/ProjectSettings.tsx
```diff
@@ -91,128 +94,133 @@ const ProjectSettings = () => {
...
+											<AutoresolveStaleErrorsForm />
+										</Stack>
+									),
+								},
+								{
+									key: 'services',
+									title: 'Services',
+									panelContent: <ServicesTable />,
+									// TODO: remove when ready to enable to customers
+									hidden: !isHighlightAdmin,
```
et, line 218: Confirmed this is not visible when I'm logged into my non highlight work email.

et: An alternative approach here would have been to just assign tabs to an array and append this tab if `isHighlightAdmin` is true. That way we wouldn't need `hidden` prop.

frontend/src/pages/ProjectSettings/GitHubSettingsModal/GitHubSettingsModal.tsx
```diff
@@ -0,0 +1,190 @@
...
+					<Select
+						aria-label="GitHub Repository"
+						className={styles.repoSelect}
+						placeholder="Search repos..."
+						onSelect={(repo: string) =>
+							form.setValue(form.names.githubRepo, repo)
+						}
+						value={form.values.githubRepo?.split('/').pop()}
+						options={githubOptions}
+						notFoundContent={<span>No repos found</span>}
```
et, line 131: I guess this can happen if they don't grant us access to any of their repos during the auth flow. Could consider calling this out somewhere.

frontend/src/pages/ProjectSettings/GitHubSettingsModal/GitHubSettingsModal.tsx
```diff
@@ -0,0 +1,190 @@
...
+				</Box>
+				{form.values.githubRepo && (
+					<Text size="small">
+						Want to double check? Check it out on{' '}
+						<a
+							href={`https://github.com/${form.values.githubRepo}/`}
+							target="_blank"
+							rel="noreferrer"
+						>
+							Github
```
et, line 157: `GitHub`

frontend/src/pages/ProjectSettings/GitHubSettingsModal/GitHubSettingsModal.tsx
```diff
@@ -0,0 +1,190 @@
...
+					/>
+
+					<Button
+						kind="danger"
+						trackingId="remove-repo-from-service"
+						size="medium"
+						onClick={() =>
+							form.setValue(form.names.githubRepo, null)
+						}
+						disabled={!form.values.githubRepo}
```
et, line 144: Not sure why but this button looks like it's disabled even when the repo is set. You have to hover over it to realize that you can indeed click on it.

Kapture 2023-08-11 at 15 28 06

et (approved):
Tried it on on the reflame preview and was able to set a repo for one of our services:

Screenshot 2023-08-11 at 3 30 42 PM

Nice work!

frontend/src/pages/ProjectSettings/ServicesTable/ServicesTable.tsx
SpennyNDaJets: Yes, so my understanding is since we are setting pagination to an object, that after should be undefined since it is not defined in the object

frontend/src/pages/ProjectSettings/ProjectSettings.tsx
SpennyNDaJets: Matching the mocks here: https://www.figma.com/file/rdBPNzn7Klk6RG1LHUCE5B/Screen-Designs?node-id=10517%3A526107&mode=dev

Didn't add the icons though

frontend/src/pages/ProjectSettings/ProjectSettings.tsx
SpennyNDaJets: Interesting... I am not actually seeing this on render or locally, so curious what you are looking at to repro

frontend/src/pages/ProjectSettings/ProjectSettings.tsx
SpennyNDaJets: Yeah, I was thinking of that approach, but thought this may be cleaner and allow us more flexibility in the future than trying to append tabs if we have multiple cases

frontend/src/pages/ProjectSettings/GitHubSettingsModal/GitHubSettingsModal.tsx
SpennyNDaJets: Yep, its also the case if they type in some nonsense or a repo that doesn't exist

Screenshot 2023-08-11 at 5 50 49 PM

frontend/src/pages/ProjectSettings/GitHubSettingsModal/GitHubSettingsModal.tsx
SpennyNDaJets: Yeah, I can take a look here

SpennyNDaJets: This just seems to be how we designed the danger button: https://github.com/highlight/highlight/blob/main/packages/ui/src/components/Button/styles.css.ts#L117

I'm not too worried here to create new styles because going to have Julian redesign this modal before rolling out

PR #6118, opened 2023-07-26
Start writing traces to Clickhouse

ccschmitz:
## Summary

Creates a new `traces` table in our Clickhouse DB and starts writing spans processed via `HandleTrace` to the table.

Screenshot 2023-07-26 at 2 55 16 PM

Future PRs will add reads for this table.

N/A

Will verify data is flowing into the new table as expected.

### DIFF
--- a/backend/otel/otel_test.go
+++ b/backend/otel/otel_test.go
@@ -84,6 +84,7 @@ func TestHandler_HandleTrace(t *testing.T) {
 		resolver: &public.Resolver{
 			ProducerQueue: &producer,
 			BatchedQueue:  &producer,
+			TracesQueue:   &producer,
 		},
 	}
 	h.HandleTrace(w, r)


backend/clickhouse/trace_row.go
```diff
@@ -0,0 +1,121 @@
...
+	TraceState string            `json:"trace_state"`
+	Attributes map[string]string `json:"attributes,omitempty"`
+}
+
+func NewTraceRow(span ptrace.Span) *TraceRow {
+	attributes := attributesToMap(span.Attributes().AsRaw())
+
+	// TODO: Use extractFields instead of copying logic here. For some reason we
+	// can't import from extract.go.
+	projectId, err := projectToInt(attributes["highlight.project_id"])
```
ccschmitz, line 51: @et n00b Go question here, but do you know why I couldn't import `extractFields` to this file? I wanted to use that to get the project and session IDs here.

et: You have a circular dependency. 

You shouldn't need to extract the projectID, it's done for you in `otel.go`.

https://github.com/highlight/highlight/blob/0a70f450b0c94ce0da69b8d56e4786c1704ef4d6/backend/otel/otel.go#L177-L178

I don't think `NewTraceRow` should have any knowledge of otel related information (e.g.  `ptrace.Span`) as that should be handled up the chain.

ccschmitz: Thanks, @et! I've refactored this so we don't reference any otel logic here and made helpers as function methods, similar to what we have in `log_row.go`. Appreciate your feedback on this!

ccschmitz:
@et @Vadman97 hold on reviewing - I need to fix up some tests and add some test coverage for a few things I added here.

backend/clickhouse/migrations/000012_create_traces_table.up.sql
```diff
@@ -0,0 +1,40 @@
...
+    INDEX idx_res_attr_key mapKeys(ResourceAttributes) TYPE bloom_filter(0.01) GRANULARITY 1,
+    INDEX idx_res_attr_value mapValues(ResourceAttributes) TYPE bloom_filter(0.01) GRANULARITY 1,
+    INDEX idx_span_attr_key mapKeys(SpanAttributes) TYPE bloom_filter(0.01) GRANULARITY 1,
+    INDEX idx_span_attr_value mapValues(SpanAttributes) TYPE bloom_filter(0.01) GRANULARITY 1,
+    INDEX idx_duration Duration TYPE minmax GRANULARITY 1
+)
+    ENGINE = MergeTree
+        PARTITION BY toDate(Timestamp)
+        ORDER BY (ServiceName, SpanName, toUnixTimestamp(Timestamp), TraceId)
+        SETTINGS index_granularity = 8192, ttl_only_drop_parts = 1;
```
ccschmitz, line 40: Most of this migration came from https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/f5275de9a1355c13d85161db6b2eae9d239f1ead/exporter/clickhouseexporter/exporter_traces.go#L177-L212.

et: Let's more closely replicate logs since we know that works correctly for our setup. Check out `backend/clickhouse/migrations/000006_create_logs_new.up.sql`

Namely, the following sticks out to me:

* the OTEL schema doesn't handle multi-tenancy, so `ProjectId` should in the `ORDER BY` since it'll always be part of our querying (see [best practices](https://clickhouse.com/docs/en/optimize/sparse-primary-indexes))
* Add a UUID column so we can handle cursor pagination (see [blog](https://www.highlight.io/blog/how-we-built-logging-with-clickhouse)).
* Merge `ResourceAttributes` and `SpanAttributes` into a single column (similar to `LogAttributes) -- would recommend calling the merged column `TraceAttributes` for consistency.
* Add a column for `ServiceVersion` (see #6117)

backend/clickhouse/trace_row.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+type Link struct {
+	TraceId    string
+	SpanId     string
+	TraceState string
+	Attributes map[string]string
+}
+
+func NewTraceRow(timestamp time.Time) *TraceRow {
```
et, line 41: `projectId` should be a param here (and not an option) since we don't want traces to be inserted accidentally without a project id present.

backend/clickhouse/trace_row.go
```diff
@@ -0,0 +1,163 @@
...
+	traceRow := &TraceRow{
+		Timestamp:   timestamp,
+		ServiceName: "",
```
et, line 44: Omit this. In Clickhouse, the default value for string columns is already an empty string ([docs](https://clickhouse.com/docs/en/sql-reference/statements/create/table#default_values)) (very similar to Go zero values).

backend/clickhouse/trace_row.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	return traceRow
+}
+
+func (t *TraceRow) WithProjectId(projectId int) *TraceRow {
```
et, line 50: Per above. Remove this option method function.

backend/clickhouse/migrations/000012_create_traces_table.up.sql
```diff
@@ -0,0 +1,39 @@
...
+    TraceState String,
+    SpanName LowCardinality(String),
+    SpanKind LowCardinality(String),
+    Duration Int64,
+    ServiceName LowCardinality(String),
+    ResourceAttributes Map(LowCardinality(String), String),
+    SpanAttributes Map(LowCardinality(String), String),
+    StatusCode LowCardinality(String),
+    StatusMessage String,
+    Events Nested (
```
et, line 18: TIL about `Nested`. Not sure if this is a new feature but it looks quite handy.

backend/worker/kafka_worker.go
```diff
@@ -106,6 +107,9 @@ func (k *KafkaBatchWorker) flush(ctx context.Context) {
 						}
 					}
 					received += len(lastMsg.PushLogs.LogRows)
+				case kafkaqueue.PushTraces:
+					traceRows = append(traceRows, lastMsg.PushTraces.TraceRows...)
+					received += len(lastMsg.PushTraces.TraceRows)
```
et, line 112: Wonder if we should make a distinction between how many logs vs how many traces are received. We obviously want to batch our writes but it feels very easy for logs and traces to create unpredictability in that behavior (for example, 99% of `received` could be traces and 1% could be logs, we flush, and then we had an unperformant write for logs.

et (commented):
Looking good! Left some feedback (especially on the schema) that I think we should address.

Also, it would be great to gate this to only projectId = 1 so we can monitor the behavior before rolling it out everyone.

backend/clickhouse/trace_row.go
ccschmitz: Ahh, right - thanks 👍

backend/worker/kafka_worker.go
ccschmitz: Yeah, I think we will want to handle traces separately. Might do something similar to what @mayberryzane did in https://github.com/highlight/highlight/pull/6133/files#diff-4af9641b11571fe8e394b9583d07b44440a1f742db21dc47196319614517f95d.

backend/clickhouse/migrations/000012_create_traces_table.up.sql
Vadman97: cc @mayberryzane for thoughts on the `partition by` and `order by`, as he's been doing some digging into clickhouse query perf.

mayberryzane: I think a custom `partition by` is only helpful if we also have a TTL on this table - are we going to add a 30 day TTL like we do for logs? May also be good to test performance - e.g. without partition, deleting will be slower but querying could be faster?
For the `order by` - I think it may make sense to have it be similar to logs (`ProjectId, Timestamp, UUID`) if the default view is going to be showing all spans in the project. Unless we're always filtering by `ServiceName` and `SpanName` but these seem like secondary filters we could add on to the default

ccschmitz: I updated the `order by` to only include `ProjectId`, `Timestamp`, and `UUID`. I also set a 30 day TTL, similar to the logs table. Let me know if you see anything else you think we should tweak here before merging.

Vadman97: could we remove `PARTITION BY` altogether and see how that performs?

backend/otel/otel.go
```diff
@@ -387,6 +422,30 @@ func (o *Handler) submitProjectLogs(ctx context.Context, projectLogs map[string]
...
+		}
+
+		traceRows := append(traceRows, traceRows...)
+
+		err := o.resolver.BatchedQueue.Submit(ctx, &kafkaqueue.Message{
+			Type: kafkaqueue.PushTraces,
+			PushTraces: &kafkaqueue.PushTracesArgs{
+				TraceRows: traceRows,
+			},
+		}, uuid.New().String())
```
Vadman97, line 439: can be `""` to allow the `Submit()` method to set the default key here

backend/worker/worker.go
```diff
@@ -482,6 +482,31 @@ func (w *Worker) PublicWorker(ctx context.Context) {
...
 	}(&wg)
+	go func(_wg *sync.WaitGroup) {
+		wg := sync.WaitGroup{}
+		wg.Add(parallelBatchWorkers)
+		buffer := &KafkaBatchBuffer{
+			messageQueue: make(chan *kafkaqueue.Message, DefaultBatchFlushSize*(parallelBatchWorkers+1)),
+		}
+		for i := 0; i < parallelBatchWorkers; i++ {
+			go func(workerId int) {
+				s, wCtx := tracer.StartSpanFromContext(ctx, "kafkaWorker", tracer.ResourceName("worker.kafka.batched.outer"))
```
Vadman97, line 493: maybe `worker.kafka.batched.outer.traces` to differentiate?

Vadman97 (approved):
🚢

backend/worker/worker.go
```diff
@@ -482,6 +482,31 @@ func (w *Worker) PublicWorker(ctx context.Context) {
...
+		wg.Add(parallelBatchWorkers)
+		buffer := &KafkaBatchBuffer{
+			messageQueue: make(chan *kafkaqueue.Message, DefaultBatchFlushSize*(parallelBatchWorkers+1)),
+		}
+		for i := 0; i < parallelBatchWorkers; i++ {
+			go func(workerId int) {
+				s, wCtx := tracer.StartSpanFromContext(ctx, "kafkaWorker", tracer.ResourceName("worker.kafka.batched.outer.traces"))
+				defer s.Finish()
+				k := KafkaBatchWorker{
+					KafkaQueue:          kafkaqueue.New(wCtx, kafkaqueue.GetTopic(kafkaqueue.GetTopicOptions{Type: kafkaqueue.TopicTypeBatched}), kafkaqueue.Consumer, nil),
```
ccschmitz, line 496: Talked w/ @Vadman97 about creating a new topic for this rather than using the same one as logs. Going to look at setting this up to avoid congestion on `TopicTypeBatched`.

ccschmitz:
Ended up creating a new `traces` topic in Kafka. Here are the settings I used:

Screenshot 2023-08-01 at 3 40 17 PM

Also ran a quick test locally to test sending messages E2E:

```go
package main

import (
	"context"
	"time"

	"github.com/highlight-run/highlight/backend/clickhouse"
	kafkaqueue "github.com/highlight-run/highlight/backend/kafka-queue"

)

func main() {
	ctx := context.TODO()

	traceRows := []*clickhouse.TraceRow{}
	timestamp := time.Now()
	traceRows = append(traceRows, clickhouse.NewTraceRow(timestamp, 1))

	kafkaTracesProducer := kafkaqueue.New(ctx, kafkaqueue.GetTopic(kafkaqueue.GetTopicOptions{Type: kafkaqueue.TopicTypeTraces}), kafkaqueue.Producer, nil)

	log.WithContext(ctx).Info("submitting message to kafka")
	if err := kafkaTracesProducer.Submit(ctx, &kafkaqueue.Message{
		Type: kafkaqueue.PushTraces,
		PushTraces: &kafkaqueue.PushTracesArgs{
			TraceRows: traceRows,
		},
	}, ""); err != nil {
		log.WithContext(ctx).WithError(err).Error("failed to submit message to kafka")
	}
}
```

Screenshot 2023-08-01 at 3 29 42 PM

Vadman97 (approved):
nice! recent changes look great

PR #6121, opened 2023-07-26
fix writing a backend error stacktrace

et:
## Summary

A regression was introduced in #6053 whereby backend errors were missing their stacktraces. This was due to the fact that we weren't plucking the data out ahead of time in the `extractFields` function.

Verified stacktrace (and other error data) is populated:
Screenshot 2023-07-26 at 12 53 15 PM

Verified we have a stacktrace with a new error in the UI:
Screenshot 2023-07-26 at 12 55 26 PM

N/A

### DIFF
--- a/backend/otel/otel.go
+++ b/backend/otel/otel.go
@@ -22,7 +22,6 @@ import (
 	"github.com/highlight-run/highlight/backend/public-graph/graph/model"
 	"github.com/highlight-run/highlight/backend/stacktraces"
 	"github.com/highlight/highlight/sdk/highlight-go"
-	hlog "github.com/highlight/highlight/sdk/highlight-go/log"
 	"github.com/openlyinc/pointy"
 	e "github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
@@ -68,7 +67,7 @@ func getBackendError(ctx context.Context, ts time.Time, fields extractedFields, 
 		TraceID:         pointy.String(traceID),
 		SpanID:          pointy.String(spanID),
 		LogCursor:       logCursor,
-		Event:           excMessage,
+		Event:           fields.exceptionMessage,
 		Type:            fields.exceptionType,
 		Source:          fields.source.String(),
 		StackTrace:      fields.exceptionStackTrace,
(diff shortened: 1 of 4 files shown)


backend/otel/otel.go
Vadman97, line 1: should we also extract the metrics' attributes (lines 88-92)?
https://github.com/highlight/highlight/pull/6121/files#diff-4f50b51c0e5994df085e64bc66af47b9aa36c0b75417787b8de7c57918fcee12R88-R92

Vadman97 (approved):
thanks for the quick fix!

sdk/highlight-go/otel.go
```diff
@@ -38,23 +39,6 @@ const MetricEvent = "metric"
...
-	ProjectIDAttribute,
-	SessionIDAttribute,
-	RequestIDAttribute,
-	SourceAttribute,
-	LogMessageAttribute,
-	LogSeverityAttribute,
-	// exception should be parsed as structured and not included as part of log attributes
-	"exception.message",
-	"exception.stacktrace",
-	"fluent.",
```
et, line 55: @Vadman97 - The code previously removed all `fluent.*` data. Not sure why we would want to do that (it could be useful?). I can re-add this logic back in if we need it.

backend/otel/otel.go
et: Good catch! Fixed in 0a70f450b

sdk/highlight-go/otel.go
Vadman97: `fluent.tag` was the one we would extract and keep. don't think there were any other attributes we need to drop (https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/b847fe7f95da3c925d75dafac5559a13bfa12e78/receiver/fluentforwardreceiver/conversion.go#L19C26-L19C36), so it's safe to remove the exclusion.

et: Gotcha. We are omitting that already so should be good 👍 
https://github.com/highlight/highlight/blob/7fb1e9b263e8e6d581ea5885934f889f144aa82f/backend/otel/extract.go#L189-L190

PR #6124, opened 2023-07-26
removed launch banner

Abhishek-More:
## Summary

Removed Launch Week Banner

## Testing

Click-testing and observing that the banner was no longer present.

### DIFF
--- a/highlight.io/components/common/Navbar/Navbar.tsx
+++ b/highlight.io/components/common/Navbar/Navbar.tsx
@@ -59,6 +59,8 @@ const Navbar = ({
 	const [isOpen, setIsOpen] = useState(false)
 	const [prevY, setPrevY] = useState(0)
 
+	hideBanner = true
+
 	const changeBackground = () => {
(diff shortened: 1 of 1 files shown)


highlight.io/components/common/Navbar/Navbar.tsx
```diff
@@ -59,6 +59,8 @@ const Navbar = ({
 	const [isOpen, setIsOpen] = useState(false)
 	const [prevY, setPrevY] = useState(0)
 
+	hideBanner = true
```
Vadman97, line 62: we could just remove the banner `<Link` but this works too :)

Vadman97 (approved):
🚢

highlight.io/components/common/Navbar/Navbar.tsx
Abhishek-More: My thinking was we should keep it so that we don't have to recreate all the banner code and styles at a later date

PR #6128, opened 2023-07-26
Add pagination and query to services

SpennyNDaJets:
## Summary
We want to add pagination to services now instead of needing to come back later and add it

1) Create a services via PSQL
2) Update the `SERVICE_LIMIT` to a number less than the number of services you created (or create more than 10)
3) Open postman to hit the `services` query
4) Hit the query with no additional parameters
5) Use the `endCursor` value in the `after` param
6) Use the `startCursor` value in the `before` param
7) Test with a `query` param

https://www.loom.com/share/bbf5a62c3f744a4fad5a0dcaefec8b14?sid=5e7e0cd1-e76c-4f6e-a1ba-fcbb957fb01a

N/A


et (approved):
Awesome 👏 
Thanks for getting this in now ❤️

PR #6137, opened 2023-07-27
fix panic when extracting metric value

et:
## Summary

A regression was introduced in #6121 such that metric values were being casted to a string and causing a panic.

Screenshot 2023-07-27 at 10 22 26 AM

Long term, we probably need to consider whether this `.(type)` syntax is actually safe to be using. It would be better to try and use builtin functions that return errors when we can't safely parse a value.

Confirmed that the panic does not occur when a new metric comes in.

N/A

### DIFF
--- a/backend/otel/extract.go
+++ b/backend/otel/extract.go
@@ -144,10 +144,19 @@ func extractFields(ctx context.Context, params extractFieldsParams) (extractedFi
 	}
 
 	if val, ok := attrs[highlight.MetricEventValue]; ok {
-		castedValue, ok := val.(float64)
+		float64Value, ok := val.(float64)
 		if ok {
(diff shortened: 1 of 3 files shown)


backend/otel/extract.go
```diff
@@ -144,7 +144,7 @@ func extractFields(ctx context.Context, params extractFieldsParams) (extractedFi
 	}
 
 	if val, ok := attrs[highlight.MetricEventValue]; ok {
-		fields.metricEventValue = val.(string)
+		fields.metricEventValue = val.(float64)
```
Vadman97, line 147: can the value ever be a string? should we still handle that case if it is?

Vadman97: cool, ty!

PR #6142, opened 2023-07-27
disable timepicker unless both a start and end time are chosen

et:
## Summary

When we added a timepicker in #5629, we didn't consider an edge case where by someone might try to choose a time without both the start/end date being selected.

Specifically, this line: https://github.com/highlight/highlight/pull/5629/files#diff-cbeeacd5af75faf0a293a5be799a26ed033c8232dbafe25e57f3713bd30bfb60R297

`selectedDates[1]` isn't valid since there is no end date chosen.

Here's a loom show how to reproduce the bug: https://www.loom.com/share/dca5353e88d74a1da9b88080deab9542

Confirmed that the timepicker cannot be clicked when an end date is not selected.
https://www.loom.com/share/22127f1f53324681bf896d756d22785e

### DIFF
--- a/packages/ui/src/components/DatePicker/PreviousDateRangePicker.tsx
+++ b/packages/ui/src/components/DatePicker/PreviousDateRangePicker.tsx
@@ -264,6 +264,11 @@ const PreviousDateRangePickerImpl = ({
 		[selectedDates[1]],
 	)
 
+	const isTimepickerDisabled = useMemo(
+		() => selectedDates.length != 2,
+		[selectedDates],
+	)
+
 	const [buttonLabel, setButtonLabel] = useState<string>(
 		getLabel({ selectedDates, presets }),
 	)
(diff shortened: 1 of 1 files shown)


Vadman97 (approved):
looks good! just a thought about the picker now - could we keep it open after picking the day range so that it is easier to pick custom times?
https://www.loom.com/share/8b22e9fb370a4e668b6ba0ea53b12127

et:
> looks good! just a thought about the picker now - could we keep it open after picking the day range so that it is easier to pick custom times?

The test plan video mentions this. My thoughts are that choosing a custom time is a secondary workflow and that we should optimize for the base case of just selecting dates (and not a custom time).

Right now, I don't think the current workflow of manually typing in a time was the right choice as it's really clunky and easy to screw up (hence the reason why we have all these [formats](https://github.com/highlight/highlight/blob/6138-logs-viewer-date-range-picker-crash-cannot-read-properties-of-undefined-reading-getdate/packages/ui/src/components/DatePicker/PreviousDateRangePicker.tsx#L22-L41)). Hence, I almost want to discourage usage of it because it likely has other bugs that we haven't found yet.

I think we should bring this up with @julian-highlight if we want to change the behavior and see if we can draw on some examples of other apps that use a timepicker and do it well.

PR #6150, opened 2023-07-28
rewrite service name if is invalid

et:
## Summary

This PR rewrites service name if it's invalid (see [slack](https://highlightcorp.slack.com/archives/C02CJANPHQS/p1690575385382839)).

Unit test added

N/A

### DIFF
--- a/sdk/highlight-ruby/highlight/lib/highlight.rb
+++ b/sdk/highlight-ruby/highlight/lib/highlight.rb
@@ -29,7 +29,6 @@ module Highlight
             @otlp_endpoint = otlp_endpoint
 
             OpenTelemetry::SDK.configure do |c|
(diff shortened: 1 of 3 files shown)


sdk/highlight-ruby/highlight/lib/highlight.rb
```diff
@@ -29,7 +29,6 @@ def initialize(project_id, otlp_endpoint=OTLP_HTTP)
             @otlp_endpoint = otlp_endpoint
 
             OpenTelemetry::SDK.configure do |c|
-                c.service_name = "highlight-sdk"
```
et, line 32: Will cut a new version of the ruby SDK when support is added to set service name by the customer.

SpennyNDaJets (approved):

Issue #6168, opened 2023-07-31
Ensure backend deploy has REACT_APP_COMMIT_SHA env var set


et:
Per #6108, we are relying on `REACT_APP_COMMIT_SHA` to be set to set the service version for backend services.
However, after a quick glance in clickhouse (`select * from logs where ProjectId = 1 order by Timestamp DESC limit 1`), it doesn't look like the column is populated correctly.

I verified that setting this env var locally does populate that column correctly so I suspect that we just need to set this in the backend deploy script.

PR #6108, opened 2023-07-25
add service to backend errors

et:
## Summary

This PR demonstrates populating service name and service version onto error_groups/objects.
I also added `WithServiceName` and `WithServiceVersion` to our go SDK (so one doesn't have to use curl) via options to `Start`. 

## How did you test this change?

https://www.loom.com/share/c73a4c3f58a74fb08a19faa1259329bb

I didn't mention it in the video but I also confirmed that we update open search correctly

Screenshot_2023-07-27_at_4_23_19_PM

## Are there any deployment considerations?

We will hold off on making a new release of the Go SDK. When we are confident this is working correctly, we will create a new git tag for customers to use (tracked in #6149)

PR #6169, opened 2023-07-31
ensure REACT_APP_COMMIT_SHA is set for backend services deploys

et:
## Summary

Sets the github deploy sha for backend services and lambdas such that the service version is correctly populated (#6108).

Unsure without actually just merging this and validating it's set in clickhouse (see linked ticket #6168). I just copied the existing pattern for setting this for the frontend deploy so I assume it'll just work.

N/A

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,7 +1,5 @@
 name: Deploy to AWS
 on:
-    pull_request:
-        types: [opened, synchronize]
     push:
         branches:
             - 'main'
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,5 +1,7 @@
 name: Deploy to AWS
 on:
+    pull_request:
+        types: [opened, synchronize]
     push:
         branches:
             - 'main'
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -62,17 +62,18 @@ jobs:
             - name: Update Lambda secrets
               env:
                   DOPPLER_TOKEN: ${{ secrets.DOPPLER_PROD_AWS_LAMBDAS_SECRET }}
+                  REACT_APP_COMMIT_SHA: ${{ github.sha }}
               run: |
                   aws lambda update-function-configuration --function-name getSessionIdsByQuery \
-                    --environment "$(doppler secrets download --no-file | jq '. + {REACT_APP_COMMIT_SHA: ${{github.sha}} }' | jq '{Variables: .}'"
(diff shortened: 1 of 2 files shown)


Vadman97 (commented):
this will make `$REACT_APP_COMMIT_SHA` available when that `run` command runs, but this won't actually bake the env var into the built images. for the lambdas, it will need to passed with the `--environment` flag to the `aws lambda update-function-configuration` call. for the main image build, we will need to update `deploy/Dockerfile.prod.arm64` to bake it as a [docker argument](https://docs.docker.com/build/guide/build-args/) into the image

et:
Thanks for the feedback @Vadman97! I made a change that I think covers this now. Mind taking another look?

.github/workflows/deploy.yml
```diff
@@ -64,15 +64,15 @@ jobs:
                   DOPPLER_TOKEN: ${{ secrets.DOPPLER_PROD_AWS_LAMBDAS_SECRET }}
               run: |
                   aws lambda update-function-configuration --function-name getSessionIdsByQuery \
-                    --environment "$(doppler secrets download --no-file | jq '{Variables: .}')"
+                    --environment "$(doppler secrets download --no-file | jq '. + {REACT_APP_COMMIT_SHA: ${{github.sha}} }' | jq '{Variables: .}'"
```
Vadman97, line 67: are the `$(` not terminated now?

Vadman97 (changes requested):

.github/workflows/deploy.yml
et: Can you propose an alternative? I don't know how to actually test a github action to verify the behavior.

Vadman97: @et You can test locally by running `echo "$(doppler secrets download --no-file | jq '. + {REACT_APP_COMMIT_SHA: ${{github.sha}} }' | jq '{Variables: .}'"` - should error

i think you need to keep the `env:` values you had, and then use the `$REACT_APP_COMMIT_SHA` value as so:

```bash
doppler secrets download --no-file | jq ". + {REACT_APP_COMMIT_SHA: $REACT_APP_COMMIT_SHA }" | jq '{Variables: .}'
```

so it would be
```bash
aws lambda update-function-configuration --function-name getSessionIdsByQuery --environment "doppler secrets download --no-file | jq \". + {REACT_APP_COMMIT_SHA: $REACT_APP_COMMIT_SHA }\" | jq '{Variables: .}'"
```

Screenshot 2023-08-01 at 10 05 46 AM

to test, you can also make this yml file run in your branch to make sure it will correctly run once merged to main. you can add the following in a temp commit, push that, then push a revert 
edit the following at the top
```yaml
on:
    ... what's there now ...
    pull_request:
        types: [opened, synchronize]
```

et:
@Vadman97 - Thanks for the tips! I confirmed the deploy passes (see the second to last commit) and removed the test commit to run on each PR (last commit).

PR #6174, opened 2023-08-01
group errors by embeddings

### LINKED ISSUE #5695: error grouping for instances without a stacktrace should perform a value similarity
first: figure out why some errors don't have a stacktrace. a simpler fix to this should be that all errors must have stacktraces.

error grouping for instances without a stacktrace should perform a value similarity check (90% of the text matches).
for example, https://app.highlight.io/1/errors/8JfiMblxeCZNXBNEUAxTAj0gwn4O and https://app.highlight.io/1/errors/jjrOKh3KgrpAdDRPpwEzec3EdX67 should be grouped together because most characters match

Vadman97:
## Summary

Groups error objects against existing error groups by using embedding similarity search.
This is currently only using one combined embeddings vector as searching more than one
cannot use a pgvector index. Will still be storing individual event/payload/stacktrace embeddings
to understand what strategy groups best.

This PR is gated for our project by the workspace settings for embeddings.
This PR also continues to track the grouping result for the existing grouping strategy
to be able to compare grouping performance.

https://www.loom.com/share/ad9a3b2b5c98432aae35512557fa1b14
New unit test

Closes #5695


ccschmitz (approved):
Code + click test look great ✨

PR #6163, opened 2023-07-31
allow service version to be queried via clickhouse

et:
## Summary

Allows searching logs for a service's version via `service_version:version`. 

## How did you test this change?

Per #6108, we are setting the service version on logs (via the Go SDK) via the `REACT_APP_COMMIT_SHA` env var. On dev using doppler (using `yarn dev:frontend` and `cd backend && make start`), this evaluates to `asdf`

Screenshot 2023-07-31 at 3 14 55 PM

Direct match works:
Screenshot 2023-07-31 at 3 33 35 PM

No match works as expected:
Screenshot 2023-07-31 at 3 11 12 PM

Wildcard works:
Screenshot 2023-07-31 at 3 11 39 PM
Note: there is a different bug on wildcard searches not highlighting correctly (#6162)

On prod, `REACT_APP_COMMIT_SHA`, evaluates to the deploy sha (see #6168). I confirmed this column is being populated:

image

PR #6179, opened 2023-08-01
add version to error instance view

et:
## Summary

This PR adds a Version field to the error instance view if defined via the SDK (see #6108).

When defined:
Screenshot_2023-08-01_at_9_40_29_AM

Note: the above pulls from the env var in doppler (see #6163). I also hardcoded a git sha to ensure it looks good (i.e. no line wrapping):
Screenshot 2023-08-01 at 9 43 19 AM

When not defined:
Screenshot 2023-08-01 at 9 40 52 AM

N/A


SpennyNDaJets (approved):
LGTM - do we have a strategy of how we will keep the env key dynamic in doppler?

PR #6182, opened 2023-08-01
queue datasync messages async

mayberryzane:
## Summary
- use the `Async` setting for the datasync producer

- ensured messages were still being submitted / processed locally

- no


Vadman97 (approved):
@mayberryzane could also try set a larger `BatchTimeout` so that writes are batched... but that won't affect the perf here since writes are async, just reduce broker load.

PR #6185, opened 2023-08-01
fix unpkg export

Vadman97:
## Summary

Unpkg does not like `.cjs` files because it does not treat them as browser-javascript.
See https://highlightcorp.slack.com/archives/C02N2AZS8BV/p1690917075492479 for full context.

Will be testing with the `e2e/html` unpkg script.

New firstload package version.


deltaepsilon (approved):

Vadman97:
unpkg e2e failing because of #6103 . merging this in to fix that

PR #6192, opened 2023-08-02
Remove committed env vars & update script

lewisl9029:
## Summary

First class support for environments just shipped in Reflame! This means env vars can be configured through the UI (currently still behind a hidden route at https://reflame.app/dashboard/app/:appId/environments), whereas previously it could only be configured through the config.

We no longer need to commit env vars into our config and no longer need to update them through a local script that pulls them through doppler, so this PR removes both.

Also added a environment selector in the Reflame command panel so you can switch between all your environments on the fly, for instance, to easily develop against a local backend: 

Screenshot 2023-08-01 at 8 48 16 PM


Vadman97 (approved):
awesome! e2e failure related to this being a fork

PR #6193, opened 2023-08-02
Moved OpenAI call for session insights to Go backend

eightants:
## Summary

Currently the session insight uses an AWS lambda function to get events for a session, and then calls the chatgpt API using the nodejs sdk provided. This is a different approach to the Error Suggestions, which uses an unofficial Go SDK for OpenAI APIs. This changes moves the session insight call to Go, so all OpenAI features can be managed in a single place. 

Tried locally and tested on a few sessions

Depending on how the AI Session Digests are being sent, this could break the functionality. We should ensure this does not break that functionality before merging


backend/event-parse/parse.go
```diff
@@ -753,3 +753,57 @@ func UnmarshallMouseInteractionEvent(data json.RawMessage) (*MouseInteractionEve
...
+
+type Event struct {
+	Type      EventType
+	Timestamp int64
+	Data      interface{}
+}
+
+func containsAnyOf(str string, substrings []string) bool {
+	for _, s := range substrings {
+		if len(s) > 0 && len(str) >= len(s) && str != s && (str == s || str[len(str)-len(s):] == s) {
```
mayberryzane, line 765: is this check different than `strings.Contains` ?

mayberryzane (approved):

backend/event-parse/parse.go
eightants: It checks if a string contains any of the substrings in the array. I think I can substitute that inner if statement with strings.contains though.

PR #6194, opened 2023-08-02
Fix height calculation for settings pages

eightants:
## Summary

Bug on Chrome. This is due to the showBanner context variable being set to true, and this is used for css classes, although banners are disabled on settings pages. 

https://www.loom.com/share/46023c7331c14f2694ca4c045fe5f3dc

Retried this interaction and saw no bottom bar.


frontend/src/__generated/rr/rr.js
SpennyNDaJets, line 1: Are these changes something we want to commit? I see this file change locally and usually ignore it since it seems unrelated

PR #6202, opened 2023-08-02
Filter out nextjs SDK noise

et:
## Summary

There's a lot of noise spamming customers due to #6181. This is a hack to filter that out until we can fix the root cause.

Confirmed this only filters out the expected strings.

N/A


SpennyNDaJets (approved):

deltaepsilon:
Thanks for this. I've been struggling to resolve these garbage errors on the SDK side.

PR #6210, opened 2023-08-03
Use Reflame's native tailwind compat

lewisl9029:
## Summary

I shipped native tailwind compat in Reflame a while back but didn't get a chance to migrate Highlight to it, until now!

This PR gets rid of all the generated tailwind CSS, so the only generated CSS left is for vanilla extract.

We also add another Reflame-specific entry point for vanilla extract for the generated CSS since we were piggybacking on the tailwind entrypoint before, and that's no longer available now that we're using the tailwind entry point for actual tailwind CSS.

Reflame previews looked good as far as I could tell. Let me know if you see any irregularities!


et (approved):
Thanks for much!

Issue #5812, opened 2023-07-04
highlight-py: Support more recent versions of fastapi


tobim:
**Is your feature request related to a problem? Please describe.**
When trying to use the python SDK in a fastapi based service:
```
Trying to add `highlight-io[FastAPI] as dependency:

Because no versions of highlight-io match >0.5.2,<0.6.0
 and highlight-io (0.5.2) depends on fastapi (>=0.92.0,<0.93.0), highlight-io (>=0.5.2,<0.6.0) requires fastapi (>=0.92.0,<0.93.0).
So, because tenant-manager depends on both fastapi (^0.95.1) and highlight-io (^0.5.2), version solving failed.
```

**Describe the solution you'd like**
Relax the version bounds in `pyproject.toml`, and verify that it works.

Vadman97:
hey @tobim , thanks for the report! we'll take this on this week - appreciate your patience. what version of fastapi are you using currently? is it `0.95.1`?

tobim:
Hi @Vadman97, it seems that the change you did is not enough.
https://github.com/highlight/highlight/blob/02f2bdbefc6e1d416b0f8d86e098464635a44271/sdk/highlight-py/pyproject.toml#L32 still only allows FastAPI between 0.92 and 0.93.
So `poetry add highlight-io[FastAPI]` fails.

Can you reopen?

tobim:
Hi @et, thanks for taking a look!
I guess `version = ">=0.92"` should work, I'll report back if  we run into trouble due to breaking changes.

et:
Hey @tobim - I relaxed the requirement and pushed a new release (`0.5.5`). Mind giving it another shot?

tobim:
Turns out we also require a newer version of uvicorn:
```
> poetry add highlight-io[FastAPI] 
Using version ^0.5.5 for highlight-io

Updating dependencies
Resolving dependencies... (4.4s)

Because no versions of highlight-io match >0.5.5,<0.6.0
 and highlight-io (0.5.5) depends on uvicorn (>=0.20,<0.21), highlight-io (>=0.5.5,<0.6.0) requires uvicorn (>=0.20,<0.21).
Because uvicorn (0.20.0) depends on uvicorn (0.20.0)
 and no versions of uvicorn match >0.20,<0.21, uvicorn (>=0.20,<0.21) requires uvicorn (0.20.0).
Thus, highlight-io (>=0.5.5,<0.6.0) requires uvicorn (0.20.0).
[...]

PR #6225, opened 2023-08-03
Handle serialized logs sent via loguru

et:
## Summary

Serialized logs sent through loguru don't look great. This PR fixes that.

`cd sdk/highlight-py/e2e/highlight_loguru && poetry run python main.py`

Confirmed extra attributes are serialized properly:

image

Bumped the python SDK version.

### DIFF
--- a/sdk/highlight-py/highlight_io/sdk.py
+++ b/sdk/highlight-py/highlight_io/sdk.py
@@ -292,10 +292,10 @@ class H(object):
                 # See: https://loguru.readthedocs.io/en/stable/api/logger.html#record
                 obj = json.loads(message)
                 extra = obj["record"]["extra"]
-                for key in extra:
-                    attributes[key] = extra[key]
+                message = obj["text"]
 
-                message = obj["message"]
+                for key, value in extra.items():
+                    attributes[key] = value
             except:
                 pass
 
--- a/sdk/highlight-py/highlight_io/sdk.py
+++ b/sdk/highlight-py/highlight_io/sdk.py
@@ -29,6 +29,7 @@ class LogHandler(logging.Handler):
     def emit(self, record: logging.LogRecord):
         ctx = contextlib.nullcontext
         span = trace.get_current_span()
+
         if span is None or not span.is_recording():
             ctx = self.highlight.trace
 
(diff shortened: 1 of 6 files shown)


sdk/highlight-py/pyproject.toml
```diff
@@ -29,7 +29,7 @@ python = "^3.8"
 blinker = { version = "^1.5", optional = true }
 django = { version = "^4.1", optional = true }
 flask = { version = "^2.2", optional = true }
-fastapi = { version = "^0.92", optional = true }
```
et, line 32: Not related but wanted to get this fix in (context: https://github.com/highlight/highlight/issues/5812#issuecomment-1665903069)

SpennyNDaJets: Sounds good! Unrelated to this PR, but curious if the other Python versions should also be `>=`?

sdk/highlight-py/highlight_io/sdk.py
```diff
@@ -284,14 +285,28 @@ def log_hook(self, span: Span, record: logging.LogRecord):
             attributes["code.filepath"] = record.pathname
             attributes["code.lineno"] = record.lineno
             attributes.update(record.args or {})
+
+            message = record.getMessage()
+            try:
```
SpennyNDaJets, line 290: Just to make sure, there isn't a wait that we can recognize when something is from loguru, and skip this if not?

SpennyNDaJets (approved):

sdk/highlight-py/pyproject.toml
et: I think the other ones make sense to use the caret since they have a major release (reaching at least v1). Don't know much about fast api but it does seem like development moves quickly and we will have rely on our users to let us know if they introduce a breaking change.

sdk/highlight-py/highlight_io/sdk.py
et: That's a great point. I didn't see anything obvious but I was only using print statements. Let me spend a little more time getting a debugger working and seeing if there's any data from  this point in the call stack that we can infer on.

et: Screenshot 2023-08-04 at 3 51 38 PM

Nothing really that helpful in this part of the call stack 😢

PR #6230, opened 2023-08-04
github workflow for ruby

et:
## Summary

This PR sets up a ruby SDK workflow. Currently it only runs the test suite but I plan on extending it for publishing new version of the SDK as well as linting.

Confirmed that the `Ruby SDK / Ruby SDK` build step runs and passes.

N/A

### DIFF
--- a/sdk/highlight-ruby/highlight/test/highlight_test.rb
+++ b/sdk/highlight-ruby/highlight/test/highlight_test.rb
@@ -2,6 +2,7 @@ require_relative "./test_helper"
 
 class HighlightTest < Minitest::Test
   def test_logger
+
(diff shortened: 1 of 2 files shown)


sdk/highlight-ruby/highlight/test/highlight_test.rb
```diff
@@ -2,6 +2,7 @@
 
 class HighlightTest < Minitest::Test
   def test_logger
+
```
et, line 5: Made this change to trigger the workflow.
See output: https://github.com/highlight/highlight/actions/runs/5757895336/job/15609618194?pr=6230

SpennyNDaJets (approved):

PR #6249, opened 2023-08-07
Fix duplicate trace rows and reenable writing trace

ccschmitz:
## Summary

Re-enables writing traces to Clickhouse and fixes a bug where we were duplicating traces.

Unit test - see the new assertion on `len(tracesMessage.PushTraces.TraceRows)`.

Will monitor Clickhouse metrics once this rolls out.


backend/otel/otel_test.go
```diff
@@ -43,77 +52,78 @@ func TestHandler_HandleLog(t *testing.T) {
...
+	})
+	assert.Truef(t, ok, "did not find a PushLogs message")
+
+	tracesMessage, ok := lo.Find(producer.messages, func(message *kafkaqueue.Message) bool {
+		return message.Type == kafkaqueue.PushTraces
+	})
+	assert.Truef(t, ok, "did not find a PushTraces message")
+
+	// 512 traces in sample file
+	assert.Equal(t, 512, len(tracesMessage.PushTraces.TraceRows))
```
ccschmitz, line 110: Added this assertion to confirm the duplication bug was fixed.

Vadman97 (approved):

ccschmitz:
@mayberryzane let me know if you think we should move to async writes now or if that's something we should consider in a follow up.

mayberryzane:
> @mayberryzane let me know if you think we should move to async writes now or if that's something we should consider in a follow up.

It's probably worth doing now since we were getting `too many parts` errors intermittently for logs even when load was low.
I did it in this PR: https://github.com/highlight/highlight/pull/6239/files, using `go-sqlbuilder` to create an insert statement and then calling `Exec` with the `async_insert` and `wait_for_async_insert` settings

mayberryzane (approved):

backend/clickhouse/traces.go
```diff
@@ -5,58 +5,86 @@ import (
...
+
+	ib := sqlbuilder.NewStruct(new(ClickhouseTraceRow)).InsertInto(TracesTable, rows...)
+	sql, args := ib.BuildWithFlavor(sqlbuilder.ClickHouse)
+
+	chCtx := clickhouse.Context(ctx, clickhouse.WithSettings(clickhouse.Settings{
+		"async_insert":          1,
+		"wait_for_async_insert": 1,
+	}))
+	fmt.Printf("::: sql (traces): %s\n", sql)
+	return client.conn.Exec(chCtx, sql, args...)
```
ccschmitz, line 87: @mayberryzane I keep hitting this error trying to write traces to Clickhouse:

```sh
ERRO[0112]/Users/chris/code/highlight/backend/worker/kafka_worker.go:293 github.com/highlight-run/highlight/backend/worker.(*KafkaBatchWorker).flushTraces() failed to batch write traces to clickhouse    error="code: 62, message: Cannot parse expression of type Array(DateTime) here: , , , , , , ), (toDateTime('2023-08-09 03:14:17'), '454360f2-6eef-406e-9ec6-01d4c0ca422c', 'b2e3dd9d868e30cc962b5d51b6c8fa91', 'e500d66b9c2a6de1', '', 1, '', '': While executing WaitForAsyncInsert"
```

I thought maybe it was writing to `Events.Timestamp`, but I tried converting that to a string even and I'm still getting that error. Do you have any ideas what might be going on here?

mayberryzane: I ran into the same thing when cutting over sessions to use async inserts, I had to change any arrays in my model to have `clickhouse.ArraySet` type:
https://sourcegraph.com/github.com/highlight/highlight/-/blob/backend/clickhouse/sessions.go?L20

ccschmitz: Ahh, nice - thank you!

PR #6254, opened 2023-08-07
add lint step to ruby sdk

et:
## Summary

This PR lints the ruby sdk using [rubocop](https://github.com/rubocop/rubocop) which is the ubiquitous ruby linter. 

I used `bundle exec rubocop -a` to autocorrect as much as possible. From there, I used manual correction on everything else (or disabled in `.rubocop.yml`). For things I wasn't sure on, I added an inline comment to ignore that specific check.

Since a lot of this PR is formatting, [remove](https://github.com/highlight/highlight/pull/6254/files?w=1) whitespace changes when reviewing.

Build passes

### DIFF
--- a/sdk/highlight-ruby/highlight/Gemfile.lock
+++ b/sdk/highlight-ruby/highlight/Gemfile.lock
@@ -1,27 +1,30 @@
 PATH
   remote: .
   specs:
-    highlight (0.1.0)
+    highlight_io (0.1.1)
       grpc (~> 1.52)
-      opentelemetry-exporter-otlp
-      opentelemetry-instrumentation-all
-      opentelemetry-sdk
-      opentelemetry-semantic_conventions
+      opentelemetry-exporter-otlp (~> 0.24.0)
+      opentelemetry-instrumentation-all (~> 0.32.0)
+      opentelemetry-sdk (~> 1.2)
+      opentelemetry-semantic_conventions (~> 1.8.0)
 
 GEM
   remote: https://rubygems.org/
   specs:
-    google-protobuf (3.22.2)
-    googleapis-common-protos-types (1.5.0)
+    ast (2.4.2)
+    google-protobuf (3.23.4)
+    googleapis-common-protos-types (1.7.0)
       google-protobuf (~> 3.14)
-    grpc (1.52.0)
-      google-protobuf (~> 3.21)
+    grpc (1.56.2)
+      google-protobuf (~> 3.23)
       googleapis-common-protos-types (~> 1.0)
+    json (2.6.3)
+    language_server-protocol (3.17.0.3)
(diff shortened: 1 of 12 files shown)


sdk/highlight-ruby/highlight/highlight.gemspec
```diff
@@ -1,30 +1,30 @@
...
   spec.version       = Highlight::VERSION
-  spec.authors       = ["Highlight"]
-  spec.email         = ["support@highlight.io"]
+  spec.authors       = ['Highlight']
+  spec.email         = ['support@highlight.io']
 
-  spec.summary       = "The Highlight SDK for Ruby"
-  spec.homepage      = "https://www.highlight.io"
-  spec.license       = "MIT"
-  spec.required_ruby_version = Gem::Requirement.new(">= 2.3.0")
```
et, line 12: I updated the min version here because this `rescue` is not valid in ruby 2.3
https://github.com/highlight/highlight/blob/et/lint-ruby/sdk/highlight-ruby/highlight/lib/highlight.rb#L45-L55

Can verify this by setting `TargetRubyVersion: 2.3` in `rubocop.yml` and running rubocop:

Screenshot 2023-08-07 at 12 32 43 PM

sdk/highlight-ruby/highlight/Gemfile.lock
```diff
@@ -1,27 +1,30 @@
...
   remote: .
   specs:
-    highlight (0.1.0)
+    highlight_io (0.1.1)
       grpc (~> 1.52)
-      opentelemetry-exporter-otlp
-      opentelemetry-instrumentation-all
-      opentelemetry-sdk
-      opentelemetry-semantic_conventions
+      opentelemetry-exporter-otlp (~> 0.24.0)
```
et, line 6: Set versioning on these dependencies. We don't need them to auto update.

sdk/highlight-ruby/highlight/README.md
```diff
@@ -1,17 +1,9 @@
 # Highlight
 
-Welcome to your new gem! In this directory, you'll find the files you need to be able to package up your Ruby library into a gem. Put your Ruby code in the file `lib/highlight`. To experiment with that code, run `bin/console` for an interactive prompt.
```
et, line 3: Cleaning up the default README from bundler's generator.

sdk/highlight-ruby/highlight/highlight.gemspec
et: 2.3 is certainly EOL: https://www.ruby-lang.org/en/downloads/branches/

As is 2.6. We may want to consider setting the min version to be 3.0 but can save that for a follow up.

deltaepsilon (approved):

PR #6256, opened 2023-08-07
Next.js major bump

deltaepsilon:
- Remove exports for the default `@highlight-run/next` and `@highlight-run/next/highlight-init` paths.
- Add `excludedHostnames` prop to `@highlight-run/next`.
- Document `excludedHostnames` and import path removal for `@highlight-run/next`
- Major version bump for `@highlight-run/next`.
- Add `excludedHostnames` to `<HighlightInit />` for `@highlight-run/remix`
- Document change and minor version bump for `@highlight-run/remix`

### Remix
image

### Next
image


docs-content/getting-started/fullstack-frameworks/next-js.md
```diff
@@ -77,6 +77,7 @@ export default function MyApp({ Component, pageProps }: AppProps) {
 	return (
 		<>
 			<HighlightInit
+				excludedHostnames={['localhost']}
```
Vadman97, line 80: in general i'd oppose adding more configuration for a subset of customers, but in this case we've had enough folks to ask for this to warrant building this in.

Vadman97 (approved):
nice! hopefully this doesn't break for any unusual setups 🤞

docs-content/getting-started/fullstack-frameworks/next-js.md
deltaepsilon: Yeah. I'd expect 100% of our clients to implement some sort of `localhost` block, if only to clean up their own logs.

jay-khatri: If we did this, won't this mean their sessions won't show up when they install Highlight for the very first time? Is this a good idea?

deltaepsilon: The `excludedHostnames` props defaults to an empty array, so they have to opt-in.

Nobody wants their `localhost` error logs streaming to Highlight in the long-term. It's important to add it to the docs so that users don't have to hunt for it or do it wrong, as we noticed in a recent bug report.

deltaepsilon: I'll add a comment to remind users to remove `localhost` from `excludedHostnames` when testing.

jay-khatri: sounds good, makes sense.

PR #6257, opened 2023-08-07
Add logging for webhooks

ccschmitz:
## Summary

We are having some reports of webhooks not behaving as expected for users. This PR adds some logging with details for successful webhook requests so we have some information to use for troubleshooting.

Local click test.

Screenshot 2023-08-07 at 3 02 22 PM
Screenshot 2023-08-07 at 3 03 05 PM

N/A - only adding some logging

### DIFF
--- a/backend/alerts/integrations/webhook/webhook.go
+++ b/backend/alerts/integrations/webhook/webhook.go
@@ -1,6 +1,7 @@
 package webhook
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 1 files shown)


ccschmitz:
@Vadman97 any thoughts on whether this is a good idea? Looking into I was investigating https://www.notion.so/runhighlight/Missing-User-properties-in-User-alerts-webhook-5bbd9125c98f47deb300752d8456bd0d?pvs=4 but didn't have much visibility into outgoing requests for webhooks.

Vadman97 (approved):

PR #6258, opened 2023-08-07
Don't require doppler for Cypress tests

ccschmitz:
## Summary

We can't run our E2E tests on PRs from external contributors because they use Doppler. This PR eliminates the dependency on Doppler, which should fix that problem.

CI passed ✅ 

N/A


packages/ui/src/components/Text/Text.stories.tsx
```diff
@@ -12,17 +12,25 @@ export default {
 } as Meta<typeof Text>
 
 export const Sizes = () => {
```
ccschmitz, line 14: Not related, but wanted to fix as they were causing noise on all PRs.

Vadman97 (approved):

PR #6263, opened 2023-08-08
cache GetProject

et:
## Summary

This PR caches `GetProject` with redis for faster lookups. Since projects don't typically change, I did set the expiration to one minute.

The motivation here is that I would like restore #6214 but that was doing many project lookups in a loop. Hence, adding caching will ensure we don't bottleneck that code path.

Additionally this fixes an issue whereby redis was not cleaning itself up on test runs (see the first failing build on this PR).

Tests pass

N/A

### DIFF
--- a/backend/redis/utils.go
+++ b/backend/redis/utils.go
@@ -502,3 +502,7 @@ func (r *Client) AcquireLock(ctx context.Context, key string, timeout time.Durat
 func (r *Client) ReleaseLock(ctx context.Context, key string) (err error) {
 	return r.redisClient.Del(ctx, key).Err()
 }
+
+func (r *Client) FlushDB(ctx context.Context) error {
+	return r.redisClient.FlushDB(ctx).Err()
+}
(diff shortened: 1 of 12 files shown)


backend/redis/utils.go
```diff
@@ -502,3 +502,7 @@ func (r *Client) AcquireLock(ctx context.Context, key string, timeout time.Durat
 func (r *Client) ReleaseLock(ctx context.Context, key string) (err error) {
 	return r.redisClient.Del(ctx, key).Err()
 }
+
+func (r *Client) FlushDB(ctx context.Context) error {
+	return r.redisClient.FlushDB(ctx).Err()
```
Vadman97, line 507: could we wrap this with `if util.IsDevOrTestEnv()` to make sure we can't accidentally wire the db in production?

Vadman97 (approved):
nice! 🚢

backend/redis/utils.go
et: good idea!
a3f6197eb

PR #6271, opened 2023-08-08
write a separate fields table in Clickhouse

mayberryzane:
## Summary
- did some benchmarking, having a separate table is significantly faster for field searches on projects with lots of fields
- use async inserts for writing sessions / fields
- reenable datasync producers

- click tested locally + benchmarked w/ 30 days worth of data for a large project in prod

- if all is looking good, will backfill existing sessions (for the past month) by writing their ids to the datasync queue

### DIFF
--- a/backend/clickhouse/sessions.go
+++ b/backend/clickhouse/sessions.go
@@ -9,7 +9,6 @@ import (
 	"github.com/ClickHouse/clickhouse-go/v2"
 	"github.com/highlight-run/highlight/backend/model"
 	"github.com/huandu/go-sqlbuilder"
-	"github.com/sirupsen/logrus"
 	"golang.org/x/sync/errgroup"
 )
 
@@ -18,9 +17,9 @@ type ClickhouseSession struct {
 	Fingerprint        int32
 	ProjectID          int32
 	PagesVisited       int32
-	ViewedByAdmins     []int32
-	FieldKeys          []string
-	FieldKeyValues     []string
+	ViewedByAdmins     clickhouse.ArraySet
+	FieldKeys          clickhouse.ArraySet
+	FieldKeyValues     clickhouse.ArraySet
 	CreatedAt          time.Time
 	UpdatedAt          time.Time
 	SecureID           string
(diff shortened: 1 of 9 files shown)


backend/clickhouse/sessions.go
```diff
@@ -48,36 +74,89 @@ func (client *Client) WriteSessions(ctx context.Context, sessions []*model.Sessi
...
+		BuildWithFlavor(sqlbuilder.ClickHouse)
+
+	var g errgroup.Group
+	g.Go(func() error {
+		return client.conn.Exec(chCtx, fieldsSql, fieldsArgs...)
+	})
+
+	g.Go(func() error {
+		return client.conn.Exec(chCtx, sessionsSql, sessionsArgs...)
+	})
```
Vadman97, line 159: neat! just checking, the context and connection are thread-safe?

Vadman97 (approved):
🎉

backend/clickhouse/sessions.go
mayberryzane: The context here should be as it's read only. I believe `client.conn` is as well (e.g. there's a `MaxOpenConns` option https://github.com/ClickHouse/clickhouse-go/blob/main/clickhouse_options.go#L137) but not 100% sure on that (I was seeing deadlocks before when both Logs and Sessions were writing in batches), will monitor this after deploying and revert if seeing the same deadlocks

Vadman97: makes sense. should we try increase the `MaxOpenConns` (defaults to 10) so that the 32 workers per ecs node can make more concurrent writes?
https://clickhouse.com/docs/en/integrations/go#connection-pooling

mayberryzane: I think because the `flush`es for the batched topic are surrounded with a mutex, there won't be more than 4 connections at a time (writing logs, traces, sessions, fields)
https://sourcegraph.com/github.com/highlight/highlight/-/blob/backend/worker/kafka_worker.go?L397
also observed from rewinding the queue on Friday that most of the time is spent waiting for async inserts to complete, I don't know if it's worth having multiple workers just to read from Kafka since they I think they spend most of their time waiting for this mutex

Vadman97: cool, good to know. thanks for looking into that. sounds like removing the mutex and writing more concurrently is a bad idea overall since we'd just be increasing that max parts value, maybe to a point that clickhouse won't be happy.

PR #6273, opened 2023-08-08
avoid writing service.* for frontend logs

et:
## Summary

We are leaking public graph service name and version into customer's frontend logs. The reason this is happening is because our OTEL endpoint serves as the backend for client SDK (i.e. the client SDK calls pub.highlight.io).

We actually already have similar code that guarded against this for leaking our public graph's OTEL information:
https://github.com/highlight/highlight/blob/6268-public-graph-service-information-leaking/backend/otel/extract.go#L263-L265
https://github.com/highlight/highlight/blob/6268-public-graph-service-information-leaking/sdk/highlight-go/otel.go#L42-L48

But we didn't update this when we introduced the service notion. This is also the reason why we were having issues when writing a service record on ingest (see [slack](https://highlightcorp.slack.com/archives/C02CJANPHQS/p1691081920973589?thread_ts=1691080057.488759&cid=C02CJANPHQS)).

* Create a new project locally (it's important for this to be different from project id = 1)
* Spin up a local react app and use this config:

```js
H.init('<NEW_PROJECT_ID>', {
    backendUrl: 'https://localhost:8082/public',
    tracingOrigins: true,
    networkRecording: {
      enabled: true,
      recordHeadersAndBody: true,
    },
});
```

Observe that we do not have the `service_name` and `service_version` populated:
Screenshot_2023-08-08_at_4_12_08_PM

### DIFF
--- a/backend/otel/extract.go
+++ b/backend/otel/extract.go
@@ -223,7 +223,7 @@ func extractFields(ctx context.Context, params extractFieldsParams) (extractedFi
 
 	if val, ok := fields.attrs[highlight.MetricEventValue]; ok {
 		float64Value, err := strconv.ParseFloat(val, 64)
-		if err != nil {
+		if err == nil {
 			fields.metricEventValue = float64Value
 		}
 		delete(fields.attrs, highlight.MetricEventValue)
--- a/backend/otel/extract.go
+++ b/backend/otel/extract.go
@@ -55,6 +55,13 @@ type extractedFields struct {
 	attrs map[string]string
 }
 
+func newExtractedFields() *extractedFields {
+	return &extractedFields{
+		source: modelInputs.LogSourceBackend,
+		attrs:  make(map[string]string),
+	}
+}
+
 type extractFieldsParams struct {
 	resource  *pcommon.Resource
 	span      *ptrace.Span
@@ -64,9 +71,7 @@ type extractFieldsParams struct {
 }
 
 func extractFields(ctx context.Context, params extractFieldsParams) (extractedFields, error) {
-	fields := extractedFields{
-		source: modelInputs.LogSourceBackend,
-	}
+	fields := newExtractedFields()
 
 	var resourceAttributes, spanAttributes, eventAttributes, scopeAttributes, logAttributes map[string]any
 	if params.resource != nil {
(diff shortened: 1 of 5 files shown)


backend/otel/syslog_test.go
```diff
@@ -1,16 +1,28 @@
...
-	assert.Equal(t, "render", attrs["hostname"])
+	fields := newExtractedFields()
+
+	fields.logBody = "<1>1 2023-07-27T05:43:22.401882Z render render-log-endpoint-test 1 render-log-endpoint-test - Render test log"
+	extractSyslog(fields)
+	assert.Equal(t, "render", fields.attrs["hostname"])
 	assert.Equal(t, "Render test log", fields.logBody)
 }
+
+func Test_extractSyslogWithStructuredData(t *testing.T) {
```
et, line 18: This is unrelated but our syslog parsing for structured data wasn't extracting data correctly. Partly because render.com doesn't support this and partly because our typing wasn't as string to catch this before.

backend/otel/extract.go
```diff
@@ -127,169 +132,156 @@ func extractFields(ctx context.Context, params extractFieldsParams) (extractedFi
...
+		}
+		delete(originalAttrs, highlight.SourceAttribute)
 	}
 
-	if val, ok := attrs[highlight.ProjectIDAttribute]; ok {
-		fields.projectID = val.(string)
-		delete(attrs, highlight.ProjectIDAttribute)
+	for k, v := range originalAttrs {
+		prefixes := []string{}
+		if fields.source == modelInputs.LogSourceFrontend {
```
et, line 159: I moved this logic up the chain (see how it was down below) such that we can chop off `service.*` before we try extracting it into the top level `serviceName`

backend/otel/extract.go
```diff
@@ -127,169 +132,156 @@ func extractFields(ctx context.Context, params extractFieldsParams) (extractedFi
...
 		spanAttributes,
 		eventAttributes,
 		scopeAttributes,
 		logAttributes,
 	)
 
-	// process potential syslog message
-	if len(fields.logBody) > 0 && fields.logBody[0] == '<' {
-		extractSyslog(&fields, attrs)
+	if val, ok := originalAttrs[highlight.DeprecatedSourceAttribute]; ok {
```
et, line 143: Moved this extraction up early such that we know if this is a frontend or backend source.

SpennyNDaJets (approved):
Code makes sense to me, and example click test looks good! Specs here 🙌

PR #6274, opened 2023-08-08
poll for new sessions, errors, and logs and display a button when new results exist

Vadman97:
## Summary

Poll for new sessions, errors, and logs and display a button when there are new results
that updates the queried time interval. 

https://www.loom.com/share/0c7ce5a684294081ba2b0743a61710d0

No

### DIFF
--- a/frontend/src/components/FeedResults/FeedResults.tsx
+++ b/frontend/src/components/FeedResults/FeedResults.tsx
@@ -7,7 +7,7 @@ import * as style from './FeedResults.css'
 
 interface Props {
 	more: number
-	type: 'sessions' | 'errors'
+	type: 'sessions' | 'errors' | 'logs'
 	onClick: () => void
 }
 
@@ -23,8 +23,8 @@ export const AdditionalFeedResults = function ({ more, type, onClick }: Props) {
 					transition={{ duration: 0.1 }}
 				>
 					<Box
-						paddingTop="8"
-						px="8"
+						paddingTop={type === 'logs' ? undefined : '8'}
+						px={type === 'logs' ? undefined : '8'}
 						width="full"
 						display="flex"
 						alignItems="center"
@@ -41,7 +41,13 @@ export const AdditionalFeedResults = function ({ more, type, onClick }: Props) {
 							<Box display="flex" alignItems="center" gap="8">
 								<IconOutlineArrowNarrowUp />
 								<Text>
-									{more} new {type}
+									{more}
+									{type === 'logs'
+										? more === 50
+											? '+'
+											: ''
+										: ''}{' '}
+									new {type}
 								</Text>
 							</Box>
 						</Button>
--- a/frontend/src/components/FeedResults/FeedResults.tsx
+++ b/frontend/src/components/FeedResults/FeedResults.tsx
@@ -1,0 +1,53 @@
+import { Button } from '@components/Button'
+import { Box, IconOutlineArrowNarrowUp, Text } from '@highlight-run/ui'
+import { AnimatePresence, motion } from 'framer-motion'
+import React from 'react'
+
+import * as style from './FeedResults.css'
+
+interface Props {
+	more: number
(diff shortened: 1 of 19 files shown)


frontend/src/pages/LogsPage/SearchForm/SearchForm.tsx
```diff
@@ -148,7 +145,7 @@ const SearchForm = ({
 					{!hideDatePicker && (
 						<PreviousDateRangePicker
 							emphasis="low"
-							selectedDates={selectedDates}
+							selectedDates={[startDate, endDate]}
```
SpennyNDaJets, line 148: Think this is a pre-existing issue, but may be making it more obvious that we set the end date when reloading the page, or fetching more results

https://github.com/highlight/highlight/assets/17744174/94a0df95-cd8e-4906-946f-4da3480e3b70

SpennyNDaJets, line 148: Think this is a pre-exisiting issue, but may be more obvious with the changes to fetch newer logs. The time frame gets updated from "Last X" to a custom time frame 

https://github.com/highlight/highlight/assets/17744174/51dda78b-ca1d-422b-a176-e6cbc7761b0e

frontend/src/components/FeedResults/FeedResults.tsx
```diff
@@ -0,0 +1,59 @@
...
+							emphasis="high"
+							trackingId="SessionsFeedMore"
+							className={style.moreButton}
+							onClick={onClick}
+						>
+							<Box display="flex" alignItems="center" gap="8">
+								<IconOutlineArrowNarrowUp />
+								<Text>
+									{more}
+									{type === 'logs'
```
SpennyNDaJets, line 45: Should we include the 50+ for more types than logs if we happen to hit that case?

SpennyNDaJets (changes requested):
Requested changes since the weird behavior, but feel free to push back

There is some weird stuff with the times when switching between pages after hitting refresh

https://github.com/highlight/highlight/assets/17744174/2d6de7db-5168-4bcd-a27f-7c0587f0896a

frontend/src/components/FeedResults/FeedResults.tsx
Vadman97: the logs query is paginated in such a way that we don't know how many more logs than 50 we have, hence i'm showing 50+. for sessions and errors, we have the exact count of total new sessions / errors.

frontend/src/pages/LogsPage/SearchForm/SearchForm.tsx
Vadman97: fixed in b2c888243c11c97daa64e53fe06959eb21343b70

Vadman97: fixed in https://github.com/highlight/highlight/commit/b2c888243c11c97daa64e53fe06959eb21343b70

Vadman97:
> Requested changes since the weird behavior, but feel free to push back
> 
> There is some weird stuff with the times when switching between pages after hitting refresh (seems like a regression from prod)
> 
>  Screen.Recording.2023-08-10.at.10.32.55.AM.mov

great catch, seems like this only happens in the production build. looking into it, will re-request review once this is fixed.

SpennyNDaJets (approved):
Thanks for fixing! Looks good to me, just saw some lint errors in the build

```
/home/runner/work/highlight/highlight/frontend/src/pages/ErrorsV2/SearchPanel/SearchPanel.tsx
  Error:   43:3  error  'searchQuery' is assigned a value but never used. Allowed unused vars must match /^_/u     @typescript-eslint/no-unused-vars
  Error:   44:3  error  'setSearchQuery' is assigned a value but never used. Allowed unused vars must match /^_/u  @typescript-eslint/no-unused-vars
```

PR #6279, opened 2023-08-09
Native SVG components for reflame

lewisl9029:
## Summary

Reflame now has native SVG components! It's API compatible with the default SVGR configuration that we use here (i.e. components live at the `ReactComponent` named export), but under the hood it generates a SVG sprite and lightweight shim components that refer to nodes within the SVG sprite (more on this approach here: https://benadam.me/thoughts/react-svg-sprites/) for much better build performance and runtime performance than SVGR, which renders all SVG nodes as React nodes.

All the icons in the new preview look identical so far, but let me know if you notice any irregularities!


Vadman97 (approved):
awesome! and the e2e test is working in forks again thanks to @ccschmitz :)

PR #6280, opened 2023-08-09
Create table states

SpennyNDaJets:
## Summary
Create the FullRow component to allow a higher level component to implement different states more easily/

Example:
```
<Table>
    <Table.Head>
         <Table.Row gridColumns={["25%", "50px", "auto"]}>
            {COLUMNS.map((column) => (
                <Table.Header>
                    {column.name}
	        </Table.Header>
            ))}
        </Table.Row>
    </Table.Head>
    <Table.Body>
        {loading && <Table.FullRow>Loading...</Table.FullRow>}
        {error && <Table.FullRow>{error}</Table.FullRow>}
    </Table.Body>
</Table>
```

1) View the Table storybook

https://www.loom.com/share/7702a1fecbff401f9606d46e0befbc53?sid=4ca408e4-a50f-429c-b793-27034809ff78

N/A - not being used


packages/ui/src/components/Table/Body/Body.tsx
```diff
@@ -1,12 +1,68 @@
...
+	</Row>
+)
+
+const EmptyBody: React.FC = () => (
+	<Row gridColumns={['100%']}>
+		<Cell justifyContent="center" my="8">
+			<Text>No results</Text>
+		</Cell>
+	</Row>
+)
```
ccschmitz, line 68: Nice to have these different states componentized, but wondering if we want to keep them hidden inside the implementation of `Table.Body`. I could see us wanting to customize these in the future. e.g. if we have an empty state, maybe we want to show a CTA to finish integrating.

What do you think about exporting these but forcing people to render them explicitly inside `Table.Body` so they can override them if needed? Also, maybe we provide something like `Table.NoContent` that provides the `Row` + `Cell` and takes in children for content to display in the center.

Nothing blocking here, just a few thoughts - take or leave them!

packages/ui/src/components/Table/Cell/Cell.tsx
```diff
@@ -1,21 +1,22 @@
 import React from 'react'
-import { Box } from '../../Box/Box'
+import { Box, BoxProps } from '../../Box/Box'
 
 import * as styles from './styles.css'
 
-export type Props = {
+export interface Props extends BoxProps {
```
ccschmitz, line 6: Nit, but what do you think about `Omit`ing the flex attributes that shouldn't be overridden?

packages/ui/src/components/Table/Table.tsx
```diff
@@ -22,9 +22,16 @@ type Props = {
 const TableComponent: React.FC<Props> = ({ children, loading, error }) => {
 	return (
 		<Box className={styles.container}>
-			{loading && 'Loading...'}
-			{!!error && error}
-			{!loading && !error && children}
+			{Children.map(children, (child: any) => {
```
ccschmitz, line 25: I think we are going to get this warning on all subsequent PRs. Perhaps this is another reason to have people explicitly render `Table.Loading` and `Table.Error` instead of trying to support the `loading` and `error` props on the top-level 🤔

ccschmitz (changes requested):
Looks great, @SpennyNDaJets! I like the composability of the components and am wondering if we want to hang on to that by avoiding the top-level `loading` and `error` props that control the main content rendering. Let me know what you think about my thoughts there. Even if we decide to keep the existing API, I think we should address the type error on `Children.map` since that will produce noise in our PRs.

ccschmitz (approved):
Looks _great_ to me!

PR #6283, opened 2023-08-09
Add additional columns to Services

SpennyNDaJets:
## Summary
Add the following fields to the `Service` table:
- BuildPrefix: path that code is moved to for deployment that is displayed on the error stacktrace
- GithubPrefix: path that is removed in deployment that is needed to fetch from GitHub appropriately
- LastSeenVersion: string representation of the version last seen in logs
- ErrorDetails: details around why a service may be in the error state

1) Test the `service` query in postman to return fields
2) Test the `editServiceGithubSettings` mutation in postman to edit fields

https://www.loom.com/share/0f9349c7cbfa46a79973c175e9a1a979?sid=c1c4ebf2-1567-487d-ade8-60c7145aac68

N/A - not being used

### DIFF
--- a/backend/model/model.go
+++ b/backend/model/model.go
@@ -2007,7 +2007,7 @@ type Service struct {
 	BuildPrefix     *string
 	GithubPrefix    *string
 	LastSeenVersion *string
-	ErrorDetails    *string
+	ErrorDetails    pq.StringArray `gorm:"type:text[]"`
 }
 
 type LogAlert struct {
--- a/backend/model/model.go
+++ b/backend/model/model.go
@@ -2000,10 +2000,14 @@ func (obj *SessionAlert) SendAlerts(ctx context.Context, db *gorm.DB, mailClient
 
 type Service struct {
 	Model
-	ProjectID      int                       `gorm:"not null;uniqueIndex:idx_project_id_name"`
-	Name           string                    `gorm:"not null;uniqueIndex:idx_project_id_name"`
-	Status         modelInputs.ServiceStatus `gorm:"not null;default:created"`
-	GithubRepoPath *string
+	ProjectID       int                       `gorm:"not null;uniqueIndex:idx_project_id_name"`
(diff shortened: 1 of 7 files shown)


backend/private-graph/graph/schema.graphqls
```diff
@@ -2116,7 +2124,13 @@ type Mutation {
 		is_opt_out: Boolean!
 		project_id: Int
 	): Boolean!
-	editService(id: ID!, project_id: ID!, github_repo_path: String): Service
+	editServiceGithubSettings(
```
SpennyNDaJets, line 2127: Renamed to github settings related since this does some other logic around setting/reseting null values

backend/model/model.go
```diff
@@ -2000,10 +2000,14 @@ func (obj *SessionAlert) SendAlerts(ctx context.Context, db *gorm.DB, mailClient
...
-	Status         modelInputs.ServiceStatus `gorm:"not null;default:created"`
-	GithubRepoPath *string
+	ProjectID       int                       `gorm:"not null;uniqueIndex:idx_project_id_name"`
+	Name            string                    `gorm:"not null;uniqueIndex:idx_project_id_name"`
+	Status          modelInputs.ServiceStatus `gorm:"not null;default:created"`
+	GithubRepoPath  *string
+	BuildPrefix     *string
+	GithubPrefix    *string
+	LastSeenVersion *string
+	ErrorDetails    *string
```
et, line 2010: Can this be an array of strings? That way we can handle multiple reasons for their being an issue.

et (approved):

backend/model/model.go
SpennyNDaJets: Just pushed up these changes in https://github.com/highlight/highlight/pull/6283/commits/1719129587f426583f5888ef1753b3a457bb2fec

et: Awesome!

PR #6284, opened 2023-08-09
ui fixes on log alerts page

mayberryzane:
## Summary
- fix bug where search box wouldn't let you enter in more than one value
- move search box to top of page
- only rerender histogram when the search form submits (avoid rerendering while the user is actively typing)

- click tested locally

- no


SpennyNDaJets (approved):
Looks good to me!

image

PR #6287, opened 2023-08-09
Fix nested and wildcard attribute matching on logs

ccschmitz:
## Summary

Fixes surfacing of matched attributes on nested/wildcard attributes and highlighting on wildcard attributes.

Screenshot 2023-08-09 at 4 09 10 PM

Click test + unit tests.

N/A


et (approved):
Tested on reflame and looks great! Thanks for fixing this.

I noticed another issue that's closely related but I confirmed it's not a regression from this PR. #6295

ccschmitz:
> I noticed another issue that's closely related but I confirmed it's not a regression from this PR. #6295

@et I took care of that in https://github.com/highlight/highlight/pull/6287/commits/09552dfaf3159b062c3d8754162082ecf280f0da.

PR #6299, opened 2023-08-10
Revert "Revert "Fix duplicate trace rows and reenable writing trace""

ccschmitz:
Reverts highlight/highlight#6297 w/ a few updates:

1. Submits a message for each trace row rather than batching up multiple trace rows in a single message.
2. Avoids creating traces for logs.


backend/worker/kafka_worker.go
```diff
@@ -284,11 +284,8 @@ func (k *KafkaBatchWorker) flushTraces(ctx context.Context) error {
 			case lastMsg = <-k.BatchBuffer.messageQueue:
 				switch lastMsg.Type {
 				case kafkaqueue.PushTraces:
-					traceRows = append(traceRows, lastMsg.PushTraces.TraceRows...)
-					received += len(lastMsg.PushTraces.TraceRows)
-				}
-				if received >= k.BatchFlushSize {
-					return
```
Vadman97, line 291: i believe we still want to keep this. what this does is make sure that if another go routine keeps adding to the `messageQueue` as we pull off of it (line 284), we'll stop once we hit the flush size

Vadman97 (approved):
🚢

Vadman97:
🎉 🚢

PR #6301, opened 2023-08-10
send a trace with blank resource attributes

et:
## Summary

This PR ensures that our public graph service information doesn't leak into a vercel log drain. The previous way of slicing this out was checking if this is a frontend log (#6273). However, that approach wasn't robust since vercel logs aren't necessarily a frontend log.

Instead, we blank out the resource data (use `StartTraceWithoutResourceAttributes` instead of `StartTrace`) at the start of the trace (instead of chopping it off at the OTEL processing side). This approach makes more sense given we understand very clearly that this was sent from a 3rd party. The same logic is applied for submitting a frontend console log (`SubmitFrontendConsoleMessages`) and for `SubmitHTTPLog`

This PR also fixes some bugs with how the vercel logs were being formatted. Here is the payload that vercel sends us:

Screenshot 2023-08-10 at 1 23 28 PM

* http.method was the status code, not the actual method
* http.method (GET|POST) wasn't even included in the log
* hardcoded `service.name` to `vercel-log-drain` (I considered using using the vercel project id but wasn't sure if that was confusing. We can revisit this if we want to adjust).
* set `service.version` to be the vercel deployment id

Here is the rest of the information in that vercel log payload:

I've also fixed an issue whereby blank values are scrubbed from the output (`code.function` is a good example of this).

See README for how to set this up locally. 

Before:
Screenshot 2023-08-10 at 1 37 32 PM

After:

Screenshot 2023-08-10 at 1 29 23 PM

### DIFF
--- a/backend/otel/extract.go
+++ b/backend/otel/extract.go
@@ -172,9 +172,7 @@ func extractFields(ctx context.Context, params extractFieldsParams) (extractedFi
 		}
 
 		for key, value := range util.FormatLogAttributes(ctx, k, v) {
-			if value != "" {
-				fields.attrs[key] = value
-			}
+			fields.attrs[key] = value
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


SpennyNDaJets (approved):
Paired with ET on this and no concerns here!

PR #6302, opened 2023-08-10
retry batched processing on failure

Vadman97:
## Summary

image
Noticed some failures to write to clickhouse due to high load which could resolve on retries.
Adding some logging here as well.

Local deploy ingesting logs.

Monitoring deploy.


mayberryzane (approved):
LGTM - could we also add some exponential backoff in case of timeout errors?

PR #6308, opened 2023-08-10
make error object group matching race-safe

Vadman97:
## Summary

Following conversation in [Slack](https://highlightcorp.slack.com/archives/C02CJANPHQS/p1691630818829679), 
this PR addresses a race condition in error grouping logic that would result in duplicate errors groups.
* Switches the redis locking implementation to the Redlock algorithm as a more robust implementation.

error grouping unit test, validating that for a burst of randomized errors on a fresh instance, all go to the same group.
Screenshot 2023-08-10 at 5 50 34 PM

old lock implementation
image

with redlock which has better correctness gurantees
Screenshot 2023-08-10 at 5 01 40 PM

No


backend/public-graph/graph/resolver.go
```diff
@@ -372,7 +372,28 @@ func (r *Resolver) getMappedStackTraceString(ctx context.Context, stackTrace []*
...
+func (r *Resolver) GetOrCreateErrorGroup(ctx context.Context, errorObj *model.ErrorObject, matchFn func() (*int, error)) (*model.ErrorGroup, error) {
+	match, err := matchFn()
+	if err != nil {
+		return nil, err
+	}
+	// no match means we are planning to create an error group
+	// we should lock the write path to make sure we do not write duplicate error groups
+	if match == nil {
+		key := fmt.Sprintf("GetOrCreateErrorGroup-project-%d", errorObj.ProjectID)
+		if mutex, err := r.Redis.AcquireLock(ctx, key, time.Minute); err == nil {
```
mayberryzane, line 384: if it isn't able to acquire the lock within the timeout, we'd expect duplicate error groups?

Vadman97: yeah, should we return an error instead so that we can try again?

Issue #6313, opened 2023-08-11
Log filters are not working as expected when multiple filters are applied


ccschmitz:
![Image](https://github.com/highlight/highlight/assets/308182/6abd52fb-c00f-42c5-b5bc-8c0513b4c014)

Note how the `code.lineno` filter is not filtering out logs that don't explicitly match.

ccschmitz:
@Vadman97 you mentioned I should tag you on this. Is it something you're planning to look into, or would you like me to?

et:
@ccschmitz - I was curious if this was a backend issue but I wrote a [test](https://github.com/highlight/highlight/pull/6314) to simulate the behavior and it passes. So I think this is a frontend issue.

Vadman97:
to add here, experienced this crash when querying for `code.lineno: *1` and updating the result
<img width="627" alt="Screenshot 2023-08-11 at 10 52 09 AM" src="https://github.com/highlight/highlight/assets/1351531/b6021a78-b41b-40bd-a0e7-7a8b5eaa169e">
<img width="647" alt="Screenshot 2023-08-11 at 10 52 03 AM" src="https://github.com/highlight/highlight/assets/1351531/22839281-3124-44bb-b57d-bc85b64dd1b1">

ccschmitz:
The issue mentioned in https://github.com/highlight/highlight/issues/6313#issuecomment-1675164274 was resolved by #6287.

ccschmitz:
# Logs SQL Debugging

Here is a query that consistently returns results that don't match all the filters:

```
level:warn code.lineno:*08 os.type:linux
```

This ended up producing the following SQL:

```sql
SELECT Timestamp, UUID, SeverityText, Body, LogAttributes, TraceId, SpanId, SecureSessionId, Source, ServiceName, ServiceVersion FROM (SELECT Timestamp, UUID, SeverityText, Body, LogAttributes, TraceId, SpanId, SecureSessionId, Source, ServiceName, ServiceVersion FROM logs WHERE ProjectId = 1 AND toUInt64(toDateTime(Timestamp)) <= ? AND toUInt64(toDateTime(Timestamp)) >= ? AND (SeverityText = ?) AND (LogAttributes[?] LIKE ? OR LogAttributes[?] = ?) ORDER BY Timestamp DESC, UUID DESC LIMIT 51) AS logs_window ORDER BY Timestamp DESC, UUID DESC
```

Here are the arguments for the SQL:

```
[...]

PR #6314, opened 2023-08-11
write a test for multiple filters

et:
## Summary

Regression test for #6313 to confirm it's not a backend issue.

Just a test so the build passes.

N/A

### DIFF
--- a/backend/clickhouse/logs_test.go
+++ b/backend/clickhouse/logs_test.go
@@ -1006,6 +1006,27 @@ func TestReadLogsWithMultipleFilters(t *testing.T) {
 	assert.Len(t, payload.Edges, 2)
 }
 
+func TestReadLogsWithMultipleFilters(t *testing.T) {
+	ctx := context.Background()
+	client, teardown := setupTest(t)
+	defer teardown(t)
+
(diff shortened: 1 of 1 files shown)


SpennyNDaJets (approved):

backend/clickhouse/logs_test.go
```diff
@@ -964,6 +964,27 @@ func TestReadLogsWithServiceVersionFilter(t *testing.T) {
...
 
+func TestReadLogsWithMultipleFilters(t *testing.T) {
+	ctx := context.Background()
+	client, teardown := setupTest(t)
+	defer teardown(t)
+
+	now := time.Now()
+	rows := []*LogRow{
+		NewLogRow(now, 1,
+			WithLogAttributes(map[string]string{"code.lineno": "162", "os.type": "linux"})),
```
SpennyNDaJets, line 975: Should we have a row that does not fall into this filter, but only one of them?

et: Good idea. Added in 965ddf2d7

Vadman97:
already in main

PR #6316, opened 2023-08-11
update logs cache keys

Vadman97:
## Summary

Digging into why logs results are sometimes incorrect and out of order, found that
the results from our `fetchMore` query are sometimes using stale variables and hence showing
previous results for a page incorrectly.

Reflame preview trying but unable to reproduce the behavior described [here](https://highlightcorp.slack.com/archives/C02N2AZS8BV/p1691471534516159).

No

### DIFF
--- a/frontend/src/util/graph.tsx
+++ b/frontend/src/util/graph.tsx
@@ -121,7 +121,13 @@ const cache = new InMemoryCache({
 		},
 		Query: {
 			fields: {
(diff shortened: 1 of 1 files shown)


et (approved):
Paired with @Vadman97  on this one. Super confusing but here are the docs we were looking at to motivate this change: https://www.apollographql.com/docs/react/pagination/key-args#setting-keyargs

PR #6324, opened 2023-08-14
backend changes for reading sessions from clickhouse

mayberryzane:
## Summary
- queries clickhouse for sessions, field suggestions, and all field types
- adds `SessionCreatedAt` to fields table to enable filtering field types and suggestions by date range
- uses the opensearch resolvers as an entrypoint, but uses the clickhouse logic if certain parameters are passed in
- will follow with a PR to enable this from the frontend (will use a local storage flag, at least until data is backfilled w/ new fields schema)

- click tested locally

- no


ccschmitz (approved):
Very cool! Code looks good and click tested as well.

PR #6338, opened 2023-08-15
accept metadata on errors across all SDKs

Vadman97:
## Summary

We support custom metadata displayed on errors in highlight, and support
metadata passed to the browser javascript client via `consumeError`.
This PR adds support for such metadata across all of our backend SDKs.
https://discord.com/channels/1026884757667188757/1136108506899034172/1136108506899034172

Unit testing for each SDK.

New versions of the node.js based SDKs, ruby SDK, and python SDK.


sdk/highlight-node/src/client.ts
```diff
@@ -110,11 +111,12 @@ export class Highlight {
 		name: string,
 		value: number,
 		requestId?: string,
-		tags?: { name: string; value: string }[],
```
ccschmitz, line 113: Do we need to mention anything about `tags` being removed on `recordMetric`? Also, are there some docs we should update?

sdk/highlight-ruby/highlight/test/highlight_test.rb
```diff
@@ -22,7 +22,7 @@ def test_record_log
 
   def test_trace
     Highlight::H.new('qe9y4yg1')
-    Highlight::H.instance.trace(nil, nil) do
+    Highlight::H.instance.trace(nil, nil, attrs: { "some.attribute" => 12 }) do
```
ccschmitz, line 25: What do you think about adding an assertion that these attributes are coming through? I might be misunderstanding the syntax here, but it looks like this invocation of `trace` is assuming `attrs` is defined as a keyword argument, but it [doesn't appear to be defined that way](https://github.com/highlight/highlight/pull/6338/files#diff-dc88d4e75408f5ac49bd6cc7c0dbbf54a2aef7faf81e7a1fc3ce9c0719575661R46).

sdk/highlight-py/highlight_io/sdk.py
```diff
@@ -146,6 +147,7 @@ def my_fn():
 
         :param session_id: the highlight session that initiated this network request.
         :param request_id: the identifier of the current network request.
+        :param attributes: additional metadata to attribute to this error.
```
ccschmitz, line 150: Just curious, why call it `attributes` here and `metadata` in the JS SDKs?

ccschmitz (changes requested):
Changes look good overall, but I think there might be a syntax error in the Ruby test. Also wondering if there are some docs that need to be updated now for `consumeError` and `recordMetric`.

sdk/highlight-node/src/client.ts
Vadman97: you're right - this is a breaking change changing the types like that so i'll refrain from making it here. we'll likely be changing the API for metrics more as we get to that part of the product.

sdk/highlight-py/highlight_io/sdk.py
Vadman97: opentelemetry calls these tags `attributes`. ideally we should refactor the JS sdks to be consistent with their naming.

sdk/highlight-ruby/highlight/test/highlight_test.rb
Vadman97: fixing that bug! what would be the best way to asset that? would i need to mock the opentelemetry sdk to see what arguments it is called with?

ccschmitz: Yeah, that's how I would expect to do it. All the testing I've done in the past was using RSpec, but you could use a syntax like:

```rb
expect_any_instance_of(OpenTelemetryTracer).to receive(:in_span).with('highlight-ctx', hash_including(
  "some.attribute": 12,
  "project_id": nil,
  ...
))`
```

ccschmitz: Also, no big deal if you don't want to add that right now. Just was thinking about how we could get a test failure to reproduce the issue if it wasn't being called correctly, but I think the test was already failing because of that.

ccschmitz (approved):

sdk/highlight-ruby/highlight/test/highlight_test.rb
Vadman97: you're right, good idea to have a test for this. added this and i think it's passing locally?

ccschmitz: Nice! One minor thing... Since it's an expectation that something _will_ be called, I think you want to make the assertion before you invoke the method. Maybe not a problem if it's already working, but might be worth calling before calling any methods.

ccschmitz: Oh, also, I don't see `rspec` in our `Gemfile`. I think we are using minitest, and I don't think it has the `expect_any_instance_of` mocking API. I wonder if this is working 🤔

Vadman97: oh :/ that makes a lot more sense. i was confused though because adding expect_any_instance_of wasn't complaining that it doesn't exist, but changing what i was asserting in existing didn't fail the test either. looking into how to mock with minitest and will make sure that it fails when i don't have the expected attributes passed

Vadman97: after bashing my head at ruby for the past 2 hours, got a test that actually fails when i assert on incorrect attributes :) 
image

PR #6343, opened 2023-08-15
record session ip addresses

Vadman97:
## Summary

We've heard a few times now that it would be useful to record the session IP address
in addition to the location. This will start recording the IP address, though it isn't displayed
yet in the UI.

Local deploy recording IP address.

No


Vadman97:
> Code looks good, should this be an opt-in feature or project level setting though? e.g. https://docs.sentry.io/platforms/python/guides/logging/enriching-events/identify-user/#:~:text=To%20opt%20out,any%20other%20logic.

@mayberryzane good idea - will make this an opt-in project setting for now for the few customers that have asked for this, since we aren't yet displaying this anyways.

PR #6350, opened 2023-08-16
remove frequency lookup from common code

mayberryzane:
## Summary
- load error group frequencies only when necessary
- issue submit button should use loading state instead of disabled

- click tested locally, validated frequency graphs were displaying

- no


Vadman97 (approved):

Vadman97:
@mayberryzane does this help with the instance counts here not showing correctly?
Screenshot 2023-08-15 at 5 07 54 PM

mayberryzane:
> does this help with the instance counts here not showing correctly?

This doesn't address that, only meant to speed up perf in places where the frequencies were not necessary (e.g. removing / attaching a comment / issue). Seems like the instance / user counts are also messed up in that view

PR #6355, opened 2023-08-16
Handle milliseconds in tsAbs param

ccschmitz:
## Summary

We have a `tsAbs` param which allows people to link to a specific point in a session. Sometimes we have a formatted timestamp in that param (`tsAbs=2023-01-01-12:00:00`) and others it's a raw millisecond value (`tsAbs=1234567890`). The client sets formatted strings and [our Slack alerts set it in milliseconds](https://github.com/highlight/highlight/blob/main/backend/worker/worker.go/#L1007).

I confirmed I can repro the error accessing sessions from a rage click alert and that it _does_ work when I open the same link in a PR preview.

N/A


SpennyNDaJets (approved):
Click test looks good on rage click alert

PR #6360, opened 2023-08-16
support huggingface hosted model

Vadman97:
## Summary

Switches our OpenAI embeddings usage to a self-hosted huggingface gte-large model.
image

Starts writing the new vectors and using them to group errors.

Locally grouping errors using new vectors.
Screenshot 2023-08-17 at 11 56 39 AM

Hosted model inference may be slower, but this is only for our project for now. Will monitor.

### DIFF
--- a/backend/public-graph/graph/resolver.go
+++ b/backend/public-graph/graph/resolver.go
@@ -458,7 +458,7 @@ func (r *Resolver) GetOrCreateErrorGroup(ctx context.Context, errorObj *model.Er
 	return errorGroup, nil
 }
 
-func (r *Resolver) GetTopErrorGroupMatchByEmbedding(ctx context.Context, combinedEmbedding, eventEmbedding, stackTraceEmbedding, payloadEmbedding model.Vector, threshold float64) (*int, error) {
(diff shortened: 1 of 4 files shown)


jay-khatri:
Pending the amount of effort it takes, I wonder is sharing some sort of graph comparing HF vs OpenAI w/ our grouping logic would be a cool things to report on in the upcoming post.

Vadman97:
> Pending the amount of effort it takes, I wonder is sharing some sort of graph comparing HF vs OpenAI w/ our grouping logic would be a cool things to report on in the upcoming post.

sg. any thoughts on what 'comparing' would mean? are you thinking inference / search performance, or 'quality' of the grouping? no idea how we could quantify the latter

backend/public-graph/graph/resolver.go
```diff
@@ -767,7 +768,7 @@ func (r *Resolver) HandleErrorAndGroup(ctx context.Context, errorObj *model.Erro
 		settings, _ = r.Store.GetAllWorkspaceSettings(ctx, workspace.ID)
 	}
 	var embedding *model.ErrorObjectEmbeddings
-	if settings != nil && settings.ErrorEmbeddingsGroup {
+	if util.IsDevEnv() || (settings != nil && settings.ErrorEmbeddingsGroup) {
```
SpennyNDaJets, line 771: Just want to double check that this is intended and not just left in from testing in dev?

Curious if we should rely on all dev envs being tied to embeddings vs allowing an engineer to edit/update a setting?

SpennyNDaJets (approved):
LGTM! Just one double check

PR #6367, opened 2023-08-17
add admin invite to setup flow

Vadman97:
## Summary

* updates the about you form to include new designs and team size dropdown
* adds an invite team page that asks about inviting your team and setting up auto join

[Reflame preview](https://preview.highlight.io/invite_team?%7Er_preview=%7B%22action%22%3A%22start%22%2C%22mode%22%3A%22production%22%2C%22region%22%3A%22sjc%22%2C%22variantId%22%3A%2201H80M04YY7DDQDJ3VF27T5EP2%22%2C%22variantData%22%3A%22%7B%5C%22type%5C%22%3A%5C%22branch%5C%22%2C%5C%22branch%5C%22%3A%5C%226059-ability-to-admins-to-your-workspace-in-setup-flow%5C%22%2C%5C%22githubOwnerName%5C%22%3A%5C%22highlight%5C%22%2C%5C%22githubRepositoryName%5C%22%3A%5C%22highlight%5C%22%7D%22%7D)

image

image

No


backend/hubspot/hubspot.go
```diff
@@ -95,7 +95,7 @@ func getDomain(adminEmail string) (domain string) {
 }
 
 type Api interface {
-	CreateContactForAdmin(ctx context.Context, adminID int, email string, userDefinedRole string, userDefinedPersona string, first string, last string, phone string, referral string) error
```
SpennyNDaJets, line 98: Any reason we removed the string types here? I noticed some of them were still present in the spec that uses HubspotMock adn the implementation below (i.e. `first` and `phone`).

Vadman97: in go `first, second string` is just shorthand for `first string, second string`. thought i would consolidate the `string` type to 'group' the variables together, since there are so many (probably would be cleaner to pass a struct around as an argument)

SpennyNDaJets: That makes sense, and thanks for the learning 🎉

PR #6408, opened 2023-08-22
ensure embeddings error grouping happens by project

Vadman97:
## Summary

Embedding-baseds grouping logic would happen across project, causing our errors to be matched
to errors from other projects with embedding writing turned on.

Ensuring errors are correctly grouped with the project filtering.
image

Testing with 2 projects locally to ensure query stays performant and works correctly with multiple partitions
image
image

No


backend/public-graph/graph/resolver.go
```diff
@@ -482,12 +482,11 @@ select eoe.%s <=> @embedding as score,
        eo.error_group_id                              as error_group_id
 from error_object_embeddings eoe
          inner join error_objects eo on eo.id = eoe.error_object_id
-order by 1
-limit 1;`, column), map[string]interface{}{
```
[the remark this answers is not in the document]
Vadman97: fixed in 53e040b
```
| QUERY PLAN |
| :--- |
| Nested Loop  \(cost=10.16..34.56 rows=3 width=12\) \(actual time=0.807..0.868 rows=16 loops=1\) |
|   Output: \(\(eoe.gte\_large\_embedding &lt;=&gt; '\[...]'::vector\) |
|         Filter: \(eoe.project\_id = 1\) |
|         Buffers: shared hit=1070 |
|   -&gt;  Index Scan using error\_objects\_pkey on public.error\_objects eo  \(cost=0.15..6.83 rows=1 width=8\) \(actual time=0.001..0.001 rows=1 loops=16\) |
|         Output: eo.id, eo.created\_at, eo.updated\_at, eo.deleted\_at, eo.organization\_id, eo.project\_id, eo.session\_id, eo.trace\_id, eo.span\_id, eo.log\_cursor, eo.error\_group\_id, eo.error\_group\_id\_alternative, eo.error\_grouping\_method, eo.event, eo.type, eo.url, eo.source, eo.line\_number, eo.column\_number, eo.os, eo.browser, eo.trace, eo.stack\_trace, eo.mapped\_stack\_trace, eo."timestamp", eo.payload, eo.environment, eo.request\_id, eo.is\_beacon, eo.service\_name, eo.service\_version |
|         Index Cond: \(eo.id = eoe.error\_object\_id\) |
|         Buffers: shared hit=32 |
| Planning Time: 0.134 ms |
| Execution Time: 0.889 ms |

```

backend/model/model.go
```diff
@@ -1522,6 +1523,37 @@ func MigrateDB(ctx context.Context, DB *gorm.DB) (bool, error) {
 		return false, e.Wrap(err, "Error setting session id sequence to 30000000")
 	}
 
+	if err := DB.Exec(`
+		CREATE TABLE IF NOT EXISTS error_object_embeddings_partitioned
+		(LIKE error_object_embeddings INCLUDING DEFAULTS INCLUDING IDENTITY)
+		PARTITION BY RANGE (project_id);
```
mayberryzane, line 1529: I think `PARTITION BY LIST` makes more sense here since we'd always want each partition to have a single project id

backend/model/model.go
```diff
@@ -1522,6 +1523,37 @@ func MigrateDB(ctx context.Context, DB *gorm.DB) (bool, error) {
...
+	partitionSize := 1
+	start := lastVal / partitionSize * partitionSize
+
+	// Make sure partitions are created for the next 1k projects
+	for i := 0; i < 1000; i++ {
+		end := start + partitionSize
+		sql := fmt.Sprintf(`
+			CREATE TABLE IF NOT EXISTS error_object_embeddings_partitioned_%d
+			PARTITION OF error_object_embeddings_partitioned
+			FOR VALUES FROM (%d) TO (%d);
```
mayberryzane, line 1547: if using `PARTITION BY LIST`:
https://www.postgresql.org/docs/current/sql-createtable.html#:~:text=)%20FOR%20VALUES%20IN%20(%27a%27%2C%20%27b%27)%3B

backend/public-graph/graph/resolver.go
```diff
@@ -477,14 +477,15 @@ func (r *Resolver) GetTopErrorGroupMatchByEmbedding(ctx context.Context, project
 	case model.ErrorGroupingMethodGteLargeEmbeddingV2:
 		column = "gte_large_embedding"
 	}
-	if err := r.DB.Raw(fmt.Sprintf(`
+	if err := r.DB.Debug().Raw(fmt.Sprintf(`
```
mayberryzane: debug

mayberryzane (approved):

backend/model/model.go
Vadman97: updated to be `PARTITION BY LIST (project_id);`

backend/model/model.go
Vadman97: updated to `FOR VALUES IN ('%d');`

PR #6417, opened 2023-08-23
Upgrade Ariakit to @ariakit/react

ccschmitz:
## Summary

Upgrades us to `@ariakit/react` from the original `ariakit` package. The library has been split up into [separate packages](https://github.com/ariakit/ariakit/tree/main/packages) and is [considered stable](https://newsletter.ariakit.org/p/stable) now, so we shouldn't have major breaking API changes in the future like we had with this one.

Here's an overview of the main changes in the API:

* All `state` props have been changed to `store` and use the [new component stores](https://ariakit.org/guide/component-stores).
* Some properties that were maintained in the component `state`s are not passed in as props to the components (e.g. `gutter` on menus).
* `values` from the component store need to be pulled from the state, which is not available directly on the store anymore and needs to be fetched via the `getState` getter.
* There was one bug I ran into where we would get an error from the [floating-ui](https://floating-ui.com/) unless we added rendered the `(Menu|Popover|Tooltip|Combobox)Arrow` components inside their respective components. The error prevented the popovers from being positioned correctly.
* Fixed up the `AdminForm` so that we weren't wrapping both name fields in the same label and updated the selects so the label can be passed directly to `Form.Select` rather than having to wrap it with a `NamedSection`, which I actually found broke the select behavior after the upgrade.
  * In general, I think we should monitor our usage of `NamedSection`. We often just want a label above a form element, not wrapping it. I'm not sure why we are using it so prolifically.

Local click test in the app + Storybook. Also click tested in a PR preview. Would appreciate if reviewer could preview changes and click test to ensure things are behaving as expected.

All client-side changes. Issues should be caught in click testing, but will monitor front end errors as well.


frontend/src/pages/Auth/AdminForm.tsx
```diff
@@ -187,91 +188,74 @@ export const AdminForm: React.FC = () => {
...
 			>
 				<AuthHeader>
 					<Text color="moderate">Tell us a bit more</Text>
 				</AuthHeader>
 				<AuthBody>
 					<Stack gap="12">
-						<Form.NamedSection
-							label="Your Name"
-							name={formState.names.role}
-						>
```
ccschmitz, line 201: Noticed this was incorrectly referencing `role` instead of `firstName`, also doesn't make sense to wrap multiple inputs with a label.

et (commented):
Thanks for tackling this upgrade. I was able to identify quite a few regressions pretty easily.

The reflame build is broken (#6500) so I did have to use the render preview. Let me know if you're able to reproduce what I'm seeing as there may be discrepancies between render/reflame.

ccschmitz:
Dang, sorry for asking for review with these easy-to-find issues. I've had a lot of iterations on this because of all the merge conflicts and obviously haven't been thoroughly click testing between fixes. My bad. Will get these fixed up and click test more thoroughly before submitting for review again!

packages/ui/src/components/Combobox/Combobox.tsx
```diff
@@ -11,9 +11,9 @@ import {
 	ComboboxGroupLabelProps as AriakitComboboxGroupLabelProps,
 	ComboboxPopover as AriakitComboboxPopover,
 	ComboboxPopoverProps as AriakitComboboxPopoverProps,
-} from 'ariakit/combobox'
```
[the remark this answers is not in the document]
ccschmitz: This was the same issue I mentioned with menus, popovers, and tooltips. Applied the same fix here:

https://github.com/highlight/highlight/pull/6417/files#diff-d0febc8dea69f6b6d226f9f6b62e9014f1bcbe93719b2819f03fb418fd23178aR33-R39

frontend/src/pages/Alerts/LogAlert/LogAlertPage.tsx
```diff
@@ -112,7 +112,7 @@ export const LogAlertPage = () => {
 		skip: !alert_id,
 	})
 
-	const form = useFormState<LogMonitorForm>({
+	const formStore = useFormStore<LogMonitorForm>({
```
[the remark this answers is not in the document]
ccschmitz: This was happening because of unnecessary `value` assignment trying to assign `value` to `formState.values.*`. This worked with the old form states, but doesn't appear to work with the new component stores.

ccschmitz:
@et this is ready for another look!

frontend/src/pages/Sessions/SessionsFeedV3/SessionQueryBuilder/components/SessionFeedConfigurationV2/SessionFeedConfigurationV2.tsx
```diff
@@ -140,7 +140,7 @@ export const DropdownMenu = function ({
 						>
 							Order
 						</Text>
-						<Menu gutter={8}>
+						<Menu>
```
Vadman97, line 143: the query builder input seems to be always focused as the input (for the key and value)

image

frontend/src/pages/Player/Toolbar/DevToolsWindowV2/DevToolsWindowV2.tsx
```diff
@@ -86,12 +86,13 @@ const DevToolsWindowV2: React.FC<
...
 	) as unknown as LogSource[]
 
-	const form = useFormState({
+	const formStore = useFormStore({
 		defaultValues: {
 			search: '',
 		},
 	})
-	const filter = form.getValue(form.names.search)
+	const formState = formStore.getState()
```
Vadman97, line 94: the dev tools filters require unfocusing to make the query

image

frontend/src/pages/Player/Toolbar/DevToolsWindowV2/DevToolsWindowV2.tsx
```diff
@@ -86,12 +86,13 @@ const DevToolsWindowV2: React.FC<
 		(level) => level !== LogSourceValue.All,
 	) as unknown as LogSource[]
 
-	const form = useFormState({
+	const formStore = useFormStore({
```
Vadman97, line 89: type and status buttons are not working

image

packages/ui/src/components/DatePicker/PreviousDateRangePicker.tsx
```diff
@@ -224,11 +224,13 @@ const PreviousDateRangePickerImpl = ({
 	const [endTimeIsValid, setEndTimeIsValid] = useState<boolean>(true)
 
 	const menu = useMenu()
+	const open = menu.getState().open
```
Vadman97, line 227: date picker choosing a custom date range is broken
image

frontend/src/pages/Alerts/ErrorAlert/ErrorAlertPage.tsx
```diff
@@ -82,7 +82,7 @@ export const ErrorAlertPage = () => {
 		? (findAlert(alert_id, alertsPayload) as any)
 		: undefined
 
-	const form = useFormState<ErrorAlertFormItem>({
+	const formStore = useFormStore<ErrorAlertFormItem>({
```
Vadman97, line 85: the alerts input box seems to have a blue focus box for me
image

frontend/src/pages/Player/Toolbar/DevToolsWindowV2/DevToolsWindowV2.tsx
ccschmitz: Fixed in [16d268f](https://github.com/highlight/highlight/pull/6417/commits/16d268ffbb810c398806b0450a490922385aeb10)

frontend/src/pages/Player/Toolbar/DevToolsWindowV2/DevToolsWindowV2.tsx
ccschmitz: Fixed in [b0cb293](https://github.com/highlight/highlight/pull/6417/commits/b0cb293f1c8667361c711d2219d4b4d04163994c) - good catch!

packages/ui/src/components/DatePicker/PreviousDateRangePicker.tsx
ccschmitz: Fixed in d6ed63c71b031b8e720838bc873218fc3c8bd45a

frontend/src/pages/Alerts/ErrorAlert/ErrorAlertPage.tsx
ccschmitz: I'm only seeing this in the Reflame preview for some reason 🤔 - not sure why this is happening, but don't think it's something that will make it to prod.

frontend/src/pages/Sessions/SessionsFeedV3/SessionQueryBuilder/components/SessionFeedConfigurationV2/SessionFeedConfigurationV2.tsx
ccschmitz: Yeah, some interesting behavior changes with these components after the upgrade... Looks like @mayberryzane is refactoring a lot of this in https://github.com/highlight/highlight/pull/6442 as well, so maybe I'll wait to merge that in to resolve this.

frontend/src/pages/Alerts/ErrorAlert/ErrorAlertPage.tsx
ccschmitz: I think this has something to do with Tailwind styles being applied that aren't applied in our app.

Screenshot 2023-09-07 at 7 47 36 AM

frontend/src/pages/Sessions/SessionsFeedV3/SessionQueryBuilder/components/SessionFeedConfigurationV2/SessionFeedConfigurationV2.tsx
Vadman97: yup, sorry for the false alarm! looks like some CSS is outdated in reflame - i don't see that blue selection on the render.com preview.

frontend/src/pages/Alerts/ErrorAlert/ErrorAlertPage.tsx
Vadman97: yeah, sorry for the false alarm - not an issue with the render.com preview. i think it's the same reflame issue that shows tge old gradient styling for the harold box (only in reflame). maybe some CSS is outdated / cached? cc @lewisl9029

frontend/src/pages/Player/Toolbar/DevToolsWindowV2/DevToolsWindowV2.tsx
Vadman97: still don't seem to be working  with b0cb293f1c8667361c711d2219d4b4d04163994c

ccschmitz: @Vadman97 looks like I fixed the placement issue with the popover, but there was another bug with triggering the popover consistently which I believe is fixed by b0b22f5bbb5fdbb1bc4aede3e393e86b17d97143. Sorry about that!

Vadman97 (approved):
looking great now!

frontend/src/pages/Alerts/ErrorAlert/ErrorAlertPage.tsx
lewisl9029: Ah yep, this definitely looks like a bug with Reflame's Tailwind compat. 

We're actually using UnoCSS under the hood since Tailwind itself is way too slow to fit in our performance budget (their new rust based tooling looks promising but isn't ready for prime time yet).

Will look into it and try to make the output closer to real Tailwind here!

PR #6418, opened 2023-08-23
feat: preserve scroll position when go back to events list

### LINKED ISSUE #6277: When clicking into the detail of an event and going back, preserve scroll position
"would be great if the scroll position was preserved when you went back to the events list since our sessions are quite long and have a lot of events, so it's hard to keep track of where you were looking at (also the ability to jump to the timestamp of an event without viewing the event's detail would work too)"

jainpawan21:
## Summary
1. Added `activeEventIndex` and `setActiveEventIndex` in PlayerUIContext
2. Used `initialTopMostItemIndex` prop of react-virtuoso library. Checkout this prop in [API reference](https://virtuoso.dev/virtuoso-api-reference/)
3. Added `IconSolidArrowCircleRight` as per design and `onClick` Handler
4. Added `searchItem` and `setSearchItem` to preserve search state

/claim #6277 
closes #6277 

https://github.com/highlight/highlight/assets/39362422/2df4e9c3-0730-4ffe-a580-20c8c963592f

### DIFF
--- a/frontend/src/pages/Player/components/EventStreamV2/EventStreamV2.tsx
+++ b/frontend/src/pages/Player/components/EventStreamV2/EventStreamV2.tsx
@@ -37,7 +37,12 @@ const EventStreamV2 = function () {
 		currentEvent,
 		setCurrentEvent,
 	} = useReplayerContext()
-	const { setActiveEvent, setRightPanelView } = usePlayerUIContext()
+	const {
+		setActiveEvent,
+		setRightPanelView,
+		activeEventIndex,
+		setActiveEventIndex,
+	} = usePlayerUIContext()
 	const [isInteractingWithStreamEvents, setIsInteractingWithStreamEvents] =
 		useState(false)
 	const [events, setEvents] = useState<HighlightEvent[]>([])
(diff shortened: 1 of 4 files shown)


Vadman97 (commented):
hey @jainpawan21 , thanks for working on this. the first part of the task looks great
> would be great if the scroll position was preserved when you went back to the events list

but there is a bit more to it
> also the ability to jump to the timestamp of an event without viewing the event's detail 

could we update the event card design to have the time and right arrow icon? clicking the right arrow icon should update the timestamp of the player without opening the event details ([figma design here](https://www.figma.com/file/rdBPNzn7Klk6RG1LHUCE5B/Screen-Designs?type=design&node-id=4451%3A207492&mode=design&t=CiOZjQGiRX6N4i9M-1)).

frontend/src/pages/Player/components/EventStreamV2/EventStreamV2.tsx
```diff
@@ -158,6 +163,7 @@ const EventStreamV2 = function () {
 							data={filteredEvents}
 							totalCount={filteredEvents.length}
```
Vadman97, line 164: awesome work here! could we also preserve the search state position when there is a event search term applied?
https://www.loom.com/share/50c911ee5236414eb478844dcc3f17c4

Vadman97 (approved):
:ship: pending the last point unresolved comment

Vadman97:
awesome work @jainpawan21 - thank you for your help!

rishi-raj-jain:
hey @jainpawan21,

Can you help me up with setting the frontend?

PR #6428, opened 2023-08-24
improve embedding grouping quality

Vadman97:
## Summary

Paired with @deltaepsilon to improve the quality of the error-tags embeddings matching.
* l2 distance appears to work better than cosine distance for this model
* ensures only similar errors from our project are matched
* reduce the number of duplicate similar errors shown
* scores similar errors by similarity

Local deploy error-tags page

No


deltaepsilon (approved):
Thanks for the fixes!

PR #6429, opened 2023-08-24
optimize trace flushing

Vadman97:
## Summary

Traces still flushing slowly, with the bottleneck being the clickhouse write.
Trying to increase the flush count in lieu of having more concurrent writes.

Traces writing locally
image

No

### DIFF
--- a/backend/clickhouse/clickhouse.go
+++ b/backend/clickhouse/clickhouse.go
@@ -111,6 +111,9 @@ func getClickhouseOptions(dbName string) *clickhouse.Options {
 			Password: Password,
 		},
 		DialTimeout: time.Duration(25) * time.Second,
+		Compression: &clickhouse.Compression{
+			Method: clickhouse.CompressionZSTD,
+		},
 	}
 
 	if useTLS() {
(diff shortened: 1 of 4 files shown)


SpennyNDaJets (approved):
LGTM!

Just to confirm my understanding, we moved the buffer out to be shared by the parallel workers so that we don't have `parallelBatchWorkers` writing to Clickhouse at the same time. In addition, we increased the buffer size by 10, which should reduce the amount of writes.

Vadman97:
> LGTM!
> 
> Just to confirm my understanding, we moved the buffer out to be shared by the parallel workers so that we don't have `parallelBatchWorkers` writing to Clickhouse at the same time. In addition, we increased the buffer size by 10, which should reduce the amount of writes.

that's correct! since those go-routines are actually not doing anything but fetching messages from kafka, there isn't a need for them (kafka is equally good at fetching bulk data across a larger number of partitions), so i'm trying to remove them

Vadman97:
> LGTM!
> 
> Just to confirm my understanding, we moved the buffer out to be shared by the parallel workers so that we don't have `parallelBatchWorkers` writing to Clickhouse at the same time. In addition, we increased the buffer size by 10, which should reduce the amount of writes.

that's correct! since those go-routines are actually not doing anything but fetching messages from kafka, there isn't a need for them (kafka is equally good at fetching bulk data across a larger number of partitions), so i'm trying to remove them

PR #6430, opened 2023-08-24
programmatic number of trace workers

Vadman97:
## Summary

Reduce trace writing batch size and make the number of workers programmatic. 

Local deploy writing traces.
image

No

### DIFF
--- a/backend/worker/worker.go
+++ b/backend/worker/worker.go
@@ -498,10 +498,10 @@ func (w *Worker) PublicWorker(ctx context.Context) {
 	}
 	wg.Add(traceWorkers)
 	for i := 0; i < traceWorkers; i++ {
-		traceBuffer := &KafkaBatchBuffer{
-			messageQueue: make(chan *kafkaqueue.Message, traceFlushSize),
-		}
 		go func(workerId int) {
+			traceBuffer := &KafkaBatchBuffer{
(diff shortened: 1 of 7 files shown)


backend/worker/worker.go
```diff
@@ -490,27 +490,33 @@ func (w *Worker) PublicWorker(ctx context.Context) {
...
-			),
-			Worker:              w,
-			BatchBuffer:         traceBuffer,
-			BatchFlushSize:      traceFlushSize,
-			BatchedFlushTimeout: DefaultBatchedFlushTimeout,
-			Name:                "traces",
+	traceFlushSize := DefaultBatchFlushSize
+	wg.Add(traceWorkers)
+	for i := 0; i < traceWorkers; i++ {
+		traceBuffer := &KafkaBatchBuffer{
```
mayberryzane, line 500: did we want a separate buffer per worker or should this be out of the loop?

mayberryzane (approved):

backend/worker/worker.go
Vadman97: i think the kafka message receiving is never the bottleneck here, so it doesn't matter if we have 1 goroutine here, or 32 goroutines sharing the buffer (bottleneck is always the single-threaded flush). i'm putting this back in the for loop so that we can set this to ie. 4 goroutines to have 4 concurrent flushes

Issue #6353, opened 2023-08-16
Update Alert Designs


jay-khatri:
- Add "create alert" flow
 - Add dropdown or select for user alert "type"
 - Alert overview page

julian-highlight:
All designs here: https://www.figma.com/file/rdBPNzn7Klk6RG1LHUCE5B/Screen-Designs?type=design&node-id=11260-320108&mode=design&t=ePfXNsMj4LBEYzT0-11

cc @eightants

PR #6432, opened 2023-08-25
Redesign for session alerts and creation flow

eightants:
## Summary

https://github.com/highlight/highlight/issues/6353

Implemented the new session alerts page with functionality, and merged the different alerts in one. 

Screenshot 2023-08-25 at 8 40 08 AM
Screenshot 2023-08-25 at 8 40 16 AM

Some stuff to fixed that is not in scope of this PR, but will need fixing:
- Form input fields, the cursor jumps to the end when trying to add text in the middle of a string
- Delete the alerts/new page, and old UI ConfigurationCard


frontend/src/pages/Alerts/components/AlertTitleField/AlertTitleField.tsx
```diff
@@ -0,0 +1,21 @@
+import { Form, FormState } from '@highlight-run/ui'
+
+import * as styles from './styles.css'
+
+const AlertTitleField = ({ form }: { form: FormState<any> }) => {
+	return (
+		<Form.Input
+			name={String(form.names.name)}
```
SpennyNDaJets, line 8: I'm a little unsure if this is pre-existing in the input, but notice that it always jumps me to the end of the string when I edit the field. For example, below I was trying to change "Alerts" to "Slack Alerts"

Uploading Screen Recording 2023-08-28 at 11.44.16 AM.mov…

SpennyNDaJets (approved):
Looks good! Left one comment below

frontend/src/pages/Alerts/components/AlertTitleField/AlertTitleField.tsx
eightants: Good catch, did some more digging on this and resolve it by setting the value field too, since this is all part of the form state.

PR #6434, opened 2023-08-25
ensure stacktraces from promises are recorded correctly

Vadman97:
## Summary

We noticed that we get browser-side errors like [this](https://app.highlight.io/1/errors/uIzopzx1FAGMCkXv0SxeoGwcK1f0?page=2&query=and%7C%7Cerror_state%2Cis%2COPEN%7C%7Cerror-field_timestamp%2Cbetween_date%2C2023-07-26T18%3A05%3A17.091Z_2023-08-25T18%3A05%3A17.091Z) recorded without a stacktrace.
The stacktrace is sent from the client as `[]` because we had fallen back to that value when the
stacktrace parsing library would error. 
That library should only error when a stack trace of `null` or `undefined` was passed in,
so we should take care of cases where that could be happening. Otherwise, we should
be explicit about raising the error about a custom-passed stacktrace.

image

New highlight.run version released.

### DIFF
--- a/sdk/client/src/index.tsx
+++ b/sdk/client/src/index.tsx
@@ -470,14 +470,7 @@ export class Highlight {
 	}
 
 	async consumeCustomError(error: Error, message?: string, payload?: string) {
-		let res: ErrorStackParser.StackFrame[] = []
-		try {
-			res = ErrorStackParser.parse(error)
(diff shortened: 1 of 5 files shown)


SpennyNDaJets (approved):
LGTM

PR #6435, opened 2023-08-25
drop errors from external hobby deployments

Vadman97:
## Summary

We've recently noticed that we have a bunch of errors for the highlight project
that seem to come from old code versions of highlight / do not have a service name associated.
We've traced this down to hobby deployments that are shipping phonehome data to our
cloud project as project 1, which means we record their errors (some of which are expected for a hobby deploy).
[This](https://app.highlight.io/1/errors/wvqgJG0ULiZ09vToq4wbnn2xH0Rb) is an example.
To filter them out, we can assume that for our production cloud deploy, all errors routed to project 1
will always be coming from one of our services, so we should check that the service name is defined.

New unit test.

No

### DIFF
--- a/backend/util/environment.go
+++ b/backend/util/environment.go
@@ -1,6 +1,9 @@
 package util
 
-import "os"
+import (
+	"os"
+	"strings"
+)
 
 var (
 	env           = os.Getenv("ENVIRONMENT")
@@ -41,6 +44,6 @@ func IsBackendInDocker() bool {
 	return InDockerGo == "true"
 }
 
-func IsOnRender() bool {
-	return DopplerConfig == "prod_aws_render"
+func IsProduction() bool {
+	return strings.HasPrefix(DopplerConfig, "prod")
 }
--- a/backend/otel/extract.go
+++ b/backend/otel/extract.go
@@ -19,6 +19,7 @@ import (
 	semconv "go.opentelemetry.io/otel/semconv/v1.17.0"
 )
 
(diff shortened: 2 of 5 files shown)


backend/util/environment.go
```diff
@@ -41,6 +44,6 @@ func IsBackendInDocker() bool {
 	return InDockerGo == "true"
 }
 
-func IsOnRender() bool {
-	return DopplerConfig == "prod_aws_render"
+func IsProduction() bool {
+	return strings.HasPrefix(DopplerConfig, "prod")
```
SpennyNDaJets, line 48: So not sure if this is relevant, but I believe some of the errors were linked to doppler configs that used docker, so would these errors still be in our project? Should they be?
Screenshot 2023-08-28 at 11 29 43 AM

SpennyNDaJets (approved):
I'm still a little hesitant to add these checks that seem like tech debt when we can filter out the errors by service name (not a perfect solution, but works for now). I don't feel strongly here and not sure the possible solution we can apply to the hobby app for those using it moving forward, so good with merging this in after addressing question below

backend/util/environment.go
Vadman97: the `IsProduction` check is just to make sure we apply this filtering logic only for our production deploy. The reason for that is simply that only our production cloud deploy is affected by this problem (where other external errors are sent to us), whereas other self-hosted stacks don't need to drop any errors from project 1.

PR #6445, opened 2023-08-28
Start writing errors in Clickhouse

mayberryzane:
## Summary
- adds error groups / objects to the datasync queue and starts writing them to clickhouse

- click tested + made sure they were written to clickhouse locally

- no, can revert if issues


backend/clickhouse/migrations/000019_create_error_objects_table.up.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE IF NOT EXISTS error_objects (
```
Vadman97, line 1: as mentioned offline, we'll want a user identifier here as well
should service name and service version be part of the schema as well?
what about custom attributes, could we store these as a map?
Screenshot 2023-08-28 at 11 30 10 AM

Vadman97 (approved):

backend/clickhouse/migrations/000019_create_error_objects_table.up.sql
mayberryzane: I think it would make the most sense to store them in a separate `error_fields` table, I may wait to do something here until we decide to support searching by these fields

mayberryzane: But yes, I plan on adding service name, version, and client_id to the error object to support those filters

PR #6450, opened 2023-08-28
deprecate LastSeenVersion from services table

et:
## Summary

We thought tracking this might be useful for debugging but instead decided against it.

Confirmed services table still loads as expected.

Will manually drop the column on prod.


SpennyNDaJets (approved):
LGTM

PR #6453, opened 2023-08-28
add service name to client config

### LINKED ISSUE #5979: Add service to client SDK

et:
## Summary

Fixes #5979

This PR wires up service information for the client SDK.

Some background:
* we already have a way of setting the `version` via the client ([docs](https://www.highlight.io/docs/sdk/client#version)).
* the result of this is stored on `Session.AppVersion`

This PR does:
* Writes `Session.AppVersion` onto `errorObject.ServiceVersion`
* Adds the ability to specify the `serviceName` via the client config
* Writes service name onto both the `session` and `errorObject` models during `ProcessPayload`

* I updated our frontend app to specify a service name (`frontend`) and a version (`1.0.0`).
* Simulated an error on the buttons page
* Confirmed the service name and version are correctly populated

Screenshot_2023-08-29_at_11_53_00_AM

* validated that serviceName is populated on the session model

Screenshot 2023-08-29 at 12 38 58 PM

* updated onboarding docs

Screenshot_2023-08-29_at_12_40_12_PM

Bumped firstload version


backend/public-graph/graph/resolver.go
```diff
@@ -2680,23 +2681,30 @@ func (r *Resolver) ProcessPayload(ctx context.Context, sessionSecureID string, e
...
+				ColumnNumber:   v.ColumnNumber,
+				OS:             sessionObj.OSName,
+				Browser:        sessionObj.BrowserName,
+				StackTrace:     &traceString,
+				Timestamp:      v.Timestamp,
+				Payload:        v.Payload,
+				RequestID:      nil,
+				IsBeacon:       isBeacon,
+				ServiceVersion: serviceVersion,
+				ServiceName:    sessionObj.ServiceName,
```
SpennyNDaJets, line 2707: Does this differ from `v.ServiceName` like for backend errors?

frontend/src/index.tsx
```diff
@@ -107,7 +107,8 @@ const options: HighlightOptions = {
 	inlineStylesheet: true,
 	inlineImages: true,
 	sessionShortcut: 'alt+1,command+`,alt+esc',
-	version: import.meta.env.REACT_APP_COMMIT_SHA || undefined,
+	version: import.meta.env.REACT_APP_COMMIT_SHA ?? '1.0.0',
```
SpennyNDaJets, line 110: Do we anticipate wanting this fallback to be updated, or more just a proof of concept that its working?

SpennyNDaJets (approved):
Looks good to me! Just a couple questions for understanding below

backend/public-graph/graph/resolver.go
et: Good question! The payload is actually different for frontend errors, they are `ErrorObjectInput` and have no concept of a service 

Screenshot 2023-08-30 at 11 12 26 AM

At this point in the code path, we're processing a session payload -- the service name and version is not available as an input here because the client isn't sending it. 

The client passes service name and version as part of the initialize session endpoint:

https://github.com/highlight/highlight/blob/a3fcd94b95f6f9757871811a4c97f8d408ddaef5/backend/public-graph/graph/resolver.go#L1049-L1123

Hence, that's why it's pulled off the session object.

PR #6459, opened 2023-08-29
Cleaning up Vercel deploy

deltaepsilon:
I had some trouble deploying to Vercel. This fixes it.

### DIFF
--- a/e2e/nextjs/package.json
+++ b/e2e/nextjs/package.json
@@ -5,10 +5,12 @@
 	"scripts": {
 		"dev": "next dev -p 3005",
 		"build": "next build",
-		"deploy": "yarn build && npx vercel build && npx vercel deploy --prebuilt",
+		"deploy": "yarn build && vercel build && vercel deploy --prebuilt",
 		"start": "next start",
-		"vercel:deploy": "yarn build && vercel build && vercel deploy --prebuilt",
 		"lint": "next lint"
+	},
+	"installConfig": {
+		"hoistingLimits": "workspaces"
 	},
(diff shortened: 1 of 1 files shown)


e2e/nextjs/package.json
```diff
@@ -5,11 +5,13 @@
...
 		"build": "next build",
-		"deploy": "yarn build && npx vercel build && npx vercel deploy --prebuilt",
+		"deploy": "yarn build && vercel build && vercel deploy --prebuilt",
 		"start": "next start",
-		"vercel:deploy": "yarn build && vercel build && vercel deploy --prebuilt",
 		"lint": "next lint"
 	},
+	"installConfig": {
+		"hoistingLimits": "workspaces"
+	},
```
Vadman97, line 14: this seems to be breaking the automatic vercel PR preview deploy

Vadman97 (approved):

e2e/nextjs/package.json
deltaepsilon: That's strange, because it fixes my local deploy.

One more mystery to untangle! 

Thanks for catching this.

deltaepsilon: @Vadman97 I did some quick looking, and I'm missing the connection. Did you see an error somewhere that queued you into `hoistingLimits` as the reason behind the failure? It looks like it was having trouble finding the an unrelated file.

image

`hoistingLimits` are localized to the folder setting them. `e2e/nextjs` shouldn't affect `highlight.io`

Issue #6465, opened 2023-08-29
Created services lack process information


et:
Looking at the records created from #6328, it seems like only the Go and Ruby SDKs are setting the `process.*` information. We need to figure out why the other SDKs aren't passing that along.

PR #6328, opened 2023-08-14
Create a service on ingest

SpennyNDaJets:

PR #6475, opened 2023-08-30
handle service updates during otel ingestion

et:
## Summary

When we create a new service during ingestion (#6328), we aren't handling updates to `process.*` information. This is important because a lot of SDKs aren't currently including that information (#6465) and when they do start including it, we should update the existing service records.

This PR changes the logic to upsert that data.

Tests added

### DIFF
--- a/backend/redis/utils.go
+++ b/backend/redis/utils.go
@@ -532,3 +532,7 @@ func (r *Client) FlushDB(ctx context.Context) error {
 	}
 	return nil
 }
+
+func (r *Client) Del(ctx context.Context, key string) (int64, error) {
+	return r.redisClient.Del(ctx, key).Result()
+}
--- a/backend/redis/utils.go
+++ b/backend/redis/utils.go
@@ -533,6 +533,9 @@ func (r *Client) FlushDB(ctx context.Context) error {
 	return nil
 }
 
-func (r *Client) Del(ctx context.Context, key string) (int64, error) {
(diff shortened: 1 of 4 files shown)


backend/redis/utils.go
```diff
@@ -532,3 +532,10 @@ func (r *Client) FlushDB(ctx context.Context) error {
 	}
 	return nil
 }
+
+func (r *Client) Del(ctx context.Context, key string) error {
```
Vadman97, line 536: will this method be used by a test for clearing a cache key?

backend/store/services.go
```diff
@@ -31,10 +35,10 @@ func (store *Store) FindOrCreateService(ctx context.Context, project model.Proje
 			service.ProcessDescription = &val
 		}
 
-		err := store.db.Where(&model.Service{
-			ProjectID: project.ID,
-			Name:      name,
-		}).FirstOrCreate(&service).Error
+		err := store.db.Clauses(clause.OnConflict{
+			Columns:   []clause.Column{{Name: "name"}, {Name: "project_id"}},
```
Vadman97, line 39: awesome, and confirmed that this is utilizing the existing `idx_project_id_name` to ensure uniqueness

Vadman97 (approved):

backend/redis/utils.go
et: Correct. The tests are down below.

PR #6495, opened 2023-09-01
Enhance logging and error handling when trying to enhance an error

SpennyNDaJets:
## Summary
Stacktraces are not being enhanced in prod due to a issue with caching. Fix the possible issue and add additional logging.

Force a server error twice, and ensure that the enhanced stacktrace is displayed both times.

N/A


backend/store/services.go
```diff
@@ -38,23 +38,22 @@ func (store *Store) UpsertService(ctx context.Context, project model.Project, na
 		err := store.db.Clauses(clause.OnConflict{
 			Columns:   []clause.Column{{Name: "name"}, {Name: "project_id"}},
 			DoUpdates: clause.AssignmentColumns([]string{"process_name", "process_version", "process_description"}),
-		}).Create(&service).Error
+		}).FirstOrCreate(&service).Error
```
SpennyNDaJets, line 41: The issue is when a service already exists, we cache the service created on line 21, which only has a few values. We still want to cache this service with all its values so it can be used by the errors.

Alternative option is to have two separate cache keys - for these functions

Vadman97: `FirstOrCreate` makes a lot of sense here, since we want this inner func to return the full service details even if the service already exists. nice job getting to the bottom of this bug!

Vadman97 (approved):

Issue #6490, opened 2023-09-01
data picker date range selection is not working on the logs page


Vadman97:
picker not accessible on the logs page
https://www.loom.com/share/e9760ad572324572b5e70bcb1c424d4a

PR #6502, opened 2023-09-01
fix: datepicker bug in selecting previous date for the date range

Sudarsh1010:
## Summary

Removed the if condition that verifies If the date picker is giving us array of Date with length of 2. Instead I added the prev value if the dates of index 0 is null or undefined or if the dates of index 1 is null or undefined

Tested on local machine and also ran the test:all script

N/A

fix: #6490 
/claim #6490 

@Vadman97


Vadman97:
looks like this is failing the reflame preview. mind running `yarn reflame-build` and committing the generated files?

PR #6514, opened 2023-09-05
update error instances table to fetch version off of error object

et:
## Summary

This PR updates the error instances table to fetch the version off of the error object instead of the session. 
Previously, backend only errors wouldn't have a version visible because they lacked a session. Per #6108, backend errors write their version to the error object and per #6453, frontend errors write the version onto the error object (in addition to the session).

* Run the go fiber e2e app ([README](https://github.com/highlight/highlight/blob/6446-ensure-error-instances-table-uses-error_object-version/e2e/go/README.md))
* Confirmed error instances table has the version shown

Screenshot 2023-09-05 at 10 58 29 AM

* Simulated a frontend error from the buttons page
* Confirmed version is visible

Screenshot 2023-09-05 at 10 56 57 AM


SpennyNDaJets (approved):
LGTM

PR #6526, opened 2023-09-05
fix panic when ingesting otel data without a project

Vadman97:
## Summary

Fix panic from log line when fields extraction would fail.

New unit test.

No


ccschmitz (approved):
Nice - thanks, Vadim!

PR #6528, opened 2023-09-05
Improve experience of viewing an error without sessions

SpennyNDaJets:
## Summary
Some errors on the instances page have a session attached, but they are not playable. Check the `excluded` value on the session to conditionally render the link

1) View an error with instances in which there it is not playable (can update an `errorObject` in Postgres by setting `excluded = true`
- [ ] Errors with excluded sessions don't have a link to a session
- [ ] Errors with sessions that are not excluded link to the session

Screenshot 2023-09-06 at 10 11 00 AM

N/A

### DIFF
--- a/frontend/src/pages/ErrorsV2/ErrorInstances/ErrorInstancesTable.tsx
+++ b/frontend/src/pages/ErrorsV2/ErrorInstances/ErrorInstancesTable.tsx
@@ -78,7 +78,7 @@ export const ErrorInstancesTable = ({ edges, searchedEmail }: Props) => {
 				let content = <>no session</>
 				let sessionLink = ''
 
-				if (session) {
+				if (session && !session.excluded) {
 					if (session.email) {
 						content = (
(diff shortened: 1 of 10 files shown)


frontend/src/pages/ErrorsV2/ErrorInstances/ErrorInstancesTable.tsx
```diff
@@ -78,7 +78,7 @@ export const ErrorInstancesTable = ({ edges, searchedEmail }: Props) => {
 				let content = <>no session</>
 				let sessionLink = ''
 
-				if (session) {
+				if (session && !session.excluded) {
```
SpennyNDaJets, line 81: Currently just showing the same message as if the session is not present, but maybe we want to display a different message. Not sure all the use cases here, but could be a callout to start recording sessions?

mayberryzane (approved):

PR #6531, opened 2023-09-06
prefer nearby brokers to avoid cross-az kafka network traffic

Vadman97:
## Summary

Turns out that we've had the ability to have our consumers prefer the nearest broker 
since we have been on [kafka 2.4 ](https://cwiki.apache.org/confluence/display/KAFKA/KIP-392%3A+Allow+consumers+to+fetch+from+closest+replica), but we have not been using the feature.

This change is based on segment io tickets https://github.com/segmentio/kafka-go/issues/415 and https://github.com/segmentio/kafka-go/issues/984 which describe a way to provide the AZ name to the client
to have the consumer groups prefer the nearest rack (avoiding cross-az traffic). 

Local deploy continues to work.
Screenshot from 2023-09-05 22-33-31

Testing the ECS rack preference is challenging locally, but should be safe to rollback with no data loss (we might just have a small delay in ingesting data).

Will be monitoring the release. Only readers should be affected, so worst case we would have an ingest delay.
Will be following up about billing costs for cross-az traffic.

### DIFF
--- a/backend/kafka-queue/kafkaqueue.go
+++ b/backend/kafka-queue/kafkaqueue.go
@@ -162,7 +162,6 @@ func New(ctx context.Context, topic string, mode Mode, configOverride *ConfigOve
 
 	pool := &Queue{Topic: topic, ConsumerGroup: groupID, Client: client}
 	if mode&1 == 1 {
-		log.WithContext(ctx).Debugf("initializing kafka producer for %s", topic)
 		pool.kafkaP = &kafka.Writer{
 			Addr:         kafka.TCP(brokers...),
 			Transport:    transport,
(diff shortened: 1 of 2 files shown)


backend/kafka-queue/kafkaqueue.go
```diff
@@ -208,6 +208,10 @@ func New(ctx context.Context, topic string, mode Mode, configOverride *ConfigOve
 			CommitInterval:        time.Second,
 			MaxAttempts:           10,
 			WatchPartitionChanges: true,
+			GroupBalancers: []kafka.GroupBalancer{
+				kafka.RackAffinityGroupBalancer{Rack: findRack()},
```
mayberryzane, line 212: if we're not doing it already somewhere else, it would be nice to add the `ReaderConfig` object as part of the "initializing kafka consumer" message so we can tell that the `findRack` is working as expected in prod

mayberryzane (approved):
🆒

PR #6539, opened 2023-09-06
remove debug canvas webworker logging in rrweb

Vadman97:
## Summary

Removes canvas web worker debug logging by default, while allowing passing
`debug: { clientInteractions: true, domRecording: true }` to enable logging.
Updates docs about manual canvas snapshot to reference the right config setting.

Reflame preview with and without H.init debug logging.
https://www.loom.com/share/8453542e758b481db81602870e1e6aff

No


SpennyNDaJets (approved):
LGTM

PR #6541, opened 2023-09-06
Fix: maven deploy undefined version in pom.xml

NgLoader:
## Summary
Fixing a maven version mismatch in deployment step.
Also moved the description to the right project.

Locally

A deployment is needed to fix this issue on maven central.

### DIFF
--- a/sdk/highlight-java/highlight-common/pom.xml
+++ b/sdk/highlight-java/highlight-common/pom.xml
@@ -9,6 +9,22 @@
 
 	<artifactId>highlight-common</artifactId>
 
+	<name>Highlight Java Common</name>
+	<url>https://github.com/highlight/highlight</url>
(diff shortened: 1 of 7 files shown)


rrweb
Vadman97, line 1: don't think this is related, and this is undoing some changes? you'll need to reset your change, do a `git submodule update --init` to update the submodule, then you shouldn't need to commit this

```bash
git reset --soft origin/main
git submodule update --init
git add -A :/
git commit -m 'Fix: maven deploy undefined version'
```

Vadman97 (approved):
should we test this by committing a tmp commit to this branch that forces the java deploy github action?

NgLoader:
> should we test this by committing a tmp commit to this branch that forces the java deploy github action?

yea

Issue #6538, opened 2023-09-06
Make the OTLP-Endpoint configurable


lksmrqrdt:
**Is your feature request related to a problem? Please describe.**
The backend of Highlight hardcoded the Open Telemetry Endpoint to a specific location, depending on the environment either `http://collector:4318` or `http://localhost:4318`. This is suboptimal, as I'm currently working on a Helm Chart for Highlight, which shouldn't rely on static hostnames.

The exact location of the issue lies here: https://github.com/highlight/highlight/blob/2ab537cbf0ae7b10215fd9b4cd26bd0dc154e0d6/backend/main.go#L237

**Describe the solution you'd like**
I would love to have an env-variable, e.g. `OTLP_ENDPOINT`, which would allow me to set up a custom endpoint, which is not `http://collector:4318` or `http://localhost:4318`.

**Describe alternatives you've considered**
For the time being, I can deviate from the usual helm template, and just adjust the service endpoint.

**Additional context**
-

Vadman97:
I think this makes sense since in the context of a helm deploy, the otlp ingest endpoint would be the collector container which would have a dynamic hostname (not the same node as the backend). Should be straightforward to add this in hobby deploy, and we can always fallback to `http://localhost:4318`  

https://github.com/highlight/highlight/blob/0dd7657bc22d56166090202f864eb9084e076233/docker/compose.hobby.yml#L18-L58

https://github.com/highlight/highlight/blob/0dd7657bc22d56166090202f864eb9084e076233/docker/.env

PR #6545, opened 2023-09-06
feat: provide a helm chart for highlight self-hosted in the hobby model

lksmrqrdt:
## Summary
Currently, deploying Highlight relies on [Docker Compose](https://github.com/highlight/highlight/blob/main/docker/compose.hobby.yml), which is totally fine for local testing, but suboptimal for deploying this on a Kubernetes Cluster.

I've tried to stay as close to the Docker Compose file as possible. The behavior of `depends_on` has been mimicked using [Init Containers](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/). The logging logic is handled using a [DaemonSet](https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/) containing Open Telemetry with a `filelog` receiver.

There are still some steps to do:
- [x] Adjust the name of the Collector Service after #6538 has been resolved
- [x] Implement Redis within the Helm Chart
- [x] Add all Init Containers for the Backend, as it currently only runs the initial migration without checking for the status of dependencies
- [ ] Fix a current issue with the migration, which occurs due to the constraints of Helm (`msg="Error creating clickhouse db instance for migrations: open //highlight/backend/clickhouse/migrations/.: no such file or directory"`)
- [ ] Finish setting up the actual backend hosting
- [ ] Add [Network Policies](https://kubernetes.io/docs/concepts/services-networking/network-policies/) to isolate private backend Services
- [ ] Make more parameters, like port numbers, configurable
- [ ] Document everything accordingly
- [ ] (Potentially) Make components optional, so you can instead e.g. use your pre-existing Postgres Database
- [ ] (Potentially) Use Sub-Charts for many components, instead of integrating them as standalone deployments
- [ ] (Potentially) Implement a GitHub Action Pipeline to build and publish a "ready-to-go" Helm Chart as Artifact

I welcome any feedback and help on this PR, as this is my very first contribution to an Open Source project. 🙏🏻

This is still a work in progress. Currently, this has only been tested by running it both on my local machine as well as in my personal AKS cluster. There are no formal tests right now, but I'm aiming for this.

This PR doesn't affect pre-existing installations.


helm/templates/influxdb-deployment.yaml
```diff
@@ -0,0 +1,44 @@
...
+        - name: influxdb
+          image: "{{ .Values.influxdb.image.repository }}:{{ .Values.influxdb.image.tag }}"
+          imagePullPolicy: {{ .Values.influxdb.image.pullPolicy }}
+          volumeMounts:
+            - name: influxdb-data
+              mountPath: /var/lib/influxdb
+          ports:
+            - containerPort: 8086
+              protocol: TCP
+          livesnessProbe:
```
anthonygregis, line 33: Should be `livenessProbe`?

lksmrqrdt: Yes, you're right. I've fixed this in the latest commit.  
Thank you!

helm/templates/postgres-deployment.yaml
anthonygregis, line 1: I'd recommend adding a subPath of `postgres` to the `postgres-data` volumeMount to prevent crashing due to lost+found directory existing in the root folder.

anthonygregis:
#6597 resolves a minor issue with the string concatenation for the migrations folder

helm/templates/backend-configmap.yaml
```diff
@@ -0,0 +1,49 @@
...
+  REACT_APP_PRIVATE_GRAPH_URI: {{ .Values.frontend.config.privateGraphUri | quote }}
+  REACT_APP_PUBLIC_GRAPH_URI: {{ .Values.frontend.config.publicGraphUri | quote }}
+  REDIS_EVENTS_STAGING_ENDPOINT: {{ include "highlight.fullname" . }}-redis:6379
+  RENDER_PREVIEW: "true"
+  SESSION_FILE_PATH_PREFIX: /tmp/
+  REACT_APP_AUTH_MODE: simple
+  IN_DOCKER: "true"
+  IN_DOCKER_GO: "true"
+  ENABLE_OBJECT_STORAGE: "true"
+  OBJECT_STORAGE_FS: "true"
```
anthonygregis, line 49: Add ON_PREM: "true" to resolve issue with opensearch setup init container

anthonygregis: Adding this triggers `projectpath.GetRoot()` to change the directory  to the `build` directory and allowing migrations to occur.

lksmrqrdt: Great observation and great work at fixing this issue at hand.  
As soon as #6538 and #6597 are merged, the requirements to complete the next major two steps should be resolved.

I'll use this information locally to focus on the remaining checkpoints later this day, as it's midnight in my time zone. Thank you once again for your great assistance in this PR. 🥇

helm/templates/backend-configmap.yaml
```diff
@@ -0,0 +1,49 @@
...
+  INFLUXDB_SERVER: {{ include "highlight.fullname" . }}-influxdb:8086
+  INFLUXDB_TOKEN: not-a-secure-token
+  KAFKA_SERVERS: {{ include "highlight.fullname" . }}-kafka:{{ .Values.kafka.config.port }}
+  KAFKA_TOPIC: dev
+  OPENSEARCH_INDEX_PREFIX: docker
+  OPENSEARCH_DOMAIN: http://{{ include "highlight.fullname" . }}-opensearch:9200
+  OPENSEARCH_DOMAIN_READ: http://{{ include "highlight.fullname" . }}-opensearch:9200
+  OPENSEARCH_PASSWORD: admin
+  OPENSEARCH_USERNAME: admin
+  OTLP_ENDPOINT: {{ include "highlight.fullname" . }}-collector:4319
```
anthonygregis, line 28: Might need to add `http://` to the front of the OTLP_ENDPOINT or it might attempt to default to localhost:8413 

Had to dig into `StartOTLP()` to discover this

helm/templates/backend-configmap.yaml
anthonygregis: Not sure if you have been working consistently with dev flag enabled but might help to flip it to production. I've noticed that when this is flipped we need to run migration either manually or via an init container. 

This becomes more of a problem when you realize the the `migrate-db` service can not actually be run unless you have already initialized all the tables once before. It looks like the runtime / service flags only get checked after all the initial services get spun up and started (ie: oauth server, directly dependent on tables existing in postgres), which becomes a blocker for running the migration service. 

This could potentially require a minor rewrite to elevate the service check above these startups, if a service like `migrate-db` is being run and not the main service there is no sense in initializing all these unused clients just to kill them afterwards.

Vadman97:
Hey @lksmrqrdt , any luck with this? Anything we could help with?

Vadman97:
hey folks, friendly bump on this @lksmrqrdt @anthonygregis

CronixZero:
Is it reasonable to expect a resolution on this pr in the next time?

Vadman97:
hey @CronixZero , our team will be taking this over soon. appreciate your patience

PR #6547, opened 2023-09-07
ignore ruby republish error

Vadman97:
## Summary

Our main branch is failing the ruby check because
the publish command does not like that we are republishing existing versions.

This makes sure that the github action will pass even if the publish fails (because the version exists).
The downside of the change is that an actual new version bump that fails to publish for another reason
will not be obvious, but it seems like a proper solution would require migrating to 

Github Action

No

No


ccschmitz (approved):
Nice!

PR #6550, opened 2023-09-07
feat: solve #6538 by fetching for the new OTLP_ENDPOINT env-variable

lksmrqrdt:
## Summary

This PR solves #6538, therefore it also solves one of the prerequisites for #6545

Locally. The changes provided by this PR are easy to verify.

If the variable isn't defined or if it is empty, the code will proceed just as before and either use `http://collector:4318` on Docker Compose environments, or `http://localhost:4318` everywhere else.  
Hence, no adjustments are needed for pre-existing installations.

No

### DIFF
--- a/backend/main.go
+++ b/backend/main.go
@@ -77,6 +77,7 @@ var (
 	stripeApiKey        = os.Getenv("STRIPE_API_KEY")
 	stripeWebhookSecret = os.Getenv("STRIPE_WEBHOOK_SECRET")
 	slackSigningSecret  = os.Getenv("SLACK_SIGNING_SECRET")
+	otlpEndpoint        = os.Getenv("OTLP_ENDPOINT")
 	runtimeFlag         = flag.String("runtime", "all", "the runtime of the backend; either 1) dev (all runtimes) 2) worker 3) public-graph 4) private-graph")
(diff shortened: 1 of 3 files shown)


backend/main.go
```diff
@@ -232,9 +233,12 @@ func main() {
 	highlight.SetProjectID("1jdkoe52")
 	if !util.IsOnPrem() && util.IsDevOrTestEnv() {
 		log.WithContext(ctx).Info("overwriting highlight-go graphql / otlp client address...")
-		highlight.SetOTLPEndpoint("http://localhost:4318")
-		if util.IsBackendInDocker() {
+		if otlpEndpoint != "" {
+			highlight.SetOTLPEndpoint(otlpEndpoint)
+		} else if util.IsBackendInDocker() {
 			highlight.SetOTLPEndpoint("http://collector:4318")
```
lksmrqrdt, line 239: We could even get rid of this part. But I didn't want to disrupt pre-existing workflows.

Vadman97: nice! yeah, technically the env var should always be set now in docker to the same thing `"http://collector:4318"`, but it's fine to keep this to be safe in case we're in docker and `OTLP_ENDPOINT` is somehow not set

Vadman97 (approved):
pulled the fork locally to confirm the dev deploy works correctly. built the backend image and ran in docker to validate the hobby deploy. otlp data working great in both, thank you for your contribution!
image

PR #6564, opened 2023-09-07
Fix commandbar not showing typeahead results

ccschmitz:
## Summary

The command bar is not currently rendering results once you start typing in the input. This is happening because we aren't subscribing to state changes on the form values that drive the results, which is fixed in this PR.

Screenshot 2023-09-07 at 3 22 46 PM

N/A

Nope!


Vadman97 (approved):
reflame clicktest working great! :pray:

PR #6565, opened 2023-09-07
remove hubspot retry looping

Vadman97:
## Summary

* Remove locking since all hubspot tasks are processed serially (same partition key).
* Remove retry logic here since the tasks are retried at the kafka level.
* Add redis caching to slow hubspot requests (`getContact` and `getCompany`)

hubspot tasks are slow with a bunch of retries to fetching contacts
image
image
image

setting invalid hubspot ids locally, then checking they are reset 
image

image

image
turned into
image

Screenshot from 2023-09-07 14-57-49
image

No

No


SpennyNDaJets (approved):
image

PR #6570, opened 2023-09-08
[hotfix] Billing page controls and error tags not working

ccschmitz:
## Summary

Some more bugs were discovered after the Ariakit upgrade.

* The billing form wasn't updating when selecting different retention periods. This was happening because we weren't subscribing to updates to the values in the form. I updated a few places where we use `formStore.getState()` to `formStore.useState()`, which will create a subscription to the store values and rerender the component every time it updates.
* The error tags page was showing an error boundary. This was happening because a `Form.Input` did not have a containing `Form` component.

Local click test.

N/A

N/A


Vadman97 (approved):
[preview](https://frontend-pr-6570.onrender.com/w/1/current-plan) clicktest looking great!

PR #6573, opened 2023-09-08
fix auth error for unauthenticated users for demo sessions / logs

mayberryzane:
## Summary
- sessions results / histogram and logs resolvers were all returning authentication errors for unauthenticated users in the demo project because they were calling `getCurrentAdmin` and returning
- this change makes an admin optional for those resolvers

- click tested locally in incognito

- no

- no


SpennyNDaJets (approved):
I think click test looks good to me (i.e. no 401 error, but also no sessions which confirmed was the case when logged in)

Screenshot 2023-09-08 at 10 59 41 AM

Vadman97:
looks like we still need this fix, @mayberryzane is this ready to merge?

Issue #6543, opened 2023-09-06
Add queries for fetching possible keys and values you can filter traces on


ccschmitz:
Similar to logs, we would like to have a typeahead that helps you build a traces query. In order to implement [the designs](https://www.figma.com/file/rdBPNzn7Klk6RG1LHUCE5B/Screen-Designs?type=design&node-id=11401%3A448848&mode=design&t=pu7S5ZT6hsNhmuBM-1) we will need queries for fetching and filtering possible keys and values for trace attributes.

PR #6575, opened 2023-09-08
Add filters on traces search

ccschmitz:
## Summary

Extracts the logs search filter components from `LogsPage` and sets them on the traces page. The filters are still pulling typeahead results from logs until we create queries for fetching keys and values (#6543).

Screenshot 2023-09-08 at 9 55 05 AM

Note that much of the diff is because of moving/renaming things to de-logify things.

Local + PR preview click test. Tried to ensure all behavior still worked still for logs search.

This is all only available for the main Highlight project right now. We should monitor performance and click test, but the changes should be isolated.

Would like to know what @julian-highlight thinks of how I set up just the table body to scroll with fixed headers, but he can take a look after this is deployed.


frontend/src/components/Search/SearchForm/SearchForm.tsx
```diff
@@ -98,42 +132,17 @@ const SearchForm = ({
 					initialQuery={initialQuery}
 					startDate={startDate}
 					endDate={endDate}
+					// TODO: Update so it's not just logs_keys
```
ccschmitz, line 135: Will handle this in a future PR.

deltaepsilon (approved):

PR #6582, opened 2023-09-10
feat: Support simple password authentication for the self-hosted deploy

### LINKED ISSUE #6483: support simple password authentication for the self-hosted deploy
* Add another auth mode alongside [SimpleAuth](https://github.com/highlight/highlight/blob/3f2065161cc25fc1af0c0a342c4df783fe37de38/backend/private-graph/graph/middleware.go#L98-L116) to the private graph middleware that checks that the user is logged in with a password set in the docker compose stack
* If a password is not set in the docker compose environment, simple auth (no auth) should remain the default
* Otherwise, the frontend would ask for the password which will be validated on the backend and create a token that can be used to authenticate further private graph requests.
* The frontend needs to be [updated](https://github.com/highlight/highlight/blob/3f2065161cc25fc1af0c0a342c4df783fe37de38/frontend/src/util/auth.tsx#L38) as well to offer a password-auth mode.

jemiluv8:
## Summary

Implemented a new Authentication mode for backend/frontend to allow anyone deploying the application to 
set a password in their environment that may be shared with "authorized" users.

## Frontend
- Support Password auth mode
- implement Password Class in frontend/src/util/auth.tsx end ensure it is used when REACT_APP_AUTH_MODE=password
  - This is similar to simple auth except that it makes a request to our public endpoint to "authenticate the" credentials given.
- On success an authorization token is set and UI works as expected 
- Add auth token to request header as is the case with firebase auth

## Backend
- implement an endpoint at /private/login to "authenticate" user using our set ADMIN_PASSWORD
- On success this endpoint returns a token and user data
- this middleware just ensures that the app may be authorized to use "protected" "endpoints" if it has a valid
   authorization token issued by us.

## Loom Demo
https://www.loom.com/share/4b5485fdfcd14805a9659c2eabe4f94e?sid=2bd72638-3584-462a-b79f-4f92861c9768

Fixes #6483
/claim #6483

- SET REACT_APP_AUTH_MODE=password in ".env"
- SET ADMIN_PASSWORD in docker/.env
- RUN the application with docker using the ./run.sh script in the docker folder
- Visit https://localhost:3000/sign_in
- Try logging in with an incorrect(not equal to ADMIN_PASSWORD) password and observe that it fails
- Try logging in with admin password and observe a successful login with any email but using the ADMIN_PASSWORD set in the previous steps succeeds

## FURTHER TESTING (TO ENSURE SimpleAuth is not broken)
- Reset REACT_APP_AUTH_MODE=simple (NO AUTH)
- Observe that you are now able to login without credentials as was the case for simple auth

A new env variable called ADMIN_PASSWORD has been introduced. This is not relevant if you do not intend to use
password auth mode. 

A new option ('password') for the existing REACT_APP_AUTH_MODE environment variable has been added.
So you can now set REACT_APP_AUTH_MODE=password along with a relevant value for ADMIN_PASSWORD
to support this feature

There are no migrations

No. There are no design changes


backend/private-graph/graph/middleware.go
```diff
@@ -109,6 +119,62 @@ func (c *SimpleAuthClient) GetUser(_ context.Context, _ string) (*auth.UserRecor
...
+			PhoneNumber: "+14081234567",
+			PhotoURL:    "https://picsum.photos/200",
+			ProviderID:  "",
+			UID:         "12345abcdef09876a1b2c3d4e5f",
+		},
+		EmailVerified: true,
+	}, nil
+}
+
+func (c *PasswordAuthClient) updateContextWithAuthenticatedUser(ctx context.Context, token string) (context.Context, error) {
```
jemiluv8, line 158: Same as implementation for SimpleAuthClient but token is validated

docker/.env
```diff
@@ -38,6 +38,8 @@ REACT_APP_PUBLIC_GRAPH_URI=https://localhost:8082/public
 REDIS_EVENTS_STAGING_ENDPOINT=redis:6379
 RENDER_PREVIEW=true
 SESSION_FILE_PATH_PREFIX=/tmp/
+REACT_APP_AUTH_MODE=password
```
jemiluv8, line 41: Intentionally setting REACT_APP_AUTH_MODE to password to facilitate testing.
I feel this ought to be the default

frontend/src/index.tsx
```diff
@@ -275,6 +275,14 @@ const AuthenticationRoleRouter = () => {
 	const isAuthLoading = authRole === AuthRole.LOADING
 	const isLoggedIn = authRole === AuthRole.AUTHENTICATED
 
+	useEffect(() => {
```
jemiluv8, line 278: Force redirect to sign_in if auth_mode is password but user does not have 'passwordToken'

frontend/src/util/auth.tsx
```diff
@@ -15,6 +17,9 @@ interface User {
 
 const fakeFirebaseUser: User = {
```
jemiluv8, line 18: I'm using fakeFirebaseUser for PasswordAuth mode as well with minor modification
I'm still considering a seperate object to be used specifically for PasswordAuth.

That would probably have been cleaner but verbose

frontend/src/util/auth.tsx
```diff
@@ -77,9 +82,82 @@ class SimpleAuth {
 	}
 }
 
+class PasswordAuth {
```
jemiluv8, line 85: Implements Password auth mode.

When initialized we assign current user only if a token exists.
(Would've been nice to check the validity of the token?)

attempts an "actual" signin using the email and password
sets token to sessionStorage on sucess.

signOut just removes the token from local storage

jemiluv8 (commented):
Self review completed.

jemiluv8:
@Vadman97, this PR is ready for review.

backend/main.go
```diff
@@ -520,6 +520,8 @@ func main() {
 			r.Use(public.PublicMiddleware)
 			r.Use(highlightChi.Middleware)
 
+			r.Post("/login", publicResolver.Login)
```
Vadman97, line 523: could we use the `PrivateGraph` instead for login? the 'public graph' is meant for our unauthenticated data ingest, while the 'private graph' is the api for retrieving data for our frontend.

the `/login` route itself should be unauthenticated on the private graph, since it is what will establish the auth token (it should be set up before the private graph middleware that makes private graph require authentication).

backend/private-graph/graph/middleware.go
Vadman97: nice!

backend/public-graph/graph/resolver.go
Vadman97, line 1: could we move this to the `private-graph/graph/resolver.go` per my comment in `backend/main.go`?

backend/public-graph/graph/resolver.go
```diff
@@ -157,6 +159,14 @@ type NetworkResource struct {
 	RequestResponsePairs RequestResponsePairs `json:"requestResponsePairs"`
 }
 
+type LoginCredentials struct {
+	Email    string `json:"email,omitempty"`
+	Password string `json:"password"`
+}
+
+var JwtAccessSecret = os.Getenv("JWT_ACCESS_SECRET")
+var ADMIN_PASSWORD = os.Getenv("ADMIN_PASSWORD")
```
Vadman97, line 168: could we move this and the `Login` method below to a separate `private-graph/graph/auth.go` (or the middleware.go) file that's part of the private-graph `graph` package? would make it cleaner for these to be next to the login handler in a separate auth related file

backend/public-graph/graph/resolver.go
```diff
@@ -3434,3 +3444,51 @@ func (r *Resolver) isExcludedError(ctx context.Context, errorFilters []string, e
 	}
 	return false
 }
+
+func (r *Resolver) Login(w http.ResponseWriter, req *http.Request) {
+	var credentials LoginCredentials
+
+	if ADMIN_PASSWORD == "" {
+		http.Error(w, "", http.StatusInternalServerError)
```
Vadman97, line 3452: good idea to check this. could we add a server-side `log.` to report that the password / simple password auth mode is not configured?

backend/public-graph/graph/resolver.go
```diff
@@ -3434,3 +3444,51 @@ func (r *Resolver) isExcludedError(ctx context.Context, errorFilters []string, e
...
+	}
+
+	if ADMIN_PASSWORD != credentials.Password {
+		http.Error(w, "invalid credentials", http.StatusBadRequest)
+		return
+	}
+
+	atClaims := jwt.MapClaims{}
+	atClaims["authorized"] = true
+	atClaims["claim"] = time.Now().Add(time.Hour * 24).Unix()
```
Vadman97, line 3473: can this be a constant near the `ADMIN_PASSWORD` that is describing that this is the password auth expiry?

frontend/src/util/auth.tsx
```diff
@@ -77,9 +82,82 @@ class SimpleAuth {
...
+	signOut(): Promise<void> {
+		sessionStorage.removeItem('passwordToken')
+		return Promise.resolve(undefined)
+	}
+}
+
 export let auth: SimpleAuth
 if (import.meta.env.REACT_APP_AUTH_MODE === 'simple') {
 	auth = new SimpleAuth()
+} else if (import.meta.env.REACT_APP_AUTH_MODE === 'password') {
```
Vadman97, line 159: could this use the new exported constant from `constants.ts`?

docker/.env
Vadman97: sounds good about making it the default, but that may require updating the e2e test to allow the cypress instance to log in with the default password.
some other issues with this:
- there are some places this is overridden, ie. `docker/env.sh`. you'll need to update that
- the password will need to be provided to the backend container for dev and hobby modes. see the `BUILD_ARGS` value, you'll need to update `backend.Dockerfile` as well
- for hobby mode, `./start-hobby.sh` should make sure the password is set to a non-default value. the shell script should exit if it is not set

frontend/src/util/auth.tsx
```diff
@@ -15,6 +17,9 @@ interface User {
 
 const fakeFirebaseUser: User = {
 	async getIdToken(): Promise<string> {
+		if (AUTH_MODE === 'password') {
+			return sessionStorage.getItem('passwordToken') || ''
+		}
```
Vadman97, line 22: clever! this allows `graph.tsx` to use this method to send the token, took me a sec to realize that

frontend/src/util/auth.tsx
Vadman97: > (Would've been nice to check the validity of the token?)

could we have the `/login` endpoint (or another one like `/user`) return the user details when a valid token is provided? that way the user here can be populated from that, instead of duplicating the demo user details (and this will allow for a future state where we may actually have more than one user with password auth)

Vadman97 (commented):
thank you for your contribution @jemiluv8 ! this is coming together, adding some initial thoughts

one other thought - would you mind recording a loom.com video testing this in your environment? would be great to see the frontend making private graph requests that are authenticated vs. that fail if you are not sending the token

backend/main.go
jemiluv8: use PrivateGraph for "login"

backend/public-graph/graph/resolver.go
jemiluv8: Done.

backend/public-graph/graph/resolver.go
jemiluv8: Done

backend/public-graph/graph/resolver.go
jemiluv8: Done

frontend/src/util/auth.tsx
jemiluv8: Done

backend/public-graph/graph/resolver.go
jemiluv8: Done

backend/main.go
```diff
@@ -450,6 +451,8 @@ func main() {
 			r.Get("/assets/{project_id}/{hash_val}", privateResolver.AssetHandler)
 			r.Get("/project-token/{project_id}", privateResolver.ProjectJWTHandler)
 
+			r.Get("/validate-token", privateResolver.ValidateAuthToken)
```
jemiluv8, line 454: And endpoint to validate token. This is currently used when PasswordAuth client on the frontend is initialized.
The intention is to assume the token is valid initially but validate in the background and log user out if we find
out token is invalid. This might be redundant because when a private graph request fails with 401, the app signs
the user out anyway and the same "effect" is achieved.

This however is prudent.

backend/main.go
```diff
@@ -441,6 +441,7 @@ func main() {
 			zapier.CreateZapierRoutes(r, db, &zapierStore, &rh)
 		})
 		r.HandleFunc("/slack-events", privateResolver.SlackEventsWebhook(ctx, slackSigningSecret))
+		r.Post(fmt.Sprintf("%s/%s", privateEndpoint, "login"), privateResolver.Login)
```
jemiluv8, line 444: Was going to put this on line 445 just before setting up the PrivateMiddleware.
Getting an error from chi that we can't set middleware before a route is setup.

This is nor readable though and it doesn't feel right to some extent.
The endpoint still works if we put it on line 445 but if someone
accidentally hits the login endpoint with a token or similar "things"
that are validated by the PrivateMiddleware, they will not be able to login.
So trade-offs perhaps?

Reference: https://github.com/go-chi/chi/issues/125

backend/private-graph/graph/auth.go
```diff
@@ -0,0 +1,127 @@
+package graph
```
jemiluv8, line 1: Introduced the auth.go file under private graph to contain auth related info.
It is cleaner, as you suggested.

I however ended up going a bit overboard with the "implementation". I tightened
up a few other things and modified the login to return user data as well as the token.

frontend/src/util/auth.tsx
jemiluv8: I updated the ```/login``` endpoint to return user data which is used instead of duplicating demo user details

frontend/src/util/auth.tsx
```diff
@@ -77,9 +92,119 @@ class SimpleAuth {
...
+				const userObject = JSON.parse(user)
+				this.currentUser = makePasswordAuthUser(userObject.email)
+			} catch (error) {
+				console.log('error setting user from session storage')
+			}
+			this.validateTokenAndInitializeUser(token)
+		}
+	}
+
+	async validateTokenAndInitializeUser(token: string): Promise<void> {
```
jemiluv8, line 118: This is probably redundant but seemed called for given my attempts
to tighten the password-auth implementation

docker/.env
jemiluv8: I just updated the cypress login e2e test to login with ADMIN_PASSWORD. It succeeded when I run it locally with "chrome" option.

I updated the BUILD_ARGS to set ADMIN_PASSWORD

Also modified the start-hobby.sh to exit with appropriate error message if ADMIN_PASSWORD is not set.

Only issue is the backend.Dockerfile. I'm not sure what to do there. Can you clarify.

Vadman97 (commented):
local click test is looking good with the default password! @jemiluv8 do you mind updating your branch to fix merge conflicts? i can take a look at any E2E test failures to get that sorted out. we might need to update some docs about this as well under `README.md#hobby-self-hosted`, [highlight.io/components/QuickstartContent/self-host/self-host.tsx](https://github.com/highlight/highlight/blob/24bd6f5ae8e3d47a010a0ae05e6febb59d0f017b/highlight.io/components/QuickstartContent/self-host/self-host.tsx#L4) and [highlight.io/components/QuickstartContent/self-host/dev-deploy.tsx](https://github.com/highlight/highlight/blob/24bd6f5ae8e3d47a010a0ae05e6febb59d0f017b/highlight.io/components/QuickstartContent/self-host/dev-deploy.tsx)

then we should be good to merge!

backend/private-graph/graph/auth.go
```diff
@@ -0,0 +1,136 @@
...
+	log "github.com/sirupsen/logrus"
+)
+
+type LoginCredentials struct {
+	Email    string `json:"email" binding:"required"`
+	Password string `json:"password" binding:"required"`
+}
+
+var ADMIN_PASSWORD = os.Getenv("ADMIN_PASSWORD")
+var ADMIN_PASSWORD_TOKEN_DURATION = time.Now().Add(time.Hour * 24).Unix()
```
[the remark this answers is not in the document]
jemiluv8: That's right. Will make update

backend/main.go
jemiluv8: As I explained in my original comment, the chi library prevents us from doing that because of its limitations on what can be done inside an ``r.Route function``. According to the [referenced](https://github.com/go-chi/chi/issues/125) github issue, we can't do the following:

```go
r.Route(privateEndpoint, func(r chi.Router) {
    r.Post("/login", privateResolver.Login)
    r.Use(private.PrivateMiddleware)
    r.Use(highlightChi.Middleware) 
}
```

Because we set up a route before setting up middleware via the ```r.Use```.
I tried it and there was an compiling the source.

I'm open to ideas here because our solution is now dependent on our preferred
"style". I know it doesn't look pretty but it kinda works also.

backend/private-graph/graph/middleware.go
```diff
@@ -109,6 +119,62 @@ func (c *SimpleAuthClient) GetUser(_ context.Context, _ string) (*auth.UserRecor
...
+func authenticateToken(tokenString string) (jwt.MapClaims, error) {
+	claims := jwt.MapClaims{}
+	_, err := jwt.ParseWithClaims(tokenString, claims, func(token *jwt.Token) (interface{}, error) {
+		return []byte(JwtAccessSecret), nil
+	})
+	if err != nil {
+		return claims, e.Wrap(err, "invalid id token")
+	}
+
+	exp, ok := claims["exp"].(float64)
```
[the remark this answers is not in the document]
jemiluv8: Updated.

backend/private-graph/graph/middleware.go
```diff
@@ -109,6 +119,62 @@ func (c *SimpleAuthClient) GetUser(_ context.Context, _ string) (*auth.UserRecor
...
+		UserInfo: &auth.UserInfo{
+			DisplayName: "Hobby Highlighter",
+			Email:       "demo@example.com",
+			PhoneNumber: "+14081234567",
+			PhotoURL:    "https://picsum.photos/200",
+			ProviderID:  "",
+			UID:         "12345abcdef09876a1b2c3d4e5f",
+		},
+		EmailVerified: true,
+	}, nil
```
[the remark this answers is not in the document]
jemiluv8: Updated.

backend/private-graph/graph/auth.go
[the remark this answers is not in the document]
jemiluv8: Done

backend/private-graph/graph/auth.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	user := &auth.UserInfo{
+		DisplayName: "Hobby Highlighter",
+		Email:       email.(string),
+		PhoneNumber: "+14081234567",
+		PhotoURL:    "https://picsum.photos/200",
+		ProviderID:  "",
+		UID:         uid.(string),
+	}
```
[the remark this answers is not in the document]
jemiluv8: Done.

jemiluv8:
@Vadman97, I resolved the issues raised from PR and made the updates to the documentation as you suggested.
I've since had challenge getting the login to work after merging main into my branch. Some private
graph requests are failing and that "crashes" the UI. So login is no longer successful. I am looking into
the issue.

Vadman97:
looks like e2e test caught a bug; looking
```
{"level":"error","service":"all-service","httpRequest":{"proto":"HTTP/1.1","remoteIP":"127.0.0.1:41094","requestID":"ubuntu/2CqAQ8hFAO-000806","requestMethod":"POST","requestPath":"/private","requestURL":"https://localhost:8082/private"},"stacktrace":"goroutine 28166 [running]:\nruntime/debug.Stack()\n\t/opt/hostedtoolcache/go/1.20.6/x64/src/runtime/debug/stack.go:24 +0x7a\ngithub.com/go-chi/chi/v5/middleware.Recoverer.func1.1()\n\t/home/runner/go/pkg/mod/github.com/go-chi/chi/v5@v5.0.7/middleware/recoverer.go:29 +0x110\npanic({0x35ff340, 0xc00447d920})\n\t/opt/hostedtoolcache/go/1.20.6/x64/src/runtime/panic.go:890 +0x262\ngithub.com/highlight-run/highlight/backend/private-graph/graph.authenticateToken({0xc001c7a460, 0xd1})\n\t/home/runner/actions-runner/_work/highlight/highlight/backend/private-graph/graph/middleware.go:137 +0x39e\ngithub.com/highlight-run/highlight/backend/private-graph/graph.(*PasswordAuthClient).updateContextWithAuthenticatedUser(0x4f4c070, {0x3cc2360, 0xc00447d710}, {0xc001c7a460, 0xd1})\n\t/home/runner/actions-runner/_work/highlight/highlight/backend/private-graph/graph/middleware.go:156 +0xa5\ngithub.com/highlight-run/highlight/backend/private-graph/graph.PrivateMiddleware.func1({0x3cbcb10, 0xc001414eb0}, 0xc0060c0700)\n\t/home/runner/actions-runner/_work/highlight/highlight/backend/private-graph/graph/middleware.go:229 +0x29a\nnet/http.HandlerFunc.ServeHTTP(0xc0000f2c00, {0x3cbcb10, 0xc001414eb0}, 0xc0060c0700)\n\t/opt/hostedtoolcache/go/1.20.6/x64/src/net/http/server.go:2122 +0x43\ngithub.com/go-chi/chi.(*Mux).ServeHTTP(0xc0011a81e0, {0x3cbcb10, 0xc001414eb0}, 0xc0060c0700)\n\t/home/runner/go/pkg/mod/github.com/go-chi/chi@v4.1.2+incompatible/mux.go:70 +0x122\ngithub.com/go-chi/chi.(*Mux).Mount.func1({0x3cbcb10, 0xc001414eb0}, 0xc0060c0700)\n\t/home/runner/go/pkg/mod/github.com/go-chi/chi@v4.1.2+incompatible/mux.go:298 +0xd2\nnet/http.HandlerFunc.ServeHTTP(0xc0005b3940, {0x3cbcb10, 0xc001414eb0}, 0xc0060c0700)\n\t/opt/hostedtoolcache/go/1.20.6/x64/src/net/http/server.go:2122 +0x43\ngithub.com/go-chi/chi.(*Mux).routeHTTP(0xc000c3ff80, {0x3cbcb10, 0xc001414eb0}, 0xc0060c0700)\n\t/home/runner/go/pkg/mod/github.com/go-chi/chi@v4.1.2+incompatible/mux.go:431 +0x2d1\nnet/http.HandlerFunc.ServeHTTP(0xc000c46c00, {0x3cbcb10, 0xc001414eb0}, 0xc0060c0700)\n\t/opt/hostedtoolcache/go/1.20.6/x64/src/net/http/server.go:2122 +0x43\ngithub.com/rs/cors.(*Cors).Handler.func1({0x3cbcb10, 0xc001414eb0}, 0xc0060c0700)\n\t/home/runner/go/pkg/mod/github.com/rs/cors@v1.7.0/cors.go:219 +0x2ec\nnet/http.HandlerFunc.ServeHTTP(0xc0005b3440, {0x3cbcb10, 0xc001414eb0}, 0xc0060c0700)\n\t/opt/hostedtoolcache/go/1.20.6/x64/src/net/http/server.go:2122 +0x43\ngithub.com/go-chi/chi/middleware.(*Compressor).Handler.func1({0x7fde2c578898, 0xc0036d8580}, 0xc0060c0700)\n\t/home/runner/go/pkg/mod/github.com/go-chi/chi@v4.1.2+incompatible/middleware/compress.go:213 +0x47f\nnet/http.HandlerFunc.ServeHTTP(0xc0005b3460, {0x7fde2c578898, 0xc0036d8580}, 0xc0060c0700)\n\t/opt/hostedtoolcache/go/1.20.6/x64/src/net/http/server.go:2122 +0x43\ngithub.com/go-chi/chi/v5/middleware.Recoverer.func1({0x7fde2c578898, 0xc0036d8580}, 0xc0060c0700)\n\t/home/runner/go/pkg/mod/github.com/go-chi/chi/v5@v5.0.7/middleware/recoverer.go:38 +0xf8\nnet/http.HandlerFunc.ServeHTTP(0xc0000f2648, {0x7fde2c578898, 0xc0036d8580}, 0xc0060c0700)\n\t/opt/hostedtoolcache/go/1.20.6/x64/src/net/http/server.go:2122 +0x43\ngithub.com/go-chi/httplog.Handler.func1.1({0x3cc07d0, 0xc002cb6380}, 0xc0060c0600)\n\t/home/runner/go/pkg/mod/github.com/go-chi/httplog@v0.2.5/httplog.go:63 +0x322\nnet/http.HandlerFunc.ServeHTTP(0xc000c807b0, {0x3cc07d0, 0xc002cb6380}, 0xc0060c0600)\n\t/opt/hostedtoolcache/go/1.20.6/x64/src/net/http/server.go:2122 +0x43\ngithub.com/go-chi/chi/v5/middleware.RequestID.func1({0x3cc07d0, 0xc002cb6380}, 0xc0060c0500)\n\t/home/runner/go/pkg/mod/github.com/go-chi/chi/v5@v5.0.7/middleware/request_id.go:76 +0x267\nnet/http.HandlerFunc.ServeHTTP(0xc0000f26c0, {0x3cc07d0, 0xc002cb6380}, 0xc0060c0500)\n\t/opt/hostedtoolcache/go/1.20.6/x64/src/net/http/server.go:2122 +0x43\ngithub.com/go-chi/chi/v5.(*ChainHandler).ServeHTTP(0xc00061c080, {0x3cc07d0, 0xc002cb6380}, 0xc0060c0500)\n\t/home/runner/go/pkg/mod/github.com/go-chi/chi/v5@v5.0.7/chain.go:31 +0x44\ngithub.com/go-chi/chi.(*Mux).ServeHTTP(0xc000c3ff80, {0x3cc07d0, 0xc002cb6380}, 0xc0060c0500)\n\t/home/runner/go/pkg/mod/github.com/go-chi/chi@v4.1.2+incompatible/mux.go:86 +0x252\nnet/http.serverHandler.ServeHTTP({0xc0004b01e0}, {0x3cc07d0, 0xc002cb6380}, 0xc0060c0400)\n\t/opt/hostedtoolcache/go/1.20.6/x64/src/net/http/server.go:2936 +0x494\nnet/http.(*conn).serve(0xc000a05b90, {0x3cc22b8, 0xc0016123c0})\n\t/opt/hostedtoolcache/go/1.20.6/x64/src/net/http/server.go:1995 +0x19b5\ncreated by net/http.(*Server).Serve\n\t/opt/hostedtoolcache/go/1.20.6/x64/src/net/http/server.go:3089 +0xa16\n","panic":"interface conversion: interface {} is float64, not int64","httpResponse":{"bytes":0,"elapsed":1.738444,"status":500},"timestamp":"2023-09-26T10:34:35.184573291-07:00","message":"Response: 500 Server Error - interface conversion: interface {} is float64, not int64"}
```

Vadman97:
@jemiluv8 https://github.com/highlight/highlight/actions/runs/6307171230/job/17150405161?pr=6582

`highlight/backend/private-graph/graph.authenticateToken({0xc001c7a460, 0xd1})\n\t/home/runner/actions-runner/_work/highlight/highlight/backend/private-graph/graph/middleware.go:137`
`Response: 500 Server Error - interface conversion: interface {} is float64, not int64`

jemiluv8:
@Vadman97, I just fixed it. Can you confirm?

backend/private-graph/graph/middleware.go
```diff
@@ -128,13 +128,13 @@ func authenticateToken(tokenString string) (jwt.MapClaims, error) {
...
 
-	exp, ok := claims["exp"].(float64)
+	exp, ok := claims["exp"]
 	if !ok {
 		return claims, e.Wrap(err, "invalid exp claim")
 	}
 
-	// Check if the current time is after the expiration
-	if time.Now().After(time.Unix(int64(exp), 0)) {
+	expClaim := int64(exp.(float64))
```
Vadman97, line 136: ah, so the claim exp is given as a `float64` by the jwt library... interesting. my bad confusing you here, thanks for fixing this!

highlight.io/components/QuickstartContent/self-host/self-host.tsx
```diff
@@ -25,6 +25,20 @@ REACT_APP_FRONTEND_URI=https://your-ip-address:3000
 				},
 			],
 		},
+		{
+			title: 'Password auth mode.',
+			content:
+				'The frontend for hobby deploy now defaults to using password auth. That uses a password set in your deploeyments `docker/.env file to authenticate users`' +
+				'Uodate the following environment variable to your preferred admin password. ',
```
Vadman97, line 32:
```suggestion
				'Update the following environment variable to your preferred admin password. ',
```

highlight.io/components/QuickstartContent/self-host/self-host.tsx
```diff
@@ -25,6 +25,20 @@ REACT_APP_FRONTEND_URI=https://your-ip-address:3000
 				},
 			],
 		},
+		{
+			title: 'Password auth mode.',
+			content:
+				'The frontend for hobby deploy now defaults to using password auth. That uses a password set in your deploeyments `docker/.env file to authenticate users`' +
```
Vadman97, line 31:
```suggestion
				'The frontend for hobby deploy now defaults to using password auth. That uses a password set in your deploeyments `docker/.env file to authenticate users`. ' +
```

backend/private-graph/graph/auth.go
```diff
@@ -20,15 +20,15 @@ type LoginCredentials struct {
 	Password string `json:"password" binding:"required"`
 }
 
-var ADMIN_PASSWORD = os.Getenv("ADMIN_PASSWORD")
-var ADMIN_PASSWORD_TOKEN_DURATION = time.Now().Add(time.Hour * 24).Unix()
+var AdminPassword = os.Getenv("ADMIN_PASSWORD")
+var ADMIN_PASSWORD_TOKEN_DURATION = time.Hour * 24
```
Vadman97, line 24: nit: rename to `AdminPasswordTokenDuration` to maintain same style for package variables as the rest of the go codebase.

Vadman97 (approved):
🚢 pending the small nits

backend/private-graph/graph/auth.go
jemiluv8: Gotcha. Will take note going forward.

backend/private-graph/graph/middleware.go
jemiluv8: Yeah. It also escaped me.

jemiluv8:
Just took care of the ``` small nits```.

docker/.env
Vadman97: cool, thanks for the updates here! regarding `backend.Dockerfile`, disregard - I thought the env var would have to be baked in to the image, but i see it's just passed as an env var via the `compose.hobby.yml` to the running backend container.

backend/main.go
Vadman97: ahh got it! no worries then, this is fine as is. thank you for the detailed explanation and the reference to the original discussion on go-chi

PR #6585, opened 2023-09-11
[hotfix] Forms using getState to access values incorrectly

ccschmitz:
## Summary

Many of our forms were using `getState` to access form store values and reference them in the UI or `onSubmit` hooks. However, these values often weren't updating the UI as expected and were still empty when we attempt to access them in our submit handlers because we were pulling them from a stale copy of the form state.

This PR updates all the places we used `formStore.getState` and makes sure we are using `useValue` and `useState` wherever we are trying to access values from the store/state in our UI or submit logic.

Local click test. You can jump to this page in a preview by going to `/invite_team` and repro the error in prod if you're logged in at https://app.highlight.io/invite_team.

Nope!


Vadman97 (approved):
clicktest looks great

Vadman97 (approved):
looks good now! thanks for the quick fix 
Screenshot 2023-09-12 at 2 30 10 PM

PR #6591, opened 2023-09-12
backend for traces search (filtering + field suggestions)

### LINKED ISSUE #6542: Add ability to filter traces on all supported attributes
Similar to logs, we would like to have the ability to filter traces by any attributes. This includes any standard and custom attributes we have for traces. We would also like to support pagination, deep linking, and any other functionality we have in logs.

Note that performance of querying logs is already not where we need to it be (#6329) so possibly look for opportunities to improve this in traces.

### LINKED ISSUE #6543: Add queries for fetching possible keys and values you can filter traces on

mayberryzane:
## Summary
- closes #6542 
- closes #6543 
- refactor log reading methods and structs to make them generic between logs and traces
- add trace-specific graphql resolvers
- use resolvers in front end code

- click tested logs and traces pages
https://www.loom.com/share/e47d626645394acab47311503bfd7ad8

- the signature of the logs graphql resolvers is staying the same but some of the type names are changing - not sure if this will cause downtime (e.g. if there's a delay between the backend and frontend deployment), but will monitor this in prod after deploying
- traces page is not released yet so no concern there

- no


backend/clickhouse/query.go
```diff
@@ -0,0 +1,394 @@
...
+	"github.com/ClickHouse/clickhouse-go/v2/lib/driver"
+	modelInputs "github.com/highlight-run/highlight/backend/private-graph/graph/model"
+	"github.com/highlight-run/highlight/backend/queryparser"
+	"github.com/huandu/go-sqlbuilder"
+	"github.com/nqd/flat"
+	"github.com/samber/lo"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+)
+
+type tableConfig[TReservedKey ~string] struct {
```
Vadman97, line 18: til about the `~` (allowing you to pass a custom type that is a string)

Vadman97 (approved):
local clicktest working great for logs and traces!
https://www.loom.com/share/627ce5b864a74c0dbdc6ddab6ce9375d

PR #6596, opened 2023-09-13
Allow filter errors by service for Clickhouse

SpennyNDaJets:
## Summary
Errors will start coming in with an associated service, and we want to allow users to filter by this service name/

1) Navigate to the errors page
2) Add a new filter
- [ ] Service filter is displayed
3) Click on the service filter
- [ ] Correct options are displayed
4) Select an option
- [ ] Correct errors are displayed

https://www.loom.com/share/Errors-unknown-column-servicename-13-September-2023-6270eed6ece14fe59f08bf94c07b4b2b?sid=a746ac6f-ebc3-456f-a65b-672b6bf78471

N/A

N/A


Vadman97 (approved):
not sure why, but it seems that my local search options don't populate
looks good otherwise!

Screenshot 2023-09-13 at 11 14 54 PM

PR #6597, opened 2023-09-13
fix: resolves issue with string replacement for migrations folder

anthonygregis:
## Summary

The old string replacement path could result in a double slash being prepended to the folder location for the clickhouse migrations folder. The new solution utilizes filepath join to cleanly build the path,

Deployed locally using docker

### DIFF
--- a/backend/clickhouse/clickhouse.go
+++ b/backend/clickhouse/clickhouse.go
@@ -3,8 +3,8 @@ package clickhouse
 import (
 	"context"
 	"crypto/tls"
-	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


backend/clickhouse/clickhouse.go
```diff
@@ -67,8 +67,9 @@ func RunMigrations(ctx context.Context, dbName string) {
 		log.WithContext(ctx).Fatalf("Error creating clickhouse db instance for migrations: %v", err)
 	}
 
+	migrationsPath := filepath.Join(projectpath.GetRoot(), "clickhouse/migrations")
```
Vadman97, line 70:
```suggestion
	migrationsPath := filepath.Join(projectpath.GetRoot(), "clickhouse", "migrations")
```

Vadman97 (approved):

PR #6599, opened 2023-09-13
fix: resolves issue with OTLP endpoint configuration

anthonygregis:
## Summary

Currently the OTLP endpoint can only be configured if it is not on premise and in a dev or test environment. This is not ideal if the environment is in docker and designed for production. I see no issue that the OTLP endpoint can not be changed if provided in the environment. This change would resolve an issue with #6545 requiring that it be designated as test or dev to work in helm.

Tested locally in docker and via helm on private EKS

Consider any potential implications for on premise usage but shouldn't be impacted since it's utilizing env for any overrides.


lksmrqrdt (commented):
Great observation this solves many issues at once. I do have some feedback, I want to mention here.

The logging should happen, regardless if there's a value for `otlpEndpoint`. Hence I would recommend moving it in front of the conditional check, instead of inside.

Having an `otlpEndpoint` defined while `ON_PREM`  is false will **always** overwrite the env, as `DevEnv` and `TestEnv` are hardcoded to be true. This can lead to unintentional behavior, hence I would recommend for the use of `else if` instead.

anthonygregis:
I didn't wanna remove those due to implications of broader impact but would it not be preferential to drop 241-247 all together for utilization via ENV? 

It's not as if these values couldn't be set via ENV when backend is utilized in docker and backend being utilized locally.

lksmrqrdt:
Yes, I agree with you here. [I left these lines there for the same reason](https://github.com/highlight/highlight/pull/6550#discussion_r1318295363), because I didn't want to disrupt any pre-existing workflows.

This should only cause issues if someone upgrades their Docker image, without upgrading their `.env`-file. The impact should be fairly limited.

anthonygregis:
I'll prepare another change that should hopefully allow for a cleaner upgrade path

anthonygregis:
This new change should allow for the removal of docker / localhost defaults and utilize the docker compose and env.sh script to change OTLP for both hobby usage and local usage

docker/compose.hobby.yml
lksmrqrdt, line 1: Just to keep everything in uniform with the other environmental variables, I think having `OTLP_ENDPOINT=http://collector:4318` in the `.env`-file and then referencing it in the list of line 18 (`x-backend-env: &backend-env`) would be a better choice.

Other than that, I'm really happy with this PR!

anthonygregis: SG to me, should be resolved.

Hg6f9:
Just to keep everything in uniform with the other environmental variables, I think having OTLP_ENDPOINT=http://collector:4318 in the .env-file and then referencing it in the list of line 18 (x-backend-env: &backend-env) would be a better choice.

Other than that, I'm really happy with this PR!

docker/env.sh
```diff
@@ -13,6 +13,7 @@ if [[ "$*" == *"--go-docker"* ]]; then
     export IN_DOCKER_GO=true
     echo "Using docker-internal infra."
 else
+    export OTLP_ENDPOINT=http://localhost:4318
```
Vadman97, line 16: in the `compose.hobby.yml` deploy, i believe this needs to be `http://collector:4318` since the collector and the backend are both running in docker and do not have the same `localhost`

Vadman97 (approved):
nice! callout for the cloud version, we'll need to add a `OTLP_ENDPOINT=https://otel.highlight.io:4318` env var

docker/env.sh
anthonygregis: Unless I'm mistaken hobby version should utilize .env which specifies `http://collection:4318` and the above is an override for local run. It only gets executed on the else statement of the if `--go-docker` check (It's provided when running `./run-hobby.sh` but not when running `./run-backend.sh`. 

Please let me know if I'm mistaken here and can modify.

Vadman97: ah that's correct, my mistake! appreciate the attentiveness here

Vadman97 (approved):

Vadman97:
just checked locally that the backend -> collector flow is working correctly in 
1. `make start` with doppler
2. `./run-backend` dev flow
3. `./run-hobby` after a local backend build
🚢

PR #6602, opened 2023-09-13
Optimize GitHub requests

SpennyNDaJets:
## Summary
Every time we enhance an error, we make a request to GitHub to get the owner's username, even though it isn't used. Move this request out of the installation request, and cache this value for a day since it should never change.

1) Create an issue locally with GitHub
2) View a newly created backend error and confirm it is still enhanced

N/A

N/A


Vadman97 (approved):
thank you!

PR #6614, opened 2023-09-14
update sveltekit docs for relative paths

Vadman97:
## Summary

Per a customer email, we've learned that sveltekit may generate relative CSS paths
for stylesheets which are not always valid (relying on some weird browser path behavior
to load correctly). This adds a note to the SvelteKit docs about disabling relative paths.

[Vercel preview](https://highlight-landing-du7eloxoc-highlight-run.vercel.app/docs/getting-started/client-sdk/sveltekit).

No

No


SpennyNDaJets (approved):
image

PR #6616, opened 2023-09-14
Send traces to Highlight + Datadog

ccschmitz:
## Summary

We are currently sending traces to Datadog for all of our manual tracing, but we want to start sending these to Highlight as well, and ultimately phase out DD for traces. This PR adds some new utility methods that can be used as alternatives to where we call `tracer.*`. All method calls to create spans will also send spans to Highlight.

Local click test. Can't test in PR preview because of the backend changes.

Here is some data from my local machine showing traces coming into Clickhouse:

Screenshot 2023-09-14 at 5 09 04 PM

We should ensure that all tracing information is landing in both places by checking Clickhouse + the DD UI.

N/A

### DIFF
--- a/backend/util/tracer.go
+++ b/backend/util/tracer.go
@@ -57,7 +57,7 @@ func StartSpanFromContext(ctx context.Context, operationName string, options ...
 		opt(&cfg)
 	}
 
-	hTracingDisabled, _ := ctx.Value(string(ContextKeyHighlightTracingDisabled)).(bool)
+	hTracingDisabled, _ := ctx.Value(ContextKeyHighlightTracingDisabled).(bool)
 	hTracingDisabled = hTracingDisabled || cfg.HighlightTracingDisabled
 	ctx = context.WithValue(ctx, ContextKeyHighlightTracingDisabled, hTracingDisabled)
 
--- a/backend/util/tracer.go
+++ b/backend/util/tracer.go
@@ -43,7 +43,9 @@ func (s *MultiSpan) SetOperationName(name string) {
 	}
 }
 
-const ContextKeyHighlightTracingDisabled = "HighlightTracingDisabled"
+type contextKeyType int
+
+const highlightTracingDisabled contextKeyType = iota
 
 func StartSpanFromContext(ctx context.Context, operationName string, options ...SpanOption) (MultiSpan, context.Context) {
 	var cfg SpanConfig
@@ -52,9 +54,9 @@ func StartSpanFromContext(ctx context.Context, operationName string, options ...
 		opt(&cfg)
 	}
 
-	hTracingDisabled, _ := ctx.Value(ContextKeyHighlightTracingDisabled).(bool)
+	hTracingDisabled, _ := ctx.Value(highlightTracingDisabled).(bool)
 	hTracingDisabled = hTracingDisabled || cfg.HighlightTracingDisabled
-	ctx = context.WithValue(ctx, ContextKeyHighlightTracingDisabled, hTracingDisabled)
+	ctx = context.WithValue(ctx, highlightTracingDisabled, hTracingDisabled)
 
 	for _, tag := range cfg.Tags {
 		ddOptions = append(ddOptions, tracer.Tag(string(tag.Key), tag.Value))
--- a/backend/util/tracer.go
+++ b/backend/util/tracer.go
@@ -13,13 +13,14 @@ import (
 )
 
 type MultiSpan struct {
-	ddSpan tracer.Span
-	hSpan  trace.Span
+	ddSpan  tracer.Span
+	hSpan   trace.Span
+	context context.Context
 }
 
 func (s *MultiSpan) Finish(err ...error) {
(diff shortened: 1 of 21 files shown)


backend/util/tracer.go
```diff
@@ -52,9 +54,9 @@ func StartSpanFromContext(ctx context.Context, operationName string, options ...
 		opt(&cfg)
 	}
 
-	hTracingDisabled, _ := ctx.Value(ContextKeyHighlightTracingDisabled).(bool)
+	hTracingDisabled, _ := ctx.Value(highlightTracingDisabled).(bool)
 	hTracingDisabled = hTracingDisabled || cfg.HighlightTracingDisabled
-	ctx = context.WithValue(ctx, ContextKeyHighlightTracingDisabled, hTracingDisabled)
+	ctx = context.WithValue(ctx, highlightTracingDisabled, hTracingDisabled)
```
ccschmitz, line 59: @Vadman97 is this right to assign an int as the key for the value? Would it be better to use a string?

Vadman97: I was getting this lint warning using a string
```
SA1029: should not use built-in type string as key for value; define your own type to avoid collisions (staticcheck)
```
apparently you're supposed to use a key type? https://stackoverflow.com/questions/40891345/fix-should-not-use-basic-type-string-as-key-in-context-withvalue-golint

ccschmitz: Yeah, I saw that as well and was just looking into it! Was wondering if we should assign a key type of string rather than `iota` though. Our current usage would translate to `WithValue(ctx, 0, hTracingDisabled`. Seems like we could accidentally override that pretty easily with another enum, right?

What do you think about this instead?

```
type contextKey string

const (
	ContextKeyHighlightTracingDisabled contextKey = "HighlightTracingDisabled"
)

func StartSpanFromContext(ctx context.Context, operationName string, options ...SpanOption) (MultiSpan, context.Context) {
	var cfg SpanConfig
	var ddOptions []tracer.StartSpanOption
	for _, opt := range options {
		opt(&cfg)
	}

	hTracingDisabled, _ := ctx.Value(string(ContextKeyHighlightTracingDisabled)).(bool)
	hTracingDisabled = hTracingDisabled || cfg.HighlightTracingDisabled
	ctx = context.WithValue(ctx, ContextKeyHighlightTracingDisabled, hTracingDisabled)

...
```

Vadman97: ooo my mistake! yeah that sounds better, feel free to make that change

backend/util/tracer.go
```diff
@@ -1,103 +1,128 @@
...
-	serverType Runtime
+type MultiSpan struct {
+	ddSpan tracer.Span
+	hSpan  trace.Span
 }
 
-func NewTracer(backend Runtime) Tracer {
-	return Tracer{serverType: backend}
+func (s *MultiSpan) Finish(err ...error) {
+	// TODO: Clean this up
```
Vadman97, line 19: anything else here? maybe if there is more than one error passed we can use the first one but `log` a warning?

Vadman97 (approved):
looks great! we should deploy this at a low-load time (afternoon US time) and see how much load on kafka + clickhouse increase. i suspect we may need to tune trace writing (maybe set up some sampling for our project?) as we should significantly increase the number of traces, but this will be a good test for our trace ingest infra. i think it's safe though in that it won't affect other ingest since kafka has a lot of headway right now, while clickhouse slowdowns should only affect trace writing (and it's ok if we have a backlog there)

ccschmitz:
> looks great! we should deploy this at a low-load time (afternoon US time) and see how much load on kafka + clickhouse increase. i suspect we may need to tune trace writing (maybe set up some sampling for our project?) as we should significantly increase the number of traces, but this will be a good test for our trace ingest infra. i think it's safe though in that it won't affect other ingest since kafka has a lot of headway right now, while clickhouse slowdowns should only affect trace writing (and it's ok if we have a backlog there)

Agreed! I'll hold off on merging until this afternoon. Will be ready to adjust our ingest and add sampling if needed. Thanks for looking this over!

backend/util/tracer.go
ccschmitz: Just added some logging in https://github.com/highlight/highlight/pull/6616/commits/1bd8daf39a9eabb469b5a345450bd3c0e6f79378.

backend/util/tracer.go
```diff
@@ -16,8 +18,9 @@ type MultiSpan struct {
 }
 
 func (s *MultiSpan) Finish(err ...error) {
-	// TODO: Clean this up
-	if len(err) > 0 && err[0] != nil {
+	if len(err) > 1 {
+		log.Warnf("Multiple errors passed to MultiSpan.Finish: %v", err)
```
Vadman97, line 22:
```suggestion
		log.WithContext(context.Background()).Warnf("Multiple errors passed to MultiSpan.Finish: %+v", err)
```

Vadman97 (approved):

PR #6635, opened 2023-09-18
Fix alert form initial state

SpennyNDaJets:
## Summary
Any value on the alert form that uses the select component is missing the initial state when configuring a form

1) Create a new session, log, and error alert - adding values to the select components
2) When editing the alert, the correct information is displayed as an initial value

https://www.loom.com/share/2a355d8a5ea84a4e8c1f4e20cc246d0d?sid=6b7edd38-7212-4487-a5f1-d83492e7b380

N/A

N/A

### DIFF
--- a/frontend/src/pages/Alerts/utils/AlertsUtils.ts
+++ b/frontend/src/pages/Alerts/utils/AlertsUtils.ts
@@ -58,14 +58,18 @@ export interface AlertForm {
  */
 export const findAlert = (
 	id: string,
+	type: 'error' | 'session',
 	alertsPayload?: GetAlertsPagePayloadQuery,
 ) => {
 	if (!alertsPayload) {
 		return undefined
 	}
 
-	const allAlerts = [
-		...alertsPayload.error_alerts,
+	if (type === 'error') {
+		return alertsPayload.error_alerts.find((alert: any) => alert?.id === id)
+	}
+
+	const possibleAlerts = [
 		...alertsPayload.new_session_alerts,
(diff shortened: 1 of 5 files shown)


frontend/src/pages/Alerts/utils/AlertsUtils.ts
```diff
@@ -58,22 +58,26 @@ export interface AlertForm {
  */
 export const findAlert = (
 	id: string,
+	type: 'error' | 'session',
```
SpennyNDaJets, line 61: Added to avoid id conflict - Alert and session alerts can have the same `id` so this could end up returning the wrong alert

SpennyNDaJets:
@ccschmitz I know you mentioned not passing in value directly, but not sure if you have another work around here?

ccschmitz:
> @ccschmitz I know you mentioned not passing in value directly, but not sure if you have another work around here?

@SpennyNDaJets I think it's different in this case because the `Select` we are using is from the antd library. It's only a code smell when using the Ariakit components from our UI package because those should automatically stay in sync if the `name` attribute is set (see the [Ariakit source](https://github.com/ariakit/ariakit/blob/f194da12038afa328873eadfe55a996cf6fac40f/packages/ariakit-react-core/src/form/form-input.ts#L51)).

ccschmitz (approved):
@SpennyNDaJets I'm sorry, I think I broke these when trying to clean up some of the issues with the Ariakit forms not working after the upgrade. This looks great - thank you for cleaning up after me!

PR #6641, opened 2023-09-18
Default Github rate limit hit cache to 1 hour

SpennyNDaJets:
## Summary
The `github-rate-limit-exceeded-%s` key is being set indefinitely, causing errors to never be enhanced when the rate limit is hit. Default this time to one hour

Test code logic in sandbox

Monitor rate limits hit

N/A

### DIFF
--- a/backend/redis/utils.go
+++ b/backend/redis/utils.go
@@ -492,7 +492,12 @@ func (r *Client) GetHubspotCompanies(ctx context.Context, companies interface{})
 }
 
 func (r *Client) SetGithubRateLimitExceeded(ctx context.Context, gitHubRepo string, expirationTime time.Time) error {
-	return r.setFlag(ctx, GithubRateLimitKey(gitHubRepo), true, time.Until(expirationTime))
+	expirationDuration := time.Until(expirationTime)
+	if expirationDuration < 0 {
+		expirationDuration = time.Hour
+	}
+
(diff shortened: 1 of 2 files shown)


backend/redis/utils.go
```diff
@@ -492,7 +492,12 @@ func (r *Client) GetHubspotCompanies(ctx context.Context, companies interface{})
 }
 
 func (r *Client) SetGithubRateLimitExceeded(ctx context.Context, gitHubRepo string, expirationTime time.Time) error {
-	return r.setFlag(ctx, GithubRateLimitKey(gitHubRepo), true, time.Until(expirationTime))
+	expirationDuration := time.Until(expirationTime)
+	if expirationDuration <= 0 {
```
Vadman97, line 496: are we sure that the `expirationTime` was in fact `0` when we were hitting that edge case, or could github have returned some duration that got formatted in a way that resulted in this being a huge number? could we add a unit test that mocks the response from github that was causing that behavior?

Vadman97 (approved):
🎉

backend/redis/utils.go
SpennyNDaJets: Added some tests - I believe this is happening if an invalid time is provided to the Redis cache. This could be nil or a time in the past.

Screenshot 2023-09-19 at 1 57 13 PM

PR #6644, opened 2023-09-18
enforce submitted timestamps +/- 1 hour from current time

mayberryzane:
## Summary
- enforcing a +/- 1 hour restriction on incoming timestamps helps for a few reasons:
  - fewer clickhouse parts for old partitions as new data stops being ingested for old partitions
  - fewer  recent parts with overlapping time ranges
  - can reliably use a date filter when populating a materialized view to avoid duplicates: https://clickhouse.com/blog/using-materialized-views-in-clickhouse#:~:text=FROM%20soruce_table%20WHERE%20date%20%3E%20%60%24todays_date%60

- made sure logs were flowing through locally

- no, can revert

- no

### DIFF
--- a/backend/otel/otel.go
+++ b/backend/otel/otel.go
@@ -390,6 +390,12 @@ func (o *Handler) HandleLog(w http.ResponseWriter, r *http.Request) {
 
 	var projectLogs = make(map[string][]*clickhouse.LogRow)
 
+	// Allow timestamps +/- 1 hour from the current time.
+	// If the timestamp is outside of this window, set it to the current time.
+	var curTime = time.Now()
+	var minTime = curTime.Add(-1 * time.Hour)
+	var maxTime = curTime.Add(time.Hour)
+
 	resourceLogs := req.Logs().ResourceLogs()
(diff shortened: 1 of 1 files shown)


backend/otel/otel.go
```diff
@@ -409,8 +415,13 @@ func (o *Handler) HandleLog(w http.ResponseWriter, r *http.Request) {
...
 				}
 
+				timestamp := fields.timestamp
+				if timestamp.Before(minTime) || timestamp.After(maxTime) {
+					timestamp = curTime
+				}
+
 				logRow := clickhouse.NewLogRow(
-					fields.timestamp, uint32(fields.projectIDInt),
+					timestamp, uint32(fields.projectIDInt),
```
Vadman97, line 424: a lot of our logs also come from the trace log events (see `HandleTrace` above). could we make this timestamp logic filter there too, perhaps by moving it into some common method? maybe `extractFields` could update the timestamp?

Vadman97 (approved):

Vadman97:
should this be merged @mayberryzane ?

Vadman97:
should we ship this?

PR #6650, opened 2023-09-18
fix log alert delete

### LINKED ISSUE #6620: error trying to delete log alert: WHERE conditions required
real user error deleting log alert

https://app.highlight.io/1/errors/AiwbzitH8kw1po5IdfEbLPlHBeXC

mayberryzane:
## Summary
- query was selecting an arbitrary alert in the project rather than by id
- refetching `GetLogAlert` after deletion was returning an error as the alert doesn't exist anymore
- fixes #6620 

- click tested locally

- no

- no

### DIFF
--- a/backend/private-graph/graph/schema.resolvers.go
+++ b/backend/private-graph/graph/schema.resolvers.go
@@ -3063,8 +3063,9 @@ func (r *mutationResolver) DeleteLogAlert(ctx context.Context, projectID int, id
 	}
 
 	alert := &model.LogAlert{}
-	if err := r.DB.Model(&model.LogAlert{Model: model.Model{ID: id}}).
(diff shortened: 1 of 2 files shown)


backend/private-graph/graph/schema.resolvers.go
```diff
@@ -3063,13 +3063,14 @@ func (r *mutationResolver) DeleteLogAlert(ctx context.Context, projectID int, id
 	}
 
 	alert := &model.LogAlert{}
-	if err := r.DB.Model(&model.LogAlert{Model: model.Model{ID: id}}).
```
Vadman97, line 3066: wonder if it's possible to set up a lint rule for non-zero-value structs passed to `.Model()` as that seems like always misguided

Vadman97 (approved):

PR #6659, opened 2023-09-19
fix demo fields

mayberryzane:
## Summary
- return specific fields for demo project

- made demo project id = 1 locally, click tested

- no

- no


Vadman97 (approved):
🙏

PR #6660, opened 2023-09-19
fix length filter, skip cache for sessions/errors query/histogram

### LINKED ISSUE #6647: session feed results cached
https://app.highlight.io/1/sessions/AnBd8R3ZHfF08JoY2bDEtwqfKvwl

mayberryzane:
## Summary
- length filter should be multiplied by 1000, max value should be treated as infinity
- parse as float because "advanced mode" passes in float values
- fetch results and histogram using 'network-only' to skip cache
- fixes #6647 

- click tested locally

- no

- no

### DIFF
--- a/backend/clickhouse/querybuilder.go
+++ b/backend/clickhouse/querybuilder.go
@@ -3,6 +3,7 @@ package clickhouse
 import (
 	"errors"
 	"fmt"
+	"math"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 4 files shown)


backend/clickhouse/querybuilder.go
```diff
@@ -228,14 +229,19 @@ func parseColumnRule(admin *model.Admin, rule Rule, projectId int, sb *sqlbuilde
...
 				return "", err
 			}
-			end, err := strconv.ParseInt(after, 10, 64)
+			start *= 1000
+			end, err := strconv.ParseFloat(after, 64)
 			if err != nil {
 				return "", err
 			}
+			end *= 1000
+			if after == "60" {
```
Vadman97, line 242: what's this hard-coded condition for?

Vadman97 (approved):
thanks for the quick fixes!

backend/clickhouse/querybuilder.go
mayberryzane: added comment - this condition exists for opensearch, if the slider is all the way to the right (at 60 mins) we should allow any greater length as well

PR #6669, opened 2023-09-20
Touchup service table

SpennyNDaJets:
## Summary
Update the services page to improve the UI and UX of integrating GitHub

[Figma designs](https://www.figma.com/file/rdBPNzn7Klk6RG1LHUCE5B/Screen-Designs?type=design&node-id=11260-325132&mode=design&t=BIFhsYxcb85J9btI-0)

1) View the services page with and without the GitHub Integration completed
2) Add the integration from the services page via the new button

https://www.loom.com/share/ace3ebd101344e90ae3ba66013dbddac

N/A

Can't hurt


julian-highlight:
Looks great @SpennyNDaJets !!! 

Are we planning to update the integration modal at some point, to match the styles to the other modals? This would be another issue, but just curious. Good progress though 💪

SpennyNDaJets:
> Are we planning to update the integration modal at some point, to match the styles to the other modals? This would be another issue, but just curious.

@julian-highlight I think we should do it in the future. I believe its coupled with the other integrations' modal styles, so would probably have to do those all together

PR #6673, opened 2023-09-20
Fix file truncation

SpennyNDaJets:
## Summary
There is some confusion for what the `/..` means in the error stacktrace. Update the stacktrace to show ".../" if there are hidden paths to the file, and display the whole path in a tooltip.

1) View an error stacktrace
- [ ] Paths with more than 5 directories are truncated
- [ ] Paths with 5 or less directories are displayed in full
- [ ] Tooltip displays whole path

Screenshot 2023-09-21 at 1 51 20 PM

N/A

Can't hurt 😄


julian-highlight:
Looks good to me @SpennyNDaJets . Is there a usecase of copying the file-path? If so we might want to add a click-to-copy interaction here as well. If that usecase is not relevant we can just keep it like it is 👍

SpennyNDaJets:
> Is there a usecase of copying the file-path? If so we might want to add a click-to-copy interaction here as well. If that usecase is not relevant we can just keep it like it is 👍

@julian-highlight it could be useful. My thought is the user would need to copy part of this code path, but it would be easier to copy and delete than try to type it out

ccschmitz (approved):

PR #6686, opened 2023-09-21
fix bug where no results for logs just shows a 0

mayberryzane:
## Summary
- fixes a "common React bug where we did something like `{ rowCount && rowCount.map(...)}`"

- tested in preview env

- no

- no


Vadman97 (approved):
🥳 🇨🇽 🚢

PR #6689, opened 2023-09-21
add materialized views to aggregate distinct keys + values, query these for logs

mayberryzane:
## Summary
- add new tables and associated materialized views for:
  - `log_attributes` - all queryable log attributes
  - `log_key_values` - distinct key value pairs and their counts
  - `log_keys` - distinct keys and their counts
- query the `log_keys` and `log_key_values` tables for logs key and value suggestions
- right now, the frontend doesn't pass the current query as a filter and is just applied on the frontend. should change this to filter down the keys / values based on the query in case there are more than 500 keys or 500 values for a key
- limits the max rows scanned to 1m to avoid an expensive operation if a key has high cardinality or there are too many keys for a project

- click tested locally, ran queries against prod

- no, tables already exist in prod so expecting the migration to have no effect

- no

### DIFF
--- a/backend/clickhouse/logs.go
+++ b/backend/clickhouse/logs.go
@@ -19,6 +19,8 @@ import (
 
 const LogsTable = "logs"
 const LogsSamplingTable = "logs_sampling"
+const LogKeysTable = "log_keys"
+const LogKeyValuesTable = "log_key_values"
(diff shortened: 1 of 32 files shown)


Vadman97 (approved):
🚢

PR #6695, opened 2023-09-22
Redirect to originally request page after sign up/in

ccschmitz:
## Summary

We noticed a couple issues with people going through the setup flow of our Vercel integration.

1. They were getting stuck in the loading state once redirect to the `/sign_in` page.
2. They weren't being redirected back to the Vercel integration page after signing up/in.

This PR fixes both issues and adds general support for redirecting to the originally request page after a user authenticates.

I also addressed a minor issue I saw on the admin form where a top border was being applied to the last name field when hovered or focused.

Local click test. I just hit the callback URL manually locally since I couldn't trigger the real auth redirect from Vercel. We should verify the behavior initiating it from Vercel once this goes live.

Need to verify the full flow starting from Vercel once deployed to production.

Nope!

### DIFF
--- a/frontend/src/pages/Auth/SignIn.tsx
+++ b/frontend/src/pages/Auth/SignIn.tsx
@@ -22,6 +22,10 @@ import React, { useCallback, useEffect } from 'react'
 import { Link, useLocation, useNavigate } from 'react-router-dom'
 
 import { useAuthContext } from '@/authentication/AuthContext'
+import {
+	AppLoadingState,
+	useAppLoadingContext,
+} from '@/context/AppLoadingContext'
 import { SIGN_UP_ROUTE } from '@/pages/Auth/AuthRouter'
 import analytics from '@/util/analytics'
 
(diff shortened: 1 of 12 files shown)


frontend/src/pages/Auth/SignIn.tsx
```diff
@@ -114,6 +120,11 @@ export const SignIn: React.FC<Props> = ({ setResolver }) => {
 	}
 
 	useEffect(() => analytics.page(), [])
+	useEffect(() => {
+		if (!loadingWorkspaceForInvite) {
+			setLoadingState(AppLoadingState.LOADED)
+		}
+	}, [loadingWorkspaceForInvite, setLoadingState])
```
ccschmitz, line 127: We needed to call `setLoadingState(AppLoadingState.LOADED)` here when redirecting from the vercel integration page. This fixes the infinite loading state.

SpennyNDaJets (approved):
LGTM!

PR #6701, opened 2023-09-24
remove opensearch client + all read/write references

mayberryzane:
## Summary
- builds on `zane/delete_sessions_ch` branch
- removes all read/write references to opensearch
- removes last case where opensearch was still used for identifier suggestions and replaces with clickhouse

- tested locally to make sure sessions were still processed and indexed in clickhouse
- tested identifier suggestion and field type endpoints

- no

- no


Vadman97 (approved):
should we remove docker / all other references to opensearch here or as another PR once we confirm we are no longer writing / reading?

PR #6713, opened 2023-09-26
introduce backend for filtering ingest

Vadman97:
## Summary

Introduces a sampling rate and ingestion exclusion filters for all 4 products.
Rate-based ingestion uses a stable ingest-item key to consistently deem 
whether a resource should be ingested or not, so that even if we replay 
the ingest queue, ingest decisions will remain the same based on client-generated keys.
Filter-based ingestion mimics the existing search logic we have but applies filters directly
as each item is ingested to avoid having to insert into clickhouse to make a filter query. This means
that the filtering logic has to be duplicated with what we have for the feeds themselves, but
the logic is shared across the 4 products.

### Sessions
* Sampling rate uses session secure ID as the key
* Exclusion filters are applied when a session is initialized or when the datasync queue has an entry (ie. new fields are appended).
### Errors
* Sampling rate uses request id, trace id, or span id. 
* Exclusion filters are applied when errors are ingested in push payload or in otel backend processing.
### Logs
* Sampling rate uses the log UUID. 
* Exclusion filters are applied in otel backend processing. 
### Traces
* Sampling rate uses the trace UUID. Because the trace UUID is consistent with all spans children of a root trace, this is effectively 'head' tracing since once we determine a root trace to be included, all child spans will also be included.
* Exclusion filters are applied in otel backend processing. 

Relates to #6749

New unit tests.
Validating sampling working correctly locally.
image

No, all new functionality is disabled by default with no way to enable this from the UI yet.

No


backend/public-graph/graph/sampling.go
```diff
@@ -0,0 +1,325 @@
...
+		key = uuid.New().String()
+	}
+
+	h := fnv.New32a()
+	if _, err := h.Write([]byte(key)); err != nil {
+		log.WithContext(ctx).WithError(err).Error("failed to calculate hash")
+		return true
+	}
+	r := rand.New(rand.NewSource(int64(h.Sum32())))
+	return r.Float64() <= rate
```
Vadman97, line 184: just use hash, no need to rng

backend/clickhouse/query.go
```diff
@@ -495,3 +498,65 @@ func KeyValuesAggregated(ctx context.Context, client *Client, tableName string,
 	span.Finish(rows.Err())
 	return values, rows.Err()
 }
+
+func matchesQuery[TObj interface{}, TReservedKey ~string](row *TObj, config tableConfig[TReservedKey], filters *queryparser.Filters) bool {
```
Vadman97, line 502: test this logic vs clickhouse filter query

backend/clickhouse/query.go
```diff
@@ -495,3 +498,65 @@ func KeyValuesAggregated(ctx context.Context, client *Client, tableName string,
...
+	v := reflect.ValueOf(*row)
+
+	rowBodyTerms := map[string]bool{}
+	if config.bodyColumn != "" && len(filters.Body) > 0 {
+		for _, field := range strings.Fields(v.FieldByName(config.bodyColumn).String()) {
```
mayberryzane, line 507: testing for case sensitivity, handling non-alpha + non-whitespace characters (like `, . ; :`)

backend/public-graph/graph/sampling.go
```diff
@@ -0,0 +1,325 @@
...
+
+	id := ptr.ToString(errorObject.RequestID)
+	if id == "" {
+		id = ptr.ToString(errorObject.TraceID)
+	}
+	if id == "" {
+		id = ptr.ToString(errorObject.SpanID)
+	}
+
+	ret := isIngestedBySample(ctx, id, settings.LogSamplingRate)
```
mayberryzane, line 100: `settings.ErrorSamplingRate`

backend/public-graph/graph/sampling.go
```diff
@@ -0,0 +1,325 @@
...
+}
+
+func (r *Resolver) IsSessionExcludedBySample(ctx context.Context, session *model.Session) bool {
+	settings, err := r.Store.GetProjectFilterSettings(ctx, session.ProjectID)
+	if err != nil {
+		log.WithContext(ctx).WithError(err).Error("failed to get project filter settings")
+		return true
+	}
+
+	ret := !isIngestedBySample(ctx, session.SecureID, settings.LogSamplingRate)
```
mayberryzane, line 142: `settings.SessionSamplingRate`

frontend/src/pages/Billing/utils/utils.ts
```diff
@@ -77,10 +77,14 @@ export const getMeterAmounts = (
 	const logsMeter = data.billingDetailsForProject?.logsMeter ?? 0
 	const logsQuota =
 		data.billingDetailsForProject?.logsBillingLimit ?? undefined
+	// TODO(vkorolik)
```
mayberryzane, line 80: `TODO`s here and in the file below

mayberryzane (approved):

mayberryzane:
I wonder if there's also a need for a rate-based sampling? e.g. X traces per minute? would be easier to set defaults across all projects for this, e.g. to limit spikes in errors

frontend/src/pages/Billing/utils/utils.ts
Vadman97: cleaned this up but leaving some TODOs in place for further changes
* traces pricing logic
* sampling UI settings

mayberryzane (approved):

backend/clickhouse/query.go
```diff
@@ -156,9 +159,10 @@ func makeSelectBuilder[T ~string](config tableConfig[T], selectStr string, proje
 	// of where to place the raw SQL later when it is being built.
 	// In this case, we are placing the marker after the `FROM` clause
 	preWheres := []string{}
+	bodyQuery := ""
 	for _, body := range filters.body {
 		if strings.Contains(body, "%") {
-			sb.Where("Body ILIKE" + sb.Var(body))
+			bodyQuery = "Body ILIKE" + sb.Var(body)
```
Vadman97, line 165: fixes existing bug with prewhere ordering

PR #6715, opened 2023-09-26
Individual trace view

ccschmitz:
## Summary

Allows you to click into a trace to view details for it and all its child spans.

https://www.loom.com/share/de41a52bee0841ca9ce43d9a9a17f3bc

Local click test. Can't test in prod because of the backend changes required to load the new UI.

This is all only available to Highlight admins and only adding reads to the DB. We should monitor performance of querying spans for some of the larger traces.

As a follow up, we should also consider renaming things from `Trace` to `Span` where appropriate. E.g. our "Traces" page shows a list of spans, and when you click into that span it loads the trace for that span.

Will have Julian look this over. I'm sure there will be some back and forth to refine the UI, but there will be a lot of iterating on this before it goes live.


frontend/src/pages/Traces/TracePanel.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { useNavigate } from 'react-router-dom'
+
+import { TracePage } from '@/pages/Traces/TracePage'
+import { useParams } from '@/util/react-router/useParams'
+
+import * as styles from './TracePanel.css'
+
+export const TracePanel: React.FC = () => {
+	const navigate = useNavigate()
+	const { project_id } = useParams<{
```
SpennyNDaJets, line 12: Not sure if there is a preference for `useParams` vs 
`const { projectId } = useProjectId()`?

frontend/src/pages/Traces/utils.ts
```diff
@@ -0,0 +1,68 @@
...
+	const hours = Math.floor(minutes / 60)
+	const days = Math.floor(hours / 24)
+
+	const dayString = days > 0 ? `${days}d` : ''
+	const hourString = hours > 0 ? `${hours % 24}h` : ''
+	const minuteString = minutes > 0 ? `${minutes % 60}m` : ''
+	const secondString = seconds > 0 ? `${seconds % 60}s` : ''
+	const millisecondString = milliseconds > 0 ? `${milliseconds % 1000}ms` : ''
+
+	if (dayString) {
```
SpennyNDaJets, line 48: Nit: Looks good! Wonder if we can simplify the if statements?
```
[dayString, hourString, minuteString, secondString, millisecondString].join(" ").trim()
```

SpennyNDaJets (approved):
Click test looks great!

Just want to check that clicking elsewhere in the table/page should close the panel?

https://github.com/highlight/highlight/assets/17744174/29d8355d-6141-4c5c-b1e9-882b9d811e31

frontend/src/pages/Traces/TracePanel.tsx
ccschmitz: Yeah, should probably be `useProjectId` unless you need other params from the URL. Thanks for mentioning!

frontend/src/pages/Traces/utils.ts
ccschmitz: Yeah, good call!

ccschmitz: Shoot, I don't think this quite accomplishes the same thing. I agree it's ugly/verbose, but I think it's fairly easy to reason about, so I'm going to leave it for now.

frontend/src/pages/Traces/TracePanel.tsx
ccschmitz: I think I'm going to merge as is and handle this in a follow up!

Issue #6716, opened 2023-09-26
@highlight-run/next/client module cannot be found


MichaelCGrant:
**Describe the bug**
When trying to import HighlightInit and ErrorBoundary from "@highlight-run/next/client" as directed by the [highlight docs](https://www.highlight.io/docs/getting-started/client-sdk/nextjs), I got the following error: "Cannot find module '@highlight-run/next/client' or its corresponding type declarations." Upon investigation, the client.d.ts file which exports the client module is not included in the package.

**To Reproduce**

1. Import HighlightInit from @highlight-run/next/client on a next.js layout
2. See error

**Expected behavior**
I should be able to import HighlightInit from @highlight-run/next/client with no errors.

**Screenshots**
![image](https://github.com/highlight/highlight/assets/93897555/04532440-3eee-439e-bed2-8b5ccaa7dc64)
![image](https://github.com/highlight/highlight/assets/93897555/1062ddc9-1d8b-4f60-94f3-10adac6f9f25)

**Additional context**
Client.d.ts is not included in package

deltaepsilon:
I'm working on it. I'm close!!!

https://github.com/highlight/highlight/pull/6712

PR #6717, opened 2023-09-26
fixes highlight/highlight#6716

MichaelCGrant:
## Summary
 added client.d.ts to "files" section of package.json


Vadman97 (commented):
thank you for the contribution @MichaelCGrant ! mind sharing how you've validated that this fixes the typescript error you were seeing? curious if you have ideas on what we can do to catch this in our test automation as well (perhaps with a `tsc` invocation that mirrors your vscode setup)

Vadman97 (approved):

PR #6727, opened 2023-09-26
Allow alerts to private channels

SpennyNDaJets:
## Summary
When an alert is created, we try to join the channel. When this channel is private, the request fails and we exit early, so not private channels will ever receive alerts.

Still send the message, even if the join conversation request fails. Keeping it consistent with the other post message sends.

1) Create a private channel
2) Add the highlight bot
3) Create an alert to point to that private channel
4) Check that the alert is created in the channel

N/A

N/A


Vadman97 (approved):
:ship: thanks for getting to the bottom of this!

PR #6728, opened 2023-09-26
allow filtering for errors with sessions

Vadman97:
## Summary

Adds the `HasSession` column to clickhouse error objects for searching for errors with sessions.
Stores the session id on clickhouse error objects in case we have a use for that.

Local deploy
Screenshot 2023-09-26 at 1 37 31 PM

Screenshot 2023-09-26 at 2 03 42 PM

Will backfill data with https://github.com/highlight/highlight/blob/main/backend/migrations/cmd/backfill-errors-clickhouse/main.go#L1

Yes @julian-highlight


backend/store/error_groups.go
```diff
@@ -43,6 +43,11 @@ func (store *Store) ListErrorObjects(errorGroup model.ErrorGroup, params ListErr
 					Where("sessions.email ILIKE ?", "%"+val[0]+"%")
 			}
 		}
+		if val, ok := filters.Attributes["has_session"]; ok {
+			if len(val) > 0 && val[0] == "true" {
+				query.Where("COALESCE(error_objects.session_id) > 0")
```
mayberryzane, line 48: do you want `COALESCE(error_objects.session_id, 0)` here? I think coalesce will still return null if its last argument is null

mayberryzane (approved):

backend/store/error_groups.go
Vadman97: thanks! forgot about that

julian-highlight:
Created a quick design here, just to make things look more cohesive: https://www.figma.com/file/rdBPNzn7Klk6RG1LHUCE5B/Screen-Designs?type=design&node-id=12035-22099&mode=design&t=0LzS3Pe1TJMfEp24-11


Low prio to actually implement this from my perspective, as your current solution works as well. This would slightly improve the UX and look of this feature. 
@Vadman97

Vadman97:
thanks @julian-highlight ! noted

PR #6731, opened 2023-09-27
fix datadog trace resource names

Vadman97:
## Summary

Noticed our datadog traces had the `resource_name` in a strange format.
Screenshot 2023-09-26 at 6 16 29 PM

Turns out the opentelemetry `attribute.Value()` returns a struct rather than a string, so an `AsString()` conversion
is needed for string tags.

Will monitor datadog traces.

No

No


ccschmitz (approved):
Good catch!

PR #6737, opened 2023-09-27
add timeout for embedding generation

Vadman97:
## Summary

The embeddings inference endpoint may slow down dramatically under heavy load but never time out.
In this case, each request takes >30 seconds and may take much longer, with the increasing load
preventing any requests from completing. This introduces a timeout to make sure we
will fail requests that take too long (longer than 1 second, where the normal p95 request time is 250ms)

local testing with a fake sleep inserted
```ERRO[0009]/Users/vkorolik/work/highlight/backend/public-graph/graph/resolver.go:768 github.com/highlight-run/highlight/backend/public-graph/graph.(*Resolver).HandleErrorAndGroup() failed to get embeddings                      error="Post \"https://rxti97k7hsleuekl.us-east-1.aws.endpoints.huggingface.cloud\": context deadline exceeded" error_object_id=0```

objects stored with classic method when the timeout is exceeded
Screenshot 2023-09-27 at 2 15 29 PM

No

No

### DIFF
--- a/backend/embeddings/embeddings.go
+++ b/backend/embeddings/embeddings.go
@@ -156,7 +156,10 @@ type HuggingfaceModelInputs struct {
 }
 
 func (c *HuggingfaceModelClient) makeRequest(ctx context.Context, b []byte) ([]byte, error) {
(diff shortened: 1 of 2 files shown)


SpennyNDaJets (approved):
LGTM!

PR #6739, opened 2023-09-27
add category to session insights emails for analytics

mayberryzane:
## Summary
- session insights emails are generated with react email so there's no sendgrid template id to use, add a category to these emails so we can filter when looking at analytics

- will validate with a dry run in prod

- no

- no


Vadman97 (approved):
ty 🚢

PR #6740, opened 2023-09-27
Update GitHub enhancement error logic

SpennyNDaJets:
## Summary
Currently, GitHub stacktraces perform by finding the first 5 files that do not contain known system level paths. This is hard to manage as different systems have different paths, and the list is ever growing. If we are missing a path, it is very likely that the service will be put into an error state due to too many errors.

Update this logic to:
1. Fetch all files that do not contain known system level paths
2. If a 404 is received from GitHub for a specific file, cache that result so we do not over fetch on errors
3. Increment an error count only if a file contains no enhancements
4. If this occurs for 5 errors straight within an hour (i.e. different errors since 404s are cached), then put the service in an error state.

1) Check that errors are still being enhanced
2) Errors with more than 5 enhanceable tracers are all enhanced
3) Files that result in errors are cached and not overfetched from GitHub

https://www.loom.com/share/6b0dea2a74f947069f0fdf48df3d00c7?sid=268cdb53-8455-4f9b-8479-8ef454b6b47a

Keep an eye on number of errors and requests in logs

N/A


backend/redis/utils.go
```diff
@@ -519,6 +523,14 @@ func (r *Client) ResetServiceErrorCount(ctx context.Context, projectId int) (int
 	return r.redisClient.Del(ctx, serviceKey).Result()
 }
 
+func (r *Client) SetGitHubFileError(ctx context.Context, gitHubRepo string, version string, fileName string) error {
+	return r.setFlag(ctx, GitHubFileErrorKey(gitHubRepo, version, fileName), true, 1*time.Hour)
```
SpennyNDaJets, line 527: How do we feel about one hour? We could do more, my only concern is increasing our redis cache size. Are there better ways to cache (i.e. store an empty file in S3)?

Vadman97: nah, redis is great for this - just a key and a boolean so it should be very efficient (i believe the key is hashed to an efficient index in redis)

hour sounds good for sure

backend/store/stacktraces.go
```diff
@@ -228,19 +215,37 @@ func (store *Store) GitHubEnhancedStackTrace(ctx context.Context, stackTrace []*
 			continue
 		}
 
-		filesEnhanced += 1
+		// check if we've previously errored on this file
+		previousError, _ := store.redis.GetGitHubFileError(ctx, *service.GithubRepoPath, *validServiceVersion, fileName)
+		if previousError {
+			newMappedStackTrace = append(newMappedStackTrace, trace)
```
Vadman97, line 221: nit: refactor this fn to return so that the `			newMappedStackTrace = append(newMappedStackTrace, trace)` logic can be reused, rather than copying it / continue everywhere we want to skip?

Vadman97 (approved):

backend/store/stacktraces.go
SpennyNDaJets: Let me know if the changes in https://github.com/highlight/highlight/pull/6740/commits/85c88b41047f94dee63de2d1139db3cab9abd4ba

Definitely looks a little cleaner - only downside I see is a little confusing returning two booleans for the enhanceable and enhanced state, but tried to call out what these meant in a comment

PR #6751, opened 2023-09-28
Fix command bar click

SpennyNDaJets:
## Summary
Ariakit's `useValue` is a hook that cannot be called inside the bodu of a function component, and causing the `onClick` method to exit early. Update to use `getValue` here, since it will fetch the value when called.

Screenshot 2023-09-28 at 1 47 29 PM

1) Open the command bar with `command-k`
2) Search for a field
3) Select the time frame
4) Click on the appropriate option
- [ ] Page is redirected
- [ ] Filter is applied
- [ ] Time is correctly accepted
- [ ] Modal disappears

https://www.loom.com/share/3c15513a77194841b5b7fd05094671e9?sid=d1213027-ef9c-4214-8824-0b35e0a0ab7d

N/A

N/A

### DIFF
--- a/frontend/src/components/CommandBar/CommandBar.tsx
+++ b/frontend/src/components/CommandBar/CommandBar.tsx
@@ -249,7 +249,7 @@ const SectionRow = ({
 				searchAttribute(attribute, {
 					withDate:
 						formStore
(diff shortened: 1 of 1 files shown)


Vadman97 (approved):
:pray:

PR #6754, opened 2023-09-28
feat: Jira Integration

### LINKED ISSUE #5787: Jira Integration
Add a Jira integration so that we can create issues straight from the Highlight UI.

Like our other issue tracker integrations, we want to be able to:
- add, configure, and remove the Jira integration from our integrations page https://github.com/highlight/highlight/blob/main/highlight.io/pages/integrations.tsx#L1 ; adding the integration will save an authorization token at the workspace level that can be used to make requests to the Jira API for creating issues
Screen Shot 2023-09-18 at 9 46 04 AM

- create a Jira issue from a session comment (a new comment on the session player or an existing comment in the right panel)
Screen Shot 2023-09-18 at 9 45 00 AM
Screen Shot 2023-09-18 at 9 50 00 AM

- create a Jira issue from an error
Screen Shot 2023-09-18 at 9 45 21 AM

- all behavior should match with the other integrations (e.g. clickup, height, github) for linking out to issues from Highlight, removing issues, etc
- for local testing, you can create your own Jira Oauth app and use your own client id / secret https://developer.atlassian.com/cloud/jira/platform/oauth-2-3lo-apps/ ; before we deploy this integration in production, you can share your integration's settings with us and we will create a Highlight Oauth app with matching settings

jemiluv8:
Add jira integration to support creating jira issues from highlight ui

fixes #5787
/claim #5787

- [x] Go to integrations page and go through the flow for adding jira integration
- [x] Verify that after authorizing our app, jira integration is marked as active
- [x] Verify that you can "remove" the jira integration  
- [x] Create jira issues from highlight ui - error comment and session comment
- [x] Ensure that clicking the link to jira issue takes you to the relevant task description

### LOOM VIDEO FOR ADDING JIRA INTEGRATION
https://www.loom.com/share/21f7ba7f112f473fab00f8497948a1c4?sid=fb098cd0-6dd4-403e-b282-ca6658f69440

### LOOM VIDEO FOR CREATING JIRA ISSUE FROM ERROR COMMENT
https://www.loom.com/share/67c9a3caa4094509af16d901285674d5?sid=d1c9603b-0f0a-4abf-81b0-8783983e412c

### LOOM VIDEO - CREATING JIRA ISSUE FROM SESSION COMMENT (New and Existing). Plus viewing issue link
https://www.loom.com/share/8758e8a519644bf983c5b0ba89df5bf5?sid=b18ce953-e6f2-427d-b8c9-7990657ce453

Two new environment variables are needed for jira integration support

- JIRA_CLIENT_ID
- JIRA_CLIENT_SECRET

The jira/atlasian app shall also need the following scopes
Screenshot 2023-10-03 at 3 24 47 AM

These can be added in app configuration under developer console
https://developer.atlassian.com/console/myapps

Yes. This requires a design review. Below are the modified views.
Screenshot 2023-10-06 at 1 35 20 AM
Screenshot 2023-10-06 at 1 35 53 AM


docker/.env
```diff
@@ -35,6 +35,8 @@ RENDER_PREVIEW=true
 SESSION_FILE_PATH_PREFIX=/tmp/
 REACT_APP_AUTH_MODE=password
 ADMIN_PASSWORD=I_AM_GROOT
+JIRA_CLIENT_ID=
+JIRA_CLIENT_SECRET=
```
jemiluv8, line 39: jira config from the console.

frontend/src/components/NewIssueModal/NewIssueModal.tsx
```diff
@@ -72,6 +74,8 @@ const NewIssueModal: React.FC<React.PropsWithChildren<NewIssueModalProps>> = ({
 	}, [defaultIssueTitle, commentText])
 
 	const [containerId, setContainerId] = useState('')
+	const [issueTypeId, setIssueTypeId] = useState('')
```
jemiluv8, line 77: For most other integrations the containerId was used to set the issueTeamId.
In the case of Jira Integration, we needed two fields: issueTypeId and issueProjectId.
This was how I set them.

frontend/src/components/NewIssueModal/NewIssueModal.tsx
```diff
@@ -102,9 +106,27 @@ const NewIssueModal: React.FC<React.PropsWithChildren<NewIssueModalProps>> = ({
 		],
 	})
 
+	const otherIssueVariables = () => {
```
jemiluv8, line 109: Only applies to jira issues.

jemiluv8 (commented):
@Vadman97, @mayberryzane I've completed a self review.
This PR is ready for review.

jay-khatri:
This does require a design review. Please add @julian-highlight.

frontend/src/pages/IntegrationAuthCallback/IntegrationAuthCallbackPage.tsx
```diff
@@ -286,14 +286,19 @@ const WorkspaceIntegrationCallback = ({
 	type: string
 	addIntegration: (code: string) => Promise<unknown>
 }) => {
+	const codeSessionStorageKey = 'integration-callback-code'
```
jemiluv8, line 289: This component and its parent are re-rendered multiple times. Don't know if that was intended.
The net effect however is that the IIFE executed in the useEffect hook for the WorkspaceIntegrationCallback
component is executed twice - calling ```await addIntegration(code) twice ```

For my workspace integration, this meant that I got two notifications - once succeeding while the other fails.
That is because my authorization_code cannot be used multiple times. The failed notification is a false negative in
this case and

So I added a way for us to store the usedAuthorization code in sessionStorage so that we prevent calling addIntegration on the same code when the component renders. Feels a bit hacky though.

I looked into other ways to ensure that the IIFE executed just once but couldn't come up with a better solution
thus far. Open to other alternatives

mayberryzane:
Hi @jemiluv8, the loom previews look good, I'm currently clicktesting these changes locally

mayberryzane (changes requested):
Hi @jemiluv8, the implementation looks pretty good and I appreciate the work! I have a few pieces of feedback:
- For the links out to Jira, could you link directly to the issue instead of to the board? For example, I'm getting linked to https://highlight-zane.atlassian.net/jira/core/projects/KAN/board?selectedIssue=KAN-4 but Jira doesn't seem to be selecting that issue. I think it would be better to link directly to the issue anyway, in this case: https://highlight-zane.atlassian.net/browse/KAN-4
- Instead of letting a user select epic, task, or subtask, I think we can always create a task and we don't need that dropdown (plus subtask is broken at the moment because I think it requires a parent task when creating)
- When creating a new issue, we should save the latest selected jira project in local storage so the user doesn't have to reselect it. Example of how we're doing this for the Clickup list: https://github.com/highlight/highlight/blob/main/frontend/src/pages/IntegrationsPage/components/ClickUpIntegration/ClickUpListSelector.tsx#L39

I think this will be good to merge after these changes pending any design feedback from @julian-highlight

jemiluv8:
ok @mayberryzane, your second suggestion regarding always creating new issues with type "task" had crossed my mind.
I just thought issue types may be created and deleted and it was possible for a jira project to not have an issue type called
task. 

Will work on them and keep you posted.

backend/private-graph/graph/resolver.go
```diff
@@ -2597,6 +2652,49 @@ func (r *Resolver) CreateHeightTaskAndAttachment(
...
+
+	if accessToken == nil {
+		return errors.New("No Jira integration access token found.")
+	}
+
+	jiraIssuePayload := jira.JiraCreateIssueFields{
+		Description: issueDescription,
+		Summary:     issueTitle,
+		Project:     jira.JiraIssueProjectData{Id: projectId},
+		IssueType:   jira.JiraIssueTypeData{Id: "10001"},
```
jemiluv8, line 2677: I am not comfortable with this but have good reason to believe this default issue type comes with every jira project
and I haven't seen a way to delete it.

backend/integrations/jira/jira.go
```diff
@@ -0,0 +1,250 @@
...
+
+	return res.Projects, nil
+}
+
+func GetJiraProjects(workspace *model.Workspace, accessToken string) ([]*modelInputs.JiraProject, error) {
+	return GetJiraIssueCreateMeta(accessToken, *workspace.JiraCloudID)
+}
+
+func MakeExternalIdForJiraTask(workspace *model.Workspace, issue *JiraIssue) string {
+	url := fmt.Sprintf("%s/browse/%s", *workspace.JiraDomain, issue.Key)
```
jemiluv8, line 206: @mayberryzane  updated external link to link directly to project as requested.

frontend/src/pages/IntegrationsPage/components/JiraIntegration/JiraProjectSelector.tsx
```diff
@@ -0,0 +1,62 @@
...
+				displayValue: (
+					<Text size="small" weight="medium">
+						{team.name} ({team.key})
+					</Text>
+				),
+			})) || []
+		)
+	}, [data?.jira_projects])
+
+	const [selectedJiraProjectId, setJiraProjectId] = useLocalStorage(
```
jemiluv8, line 29: Saving default project Id to set on the new task modal.

jemiluv8 (commented):
@mayberryzane  I've made all changes requested.

Overall it seems to have simplified the task and reduced the amount of 
changes I had to make to get this working.

frontend/src/pages/IntegrationAuthCallback/IntegrationAuthCallbackPage.tsx
mayberryzane: Could the rendering multiple times be a side effect of how we use `React.StrictMode` in our dev environments? Anyway I think this is an ok workaround

mayberryzane (approved):
🚢 changes look good to me, thanks for addressing the feedback! I will merge it in / handle deployment. Thank you!

backend/private-graph/graph/resolver.go
mayberryzane: I think this sounds good to me for a first version - longer term it might make sense to have a "task type" dropdown as part of the integration settings page or something like that

PR #6762, opened 2023-09-29
Fix ariakit menu z-index

SpennyNDaJets:
**Suggestion: Review hiding whitespace**

## Summary
Menu's are being displayed underneath relatively positioned components rendered further down in the down. Render the aria kit menu in a portal that will render the menu in the bottom of the dom.

Screenshot 2023-09-29 at 11 12 53 AM

1) View the errors page
2) Make your screen width small enough so the error title is as close to the left panel as it can be
3) Open the segments popover
4) Scroll up and down with the popover open
- [ ] No components in the error details are displayed over the content
5) Add a filter and open the combobox to select a value
6) Scroll up and down with the combobox open
- [ ] No components in the error details are displayed over the content
7) Repeat for other menus (e.g. header menus)
- [ ] No difference in behavior/accessibility
- [ ] Menus are rendered over all other content

N/A

N/A

### DIFF
--- a/packages/ui/src/components/ComboboxSelect/ComboboxSelect.tsx
+++ b/packages/ui/src/components/ComboboxSelect/ComboboxSelect.tsx
@@ -10,6 +10,7 @@ import {
 	SelectLabel,
 	SelectPopover,
 	PopoverArrow,
+	Portal,
 } from '@ariakit/react'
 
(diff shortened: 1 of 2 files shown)


Vadman97 (approved):
nice! great to fix this in the base ui package component

PR #6764, opened 2023-09-29
save session properties as fields

mayberryzane:
## Summary
- this was inadvertently removed when removing opensearch
- session properties should be converted into fields and saved in the `session_fields` table

- tested `country is United States` filter locally
- tested backfill script locally and verified more sessions were added with `country is United States`

- will run backfill script after deploying

- no

### DIFF
--- a/backend/migrations/cmd/backfill-session-properties-clickhouse/main.go
+++ b/backend/migrations/cmd/backfill-session-properties-clickhouse/main.go
@@ -1,0 +1,70 @@
+package main
+
+import (
+	"context"
+	"os"
+
+	kafka_queue "github.com/highlight-run/highlight/backend/kafka-queue"
+	"github.com/highlight-run/highlight/backend/model"
+	public "github.com/highlight-run/highlight/backend/public-graph/graph"
+	"github.com/highlight-run/highlight/backend/redis"
+	"github.com/highlight-run/highlight/backend/store"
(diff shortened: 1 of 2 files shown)


backend/migrations/cmd/backfill-session-properties-clickhouse/main.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	var sessionIds []int
+	if err := db.Raw(`
+		SELECT id
+		FROM sessions
+		WHERE created_at >= now() - interval '5 days'
+	`).Scan(&sessionIds).Error; err != nil {
+		log.WithContext(ctx).Fatal(err)
+	}
```
Vadman97, line 49: should this `FindInBatches` to avoid loading the massive int array in memory? guessing we'd ultimately want to do this for all sessions

Vadman97 (approved):
🚢

backend/migrations/cmd/backfill-session-properties-clickhouse/main.go
mayberryzane: yeah that would probably be better, but I've used this script on 3 months of sessionids / errorids before and it's worked fine. This script is fine to run on just 5 days worth of sessions because these fields were being indexed prior to turning off opensearch writes (because this code was removed)

PR #6769, opened 2023-09-29
add materialized view for traces perf

mayberryzane:
## Summary
- similar to logs, add materialized views for trace attributes, keys, values for better performance for search suggestions

- confirmed that all tables were being populated locally

- no

- no

### DIFF
--- a/backend/clickhouse/migrations/000043_create_traces_sampling_table.up.sql
+++ b/backend/clickhouse/migrations/000043_create_traces_sampling_table.up.sql
@@ -1,0 +1,34 @@
+CREATE TABLE IF NOT EXISTS traces_sampling (
+    Timestamp DateTime64(9),
+    UUID UUID,
+    TraceId String,
+    SpanId String,
+    ParentSpanId String,
+    ProjectId UInt32,
+    SecureSessionId String,
+    TraceState String,
+    SpanName LowCardinality(String),
+    SpanKind LowCardinality(String),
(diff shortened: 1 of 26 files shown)


backend/clickhouse/migrations/000043_create_traces_sampling_table.up.sql
```diff
@@ -0,0 +1,34 @@
...
+        SpanId String,
+        TraceState String,
+        Attributes Map(LowCardinality(String), String)
+    )
+) ENGINE = MergeTree
+ORDER BY (
+        ProjectId,
+        toStartOfHour(Timestamp),
+        cityHash64(UUID)
+    ) SAMPLE BY cityHash64(UUID) TTL toDateTime(Timestamp) + toIntervalDay(30);
```
Vadman97, line 34: should we be doing this with `SpanName` from the start?

Vadman97 (approved):

backend/clickhouse/migrations/000043_create_traces_sampling_table.up.sql
mayberryzane: I was thinking some more about using `SpanName` as the primary key for the main `traces` table e.g. `ORDER BY (ProjectId, SpanName, Timestamp, UUID)`, but unless we're always filtering by `SpanName`, it will end up slowing things down

PR #6770, opened 2023-09-29
remove opensearch references in docker scripts

mayberryzane:
## Summary
- we're not reading/writing to opensearch anymore, this removes the remaining references in the docker scripts

- tested `start-infra` script locally

- no

- no


Vadman97 (approved):
🙏 less RAM usage for the hobby instance

docker/env.sh
Vadman97, line 1: there are references to other env vars in `.env` - mind getting rid of those too?

PR #6781, opened 2023-10-02
fix duplicate errors

Vadman97:

### DIFF
--- a/backend/util/tracer-graphql.go
+++ b/backend/util/tracer-graphql.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/99designs/gqlgen/graphql"
 	log "github.com/sirupsen/logrus"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
 )
 
 type Tracer struct {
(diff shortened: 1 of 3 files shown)


SpennyNDaJets (approved):
LGTM!

Just to make sure I'm understanding the changes:
1) Try to match with embeddings. If no match, create or match via classic
  a) before we were using classic before we knew if there was an embedded match
2) Update our graphql tracer to just send traces to Datadog, since these are already being reported by the SDK
3) Not sure I understand the SDK tracer changes

sdk/highlight-go/tracer.go
```diff
@@ -79,12 +73,6 @@ func (t Tracer) InterceptField(ctx context.Context, next graphql.Resolver) (inte
 // InterceptResponse instruments timing, payload size, and error information
 // of the response handler. The metric is grouped by the corresponding operation name.
 func (t Tracer) InterceptResponse(ctx context.Context, next graphql.ResponseHandler) *graphql.Response {
-	// if we don't have a highlight session in the context, no point trying to
-	// instrument since we won't be able to store the metric
-	if _, _, err := validateRequest(ctx); err != nil {
```
Vadman97, line 84: this comment is out of date, from back when we were using this code path to record session metrics. traces can be recorded for a project without a session / request id.

backend/util/tracer-graphql.go
```diff
@@ -73,17 +74,10 @@ func (t Tracer) InterceptResponse(ctx context.Context, next graphql.ResponseHand
...
+	span, ctx := tracer.StartSpanFromContext(ctx, "graphql.operation", tracer.ResourceName(opName))
+	span.SetTag("backend", t.serverType)
 	resp := next(ctx)
-	if resp != nil {
-		var errs []error
-		for _, err := range resp.Errors {
-			errs = append(errs, err)
-		}
-		span.Finish(errs...)
-	}
```
Vadman97, line 86: don't record response errors because gqlgen doesn't have a stack trace for them, and they are duplicates of the `field` errors that are reported above in `InterceptField`

backend/util/tracer-graphql.go
```diff
@@ -32,18 +33,18 @@ func (t Tracer) Validate(graphql.ExecutableSchema) error {
 }
 
 func (t Tracer) InterceptField(ctx context.Context, next graphql.Resolver) (interface{}, error) {
-	start := time.Now()
 	// taken from: https://docs.datadoghq.com/tracing/setup_overview/custom_instrumentation/go/#manually-creating-a-new-span
 	fc := graphql.GetFieldContext(ctx)
-	fieldSpan, ctx := StartSpanFromContext(ctx, "operation.field", ResourceName(fc.Field.Name))
-	fieldSpan.SetAttribute("field.type", fc.Field.Definition.Type.String())
+	fieldSpan, ctx := tracer.StartSpanFromContext(ctx, "operation.field", tracer.ResourceName(fc.Field.Name))
```
Vadman97, line 38: switch to only datadog tracing in this highlight-only util file. eventually, this file can be dropped. highlight gqlgen tracing happens via the `sdk/highlight-go/tracer.go` file

Vadman97:
that's right @SpennyNDaJets ! added a review comment to the sdk changes - tldr we can record gql traces now even without a session / request context

PR #6783, opened 2023-10-02
Fix aria kit menu z index

SpennyNDaJets:
## Summary
Had to https://github.com/highlight/highlight/pull/6762 since it caused issues with popovers rendering under the session search panel. Update the Error search panel to have a higher zIndex to match the session search panel.

Note: an alternative solution is to remove the zIndex of the session search panel, but unsure how this could break things

1) View the sessions page
2) Open the popover by search for filters
- [ ] Popover is rendered above to the side panel
- [ ] Popover is rendered above any session details
3) View the errors page
4) Open the popover by search for filters
- [ ] Popover is rendered above to the side panel
- [ ] Popover is rendered above any session details

Screenshot 2023-10-02 at 4 32 33 PM

N/A

N/A

### DIFF
--- a/frontend/src/pages/ErrorsV2/ErrorsV2.tsx
+++ b/frontend/src/pages/ErrorsV2/ErrorsV2.tsx
@@ -141,7 +141,11 @@ export default function ErrorsV2() {
 			</Helmet>
 
 			<Box cssClass={styles.container}>
-				{!isBlocked && <SearchPanel />}
+				{!isBlocked && (
+					<Box cssClass={styles.searchPanelContainer}>
+						<SearchPanel />
+					</Box>
+				)}
 
 				<div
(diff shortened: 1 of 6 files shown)


frontend/src/pages/ErrorsV2/ErrorsV2.tsx
```diff
@@ -141,7 +141,11 @@ export default function ErrorsV2() {
 			</Helmet>
 
 			<Box cssClass={styles.container}>
-				{!isBlocked && <SearchPanel />}
+				{!isBlocked && (
+					<Box cssClass={styles.searchPanelContainer}>
```
SpennyNDaJets, line 145: Change that differs from reverted PR

Vadman97 (approved):
looking great in the render.com preview

PR #6786, opened 2023-10-02
Fix gif on privacy docs

jay-khatri:


scripts/pgclirc
Vadman97, line 1: unrelated file

Vadman97 (approved):

Vadman97 (changes requested):
not working, i think because of CORS?

image

PR #6805, opened 2023-10-05
ensure error embeddings partitioned table runs migration without table level lock

Vadman97:
## Summary

Ensures that partitions of `error_object_embeddings_partitioned` are attached 
[without requiring a table lock](https://www.postgresql.org/docs/current/ddl-partitioning.html#DDL-PARTITIONING-DECLARATIVE-MAINTENANCE).
Automates creating L2 distance indexes for new partitions.

Local migration with debug
```
2023/10/05 11:45:39 /home/vkorolik/work/highlight/backend/model/model.go:1595
[1.964ms] [rows:0] 
			CREATE TABLE IF NOT EXISTS error_object_embeddings_partitioned_1007
			(LIKE error_object_embeddings_partitioned INCLUDING ALL);
		

2023/10/05 11:45:39 /home/vkorolik/work/highlight/backend/model/model.go:1602
[149.305ms] [rows:0] 
			CREATE INDEX ON error_object_embeddings_partitioned_1007
			USING ivfflat (gte_large_embedding vector_l2_ops) WITH (lists = 1000);
		

2023/10/05 11:45:39 /home/vkorolik/work/highlight/backend/model/model.go:1611
[2.417ms] [rows:0] 
			ALTER TABLE error_object_embeddings_partitioned 
			ATTACH PARTITION error_object_embeddings_partitioned_1007
			FOR VALUES IN ('1007');
```

No

No

### DIFF
--- a/backend/model/model.go
+++ b/backend/model/model.go
@@ -1591,7 +1591,7 @@ func MigrateDB(ctx context.Context, DB *gorm.DB) (bool, error) {
 	DB.Raw("select split_part(relname, '_', 5) from pg_stat_all_tables where relname like 'error_object_embeddings_partitioned%' order by relid desc limit 1").Scan(&lastCreatedPart)
 
 	// Make sure partitions are created for the next 1k projects, starting with the next partition needed
-	for i := lastCreatedPart + 1; i < lastVal+1002; i++ {
+	for i := lastCreatedPart + 1; i < lastVal+1000; i++ {
 		if err := DB.Exec(fmt.Sprintf(`
 			CREATE TABLE IF NOT EXISTS error_object_embeddings_partitioned_%d 
 			PARTITION OF error_object_embeddings_partitioned
--- a/backend/model/model.go
+++ b/backend/model/model.go
@@ -1575,7 +1575,7 @@ func MigrateDB(ctx context.Context, DB *gorm.DB) (bool, error) {
 
 	if err := DB.Exec(`
 		CREATE TABLE IF NOT EXISTS error_object_embeddings_partitioned
-		(LIKE error_object_embeddings INCLUDING DEFAULTS INCLUDING IDENTITY)
+		(LIKE error_object_embeddings INCLUDING ALL)
 		PARTITION BY LIST (project_id);
 	`).Error; err != nil {
 		return false, e.Wrap(err, "Error creating error_object_embeddings_partitioned")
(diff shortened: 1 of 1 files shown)


backend/model/model.go
```diff
@@ -1590,17 +1590,38 @@ func MigrateDB(ctx context.Context, DB *gorm.DB) (bool, error) {
...
 		}
+
+		if err := DB.Exec(fmt.Sprintf(`
+			CREATE INDEX ON error_object_embeddings_partitioned_%d
+			USING ivfflat (gte_large_embedding vector_l2_ops) WITH (lists = 1000);
+		`, i)).Error; err != nil {
+			return false, e.Wrapf(err, "Error creating index error_object_embeddings for index %d", i)
+		}
+
+		// in case this partition was already attached by a previous failed migration, try detach first
```
mayberryzane, line 1610: to minimize locking on the table, would probably be better to only attach partition and ignore any "partition already attached" error

backend/model/model.go
```diff
@@ -1590,17 +1590,38 @@ func MigrateDB(ctx context.Context, DB *gorm.DB) (bool, error) {
...
 	DB.Raw("select split_part(relname, '_', 5) from pg_stat_all_tables where relname like 'error_object_embeddings_partitioned%' order by relid desc limit 1").Scan(&lastCreatedPart)
 
-	// Make sure partitions are created for the next 1k projects
-	for i := lastCreatedPart; i < lastVal+1000; i++ {
-		sql := fmt.Sprintf(`
+	// Make sure partitions are created for the next 1k projects, starting with the next partition needed
+	for i := lastCreatedPart + 1; i < lastVal+1000; i++ {
+		if err := DB.Exec(fmt.Sprintf(`
 			CREATE TABLE IF NOT EXISTS error_object_embeddings_partitioned_%d
 			PARTITION OF error_object_embeddings_partitioned
```
mayberryzane, line 1597: I think this will require the same access exclusive lock and create it as a partition. I think `CREATE TABLE LIKE` would be better here https://www.postgresql.org/docs/current/ddl-partitioning.html#:~:text=parent%20table%27s%20definition%3A-,CREATE%20TABLE%20measurement_y2008m02,-(LIKE%20measurement%20INCLUDING

mayberryzane (approved):

PR #6816, opened 2023-10-07
Fix dialog story & try out reflame test update

lewisl9029:
## Summary

New version of Reflame tests can now reuse results from a previous test when it finds one with an identical dependency graph, a critical cost-savings measure for large test suites.

I also updated the test results display to use markdown content with embedded images directly instead of the images field in github checks, and split tests results into failed, completed, skipped sections, which should make test results a lot easier to digest.

While testing this out I actually found a bug with one of the stories and fixed it here. :)

Check out the new test results display [here](https://github.com/highlight/highlight/pull/6816/checks?check_run_id=17488390451), and let me know what you think!


Vadman97 (approved):
:pray: ty @lewisl9029 !

PR #6836, opened 2023-10-10
Blog: Vercel Edge Runtime Support

deltaepsilon:
image


ccschmitz (approved):
Great article, @deltaepsilon! Nice breakdown of the problem and work you did to get through it.

A few thoughts, but take or leave them because I think this is already solid.

1. Is this meant to be an announcement that we support the edge runtime? If yes, I'm wondering if we want to be more upfront/splashy about the fact Highlight now supports this.
2. It sounded like we're one of the only people supporting this, is that right? Is there anything noteworthy about that which we should call out? Maybe that's not something to emphasize in this article, but more broadly in our marketing.
3. Maybe this is the right tack to take, but we mention the effort that went into forking + patching things. My mind always thinks "maintenance nightmare" when I hear this, which I know you address later on, but wondering if we want to hold back on some of the gory details.

Again, great article, and feel free to take or leave any of those comments!

ccschmitz:
@deltaepsilon sorry, left a few comments but didn't realize it was set to auto merge - my bad 😸

Issue #6825, opened 2023-10-09
Fix filter highlighting on log alerts


ccschmitz:
![Image](https://github.com/highlight/highlight/assets/308182/a1493b35-1dff-48df-89aa-231f0f4e3456)

Also an issue with long search strings.

![Image](https://user-images.githubusercontent.com/17744174/274305322-eed76f32-88b2-442d-8d9f-fed9ab055694.png)

PR #6840, opened 2023-10-10
Traces table refinements

ccschmitz:
## Summary

* Cleans up some table of the interactive table styles on rows and cells
* Tightens up the content on the traces table and fixes some styling issues
* Ensures we don't lose search params when viewing individual traces

Screenshot 2023-10-11 at 6 26 03 AM

All client-side changes only and the page is only visible to Highlight admins.

Will have Julian click test when this goes out!


frontend/package.json
```diff
@@ -144,9 +144,9 @@
 		"@types/validator": "^13.1.3",
 		"@typescript-eslint/eslint-plugin": "^5.46.1",
 		"@typescript-eslint/parser": "^5.46.1",
-		"@vanilla-extract/css": "^1.11.1",
-		"@vanilla-extract/esbuild-plugin": "^2.2.2",
-		"@vanilla-extract/vite-plugin": "^3.8.2",
+		"@vanilla-extract/css": "^1.13.0",
+		"@vanilla-extract/esbuild-plugin": "^2.3.0",
+		"@vanilla-extract/vite-plugin": "^3.9.0",
```
ccschmitz, line 149: Was attempting to fix HMR when making changes to styles in the UI package and upgraded these packages as part of it. I didn't get it fixed, but figured it couldn't hurt to leave upgrades in.

packages/ui/tsconfig.json
```diff
@@ -9,7 +9,7 @@
 		"types": ["vitest/globals"],
 		"baseUrl": "src",
 		"paths": {
-			"@/*": ["./src/*"],
+			"@/*": ["./*"],
```
ccschmitz, line 12: Was also trying to fix import aliases. This got it working in the editor, but our builds were failing to resolve the aliases.

SpennyNDaJets (approved):
Looks great to me! Click tested a few instances of the table to check that it works!

SpennyNDaJets:
@ccschmitz possibly unrelated but noticed this weird grey shading when searching for a field
Screenshot 2023-10-11 at 10 16 29 AM

ccschmitz:
@SpennyNDaJets thanks for the review and mentioning that! I think that is a bug with highlighting search terms. I can look into this in https://github.com/highlight/highlight/issues/6825!

PR #6842, opened 2023-10-11
introduce UI for filtering ingest

Vadman97:
## Summary

Introduces a UI for filtering the 4 products.
* Set up an ingestion exclusion query
* Set up an ingestion sampling rate
* Set up a minute rate limit

This PR encompasses frontend changes tested locally. Backend changes tested separately in #6713
https://localhost:3000/1/settings/filters/

No

UI will be updated with design updates from @julian-highlight post launch week.


backend/clickhouse/traces.go
```diff
@@ -310,17 +311,23 @@ func (client *Client) ReadTracesMetrics(ctx context.Context, projectID int, para
 		}
 	}
 
+	groupStr := ""
+	for _, group := range groupBy {
+		groupStr += fmt.Sprintf("toString(TraceAttributes['%s']), ", group)
```
mayberryzane, line 316: I think this is vulnerable to sql injection b/c group comes directly from user input - can we hardcode `TraceAttributes['ingested']` or use `sb.Var` here?

mayberryzane (changes requested):
backend code lgtm other than a comment, and frontend looked good in the demo this morning!

.github/workflows/deploy.yml
```diff
@@ -332,6 +332,13 @@ jobs:
                   aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
                   aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
                   aws-region: us-east-2
+
+            - name: Login to Docker Hub
```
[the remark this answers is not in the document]
Vadman97: sorry, will separate it. this is to avoid some CI failures I noticed when the docker API rate limits us (using our paid logged in account avoids the rate limit)

backend/clickhouse/traces.go
Vadman97: good catch, i actually had tested for this and wasn't able to perform an attack (clickhouse detects that multiple commands are trying to run as part of one query and fails) but i've refactored this to use `sb.Var`

mayberryzane (approved):

PR #6853, opened 2023-10-12
support user bearer token auth

Vadman97:
## Summary

Extends our OAuth server implementation to support `client_credentials` oauth bearer tokens.
Tokens can be associated with an `AdminID` to support OAuth client authentication as that user.
Adds a python OAuth client example for using the API.

Running the new `sdk/highlight-py/e2e/highlight_oauth/main.py` 
with `HIGHLIGHT_OAUTH_CLIENT_ID` and `HIGHLIGHT_OAUTH_CLIENT_SECRET` set
to values inserted into load `o_auth_secret_stores` table.

without the `o_auth_operations` entry for `GetSessionsClickhouse`, permission is denied.
Screenshot 2023-10-12 at 3 05 48 PM

No

No


SpennyNDaJets (commented):
Code and test look good to me!

When I was click testing on the app, I noticed your branch was returning 500s when fetching for the session contents. I confirmed that main was successfully fetching these, but not sure if this is related to your changes
Screenshot 2023-10-13 at 10 32 23 AM

Vadman97:
> Code and test look good to me!
> 
> When I was click testing on the app, I noticed your branch was returning 500s when fetching for the session contents. I confirmed that main was successfully fetching these, but not sure if this is related to your changes Screenshot 2023-10-13 at 10 32 23 AM

You're right - I think making all private graph endpoints require auth breaks some of the self-hosted features (the local resource urls on `/private/direct/...` do not have the token added so they are failing auth.

That change isn't related to the rest of the change so i'll revert it.

Vadman97:
@SpennyNDaJets fixed now. troubleshooting the new E2E test.

Screenshot 2023-10-13 at 10 32 55 AM

SpennyNDaJets (approved):
LGTMPS

PR #6854, opened 2023-10-12
Fix log alert saving logic to use update, prevent duplication

eightants:
## Summary

Fix: https://www.loom.com/share/2020b947f0fc4dd28eb9e4ee9507388b

Did some deep dive into it because I had trouble repo'ing. It looks like it only affects Log Alerts, as I tried it on all alerts in prod and only that one cause duplication. 

Solution: Change save() to updates() because save creates a new object if it doesn't exist already for a primary key. 
Also bug fix by adding a redirect to alerts home after updating log alert

### DIFF
--- a/backend/private-graph/graph/schema.resolvers.go
+++ b/backend/private-graph/graph/schema.resolvers.go
@@ -3070,7 +3070,7 @@ func (r *mutationResolver) UpdateLogAlert(ctx context.Context, id int, input mod
 
 	if err := r.DB.Model(&model.LogAlert{Model: model.Model{ID: id}}).
 		Where("project_id = ?", input.ProjectID).
-		Save(alert).Error; err != nil {
+		Updates(alert).Error; err != nil {
 		return nil, e.Wrap(err, "error updating log alert")
 	}
 
--- a/frontend/src/pages/Alerts/LogAlert/LogAlertPage.tsx
+++ b/frontend/src/pages/Alerts/LogAlert/LogAlertPage.tsx
@@ -348,6 +348,7 @@ export const LogAlertPage = () => {
 							})
 								.then(() => {
 									message.success(`Log alert ${createStr}d!`)
+									navigate(`/${project_id}/alerts`)
 								})
 								.catch(() => {
 									message.error(


backend/private-graph/graph/schema.resolvers.go
```diff
@@ -3070,7 +3070,7 @@ func (r *mutationResolver) UpdateLogAlert(ctx context.Context, id int, input mod
 
 	if err := r.DB.Model(&model.LogAlert{Model: model.Model{ID: id}}).
 		Where("project_id = ?", input.ProjectID).
```
Vadman97, line 3072: is the problem that we should be searching by alert id? `Where("id = ", input.ID)`, that way GORM will know that this is an existing entry

frontend/src/pages/Alerts/LogAlert/LogAlertPage.tsx
Vadman97, line 1: for some reason `useUpdateLogAlertMutation` doesn't actually refetch the log alerts payload, so the updates don't show up on the alerts page. could we check that this is fixed?

backend/private-graph/graph/schema.resolvers.go
eightants: Doesn't `{Model: model.Model{ID: id}})` already match the alert id? That's how the other alerts are being updated and it works for those.

Vadman97: oh interesting, TIL

frontend/src/pages/Alerts/LogAlert/LogAlertPage.tsx
Vadman97: we can do this separately too if you just want to get this in

Vadman97 (approved):

frontend/src/pages/Alerts/LogAlert/LogAlertPage.tsx
eightants: Yeah can get this other change in the next PR. I'm not too sure why it wouldn't refetch since the other alerts do, and it's the same logic.

PR #6855, opened 2023-10-12
update style of traces histogram / latency chart to match designs

mayberryzane:
## Summary
- updates traces page to match the figma design https://www.figma.com/file/rdBPNzn7Klk6RG1LHUCE5B/Screen-Designs?node-id=12256%3A405310&mode=dev

- click tested locally

- no

- no


frontend/src/pages/Traces/LatencyChart.tsx
```diff
@@ -0,0 +1,147 @@
...
+						animationDuration={100}
+						dot={false}
+					/>
+				))}
+				{/* {referenceAreaProps && (
+            <ReferenceArea
+                {...referenceAreaProps}
+                isFront
+            />
+        )} */}
```
ccschmitz, line 143: Is this coming back at some point, or do you think we should clean up entirely?

ccschmitz (approved):
Looks great!

One minor thing I noticed is that I think [this offset](https://github.com/highlight/highlight/blob/dfaa3c0d70e0372974c7b9a95cd5b715f61fb1f1/frontend/src/pages/Traces/TracesList.tsx#L60) we have on the traces table is now off a little and the last item isn't being fully shown (note the icon popping out in the bottom left of the screenshot).

Screenshot 2023-10-12 at 5 04 21 PM

Probably not a great way to handle that in the first place, and something I can look at in a future PR if you don't want to worry about it here 😅 

Overall, it's looking great!

frontend/src/pages/Traces/LatencyChart.tsx
mayberryzane: nope! deleted in latest commit

PR #6865, opened 2023-10-13
add backend setup logic for traces, add quickstart guide for go

mayberryzane:
## Summary
- adds logic to private graph / trace ingest to record and query when backend traces are being sent
- add traces quickstart step guarded by `isHighlightAdmin`

- clicktested locally

- will add guide for OLTP and add to docs page later

- no


Vadman97 (approved):
🚢

PR #6875, opened 2023-10-17
show error boundary feedback on error instances

### LINKED ISSUE #6642: show error boundary feedback on error instances
we currently associate error boundary feedback with a session.
we should surface it on the error instance view via that association per [the following design.](https://www.figma.com/file/rdBPNzn7Klk6RG1LHUCE5B/Screen-Designs?type=design&node-id=11613%3A575399&mode=dev)

jemiluv8:
## Summary
Adds support for querying session comments on the error_instance
Shows error boundary session comment feedback on UI as [per design](https://www.figma.com/file/rdBPNzn7Klk6RG1LHUCE5B/Screen-Designs?node-id=11613%3A575444&mode=dev)

I made a page crush and the error boundary feedback component was shown
I filled the form and submitted
On the error instance, I checked to make sure the feedback is shows
And the link on the feedback component takes me to the related session

No

Yes

## LOOM VIDEO
This shows the feedback on the error instance under user details
Also demonstrates the link to related session
https://www.loom.com/share/b357e312b57d4b5cb362ec4c816d029e?sid=1fd0eac7-b127-4030-9f25-0da9597f4eaf

### SCREENSHOT
Screenshot 2023-10-17 at 2 30 37 PM

/claim #6642
fixes #6642

### DIFF
--- a/backend/private-graph/graph/schema.graphqls
+++ b/backend/private-graph/graph/schema.graphqls
@@ -81,7 +81,7 @@ type Session {
 	deviceMemory: Int
 	last_user_interaction_time: Timestamp!
 	chunked: Boolean
-	session_comments: [SessionComment]
+	session_comments: [SessionComment]!
 }
 
 type SessionInterval {
--- a/backend/private-graph/graph/schema.graphqls
+++ b/backend/private-graph/graph/schema.graphqls
@@ -81,6 +81,7 @@ type Session {
 	deviceMemory: Int
 	last_user_interaction_time: Timestamp!
 	chunked: Boolean
+	session_comments: [SessionComment]
 }
 
 type SessionInterval {
--- a/backend/private-graph/graph/schema.graphqls
+++ b/backend/private-graph/graph/schema.graphqls
@@ -81,7 +81,7 @@ type Session {
 	deviceMemory: Int
 	last_user_interaction_time: Timestamp!
 	chunked: Boolean
-	session_comments: [SessionComment]
+	session_feedback: [SessionComment!]
 }
 
 type SessionInterval {
--- a/backend/private-graph/graph/schema.graphqls
+++ b/backend/private-graph/graph/schema.graphqls
@@ -81,7 +81,7 @@ type Session {
 	deviceMemory: Int
 	last_user_interaction_time: Timestamp!
 	chunked: Boolean
-	session_comments: [SessionComment]!
+	session_comments: [SessionComment]
 }
 
 type SessionInterval {
--- a/backend/private-graph/graph/schema.resolvers.go
+++ b/backend/private-graph/graph/schema.resolvers.go
@@ -3117,7 +3117,7 @@ func (r *mutationResolver) UpdateLogAlert(ctx context.Context, id int, input mod
 
 	if err := r.DB.Model(&model.LogAlert{Model: model.Model{ID: id}}).
 		Where("project_id = ?", input.ProjectID).
-		Save(alert).Error; err != nil {
+		Updates(alert).Error; err != nil {
 		return nil, e.Wrap(err, "error updating log alert")
 	}
 
--- a/backend/private-graph/graph/schema.resolvers.go
+++ b/backend/private-graph/graph/schema.resolvers.go
@@ -7876,7 +7876,7 @@ func (r *sessionResolver) SessionComments(ctx context.Context, obj *model.Sessio
 	}
 	sessionComments := []*model.SessionComment{}
 
-	if err := r.DB.Preload("Attachments").Preload("Replies").Where(model.SessionComment{SessionId: s.ID, SessionCommentType: SessionCommentType.Feedback}).Order("timestamp asc").Find(&sessionComments).Error; err != nil {
+	if err := r.DB.Preload("Attachments").Preload("Replies").Where(model.SessionComment{SessionId: s.ID}).Order("timestamp asc").Find(&sessionComments).Error; err != nil {
 		return nil, e.Wrap(err, "error querying session comments for session")
 	}
 	return sessionComments, nil
--- a/backend/private-graph/graph/schema.resolvers.go
+++ b/backend/private-graph/graph/schema.resolvers.go
@@ -7864,22 +7864,22 @@ func (r *sessionResolver) DeviceMemory(ctx context.Context, obj *model.Session) 
 	return deviceMemory, nil
 }
 
-// SessionComments is the resolver for the session_comments field.
-func (r *sessionResolver) SessionComments(ctx context.Context, obj *model.Session) ([]*model.SessionComment, error) {
+// SessionFeedback is the resolver for the session_feedback field.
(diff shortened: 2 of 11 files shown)


backend/private-graph/graph/schema.resolvers.go
```diff
@@ -7864,6 +7864,24 @@ func (r *sessionResolver) DeviceMemory(ctx context.Context, obj *model.Session)
 	return deviceMemory, nil
 }
 
+// SessionComments is the resolver for the session_comments field.
+func (r *sessionResolver) SessionComments(ctx context.Context, obj *model.Session) ([]*model.SessionComment, error) {
```
jemiluv8, line 7868: Resolver to return session comments preloaded with attachments and replies.
Not currently using the attachments and replies on UI at the moment. Have to consider removing the
preload.

jemiluv8 (commented):
Self review completed.

julian-highlight:
> @mayberryzane, @julian-highlight I've completed a self review.

Yo! Great work. A bit of feedback regarding the front end:
- Can we make sure to use the right spacing, fonts & colors for the displayed feedback:
image
- The boundary feedback should have 'static/surface/nested' as it's background color

Apart from that, this looks great! @jemiluv8

jemiluv8:
Ok @julian-highlight. Will make changes and update you.

jemiluv8:
Hello @julian-highlight, below are some of the issues I fixed. 

- [x] set background of feedback component to static/surface/nested
- [x] fix font size of actual feedback text to 13px and set its color to static/default/content 
- [x] fix padding of main container 8 12 12 12
- [x] fix gap between "Feedback #0X" and the timestamp to 6px

I believe that is all the issues covered. Let me know if I've overlooked something.

Screenshot 2023-10-18 at 3 47 04 PM

Screenshot 2023-10-18 at 3 54 29 PM

backend/private-graph/graph/schema.resolvers.go
```diff
@@ -7864,6 +7864,24 @@ func (r *sessionResolver) DeviceMemory(ctx context.Context, obj *model.Session)
...
+		sessionComments := []*model.SessionComment{}
+		return sessionComments, nil
+	}
+	s, err := r.canAdminViewSession(ctx, obj.SecureID)
+	if err != nil {
+		return nil, err
+	}
+	sessionComments := []*model.SessionComment{}
+
+	if err := r.DB.Preload("Attachments").Preload("Replies").Where(model.SessionComment{SessionId: s.ID, Type: model.SessionCommentTypes.FEEDBACK}).Order("timestamp asc").Find(&sessionComments).Error; err != nil {
```
jemiluv8, line 7879: Query session comments with
```golang
Type: model.SessionCommentTypes.FEEDBACK
```

frontend/src/pages/Sessions/SessionsFeedV3/MinimalSessionCard/utils/utils.ts
```diff
@@ -2,8 +2,13 @@ import { Maybe, Session } from '@graph/schemas'
 import { H } from 'highlight.run'
 import validator from 'validator'
 
+type SessionWithIdentityInformation = Pick<
+	Session,
+	'user_properties' | 'identifier' | 'fingerprint'
+>
+
 export const getIdentifiedUserProfileImage = (
-	session?: Maybe<Partial<Session>>,
```
jemiluv8, line 6: Got a weird typescript error after the following field was added to the Session type in the frontend ```generated/schemas.tsx```

```typescript
session_comments?: Maybe<Array<Maybe<SessionComment>>>
```

This was the "solution/workaround" that satisfied the ts compiler.

julian-highlight:
> Hello @julian-highlight, below are some of the issues I fixed.
> 
> * [x]  set background of feedback component to static/surface/nested
> * [x]  fix font size of actual feedback text to 13px and set its color to static/default/content
> * [x]  fix padding of main container 8 12 12 12
> * [x]  fix gap between "Feedback #0X" and the timestamp to 6px
> 
> I believe that is all the issues covered. Let me know if I've overlooked something.
> 
> Screenshot 2023-10-18 at 3 47 04 PM Screenshot 2023-10-18 at 3 54 29 PM

sorry for the picky comments but can we also center align the 'go to' tag with the text on the left?
The border seems to be overlapped by the new background color as well. (Visible on the corners at the bottom). 

If you could resolve these two issues I'm entirely happy from a design standpoint! :) 
* [ ] Center align 'go to' action
* [ ] Fix overlapping background on the bottom corners

jemiluv8:
@julian-highlight, I don't mind the "picky" comments.
I'll make the changes and update.

julian-highlight:
@jemiluv8 Looks great! Thanks for the effort. 

@mayberryzane design looks good now!

mayberryzane:
hi @jemiluv8 , taking a look at this today

backend/private-graph/graph/schema.resolvers.go
mayberryzane: could you rename the field from `session_comments` to `session_feedback` so it's clear that only feedback is included in the results? I also think you can omit `.Preload("Attachments").Preload("Replies")` since I don't think those fields are used for feedback?

backend/private-graph/graph/schema.graphqls
```diff
@@ -81,6 +81,7 @@ type Session {
 	deviceMemory: Int
 	last_user_interaction_time: Timestamp!
 	chunked: Boolean
+	session_comments: [SessionComment]
```
mayberryzane, line 84: could you change the type of this field to `[SessionComment!]!`? Will make it easier in the frontend code if we don't have to check that the array exists and the elements aren't null. I don't think there's a case where we need a null result because we could just return an empty array instead.

mayberryzane (changes requested):
Thanks @jemiluv8 , I have two small pieces of feedback, the rest of the PR looks good to me!

jemiluv8:
ok @mayberryzane. I'll make those changes now.

backend/private-graph/graph/schema.graphqls
jemiluv8: I'm leaning towards ```session_feedback: [SessionComment!]``` instead. The is because 

- there are other places that query session without session_feedback so in those cases session feedback is not returned by their respective graph queries at all
- I get ts errors involving other components that use the Session type. Those components do not query the session_feedback on their sessions so that the field is indeed missing.

backend/private-graph/graph/schema.resolvers.go
jemiluv8: I've made the relevant changes.

jemiluv8:
@mayberryzane  I've applied the renaming from session_comments to session_feedback. 
Also updated the query type to return ```[SessionComment!]``` instead of ```[SessionComment!]!``` because doing so will
break other places where the shared session is used. At the moment we only query session_feedback

mayberryzane (approved):
LGTM! Thanks for addressing the feedback. Running the Vercel check and will merge in once all checks finish

Issue #5948, opened 2023-07-16
Allow Client to Customize and Sanitize Network Request and Response


jdorleans:
**The Problem**
While Highlight does provide whitelisting to [redact network header and body](https://www.highlight.io/docs/getting-started/client-sdk/replay-configuration/recording-network-requests-and-responses#redacting-headers-and-bodies), the current functionality is very limiting. It is a non-trival job to whitelist all sort of data types and keys for PIIs, tokens, passwords and so on by hand. It's specially harder when using graphql due to the single endpoint nature of the protocol.

**A Solution**
Instead of having a rigid list of keywords and url rules, I'd like to be able to modify the entire network data to my liking based on my application needs and privacy rules. One good example is [LogRocket network sanitizer functions](https://docs.logrocket.com/v1.0/reference/network) where it allows for the client application to customize the request/response right before sending any data to its platform.

One can for instance, modify data from request/response headers, body etc as well as prevent the entire request/response to even be logged. Here's a simple example from LogRocket (please check link above for more):

```
LogRocket.init(YOUR_APP_ID, {
  network: {
    requestSanitizer: request => {
      // whatever if logic based on request data
      if (request.url.toLowerCase().indexOf('ignore') !== -1) {
        // ignore the entire request/response pair (no network logs)
        return null;
      }
     
      // otherwise, log the request normally
      return request;
    },
  },
});

LogRocket.init(YOUR_APP_ID, {
  network: {
    requestSanitizer: request => {
      // whatever if logic based on request data
      if (request.url.toLowerCase().indexOf('ignore') !== -1) {
        // remove the request body entirely
        request.body = null;

        // alternatively, we can modify the body with a custom logic
        request.body = mySanitizer(request.body);
      }

      // log the modified request
      return request;
    },
  },
});

// same logic applies for headers and the entire response object
```

**Notes**
My team and I have considered moving from LogRocket to Highlight, however this missing capability prevents us to do so.

**Related to**
This request might be related to:
- https://github.com/highlight/highlight/issues/3782
- https://github.com/highlight/highlight/issues/4060

PR #6892, opened 2023-10-19
Allow users to pass in sanitize method for XHR/fetch requests

SpennyNDaJets:
## Summary
While Highlight does provide whitelisting to [redact network header and body](https://www.highlight.io/docs/getting-started/client-sdk/replay-configuration/recording-network-requests-and-responses#redacting-headers-and-bodies), the current functionality is very limiting. It is a non-trival job to whitelist all sort of data types and keys for PIIs, tokens, passwords and so on by hand. It's specially harder when using graphql due to the single endpoint nature of the protocol.

Instead of having a rigid list of keywords and url rules, I'd like to be able to modify the entire network data to my liking based on my application needs and privacy rules. Allow for the client application to customize the request/response right before sending any data to its platform.

One can for instance, modify data from request/response headers, body etc as well as prevent the entire request/response to even be logged.

1) Add in a `requestResponseSanitizer` to the `networkRecording` hash in the frontend's Highlight options. Example:
```
requestResponseSanitizer: (pair) => {
    if (pair.request.verb === 'POST') {
        console.log('blocking request', pair)
	return null
    } else {
	console.log('allowing request', pair)
	pair.request.headers ||= {}
	pair.request.headers['editTest'] = 'true'
    }

    return pair
}
```
2) Build the SDK and start a Highlight session
3) Click around to get different types of requests
4) Close the session, and let it process
5) View the processed sessions network logs
- [ ] Requests ignored appropriately
- [ ] Requests updated appropriately
- [ ] No known missing requests

https://www.loom.com/share/47ce07c349f14e9aa3fb802b21221167?sid=7153b018-f6ba-46a5-bff8-6c635b269607

Monitor that Highlight records properly without the `requestResponseSanitizer` method

N/A

### DIFF
--- a/docs-content/getting-started/3_client-sdk/7_replay-configuration/recording-network-requests-and-responses.md
+++ b/docs-content/getting-started/3_client-sdk/7_replay-configuration/recording-network-requests-and-responses.md
@@ -2,7 +2,7 @@
 title: Recording Network Requests and Responses
 slug: recording-network-requests-and-responses
 createdAt: 2021-09-14T00:14:21.000Z
-updatedAt: 2022-03-07T22:43:17.000Z
+updatedAt: 2023-10-25T15:40:08.873Z
 ---
 
 Highlight out of the box shows you all the network requests durations, response codes, and sizes for a session. If you'd like more data such as the headers and bodies, you can enable recording of network requests and responses by setting `networkRecording.recordHeadersAndBody` (see [NetworkRecordingOptions](../../../sdk/client.md#Hinit)) to `true` when initializing Highlight.
@@ -59,6 +59,38 @@ If you are dealing with sensitive data or want to go the allowlist approach then
 You can also redact specific headers by using `networkRecording.networkHeadersToRedact` and redact specific keys in the request/response body with `networkRecoding.networkBodyKeysToRedact`.
 
 This configuration is only available for `highlight.run` versions newer than `4.1.0`.
+
+## Custom Sanitizing of Response and Requests
+
+If you need more granular control over managing and ignoring data in request and responses, then you can create a sanitize function to process every recorded request/response pair. This is configured by using the `networkRecording.requestResponseSanitizer` option.
+
+The `networkRecording.requestResponseSanitizer` method receives a RequestResponsePair, and should return an object of the same type or a `null` value. Returning a `null` value means that Highlight will drop the request, and no related network logs will be seen in the session replay.
+
+Dropping logs is not recommended unless necessary, as it can cause issues with debugging due to the missing requests. Rather, it is recommended to delete or redact header and body fields in this method.
+
+This configuration is only available for `highlight.run` versions newer than `8.1.0`.
+
+```typescript
+H.init('<YOUR_PROJECT_ID>', {
+  networkRecording: {
+    enabled: true,
+    recordHeadersAndBody: true,
+    requestResponseSanitizer: (pair) => {
+      if (pair.request.url.toLowerCase().indexOf('ignore') !== -1) {
+        // ignore the entire request/response pair (no network logs)
+        return null
+      }
+	 
+      if (pair.response.body.indexOf('secret') !== -1) {
+        // remove the body in the response
+        delete pair.response.body;
+      }
+	 
+      return pair
+    }
+  },
+})
+```
 
 ## API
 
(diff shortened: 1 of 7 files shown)


SpennyNDaJets:
@Vadman97 one call out that I wanted to make from what we discussed:

Opted that when a user returns `null` from this method, the request is not logged at all. I believe this is
1) More in line with the feature ask in https://github.com/highlight/highlight/issues/5948
2) Allows the user more flexibility - they can easily redact any data they feel fit, and should be allowed to ignore requests all together if that is their case.

docs-content/getting-started/3_client-sdk/7_replay-configuration/recording-network-requests-and-responses.md
```diff
@@ -60,6 +60,38 @@ You can also redact specific headers by using `networkRecording.networkHeadersTo
 
 This configuration is only available for `highlight.run` versions newer than `4.1.0`.
 
+## Custom Sanitizing of Response and Requests
+
+If you need more granular control over managing and ignoring data in request and responses, then you can create a sanitize function to process every recorded request/response pair. This is configured by using the `networkRecording.requestResponseSanitizer` option.
+
+The `networkRecording.requestResponseSanitizer` method receives a RequestResponsePair, and should return an object of the same type or a `null` value. Returning a `null` value means that Highlight will drop the request, and no related network logs will be seen in the session replay.
```
deltaepsilon, line 67:
```suggestion
The `networkRecording.requestResponseSanitizer` method receives a Request/Response pair, and should return an object of the same type or a `null` value. Returning a `null` value means that Highlight will drop the request, and no related network logs will be seen in the session replay.
```

docs-content/getting-started/3_client-sdk/7_replay-configuration/recording-network-requests-and-responses.md
```diff
@@ -60,6 +60,38 @@ You can also redact specific headers by using `networkRecording.networkHeadersTo
 
 This configuration is only available for `highlight.run` versions newer than `4.1.0`.
 
+## Custom Sanitizing of Response and Requests
+
+If you need more granular control over managing and ignoring data in request and responses, then you can create a sanitize function to process every recorded request/response pair. This is configured by using the `networkRecording.requestResponseSanitizer` option.
```
deltaepsilon, line 65:
```suggestion
Create a sanitize function to gain granular control of the data that your client sends to Highlight. The sanitize function is defined in the second argument of `H.init` under `networkRecording.requestResponseSanitizer`.
```

sdk/client/src/listeners/network-listener/utils/utils.ts
```diff
@@ -31,10 +32,50 @@ type PerformanceResourceTimingWithRequestResponsePair =
...
+	) => RequestResponsePair | null
+}
+
+const sanitizeRequestResponsePair = (
+	pair: RequestResponsePair,
+	{
+		headersToRedact,
+		headersToRecord,
+		requestResponseSanitizer,
+	}: SanatizeOptions,
```
deltaepsilon, line 49: Spelling: `SanitizeOptions`

deltaepsilon (approved):

sdk/client/src/listeners/network-listener/utils/utils.ts
```diff
@@ -108,26 +149,49 @@ export const matchPerformanceTimingsWithRequestResponsePair = (
...
+					performanceTiming.requestResponsePair
+
+				if (requestResponsePair) {
+					requestResponsePair = sanitizeRequestResponsePair(
+						performanceTiming.requestResponsePair,
+						sanitizeOptions,
+					)
+
+					// ignore request if it was filtered out by the user defined sanitizer
+					if (!requestResponsePair) {
```
Vadman97, line 167: cool, so returning `false`, `undefined,` or any other null-ish value would work

Vadman97 (approved):
looks good, but i'm struggling to get recording working - all network requests seem to be blank for config
```

		requestResponseSanitizer: (pair) => {
			if (Math.random() < 0.1) return null;
			return {
				...pair,
				response: {
					...pair.response,
					headers: {hello: 'world'},
				}
			}
		}
```
might be missing something, not sure
Screenshot 2023-10-25 at 8 21 11 PM

SpennyNDaJets:
> d19676f

Good catch! I had a little regression in one of the refactors, but should be addressed now
Screenshot 2023-10-26 at 10 47 43 AM

PR #6897, opened 2023-10-19
cleanup internal traces

Vadman97:
## Summary

Noticed with @ccschmitz that we have some traces (`highlight/chi`) that do not have
the parent span recorded. We suspect this is because we are sampling by trace id, but our
middleware ordering did not propagate the trace id correctly.

Renaming some middleware span naming to be more consistent with the rest of our traces.

I've also refactored the private graph tracing per the [DD docs for resource name](https://docs.datadoghq.com/tracing/services/resource_page/), 
where we have some confusion with how we name traces. 

Local deploy showing correct propagation of the trace.
image

No

No

### DIFF
--- a/backend/main.go
+++ b/backend/main.go
@@ -405,6 +405,7 @@ func main() {
 		}
 
 		r.Route("/oauth", func(r chi.Router) {
+			r.Use(highlightChi.Middleware)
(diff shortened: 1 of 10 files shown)


ccschmitz (approved):
Awesome - thank you!

PR #6900, opened 2023-10-20
fix forceNew maintaining user

Vadman97:
## Summary

The client would be resetting the user identifier of the new session when
`H.init({forceNew: true})` is set, since the new session is likely going to `H.identify` as
a different user, but we only maintain the first user identifier.

Local deploy.

Changeset

No

### DIFF
--- a/.changeset/two-dots-occur.md
+++ b/.changeset/two-dots-occur.md
@@ -1,0 +1,5 @@
+---
+'highlight.run': patch
+---
+
+make H.init({forceNew: true}) reset the user identifier of the new session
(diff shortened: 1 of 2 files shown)


SpennyNDaJets (approved):
Code looks good!

Curious, how do we identify the user on a reset when forceNew is set?

Vadman97:
> Code looks good!
> 
> Curious, how do we identify the user on a reset when forceNew is set?

The customer would have to call `H.identify()` after `H.init({forceNew: true})` to set the new user's identity.

PR #6915, opened 2023-10-21
search sessions / errors by service name / version

mayberryzane:
## Summary
- adds service name + version to session search UI
- adds service version to error search UI
- indexes service name and version as session fields when a session is initialized
- Q: is a session's app version equivalent to service version? the way I implemented it it's pretty much an alias, do we want to remove app version from the query builder?
- https://www.loom.com/share/55eeabed43a84e81a263936bac848294

- clicktested locally

- can backfill service name and version fields for existing sessions if we want to 

- no

### DIFF
--- a/backend/clickhouse/sessions.go
+++ b/backend/clickhouse/sessions.go
@@ -4,9 +4,10 @@ import (
 	"context"
 	"errors"
 	"fmt"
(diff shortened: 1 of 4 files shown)


Vadman97 (approved):
thanks! should be fine to not backfill

PR #6917, opened 2023-10-21
use rollup to bundle @highlight-run/node sdk

Vadman97:
## Summary

It appears that our node sdk has been broken because `@opentelemetry/context-async-hooks`
does not support ESM builds (which we started to support when we refactored the SDK for next edge runtime).

This change updates the `@highlight-run/node` build to make sure
the dynamic require can be handled correctly in ESM environments.
This involves changing the build to rollup which will bundle all dependencies by default
and will ensure that CJS dependencies are converted to ESM for compatibility with all environments.

Switches to `proto` trace export which seems to be more reliable in my testing for both edge and serverless.

New e2e express example
before any changes, failing
image

after changes to the `@highlight-run/node` sdk only
image

E2E next app working: https://vercel.live/link/nextjs-highlight-demo-h31yfphet-highlight-run.vercel.app?via=deployment-domains-list-commit

image

Changeset

No


deltaepsilon (approved):

e2e/express/package.json
```diff
@@ -0,0 +1,17 @@
+{
+	"name": "e2e-express",
```
Vadman97, line 2: @deltaepsilon the reason i'm naming this `e2e-express` is that `express` collides with the `express` js sdk itself.

PR #6934, opened 2023-10-24
Add trace information to network request panel

ccschmitz:
## Summary

The main update in this PR is adding the trace flame graph and span details on the network resource panel. There are a number of other updates to traces and the flame graph added to support this.

* Moves the data to display a trace into a `TraceContext` which provides more flexibility for building UI with the trace data (e.g. allowed me to display span attributes and a flame graph separately on the network request panel).
* Fixes an issue with the time display of ticks for longer durations + updates the numbers to use scientific notation, which is easier to parse.
* Assigns `startTime` to all spans on the server which is the unix time in nanoseconds. This makes our flame graph span placement more precise since we could only get to millisecond precision when parsing `timestamp` in the client.
* Hides `children` in the span attributes, fixing some UI jank when updating the span attributes in the JSON viewer.
* Throttles the zoom handler and removes the scaling factor which makes it easier/faster to zoom into the area of the graph you want to see.

https://www.loom.com/share/144c34193ea149c4a309b3b817860bae

Local click test. Can't be click tested in the PR previews because of the backend changes.

Only backend change is adding `startTime` on spans. This should be fairly safe, but will click test to confirm as this rolls out.

N/A - no design changes in this PR.


SpennyNDaJets (approved):
Code looks good to me! Just wanted to call out a couple things I noticed from click testing, but not sure if this is a local setup error or due to previous data formats?

1. Hit a type error when attempting to scroll - did not happen everytime though
2. "Invalid date" string in trace sidebar


Screenshot 2023-10-25 at 12 56 34 PM

ccschmitz:
@SpennyNDaJets thank you for catching those issues! I addressed both and cleaned up a bunch of UI to better match [the latest mocks](https://www.figma.com/file/rdBPNzn7Klk6RG1LHUCE5B/Screen-Designs?type=design&node-id=12576%3A20863&mode=design&t=1WsLdq4ODD2Boskj-1) (added prev/next buttons + hotkeys).

Screenshot 2023-10-25 at 3 03 40 PM

SpennyNDaJets (approved):
image

PR #6946, opened 2023-10-25
Support NodeNext module resolution

### LINKED ISSUE #6798: Types don't work with ESM
# Describe the bug
The types are incompatible with ESM because the imports in the `.d.ts` files aren't written in a way that ESM understands.

# To Reproduce

1. Create a Remix project (`npx create-remix@latest`).
2. Ensure `package.json` has `"type": "module"`.
3. Ensure `tsconfig.json` has `"moduleResolution": "nodenext"` (which will make it resolve the types the ES Modules way).
4. Follow the installation instructions of highlight.io for Remix.
5. See the type errors (as showed in the screenshot).

# Expected behavior
All relative paths in the `.d.ts` generated by highlight.io should have the `.js` extension (yes, it must be `.js`, not `.ts`) and never point to a folder.

In other words, this is compatible with ESM:
✅ import foo from './some-file.js
✅ import foo from './some-folder/index.js

And this is incompatible:
❌ import foo from './some-file'
❌ import foo from './some-folder'

# Screenshots
image

image

The imports above should be changed to:
```diff
-import type { HighlightOptions, HighlightPublicInterface } from '../../client/src/types/types';
-import { HighlightSegmentMiddleware } from './integrations/segment';
-import configureElectronHighlight from './environments/electron';
+import type { HighlightOptions, HighlightPublicInterface } from '../../client/src/types/types.js';
+import { HighlightSegmentMiddleware } from './integrations/segment.js';
+import configureElectronHighlight from './environments/electron.js';
```

# Additional context
It's the same as what was happening in https://github.com/rtivital/mantine-v7/issues/13. Also read https://www.typescriptlang.org/docs/handbook/esm-node.html.

deltaepsilon:
Closes #6798 

## Summary

I'm uncertain if adding these `.js` file suffixes hurts anything. It does add `NodeNext` support for TypeScript module resolution.


sdk/firstload/src/index.tsx
```diff
@@ -14,21 +14,24 @@ import {
...
 import firstloadVersion from './__generated/version'
 import {
 	SessionData,
 	getPreviousSessionData,
-} from '@highlight-run/client/src/utils/sessionStorage/highlightSession'
+} from '@highlight-run/client/src/utils/sessionStorage/highlightSession.js'
 import { initializeFetchListener } from './listeners/fetch'
 import { initializeWebSocketListener } from './listeners/web-socket'
-import { listenToChromeExtensionMessage } from './browserExtension/extensionListener'
+import { listenToChromeExtensionMessage } from './browserExtension/extensionListener.js'
```
Vadman97, line 34: how does this work for the files that are imported? ie. `./listeners/fetch/index.ts` imports more files without a `.js` extension

sdk/firstload/src/index.tsx
```diff
@@ -14,21 +14,24 @@ import {
...
+	MixpanelAPI,
+	setupMixpanelIntegration,
+} from './integrations/mixpanel.js'
 
-import { FirstLoadListeners } from '@highlight-run/client/src/listeners/first-load-listeners'
-import { GenerateSecureID } from '@highlight-run/client/src/utils/secure-id'
-import { HighlightSegmentMiddleware } from './integrations/segment'
-import configureElectronHighlight from './environments/electron'
+import { FirstLoadListeners } from '@highlight-run/client/src/listeners/first-load-listeners.js'
+import { GenerateSecureID } from '@highlight-run/client/src/utils/secure-id.js'
```
Vadman97, line 24: how does this work if the `.js` files do not exist until the build?

sdk/firstload/src/index.tsx
```diff
@@ -14,21 +14,24 @@ import {
...
+import { HighlightSegmentMiddleware } from './integrations/segment.js'
+import configureElectronHighlight from './environments/electron.js'
 import firstloadVersion from './__generated/version'
 import {
 	SessionData,
 	getPreviousSessionData,
-} from '@highlight-run/client/src/utils/sessionStorage/highlightSession'
+} from '@highlight-run/client/src/utils/sessionStorage/highlightSession.js'
```
Vadman97, line 31: is there is no way to set something in our tsconfig to do this across the entire package?

Vadman97 (changes requested):
was this tested with both `node` and `nodenext` typescript module resolution?

sdk/firstload/src/index.tsx
deltaepsilon: VSCode does not complain. Apparently this is a common issue?

sdk/firstload/src/index.tsx
deltaepsilon: Stackoverflow indicates that it's either done in-code or by regexp replacing:

https://stackoverflow.com/questions/62619058/appending-js-extension-on-relative-import-statements-during-typescript-compilat

sdk/firstload/src/index.tsx
deltaepsilon: I just added `*/index.js` to these imports and it works great in both `node` and `nodenext`.

Vadman97 (approved):
needed to add 8b867a50330231497bc1e47d740acfea3aa94290 to make sure remix sdk was actually built as part of `yarn build:sdk`, but the types look good on my machine :ship:

deltaepsilon:
> needed to add [8b867a5](https://github.com/highlight/highlight/commit/8b867a50330231497bc1e47d740acfea3aa94290) to make sure remix sdk was actually built as part of `yarn build:sdk`, but the types look good on my machine 🚢

Great catch!

PR #6968, opened 2023-10-26
update copy for demo project email gate

Vadman97:

Screenshot 2023-10-25 at 8 05 46 PM

No

### DIFF
--- a/frontend/src/components/DemoModal/DemoModal.tsx
+++ b/frontend/src/components/DemoModal/DemoModal.tsx
@@ -4,7 +4,6 @@ import {
 	Box,
 	Form,
 	IconSolidExclamation,
-	Tag,
 	Text,
 	useFormStore,
 } from '@highlight-run/ui'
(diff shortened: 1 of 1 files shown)


frontend/src/components/DemoModal/DemoModal.tsx
```diff
@@ -24,6 +23,9 @@ export const DemoModal = () => {
 		if (!(await formStore.validate())) {
 			return
 		}
+		if (email.indexOf('@gmail.') !== -1) {
+			return
```
SpennyNDaJets, line 27: Not blocking, but maybe we create a card to give the user an error for feedback in this case?

SpennyNDaJets (approved):

frontend/src/components/DemoModal/DemoModal.tsx
Vadman97: will update with an error tag at the bottom of the card.

PR #6969, opened 2023-10-26
Capture spans for DB operations

ccschmitz:
## Summary

Adds a GORM plugin to the Highlight Go SDK that creates spans for all DB operations performed through GORM.

Screenshot 2023-10-31 at 8 38 43 PM

In order for these new spans to be associated with the correct parent spans we also need to pass the context in to GORM using `WithContext`, which was updated in a lot of places, but not everywhere. We also needed to add some span attributes manually to have them recognized as DB operations in the client (`db.system`). We will follow add a linting rule and to ensure `WithContext` is called in https://github.com/highlight/highlight/issues/7047.

Local click test. Can't click test in Reflame preview because of the backend changes.

Should monitor for any errors and view individual traces that include DB operations to ensure spans are being displayed correctly on flame graphs.

Have been working w/ Julian on the flame graph styles. Have some open questions, like how we should display errors on spans, but nothing we can't handle in follow up PRs.


frontend/src/pages/Traces/TraceFlameGraph.tsx
```diff
@@ -194,7 +194,7 @@ export const TraceFlameGraph: React.FC = () => {
 								span={span}
 								depth={0}
 								height={height}
-								width={width}
+								width={width - outsidePadding * 2}
```
ccschmitz, line 197: Fixes a bug in the `offsetX` logic for spans.

frontend/src/pages/Traces/TraceFlameGraphNode.tsx
```diff
@@ -46,10 +86,21 @@ export const TraceFlameGraphNode = memo<Props>(
...
+			? 'yellow'
+			: isFrontendSpan
+			? 'green'
+			: hasError
+			? 'red'
+			: 'purple'
+		const theme = spanThemes[themeName] ?? spanThemes['purple']
+		const fill = isSelectedSpan ? theme.selectedBackend : theme.background
+		const color = isSelectedSpan ? theme.selectedColor : theme.color
+		const stroke = isSelectedSpan ? theme.selectedBackend : theme.border
```
ccschmitz, line 103: Will need to follow up on how we colorize these spans. Will evolve this in future PRs.

sdk/highlight-go/gorm.go
```diff
@@ -0,0 +1,169 @@
+package highlight
```
ccschmitz, line 1: @Vadman97 is this where you would put this code? Thought about putting it into a separate package, similar to what we did with `hlog`.

sdk/highlight-go/highlight.go
```diff
@@ -243,3 +244,19 @@ func shutdown() {
 	}
 	state = stopped
 }
+
+func SetupGormTracing(ctx context.Context, db *gorm.DB, attrs []attribute.KeyValue) error {
```
ccschmitz, line 248: @Vadman97 is there a convention for naming things like this? Curious what you think the DX should be for setting up the GORM plugin.

Vadman97: what if we move to the following 

```go
	highlight.Start(
		highlight.WithServiceName(serviceName),
		highlight.WithServiceVersion(os.Getenv("REACT_APP_COMMIT_SHA")),
                 highlight.WithGORMTracing(db, attrs...)
	)
```
with this method being
```
func WithGORMTracing(db *gorm.DB, attrs attribute.KeyValue...) Option {
...
}
```

looking at gorms docs, i think they prefer `GORM` to `Gorm`?

sdk/highlight-go/highlight.go
```diff
@@ -13,6 +13,7 @@ import (
 	"github.com/pkg/errors"
 	"go.opentelemetry.io/otel/attribute"
 	semconv "go.opentelemetry.io/otel/semconv/v1.4.0"
+	"gorm.io/gorm"
```
Vadman97, line 16: does this mean gorm is a dependency of our SDK now?

sdk/highlight-go/gorm.go
```diff
@@ -0,0 +1,169 @@
...
+
+// WithGormDBName configures a db.name attribute.
+func WithGormDBName(name string) PluginOption {
+	return func(p *otelPlugin) {
+		p.attrs = append(p.attrs, semconv.DBNameKey.String(name))
+	}
+}
+
+// WithoutGormQueryVariables configures the db.statement attribute to exclude query variables
+func WithoutGormQueryVariables() PluginOption {
```
Vadman97, line 165: are these options something that should be part of the `WithGORMTracing` options?

sdk/highlight-go/gorm.go
Vadman97: having just discussed it, i feel like having everything in the same `highlight` package is an easier experience for customers to set up? i think this is fine and i can take an action item to move our logrus instrumentation into the same.

backend/main.go
```diff
@@ -275,6 +277,14 @@ func main() {
 		log.WithContext(ctx).Fatalf("Error setting up DB: %v", err)
 	}
 
+	attrs := []attribute.KeyValue{
+		attribute.String(highlight.ProjectIDAttribute, highlight.GetProjectID()),
+		semconv.ServiceNameKey.String("gorm"),
```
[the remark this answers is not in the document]
ccschmitz: Oh yeah, I actually meant to remove this after we settled on sniffing out some of the trace attributes to determine how to highlight things in the client. Thanks for calling this out!

sdk/highlight-go/highlight.go
ccschmitz: Ah, yeah, I guess it would be. I see your comment about putting everything in the `highlight` package, so I guess that would mean we need to make GORM, Logrus, etc. deps of this now, right?

Vadman97: https://github.com/highlight/highlight/blob/ad309bf27a6b7c4920e8cd091e6813f3ae098f9c/sdk/highlight-go/go.mod#L1-L20

actually it doesn't matter whether these are in the same package or not because they will be part of the same module and thus share dependencies. i'm just worried what impact this will have if a customer on gorm v2 uses our sdk depending on v1, or vice versa, but maybe that's something we can test? what happens with our backend (on gorm v1) if we make this sdk use gorm v2?

backend/main.go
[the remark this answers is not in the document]
Vadman97: cool!

backend/go.mod
```diff
@@ -28,6 +28,8 @@ replace github.com/miekg/dns => github.com/miekg/dns v1.1.25
 
 replace github.com/gin-gonic/gin => github.com/gin-gonic/gin v1.7.7
 
+replace github.com/highlight/highlight/sdk/highlight-go => ../sdk/highlight-go
```
ccschmitz, line 31: @Vadman97 I think this fixed the issue we were running into. Props to @mayberryzane for finding this!

Vadman97: amazing, and TIL!

PR #6977, opened 2023-10-26
Fix relative stylesheets

SpennyNDaJets:
## Summary
Some CSS stylesheets have relative path references from which they need to be fetched. Previously, this would be fetched from `app.highlight.io`, but we should use the url of the site in the snapshot.

Screenshot 2023-10-19 at 2 13 52 PM

I am not currently able to test on their site (waiting for an invite), but confirmed that Highlight sessions with absolute urls are not broken.

N/A

N/A

### DIFF
--- a/backend/event-parse/parse.go
+++ b/backend/event-parse/parse.go
@@ -101,13 +101,16 @@ const (
 	MaxAssetSize = 3 * 1e9
 )
 
-type fetcher interface {
-	fetchStylesheetData(string) ([]byte, error)
-}
+func (s *Snapshot) fetchStylesheetData(href string) ([]byte, error) {
+	u, err := url.Parse(href)
+	if err != nil {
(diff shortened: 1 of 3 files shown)


SpennyNDaJets:
Curious if y'all have any thoughts on how to test this

backend/event-parse/parse.go
```diff
@@ -368,7 +388,7 @@ func (s *Snapshot) InjectStylesheets() error {
 		if !ok || !strings.Contains(href, "css") {
 			continue
 		}
-		data, err := fetch.fetchStylesheetData(href)
+		data, err := fetch.fetchStylesheetData(href, s)
 		if err != nil {
 			continue
```
Vadman97, line 393: would it be helpful to add some backend logging here if the request fails? now that there are more error cases here

Vadman97 (approved):
awesome! thanks for helping w this!

PR #6978, opened 2023-10-26
fix trace ID encoding

Vadman97:
## Summary

x-highlight-request network id encoding to trace id on spans would not encode
capital letters. Switch this to a encoding method that captures the entire a-z A-Z 0-9 range.

Traces captured locally.
Screenshot 2023-10-26 at 5 38 37 PM

No

No

### DIFF
--- a/sdk/highlight-go/otel.go
+++ b/sdk/highlight-go/otel.go
@@ -2,7 +2,6 @@ package highlight
 
 import (
 	"context"
-	"encoding/base64"
 	"fmt"
 	"net/url"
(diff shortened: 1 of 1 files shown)


ccschmitz (approved):
Incorporated these changes into my branch adding the GORM spans. This seemed to preserve the existing behavior, but I didn't see the GORM spans being associated with the correct trace still. Not sure if you want to look further into that or address as a follow up since this fixes some other issues.

Thanks so much for digging into this!

PR #6984, opened 2023-10-27
Join workspace fixes

ccschmitz:
## Summary

1. Fixes a bug where the join workspace submit button was disabled until you interacted with the form. It would look like you had a workspace selected to join, but the button was disabled until you interacted with the workspace dropdown, which caused confusion for people joining workspaces.
1. Updates the CTA to dismiss the join workspace page when your able to join a workspace based on your email domain.
1. Fixes a bug where the `allowed_auto_join_email_origins` field wasn't being populated even when the checkbox was selected on the invite team page. This was because the server expected a string that could be parsed as JSON - an array of allowed email domains.
1. Updates the loading state of the form submit button on the sign up form so that it is always dismissed after a successful submission. It was previously flickering the disabled state, but clickable again before transitioning to the next page.

Local click test with a new account. Used https://www.guerrillamail.com for generating some new email addresses with matching domains and confirming they were updated as expected.

Should monitor folks signing up and joining a workspace to ensure they aren't running into these issues anymore.

Nope!

### DIFF
--- a/frontend/src/pages/Auth/InviteTeam.tsx
+++ b/frontend/src/pages/Auth/InviteTeam.tsx
@@ -128,15 +128,17 @@ export const InviteTeamForm: React.FC = () => {
 						},
 					}),
 				)
-				if (adminRole === AdminRole.Admin) {
+				if (
+					adminRole === AdminRole.Admin &&
+					formState.values.autoJoinDomain
+				) {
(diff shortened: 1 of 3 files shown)


Vadman97 (approved):
reflame preview looks good! thank you for the fixes 🙏

Issue #6898, opened 2023-10-19
Text is cut off in the pill component.


jay-khatri:
Text is cut off in the pill component.

<img width="1253" alt="image" src="https://github.com/highlight/highlight/assets/20292680/d9f1b510-0a16-460f-ab9e-ec67b3823bad">

https://app.highlight.io/1/errors/pOJIgQ1seJUQIxyV4sqQyeGIwgns?page=1&query=and%7C%7Cerror_state%2Cis%2COPEN%7C%7Cerror-field_timestamp%2Cbetween_date%2C2023-09-19T21%3A52%3A30.788Z_2023-10-19T21%3A52%3A30.788Z

jay-khatri:
see the "f" in frontend.

PR #6985, opened 2023-10-27
Clean up error tags component

ccschmitz:
## Summary

We had [a ticket](https://github.com/highlight/highlight/issues/6898) for fixing an issue where text was being cut off. I dug in to investigate and found it was fixed in https://github.com/highlight/highlight/pull/6936, but realized the only reason we needed the fix was because we had an unnecessary `<Text>` element wrapping the contents of the tag. This PR cleans up the markup since `Tag` already wraps the contents with `<Text>`:

https://github.com/highlight/highlight/blob/c699d1944192267691b2ef5ff5a19bb2facb3f3f/packages/ui/src/components/Tag/Tag.tsx#L72-L83

Local click test.

N/A - small client-side tweak.

Nope!

### DIFF
--- a/frontend/src/pages/ErrorsV2/ErrorTag/ErrorTag.tsx
+++ b/frontend/src/pages/ErrorsV2/ErrorTag/ErrorTag.tsx
@@ -33,6 +33,7 @@ const ErrorTag = React.memo(
 					kind="secondary"
 					size="medium"
 					shape="basic"
+					lines="1"
 				>
 					{errorGroup.serviceName && errorGroup.serviceName != ''
 						? errorGroup.serviceName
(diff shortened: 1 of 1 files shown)


frontend/src/pages/ErrorsV2/ErrorTag/ErrorTag.tsx
```diff
@@ -33,31 +33,23 @@ const ErrorTag = React.memo(
 					kind="secondary"
 					size="medium"
 					shape="basic"
-					lines="1"
```
SpennyNDaJets, line 36: I think the `lines="1"` was added here incase the tag text was really long. Do we know if a long tag text will cause a line break or not?

SpennyNDaJets (approved):
Looks good to me!

frontend/src/pages/ErrorsV2/ErrorTag/ErrorTag.tsx
ccschmitz: Yeah, I suppose that could technically be something longer. Just added this back and truncated the error tag `Tag` as well. The text clipping is still fixes here - the problem was with the `<Text>` element we were adding because `lines="1"` was being passed to the `<Text>` inside of `Tag` but not applied on the one we added in this component.

Issue #6156, opened 2023-07-31
Billing logic for Traces


jay-khatri:
Need to add backend and UI for billing on traces.
```[tasklist]
### Tasks
- [x] Add billing log for traces, similar to logs, errors, and sessions
- [ ] Add pricing to website
```

PR #6989, opened 2023-10-27
billing for traces

mayberryzane:
## Summary
- adds billing for traces
- price and default included amounts match billing for logs
- doesn't allow a max spend for traces as this functionality is going away for the other products
- addresses the app changes for #6156, need to make changes to the website too

- click tested locally, ran `make report-stripe-usage` locally

- added traces product to prod stripe: https://dashboard.stripe.com/products/prod_Oti9dSGXgkhS6T

- no

### DIFF
--- a/backend/model/model.go
+++ b/backend/model/model.go
@@ -291,6 +291,7 @@ type Workspace struct {
 	MonthlyMembersLimit         *int
 	MonthlyErrorsLimit          *int
 	MonthlyLogsLimit            *int
+	MonthlyTracesLimit          *int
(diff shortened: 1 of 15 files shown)


deltaepsilon (approved):

backend/model/model.go
```diff
@@ -291,6 +291,7 @@ type Workspace struct {
 	MonthlyMembersLimit         *int
 	MonthlyErrorsLimit          *int
 	MonthlyLogsLimit            *int
+	MonthlyTracesLimit          *int
```
Vadman97, line 294: could we update it with a customer trace price override to support #7006?

PR #6991, opened 2023-10-27
ensure only one error is sent for the trace context manager

Vadman97:
## Summary

Turns out the opentelemetry `self.tracer.start_as_current_span` contextmanager
already reports an exception if the inner contents raise an error. Since we want
to `raise` there to propagate the exception to the outer web framework, we
set `record_exception=False` to avoid having the opentelemetry context manager
reporting the exception event a second time (in addition to our manual time that reports custom attributes).

Unit test
Local testing

one error recorded with fix, two duplicate errors without the fix
Screenshot 2023-10-27 at 4 21 44 PM

New python patch version.

No

### DIFF
--- a/e2e/express/src/index.mjs
+++ b/e2e/express/src/index.mjs
@@ -1,16 +1,25 @@
-/* const express = require('express')
-const highlight = require('@highlight-run/node')
-const H = highlight.H */
-
 import express from 'express'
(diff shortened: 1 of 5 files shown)


SpennyNDaJets (approved):
Makes sense to me!

PR #7006, opened 2023-10-31
support custom and graduated pricing

### LINKED ISSUE #6381: support custom usage-based overage prices

Vadman97:
## Summary

Adds support for custom overage price IDs that will be used when attaching
overage invoice items. Can work alongside the existing pricing model or can replace
existing default prices.

* new workspace model columns allow setting a custom overage price id
* the billing logic now supports metered prices (recurring prices) that are usage-billed.

Closes #6381

Screenshot from 2023-10-31 13-35-42

tested on a stripe subscription that previously had a default non-graduated invoice item for sessions to make sure the new path will remove the existing invoice item.

No

No

### DIFF
--- a/backend/pricing/pricing.go
+++ b/backend/pricing/pricing.go
@@ -443,7 +443,7 @@ func FillProducts(stripeClient *client.API, subscriptions []*stripe.Subscription
 }
 
 // Returns the Stripe Prices for the associated tier and interval
-func GetStripePrices(stripeClient *client.API, productTier backend.PlanType, interval model.PricingSubscriptionInterval, unlimitedMembers bool, retentionPeriod *backend.RetentionPeriod) (map[model.PricingProductType]*stripe.Price, error) {
+func GetStripePrices(stripeClient *client.API, workspace *model.Workspace, productTier backend.PlanType, interval model.PricingSubscriptionInterval, unlimitedMembers bool, retentionPeriod *backend.RetentionPeriod) (map[model.PricingProductType]*stripe.Price, error) {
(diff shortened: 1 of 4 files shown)


backend/pricing/pricing.go
```diff
@@ -651,9 +668,21 @@ func (w *Worker) reportUsage(ctx context.Context, workspaceID int, productType *
 	}
 
 	invoiceLines := map[model.PricingProductType]*stripe.InvoiceLine{}
-	for _, line := range invoice.Lines.Data {
+	for _, line := range lo.FindUniquesBy(invoice.Lines.Data, func(item *stripe.InvoiceLine) string {
```
mayberryzane, line 671: what is the `FindUniquesBy` for here?

backend/pricing/pricing.go
```diff
@@ -651,9 +668,21 @@ func (w *Worker) reportUsage(ctx context.Context, workspaceID int, productType *
...
+		return item.InvoiceItem
+	}) {
 		productType, _, _, _, _ := GetProductMetadata(line.Price)
 		if productType != nil {
+			if invoiceLines[*productType] != nil {
```
mayberryzane, line 676: `invoiceLines[*productType] != nil` wouldn't be true until it's set on 686 right?

mayberryzane (approved):

backend/pricing/pricing.go
Vadman97: correct, this deletes duplicate line items for old prices for a given product

Issue #7010, opened 2023-10-31
SSR error catch not working in nextjs (old pages not appDir)


lveillard:
**Describe the bug**
Hello! I'm following the instructions to catch SSR errors but for some reason they don't seem to work, at least in development

**To Reproduce**
- Custom _error page:
```
import NextError from 'next/error';
import type { HighlightErrorProps } from '@highlight-run/next/client';
import { H, getHighlightErrorInitialProps } from '@highlight-run/next/client';

export default function CustomError({ errorMessage, statusCode }: HighlightErrorProps) {
	/// needs to be init again because this is SSR
	H.init(process.env.NEXT_PUBLIC_HIGHLIGHT_PROJECT_ID);
	H.consumeError(new Error(errorMessage));

	return <NextError statusCode={statusCode} />; // Render default Next error page
}

CustomError.getInitialProps = getHighlightErrorInitialProps;
```
Then throw an error in getServerSideProps
```
export const getInitialProps = async (context) => {
	throw new Error('Forced SSR Error');
};

export default MyPage;
```
**versions**
- "@highlight-run/next": "^7.0.0",
- "next": "13.5.4",

**Expected**
- The error should be catched (the other errors do)

**Documentation**
I find there are 2 key things missing in the nextjs documentaiton
- how to test this particular SSR error catch
- how to build sourcemaps specifically for nextjs (or even having a tool like sentry to automatically generate and send them)

Vadman97:
Hey @lveillard , can you confirm that you experience this issue with "@highlight-run/next": "7.0.2"? "7.0.1" introduced some important bugfixes to the server-side capture

lveillard:
> Hey @lveillard , can you confirm that you experience this issue with "@highlight-run/next": "7.0.2"? "7.0.1" introduced some important bugfixes to the server-side capture

Not getting any errors with 7.0.2 neither but maybe getServerSide props is not supposed to be catched like this?

Also, not sure if related but with 7.0.2 as well, creating my page-router-highlight-config.ts i get another error

Unable to resolve path to module '@highlight-run/next/server'.eslint[import/no-unresolved](https://github.com/import-js/eslint-plugin-import/blob/v2.29.0/docs/rules/no-unresolved.md)
[...]

Issue #6839, opened 2023-10-10
RangeError: Maximum call stack size exceeded when logging Twilio Verification object


adymo:
**Describe the bug**

If Highlight is initialized, then the following code
```
  const verification = await twilio(sid, auth).verify.v2.services(verifySid).
    verifications.create({to: "<phone>",channel: 'sms'});
  console.log(verification);
```
fails to capture `console.log(verification)` output with 
```
~/projects/twilio_highlight_test/node_modules/@highlight-run/node/dist/index.js:120195
  function replacer(input) {
                   ^

RangeError: Maximum call stack size exceeded
    at replacer (~/projects/twilio_highlight_test/node_modules/@highlight-run/node/dist/index.js:120195:20)
    at replacer (~/projects/twilio_highlight_test/node_modules/@highlight-run/node/dist/index.js:120199:11)
    at replacer (~/projects/twilio_highlight_test/node_modules/@highlight-run/node/dist/index.js:120199:11)
...
```

**To Reproduce**
Here's the sample project you can use to reproduce the bug. You do need Twilio and Twilio Verify credentials to make it work.
[twilio_highlight_test.tar.gz](https://github.com/highlight/highlight/files/12861004/twilio_highlight_test.tar.gz)

**Expected behavior**
Twilio Verification object isn't that large to exhaust the stack size, so I wouldn't expect this to happen.

Here's a sample console.log output:
```
{
  sid: 'AAABBBAAABBBAAABBBAAABBB0123456789',
  serviceSid: 'AAABBBAAABBBAAABBBAAABBB0123456789',
  accountSid: 'AAABBBAAABBBAAABBBAAABBB0123456789',
  to: '+16500000000',
  channel: 'sms',
  status: 'pending',
  valid: false,
  lookup: { carrier: null },
  amount: null,
  payee: null,
  sendCodeAttempts: [
    {
      attempt_sid: 'AAABBBAAABBBAAABBBAAABBB0123456789',
      channel: 'sms',
      time: '2023-10-10T18:57:40.000Z'
    },
    {
      attempt_sid: 'AAABBBAAABBBAAABBBAAABBB0123456789',
      channel: 'sms',
      time: '2023-10-10T18:58:05.000Z'
    },
    {
      attempt_sid: 'AAABBBAAABBBAAABBBAAABBB0123456789',
      channel: 'sms',
      time: '2023-10-10T18:59:59.000Z'
    }
  ],
[...]

PR #7039, opened 2023-10-31
fix node sdk timeouts

### LINKED ISSUE #6987: @highlight-run/node times out exporting data
**Describe the bug**
```
error: "highlight-node failed to flush:  Error [OTLPExporterError]: Request Timeout" 
```

Screenshot 2023-10-27 at 2 04 23 PM

Vadman97:
## Summary

Our node SDK was throwing red-herring errors while shipping logs / traces because of
the interval flush logic. This logic was not needed as OTEL has it's own loop for flushing data.

* fixes recursion error during stringification of certain objects #6839 
* ensures console attributes are correctly reported correctly as structured attributes
* ensures we fully rely on bundling opentelemetry to help with #7010

Closes #6987

tested locally by ingesting a network latency `tc qdisc add dev lo root netem delay 10000ms`

before
Screenshot from 2023-10-31 14-25-11

after
Screenshot from 2023-10-31 14-29-55

Logs coming through correctly locally

Screenshot from 2023-10-31 16-03-25

tested in next js edge runtime as well

No

Changeset

### DIFF
--- a/sdk/highlight-node/src/hooks.ts
+++ b/sdk/highlight-node/src/hooks.ts
@@ -36,6 +36,7 @@ interface ConsolePayload {
 	level: keyof typeof ConsoleLevels
 	message: string
 	stack: object
+	attributes?: { [key: string]: any }
 }
 
 type ConsoleFn = (...data: any) => void
@@ -103,6 +104,10 @@ export function hookConsole(
 							typeof o === 'object' ? safeStringify(o) : o,
 						)
 						.join(' '),
+					attributes: data
+						.slice(1)
+						.filter((d) => typeof d === 'object')
+						.reduce((a, b) => ({ ...a, ...b }), {}),
 					stack: o.stack,
 				})
 			}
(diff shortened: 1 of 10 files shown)


sdk/highlight-node/src/hooks.ts
```diff
@@ -33,18 +36,22 @@ interface ConsolePayload {
...
 }
 
 type ConsoleFn = (...data: any) => void
 
 let consoleHooked = false
 
 export function safeStringify(obj: any): string {
-	function replacer(input: any): any {
+	function replacer(input: any, depth?: number): any {
+		if ((depth ?? 0) > MAX_RECURSION) {
```
SpennyNDaJets, line 48: Curious the impact if something where to be over the `MAX_RECURSION` value? The value seems low to the minimum browser results.

SpennyNDaJets (approved):
Code looks good!

sdk/highlight-node/src/hooks.ts
Vadman97: I think it's unlikely an object someone wants to serialize intentionally will be deeper than 128 children, while keeping the value low means it's faster to detect a cycle in the object.

PR #7055, opened 2023-11-02
Separate right panel open state from comment mode logic

ccschmitz:
## Summary

Allows the comment panel to be displayed regardless of the value of `showRightPanel` and disables the toggle next to the player mode switch when in comment mode. Also, swaps the order of inspect and comment mode in the player mode toggle.

https://www.loom.com/share/e6880ef08fa74343a5790cc7635e88c1

Local and PR preview click test.

N/A - relatively small client-side change.

I worked w/ Julian on determining the exact UX we wanted here. He has plans for a theater mode that will improve the UX here even further 🥳


SpennyNDaJets (approved):
Code looks good, and click test works as expected!

PR #7057, opened 2023-11-02
Add headers to the logs table

SpennyNDaJets:
## Summary
Update the logs table to use our common table component, and provide the table with headers to match the style of the traces table.

Note: this still uses the virtual table - we may want to consider building a wrapper around our table component to handle these types of table when converting the traces table.

1) Load the logs table
- [ ] Confirm more logs can be fetched via the top row notification
- [ ] Confirm more loads are loaded when scrolling to the bottom
- [ ] Confirm logs row can be opened and closed 
- [ ] Confirm data in logs can be be copied or used to filter appropriately
2) Copy a link to a log, and navigate to the url
- [ ] Confirm the logs is scrolled down to
- [ ] Confirm the log is expanded by default

https://www.loom.com/share/26413f3719734be4be98cea85ab478b4?sid=60d88918-33fe-4878-a2e1-8c9aa8b0ea5e

N/A

Yes


frontend/src/pages/LogsPage/LogsTable/LogsTable.tsx
```diff
@@ -218,101 +275,89 @@ const LogsTableInner = ({
...
+									clearMoreLogs()
+									handleAdditionalLogsDateChange()
+								}}
+							/>
+						</Box>
+					</Table.Row>
+				)}
+			</Table.Head>
+			<Table.Body
+				forwardRef={bodyRef}
```
ccschmitz, line 301: Nit, but what do you think about using `forwardRef` inside `Table.Body` to pass the ref rather than a custom prop?

packages/ui/src/components/Table/Header/Header.tsx
```diff
@@ -1,11 +1,11 @@
 import React from 'react'
-import { Box } from '../../Box/Box'
+import { Box, BoxProps } from '../../Box/Box'
```
ccschmitz, line 2: Think the GitHub Actions comment is right here - could we remove this?

ccschmitz (approved):
Code + click test look good! I think my main question is around the UX of displaying the expanded log details within the log message table cell. Wondering if we can come up with a better treatment for that.

Has @julian-highlight had a chance to click test and provide any feedback?

frontend/src/pages/LogsPage/LogsTable/LogsTable.tsx
```diff
@@ -128,37 +141,80 @@ const LogsTableInner = ({
...
+											)
+										},
+									)}
+								</Box>
+							)}
+						<LogDetails
+							matchedAttributes={matchedAttributes}
+							row={row}
+							queryTerms={queryTerms}
+						/>
```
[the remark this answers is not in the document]
SpennyNDaJets: Talked with Julian, I'm going to merge this in as is, and we can come back and address depending on our final solution - i.e. open below vs in left panel, etc

PR #7063, opened 2023-11-02
Add Reflame interaction test

lewisl9029:
## Summary

This PR adds an example Reflame interaction test for the ComboboxSelect component.

We can turn a screenshot test (i.e. a React component/element exported with a name that ends in `_test`) into a interaction test by assigning a `.run` function to it. 

This function will receive the `screen` and `user` named args from [testing library](https://testing-library.com/docs/) that we can use to select elements and interact with them. It will also receive a `captureScreenshot` function that will take a screenshot to be compared to previous reference screenshots captured in the default branch.

A simple interaction test for a Switch component could look something like this: 
```js
export const Switch_test = <Switch />
Switch_test.run = async ({ user, screen, captureScreenshot }) => {
  // calling captureScreenshot here for demonstration, but in practice I'd recommend just exporting a separate SwitchDefault_test element
  await captureScreenshot({ name: 'before click'})
  const switch = await screen.findByRole('switch')
  await user.click(switch)
  // we can omit this because Reflame will automatically take a screenshot at the end of the test
  // await captureScreenshot({ name: 'after click'})
}
```

For a more involved example, see the ComboboxSelect test in this PR.

Reflame creates live deployments for every test, and since testing library is designed to run as client-side JS in the browser, we can inspect our tests running in the browser and debug them with the browser dev tools we all know and love, just by clicking a link! See for yourself by clicking any of the inspect links in https://github.com/highlight/highlight/pull/7063/checks?check_run_id=18314391807

Give it a shot on any other components you'd like to test and let me know what you think!

P.S. now that we have full Reflame screenshot test coverage for all the stories, we can probably remove a bunch of vitest tests in the ui package that just assert existence of some text, since the screenshot tests will check for that and much more. Happy to make that change in another PR if you'd like.

### DIFF
--- a/packages/ui/src/components/ComboboxSelect/ComboboxSelect.tsx
+++ b/packages/ui/src/components/ComboboxSelect/ComboboxSelect.tsx
@@ -207,6 +207,8 @@ export const ComboboxSelect_test = () => {
 	)
 }
 
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
 ComboboxSelect_test.run = async ({ user, screen, captureScreenshot }) => {
 	const combobox = await screen.findByRole('combobox')
 	await captureScreenshot(combobox, { name: 'initial' })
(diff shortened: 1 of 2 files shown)


Vadman97 (approved):
:tada:

Vadman97 (approved):
:tada:

PR #7064, opened 2023-11-02
create reporting schema with view in postgres

joshdirkx:
## Create Reporting Schema + Error & Session Usage View

This query will allow us to track week over week usage of our error and session products more easily

@Vadman97 & @mayberryzane were nice enough to make sure my SQL was correct

No

No


Vadman97 (approved):
🎉

deltaepsilon (approved):

PR #7070, opened 2023-11-02
Node SDK Resource merge

deltaepsilon:
It looks like I broke a resource merge a few months back. It didn't appear to break anything, because we were setting session data on every span manually.

BUT, with this resource fixed, we can get automatic tracing data straight from Next.js.

HT to @Vadman97 for finding the bug!

### DIFF
--- a/e2e/nextjs/next.config.mjs
+++ b/e2e/nextjs/next.config.mjs
@@ -3,7 +3,6 @@ import { withHighlightConfig } from '@highlight-run/next/config'
 
 const nextConfig = {
 	experimental: {
-		appDir: true,
 		instrumentationHook: false,
 	},
 	productionBrowserSourceMaps: true,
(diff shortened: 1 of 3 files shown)


e2e/nextjs/next.config.mjs
```diff
@@ -3,7 +3,6 @@ import { withHighlightConfig } from '@highlight-run/next/config'
 
 const nextConfig = {
 	experimental: {
-		appDir: true,
```
Vadman97, line 6: don't we use app dir in the e2e next example? should we keep this true?

Vadman97 (approved):

e2e/nextjs/next.config.mjs
deltaepsilon: This is no longer necessary in Next 14.

PR #7078, opened 2023-11-03
Playing around with parent spans

deltaepsilon:
## Summary

We need to let `@highlight-run/node` users create their own spans... but with Highlight's custom attributes set automatically.


Vadman97 (approved):
nice!

PR #7082, opened 2023-11-04
refactor node sdk highlight context types

Vadman97:
## Summary

Removes our usage of `global.__HIGHLIGHT__` which does not work in edge runtime.
Consolidates duplicate types that expose the highlight session / request context.

Vercel deploy of highlight.io edge runtime methods.

Changeset

No

### DIFF
--- a/.changeset/quick-planes-hug.md
+++ b/.changeset/quick-planes-hug.md
@@ -1,5 +1,0 @@
----
-'@highlight-run/next': patch
----
-
-fix incorrect setting of highlight context in edge runtime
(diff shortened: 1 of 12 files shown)


deltaepsilon (approved):
This is effectively the same thing I did... with more cleanup!

I found out that Edge polyfills AsyncLocalStorage, so we could use it in Edge if we think it makes a difference.

Vadman97:
> This is effectively the same thing I did... with more cleanup!
> 
> I found out that Edge polyfills AsyncLocalStorage, so we could use it in Edge if we think it makes a difference.

good call @deltaepsilon , converted to asynclocalstorage in edge now

PR #7100, opened 2023-11-06
Repair Next.js Edge SDK

deltaepsilon:
## Summary

Edge doesn't provide `global`, so we're using `AsyncLocalStorage` instead.  

Testing also revealed that the header parsing logic was inverted, causing headers to get lost on all Edge requests. 

Deploys to Vercel

No.

Batch bump to Next SDK

### DIFF
--- a/docs-content/getting-started/fullstack-frameworks/next-js/2_page-router.md
+++ b/docs-content/getting-started/fullstack-frameworks/next-js/2_page-router.md
@@ -139,11 +139,52 @@ function ThrowerOfErrors({
 }
 ```
 
-## Catch server-side render (SSR) errors (optional)
+## Enable server-side tracing
+
+We use `experimental.instrumentationHook` to capture [Next.js's automatic instrumentation](https://nextjs.org/docs/app/building-your-application/optimizing/open-telemetry). This method captures detailed API route tracing as well as server-side errors.
+
+1. Enable `experimental.instrumentationHook` in `next.config.js`.
+```javascript
+// next.config.mjs
+import { withHighlightConfig } from '@highlight-run/next/config'
+
+const nextConfig = {
+	experimental: {
+		instrumentationHook: true,
+	},
+	// ...additional config
+}
+
+export default withHighlightConfig(nextConfig)
+```
+
+2. Call `registerHighlight` in `instrumentation.ts`
+```jsx
+// instrumentation.ts
+import { CONSTANTS } from './constants'
+
+export async function register() {
+	const { registerHighlight } = await import('@highlight-run/next/server')
+
+	registerHighlight({
+		projectID: CONSTANTS.NEXT_PUBLIC_HIGHLIGHT_PROJECT_ID,
+		serviceName: 'my-nextjs-backend',
+	})
+}
+```
+
+3. App Router instrumentation requires `app/instrumentation.ts` to be defined, so re-export your handler from `./instrumentation.ts`
+```typescript
+// src/instrumentation.ts:
+export { register } from '../instrumentation'
+
+```
+
+## Catch server-side render (SSR) errors
 
 Page Router uses [pages/_error.tsx](https://nextjs.org/docs/pages/building-your-application/routing/custom-error#more-advanced-error-page-customizing) to send server-side render errors to the client. We can catch and consume those errors with a custom error page.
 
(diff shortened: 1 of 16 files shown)


docs-content/getting-started/fullstack-frameworks/next-js/2_page-router.md
```diff
@@ -139,11 +139,52 @@ function ThrowerOfErrors({
 }
 ```
 
-## Catch server-side render (SSR) errors (optional)
+## Enable server-side tracing
```
deltaepsilon, line 142: It turns out that `instrumentation.ts` captures traces, but only in certain cases can it capture SSR errors. So `_error.tsx` and `error.tsx` are still the most reliable ways to capture SSR errors.

In short, `instrumentation.ts` doesn't currently catch Edge SSR. Our `registerHighlight` handler is currently empty in Edge environments, but OpenTelemetry doesn't really work in Edge anyway, so I don't suspect that Next.js instruments Edge routes anyway.

e2e/nextjs/instrumentation.ts
```diff
@@ -2,13 +2,11 @@
 import { CONSTANTS } from '@/constants'
 
 export async function register() {
-	if (process.env.NEXT_RUNTIME === 'nodejs') {
```
deltaepsilon, line 5: This check is no longer necessary due to better SDK structure.

e2e/nextjs/pages/redirect.tsx
```diff
@@ -5,7 +5,7 @@ type Props = {
 }
 export default function RedirectPage({ shouldRedirect }: Props) {
 	if (shouldRedirect) {
-		return redirect(`/isr`)
+		return redirect(`/ssr`)
```
deltaepsilon, line 8: `/isr` was just confusing. Switching to `/ssr` across the board.

sdk/highlight-next/src/util/highlight-edge.ts
```diff
@@ -21,6 +22,7 @@ let globalRequestMetadata: RequestMetadata = {
 	requestId: '',
 }
 let executionContext: ExtendedExecutionContext
+const asyncLocalStorage = new AsyncLocalStorage<HighlightContext>()
```
deltaepsilon, line 25: Copy/paste from the `@highlight-run/node` implementation. It worked first try.

sdk/highlight-next/src/util/types.ts
```diff
@@ -71,4 +71,4 @@ export interface HighlightInterface {
 	flush: () => Promise<void>
 }
 
-export const HIGHLIGHT_REQUEST_HEADER = 'X-Highlight-Request'
+export const HIGHLIGHT_REQUEST_HEADER = 'x-highlight-request'
```
deltaepsilon, line 74: I'm not sure if capitalization hurt here, but it definitely didn't help!

sdk/highlight-next/src/util/with-highlight-edge.ts
```diff
@@ -19,7 +19,7 @@ export function Highlight(env: HighlightEnv) {
 		) {
 			H.initEdge(request, env, event)
 			const headers: IncomingHttpHeaders = {}
-			request.headers.forEach((k, v) => (headers[k] = v))
+			request.headers.forEach((k, v) => (headers[v] = k))
```
deltaepsilon: This was baffling. IT took me a while to realize that the keys and values were flipped!

deltaepsilon: See https://developer.mozilla.org/en-US/docs/Web/API/Headers/forEach

sdk/highlight-next/src/util/with-highlight-edge.ts
```diff
@@ -19,7 +19,7 @@ export function Highlight(env: HighlightEnv) {
 		) {
 			H.initEdge(request, env, event)
 			const headers: IncomingHttpHeaders = {}
-			request.headers.forEach((k, v) => (headers[k] = v))
+			request.headers.forEach((v, k) => (headers[k] = v))
```
deltaepsilon, line 22: This was baffling at first, but MDN explains the flipped args:

https://developer.mozilla.org/en-US/docs/Web/API/Headers/forEach

Vadman97 (approved):

PR #7101, opened 2023-11-07
Fix new user alert

SpennyNDaJets:
## Summary
When a session session flips from excluded to viewable, this typically happens before a user has been identified, resulting in the new user alert never being sent. Second, at times we call identify on a session multiple times (at least experienced this locally), and the identify session is called on subsequent times, we incorrectly find the current session and believe it to be a previous session.

This PR fixes these issues to start sending the new user alert consistently.

1. Create a two new alerts for the project and report to Slack, discord, and/or email
- New session alert
- New user alert
2) Open up a incognito browser and create a new account
- [ ] New session alert is called once
- [ ] New user alert is called once

Screenshot 2023-11-07 at 10 39 21 AM

N/A

N/A


backend/public-graph/graph/resolver.go
```diff
@@ -2900,7 +2900,7 @@ func (r *Resolver) ProcessPayload(ctx context.Context, sessionSecureID string, e
 	}
 
 	// if the session changed from excluded to not excluded, it is viewable, so send any relevant alerts
-	if sessionObj.Excluded && !excluded {
+	if !excluded && (sessionObj.Excluded || (sessionObj.FirstTime != nil && *sessionObj.FirstTime)) {
```
SpennyNDaJets, line 2903: @mayberryzane noticed these changes were added to avoid spending too much CPU in `SendSessionInitAlert`. I think this would fire every payload where the session is a user's first session, but still excludes other following sesssions

Vadman97: looks good, you're just including first time sessions to this branch

Vadman97 (approved):

PR #7105, opened 2023-11-08
spinning startup banner

Vadman97:


julian-highlight:
yo! Can we make it have less padding on top and bottom? Generally I'd prefer a looping animation (just like on billboards) here.

Reference: 
image

@Vadman97

julian-highlight:
@Vadman97 do you think adding spaces or '+++' in between looping text would help readability here? Maybe it's also a bit to fast?

PR #7106, opened 2023-11-08
Add missing meta and link tags to reflame deploy

lewisl9029:
## Summary

This PR adds all the link/meta tags currently missing from the Reflame deployment, except the web manifest link which needs some special handling to work as expected.

I inspected the HTML response in the preview and verified all the expected tags are present. 

Could use a render preview to verify my changes didn't break anything there.


.reflame.config.jsonc
```diff
@@ -58,6 +59,94 @@
...
+      "href": "https://fonts.gstatic.com",
+      "crossorigin": true
+    }
+    // FIXME: this doesn't actually work yet
+    // Need to rewrite icon urls within the manifest
+    // Probably need a first class webmanifest config api instead
+    // {
+    //   "rel": "manifest",
+    //   "href": "/manifest.json"
+    // }
```
lewisl9029, line 148: Will follow up with a separate PR for this. Curious if this would be a blocker for going production?

frontend/index.html
```diff
@@ -7,22 +7,36 @@
 		<link
 			rel="apple-touch-icon"
 			sizes="180x180"
-			href="/apple-touch-icon.png"
+			href="./src/apple-touch-icon.png"
```
lewisl9029, line 10: Moving images from public to src so reflame can access them.

frontend/index.html
```diff
@@ -70,35 +84,9 @@
 		/>
 		<meta property="twitter:image" content="/hero.png" />
 
-		<link rel="apple-touch-icon" href="/logo192.png" />
-		<!--
-      manifest.json provides metadata used when your web app is installed on a
-      user's mobile device or desktop. See https://developers.google.com/web/fundamentals/web-app-manifest/
-    -->
-		<link rel="manifest" href="/manifest.json" />
-		<link rel="preconnect" href="https://fonts.googleapis.com" />
```
lewisl9029, line 79: Ended up removing this preconnect since there's a stylesheet link to the same origin 2 lines below, which would make the connection anyways. As far as I can tell, this isn't saving us any time.

The preconnect below to fonts.gstatic.com on the other hand does help because without it, we'd only start connecting after loading https://fonts.googleapis.com/css2?family=Roboto+Mono&display=swap in a waterfall.

frontend/index.html
```diff
@@ -7,22 +7,36 @@
...
+      manifest.json provides metadata used when your web app is installed on a
+      user's mobile device or desktop. See https://developers.google.com/web/fundamentals/web-app-manifest/
+    -->
+		<link rel="manifest" href="/manifest.json" />
+		<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
+		<link
+			href="https://fonts.googleapis.com/css2?family=Roboto+Mono&display=swap"
+			rel="stylesheet"
 		/>
-		<link rel="manifest" href="/site.webmanifest" />
```
lewisl9029, line 24: Removed this webmanifest because it seems to be mostly empty, and there's another manifest at manifest.json that's referred to later on and actually populated.

frontend/index.html
```diff
@@ -70,35 +84,9 @@
...
-		<link rel="manifest" href="/manifest.json" />
-		<link rel="preconnect" href="https://fonts.googleapis.com" />
-		<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
-		<link
-			href="https://fonts.googleapis.com/css2?family=Roboto+Mono&display=swap"
-			rel="stylesheet"
-		/>
 		<title>highlight.io</title>
 		<!-- Google Tag Manager -->
-		<script>
```
lewisl9029, line 87: Moved this over to gtm.js similar to other third party scripts.

frontend/index.html
```diff
@@ -7,22 +7,36 @@
 		<link
 			rel="apple-touch-icon"
 			sizes="180x180"
-			href="/apple-touch-icon.png"
+			href="./src/apple-touch-icon.png"
 		/>
+		<link rel="apple-touch-icon" href="./src/logo192.png" />
```
lewisl9029, line 12: Reordered these and found out we seem to have 2 apple-touch-icons at very similar sizes. Maybe we could remove one?

.reflame.config.jsonc
lewisl9029: Nvm, implemented this earlier today and updated this PR directly: https://github.com/highlight/highlight/pull/7106/commits/8ce4b81f8487baafa9a731d38280ceb58eb8aba7

Preview is now installable. :)
Screenshot 2023-11-08 at 7 39 37 PM

frontend/index.html
Vadman97: sweet, thanks for catching this

Vadman97 (approved):
great, thanks for your help here @lewisl9029 !

PR #7109, opened 2023-11-08
Cleaning up src/instrumentation.ts docs

deltaepsilon:
## Summary

This PR was originally just a docs update, but it spiraled a bit when I discovered that the `@highlight-run/next/ssr` package was broken. I also found a circular dependency in `@highlight-run/node`, which was an easy fix.

So this PR now includes a docs update and two patches. 

# Original PR

A more careful reading of the current Next.js docs shows that we only need a single `instrumentation.ts` file, not two as I had previously imagined.

image


SpennyNDaJets (approved):
LGTM

PR #7112, opened 2023-11-08
Support NOT filters for logs and traces

ccschmitz:
## Summary

Adds support for `NOT` clauses on logs and traces by prefixing attributes with `-`. Also updates the docs and adds a link to view the docs for the search specification.

https://www.loom.com/share/e8482f8dc3f54d27bf0caaac330e879c

Note that this maintains the ability to assign multiple filters on the same attribute as OR queries, but any NOT filters will be applied using an AND.

Local click test.

Should keep an eye on performance of queries, especially on traces, otherwise just need to ensure filters continue to work as expected.

N/A

### DIFF
--- a/backend/clickhouse/logs_test.go
+++ b/backend/clickhouse/logs_test.go
@@ -1336,6 +1336,60 @@ func Test_LogMatchesQuery_ClickHouse(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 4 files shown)


SpennyNDaJets (approved):

PR #7113, opened 2023-11-08
grafana datasource v1

### LINKED ISSUE #7011: add grafana datasource plugin v1
- customer POC requirements
    - graph of top api latencies over time
    - graph of number of request errors over time, grouped by endpoint
        - distribution of HTTP status codes over time

Plugin spec

- drop down of the product you want to query, starting w `traces`
- text input for a query. can use the same syntax as on [app.highlight.io](http://app.highlight.io/)?
- picker for fields to query
- picker for filters
- picker for group by

mayberryzane:
## Summary
- adds a grafana data source for Highlight which uses the `traces_metrics` gql resolver
- allows the user to select the table, metric, column, group by, and query
- closes #7011
- https://www.loom.com/share/4c14dc5c38514312a26048c0027eeff5

- click tested locally

- no

- no


sdk/highlight-grafana-datasource/src/datasource.ts
```diff
@@ -0,0 +1,192 @@
...
+                },
+              },
+              column: target.column,
+            },
+            query: GET_TRACES_METRICS,
+          }),
+          {}
+        );
+        if (response.errors?.length) {
+          throw response.errors.map((e) => JSON.stringify(e)).join(', ');
```
Vadman97, line 117: should we pull this out into some function that is driven by the query type? if we have the dropdown for something other than traces, how would we choose what query we make here?

sdk/highlight-grafana-datasource/src/components/QueryEditor.tsx
```diff
@@ -0,0 +1,81 @@
...
+    { value: 'ServiceVersion', label: 'service_version' },
+    { value: 'http.status_code', label: 'http.status_code' },
+  ];
+
+  return (
+    <div className="gf-form">
+      <Field label="Table">
+        <Select value={table} onChange={onTableChange} options={tableOptions} />
+      </Field>
+      <Field label="Metric" error="why isn't this displayed???">
```
Vadman97, line 65: `error` value from testing?

sdk/highlight-grafana-datasource/src/README.md
```diff
@@ -0,0 +1,49 @@
+<!-- This README file is going to be the one displayed on the Grafana.com website for your plugin. Uncomment and replace the content here before publishing.
```
Vadman97, line 1: could we have a description of highlight / the plugin for publishing in the future?

package.json
Vadman97, line 1: could the `test:all` script run the grafana datasource tests?

Vadman97 (approved):

sdk/highlight-grafana-datasource/src/datasource.ts
mayberryzane: yeah I think having a mapping from `target.table` to the function returning a data frame is probably the way to go, can do this when we add another table to the dropdown

sdk/highlight-grafana-datasource/src/README.md
mayberryzane: 👍 will get a PR out tomorrow with this + grafana datasource tests

PR #7114, opened 2023-11-08
Add CI check for WithContext on GORM calls

ccschmitz:
## Summary

Adds a new CI rule checking that we call `WithContext` everywhere we perform certain GORM operations and fixes all the violations. Currently ignoring many areas of the codebase and many violations were fixed by using `context.TODO()`. We should look at filling in the missing contexts and expanding coverage of the linting rule over time, but the key areas of our background codebase are covered with this.

Ran in CI + click tested locally.

Should keep an eye on any new errors, but should be a fairly safe change if builds are clean.

N/A

### DIFF
--- a/backend/pricing/pricing.go
+++ b/backend/pricing/pricing.go
@@ -30,12 +30,12 @@ const (
 )
 
 func GetWorkspaceMembersMeter(DB *gorm.DB, workspaceID int) int64 {
-	return DB.Model(&model.Workspace{Model: model.Model{ID: workspaceID}}).Association("Admins").Count()
+	return DB.WithContext(context.TODO()).Model(&model.Workspace{Model: model.Model{ID: workspaceID}}).Association("Admins").Count()
 }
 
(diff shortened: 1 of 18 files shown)


backend/pricing/pricing.go
```diff
@@ -30,12 +30,12 @@ const (
 )
 
 func GetWorkspaceMembersMeter(DB *gorm.DB, workspaceID int) int64 {
-	return DB.Model(&model.Workspace{Model: model.Model{ID: workspaceID}}).Association("Admins").Count()
+	return DB.WithContext(context.TODO()).Model(&model.Workspace{Model: model.Model{ID: workspaceID}}).Association("Admins").Count()
```
SpennyNDaJets, line 33: Just curious: whats the impact on the trace when we use `context.TODO()`?

SpennyNDaJets (approved):
image

backend/pricing/pricing.go
ccschmitz: I think it's equivalent to the current behavior where it creates a new trace + context when we don't pass one, but I'm not 100% certain. Curious if @Vadman97 has thoughts on this!

PR #7126, opened 2023-11-10
remove intercom

Vadman97:
## Summary

Replace intercom with discord callouts.
This change allows us to centralize the support to discord while maintaining
code to request help in the app.

Reflame preview.

No

No


frontend/src/util/window.ts
```diff
@@ -1,42 +1,16 @@
...
 	return FRONTEND_URI
 }
 
-interface IntercomInit {
-	admin?: Admin
+export const showSupportBubble = () => {
+	window.open('https://highlight.io/community', '_blank')
 }
 
-export function loadIntercom({ admin }: IntercomInit = {}) {
```
SpennyNDaJets, line 15: Looks like we are still trying to import this function from a couple files
- frontend/src/index.tsx
- frontend/src/routers/AppRouter/AppRouter.tsx

frontend/src/util/window.ts
```diff
@@ -1,42 +1,16 @@
...
-		})
-}
-
-export const showIntercomBubble = () => {
-	window.Intercom('update', { hide_default_launcher: false })
-}
-
-export const showIntercomMessage = (message = '') => {
-	window.Intercom('showNewMessage', message)
+export const showSupportMessage = async (_ = '') => {
```
SpennyNDaJets, line 13: Looks like we aren't using the message that is provided in the method call (e.g. `'Hi! I would like to use the session export feature.'`), and wanted to make sure that this was intended

SpennyNDaJets (changes requested):
Requesting changes for the export errors, but also added a question around the message

frontend/src/util/window.ts
Vadman97: whoops, good catch!

frontend/src/util/window.ts
Vadman97: yeah, wasn't sure what to best do with it; going to add it to the popup

Vadman97:
Screenshot 2023-11-20 at 3 53 23 PM
Screenshot 2023-11-20 at 3 53 34 PM

SpennyNDaJets (approved):
LGTM!

PR #7148, opened 2023-11-13
support esm and cjs properly in highlight.run

### LINKED ISSUE #4933: next.js sdk should export `H` object from `highlight.run`

Vadman97:
## Summary

Add rollup plugins to fix issue with next.js 14 where the @highlight-run/next
could not import the highlight.run SDK.

* Changes our @highlight-run/next build to use rollup for consistency
* Changes our @highlight-run/node exports to prefer cjs for Next

Closes #4933

vercel preview https://vercel.com/highlight-run/nextjs-highlight-demo/GRBpNDQRVLXsi6mRwAmUWrH2wEVz

Next.js 14
Screenshot 2023-11-15 at 11 03 31 AM

Screenshot 2023-11-15 at 11 09 02 AM

express.js CJS usage still ok
Screenshot 2023-11-16 at 10 44 23 AM


deltaepsilon:
There's a whole thread on the Next.js repo about this. It's not just us that ran into the bug:

https://github.com/vercel/next.js/issues/57964

I'm leaning toward waiting for a fix from Next, because they broke a lot of external dependencies. This PR repairs `highlight.run`, but the same error occurs for `@highlight-run/node`, which means we're still broken 😢

deltaepsilon (approved):

PR #7150, opened 2023-11-14
feat: Microsoft Teams Integration

### LINKED ISSUE #6858: Microsoft Teams Integration
Add a microsoft teams integration for issue creation.

https://github.com/infracloudio/msbotbuilder-go

Like our other notification integrations (Slack / Discord), we want to be able to:
- all behavior should match with the other integrations (e.g. slack, discord) for comment notification, alert notifications, alert creation, etc.
- we should allow saving the bot configuration on each alert type or other notification destination. follow the code where other discord / slack configuration is defined.
- for local testing, you can create your own Microsoft Teams Bot (https://learn.microsoft.com/en-us/microsoftteams/platform//sbs-gs-javascript?tabs=vscode%2Cvsc%2Cviscode#register-your-web-service-with-the-bot-framework) and use your own client id / secret; before we deploy this integration in production, you can share your integration's settings with us and we will create a Highlight Oauth app with matching settings

Bounty: $200
cc @jemiluv8

jemiluv8:
## Summary
Microsoft teams bot integration
fixes #6858 

ROADMAP
- [X] Support adding/removing ms teams integration - this just sets up the tenantId and gives our microsoft application access to the tenants ms teams and other related resources
- [x] Support sending notifications via microsoft teams channels for all alert types - session, log, error.
- [x] Figure out bot registration issues for tenant
- [x] Implement support for fetching actual channels on team the bot is installed on
- [x] Test support for all alert types - session, log and error. Seek design comment if any.
- [ ] Final testing and delivery
- [ ] Assist in publishing the bot to the msteams store

### Highlight UI Integration
- [ ] Visit integrations page
- [ ] Observe a new integration titled, Microsoft Teams
- [ ] Toggle the Microsoft teams integration
- [ ] Follow the redirect to authorize app permission
- [ ] Observe the integration is successfully "synced" with highlight
- [ ] Observe that you are able to "remove" the microsoft teams integration

### Bot Integration
- Install the highlight notifications bot to any team under the tenant for which the integration was authoriized in the step above
- Once installed, a message is sent to the teams channel where the app was installed
- If no such message was sent in a reasonable amount of time, it is possible the "Highlight UI Integration" didn't work. You might have to go back to step one and start over

### Setting Up the bot
- Register a bot with the microsoft bot framework https://dev.botframework.com/bots
- Under the bot's settings, set URL to point to either a local instance via (ngrok) or a hosted one
- Send a "ping" message to test that the url was set up correctly - if the setup was done well, you'll receive a response.

### Teams App Installation 
To install the teams bot - before it is published - follow the instructions [here](https://learn.microsoft.com/en-us/microsoftteams/platform//sbs-gs-javascript?tabs=vscode%2Cvsc%2Cviscode&tutorial-step=1#register-your-web-service-with-the-bot-framework)

A sample manifest can be found at https://drive.google.com/file/d/1JRp6vT_2wRaDu4bMBsrOdH_M5zcY5ApP/view?usp=sharing. That is specific to my app so you'd need to unzip the file, make your bot specific modifications and then zip it.

New Environment Variables
MICROSOFT_TEAMS_BOT_PASSWORD (This is the same as the client secret generated for you microsoft application)
MICROSOFT_TEAMS_BOT_ID

### Api Permissions
Ensure the Application has the following api permissions
Screenshot 2024-01-02 at 5 02 43 PM

### New Workspace Fields
MicrosoftTeamsTenantId *string
MicrosoftTeamsChannels *string

### Complete Demo
The link below demonstrates integration and bot setup.
We cannot initiate bot installation from the UI until the bot is "published" to the microsoft store.
https://drive.google.com/file/d/1jR_2o1LkCcjH-Q8LQgD4qIEUo6k4Re1g/view?usp=sharing

### Sample Messages
Screenshot 2024-01-03 at 5 59 51 PM
Screenshot 2024-01-03 at 6 00 02 PM
Screenshot 2024-01-03 at 6 00 24 PM
Screenshot 2024-01-03 at 6 00 45 PM

Not yet

Screenshot 2023-11-14 at 3 34 45 PM


backend/alerts/integrations/microsoft-teams/message_templates.go
```diff
@@ -0,0 +1,349 @@
...
+	"actions": [
+		{
+			"type":  "Action.OpenUrl",
+			"title": "{{.ActionTitle}}",
+			"url":   "{{.ActionURL}}"
+		}
+	]
+}`)
+
+var ErrorAlertMessageTemplate = []byte(`{
```
jemiluv8, line 248: Most messages sent by the ms teams bot use adaptive cards.
Even though they need to be valid json, I treat them like templates.
The "rendered" template must however be valid json.

backend/alerts/integrations/microsoft-teams/messages_test.go
```diff
@@ -0,0 +1,225 @@
+package microsoft_teams
```
jemiluv8, line 1: I'm taking a leaf out of the example set by the discord implementation.
This is probably the fastest and most effective way to test the ms teams
bot.

Successfully running this test validates about 60+% of the work for this PR.

backend/alerts/integrations/microsoft-teams/microsoft_teams.go
```diff
@@ -0,0 +1,203 @@
...
+
+	err = json.Unmarshal(b, &unmarshalled)
+	if err != nil {
+		return unmarshalled, errors.Wrap(err, "error unmarshaling microsoft api response"+string(b))
+	}
+
+	return unmarshalled, nil
+}
+
+func GetTeamsChannel(workspace *model.Workspace) ([]*model.MicrosoftTeamsChannel, error) {
```
jemiluv8, line 191: I chose to save the ms teams channels on the workspace as a mapping of teamId to list of channels

backend/alerts/integrations/microsoft-teams/templates_test.go
```diff
@@ -0,0 +1,124 @@
+package microsoft_teams
```
jemiluv8, line 1: My treatment of adaptive cards can be error prone. 
Tests need to exist of all templates.

backend/main.go
```diff
@@ -417,7 +417,9 @@ func main() {
 			zapier.CreateZapierRoutes(r, db, &zapierStore, &rh)
 		})
 		r.HandleFunc("/slack-events", privateResolver.SlackEventsWebhook(ctx, slackSigningSecret))
+		r.Post(fmt.Sprintf("%s/%s", privateEndpoint, "microsoft-teams/bot"), privateResolver.MicrosoftTeamsBotEndpoint)
```
jemiluv8, line 420: An endpoint it needed by the microsoft bot framework when setting up a bot.
This is used by the framework to send messages to the bot.

All request validation is done by the msbotbuilder framework

backend/private-graph/graph/resolver.go
```diff
@@ -2105,6 +2123,49 @@ func (r *Resolver) AddSlackToWorkspace(ctx context.Context, workspace *model.Wor
 	return nil
 }
 
+func (r *Resolver) RemoveMicrosoftTeamsFromWorkspace(workspace *model.Workspace, projectID int) error {
+
+	if err := r.DB.Transaction(func(tx *gorm.DB) error {
```
jemiluv8, line 2128: We remove the tenant integration and the channels saved on the workspace.
We then remove teams channels to notify on all alerts.

backend/private-graph/graph/resolver.go
```diff
@@ -3667,3 +3728,141 @@ func (r *Resolver) GetJiraProjects(
...
+
+	bot, err := microsoft_teams.NewMicrosoftTeamsBot("common")
+
+	if err != nil {
+		log.WithContext(ctx).Error(e.Wrap(err, "error making ms bot adapter"))
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	act, err := bot.ParseRequest(ctx, req)
```
jemiluv8, line 3744: This effectively "validates" and authenticates the request using our client id and secret.
It also extracts the payload into a friendly form that we are able to interact with.

backend/private-graph/graph/resolver.go
```diff
@@ -3667,3 +3728,141 @@ func (r *Resolver) GetJiraProjects(
...
+	if err != nil {
+		log.WithContext(ctx).Error(e.Wrap(err, "error parsing bot framework request"))
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	if act.Type == schema.ConversationUpdate {
+		if len(act.MembersRemoved) > 0 {
+			// identify if our bot is part of the id
+			for _, memberRemoved := range act.MembersRemoved {
```
jemiluv8, line 3755: MS teams bot can be installed in multiple "teams/groups" for the same integration (aka tenantID).
So when the bot is removed, we just remove the channels that had previously been saved
belonging to the team/group removed.

jemiluv8 (commented):
Self Review completed

frontend/src/pages/Alerts/ErrorAlert/ErrorAlertPage.tsx
```diff
@@ -275,6 +288,7 @@ export const ErrorAlertPage = () => {
 						}
 
 						if (isCreate) {
+							console.log('Input', input)
```
mayberryzane, line 291: can remove this console.log

frontend/src/pages/Setup/SetupRouter/AlertsSetup.tsx
```diff
@@ -380,6 +405,7 @@ const AlertPicker = function ({
 				})
 				channelID = data?.upsertDiscordChannel.id ?? ''
 			}
+			// TODO: do we have to do same for microsoft_teams platform?
```
mayberryzane, line 408: is this comment still relevant?

mayberryzane (changes requested):
sorry for the delay here! code looks good, just a few small items and will click test tomorrow

backend/alerts/alerts.go
```diff
@@ -77,6 +78,27 @@ func SendErrorAlert(ctx context.Context, event SendErrorAlertEvent) error {
...
+			return nil
+		}
+
+		bot, err := microsoft_teams.NewMicrosoftTeamsBot(*event.Workspace.DiscordGuildId)
+		if err != nil {
+			return err
+		}
+
+		payload = attachReferrerToErrorAlertPayload(ctx, payload, routing.Discord)
+		for _, channel := range event.ErrorAlert.DiscordChannelsToNotify {
```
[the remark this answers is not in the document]
jemiluv8: Correct.

backend/alerts/alerts.go
```diff
@@ -77,6 +78,27 @@ func SendErrorAlert(ctx context.Context, event SendErrorAlertEvent) error {
 		return nil
 	})
 
+	g.Go(func() error {
+		if !isWorkspaceIntegratedWithMicrosoftTeams(*event.Workspace) {
+			return nil
+		}
+
+		bot, err := microsoft_teams.NewMicrosoftTeamsBot(*event.Workspace.DiscordGuildId)
```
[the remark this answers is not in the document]
jemiluv8: Correct

frontend/src/pages/Setup/SetupRouter/AlertsSetup.tsx
jemiluv8: No. Its been removed now.

jemiluv8:
I've made changes requested.

mayberryzane:
Hi @jemiluv8, thank you. I set up the integration locally and it was mostly looking good. I had 2 questions:
1. After integrating with Highlight, would it be possible to retrieve all available Teams channels whenever the alerts page is loaded instead of when the bot is installed? It's a little clunky at the moment because if a bot is added before Highlight is integrated, the alerts UI shows no available channels. Does this also mean we would have to re-install the bot in Teams whenever new channels are added?
2. How can a user delete the bot/app in Teams after it is installed and added to a team? Is there anything we can do to list the team here under "Recently Used" on the "Manage your apps" page?
Screen Shot 2024-01-22 at 5 40 58 PM

jemiluv8:
> How can a user delete the bot/app in Teams after it is installed and added to a team? 

I'm assuming you want the bot to be uninstalled on all Microsoft teams under the tenant whenever the highlight UI integration is "removed". I have to look into this. I expect there could be an api call that can do this.

At the moment, since we support installing the bot on multiple "teams" under the same "tenant", we remove the teams integration on highlight when the bot is removed and there are no longer other bot installations for the tenant. 

_So when we do a teams integration and install the bot on one "team", uninstalling the bot on the team will also "remove" the highlight integration._

> Is there anything we can do to list the team here under "Recently Used" on the "Manage your apps" page?

I'm not sure I understand. Based on the screenshot below, it seems the team is listed there or I'm missing something.
Screenshot 2024-01-23 at 11 43 31 AM

mayberryzane:
@jemiluv8 
> Installing the bot before highlight UI integration is a no op because there is no way we can associate the "team" the bot was installed on and the highlight workspace/account. So our "reasonable" constraint is to ensure that the integration is done on highlight UI before the bot is installed.

If the change was made to fetch all available channels instead of updating that channels list, wouldn't this be possible? Whenever a user is on the alerts page, we could fetch all teams available to the bot under the Teams tenant id associated with the Highlight workspace, then fetch all channels in each of those teams. I think for maintainability, it will be a lot easier if we're fetching the channels list whenever the user is on the alerts screen. In most cases, the bot will only be added to one or two teams, and updating an alert is an infrequent action, so if it takes a second or two to load all channels, this is still a better UX than showing an outdated channel list or requiring the bot to be installed. Plus, fetching the channels for each team can be parallelized. Thanks for including all the details in your caveat.

For deleting, maybe I hit some weird bug on my end with the MacOS version of Teams, but it's preventing me from setting up the Highlight integration after disabling it in the Highlight UI because I am unable to uninstall / reinstall the bot. I think if the above change is possible, this won't be an issue. Thank you!

jemiluv8:
> @jemiluv8
> 
> > Installing the bot before highlight UI integration is a no op because there is no way we can associate the "team" the bot was installed on and the highlight workspace/account. So our "reasonable" constraint is to ensure that the integration is done on highlight UI before the bot is installed.
> 
> If the change was made to fetch all available channels instead of updating that channels list, wouldn't this be possible? Whenever a user is on the alerts page, we could fetch all teams available to the bot under the Teams tenant id associated with the Highlight workspace, then fetch all channels in each of those teams. I think for maintainability, it will be a lot easier if we're fetching the channels list whenever the user is on the alerts screen. In most cases, the bot will only be added to one or two teams, and updating an alert is an infrequent action, so if it takes a second or two to load all channels, this is still a better UX than showing an outdated channel list or requiring the bot to be installed. Plus, fetching the channels for each team can be parallelized. Thanks for including all the details in your caveat.
> 
> For deleting, maybe I hit some weird bug on my end with the MacOS version of Teams, but it's preventing me from setting up the Highlight integration after disabling it in the Highlight UI because I am unable to uninstall / reinstall the bot. I think if the above change is possible, this won't be an issue. Thank you!

Agreed. Will make sure we always fetch the ms teams channels when on the alerts page. I will also parallelize the api requests.

mayberryzane:
@jemiluv8 also just pushed a few copy changes and renamed all usages of `MICROSOFT_TEAMS_CLIENT_ID` to `MICROSOFT_TEAMS_BOT_ID` - was there a reason to have separate environment variables here or is ok for them to be the same? My understanding is that these are the same IDs, feel free to revert that if it's an issue

jemiluv8:
> @jemiluv8 also just pushed a few copy changes and renamed all usages of `MICROSOFT_TEAMS_CLIENT_ID` to `MICROSOFT_TEAMS_BOT_ID` - was there a reason to have separate environment variables here or is ok for them to be the same? My understanding is that these are the same IDs, feel free to revert that if it's an issue

I've seen, reviewed and agree with the changes. They are the same IDs.

backend/alerts/integrations/microsoft-teams/microsoft_teams.go
```diff
@@ -0,0 +1,234 @@
...
+
+	err = json.Unmarshal(b, &unmarshalled)
+	if err != nil {
+		return unmarshalled, errors.Wrap(err, "error unmarshaling microsoft api response"+string(b))
+	}
+
+	return unmarshalled, nil
+}
+
+func GetMicrosoftTeamsChannelSuggestions(workspace *model.Workspace) ([]*model.MicrosoftTeamsChannel, error) {
```
jemiluv8, line 194: This feels overkill and I also return the first error we encounter.
Below is a simpler non-concurrent version I wrote

```go
func GetTeamsChannel(workspace *model.Workspace) ([]*model.MicrosoftTeamsChannel, error) {
	allChannels := []*model.MicrosoftTeamsChannel{}
	teamsGroups := GetMicrosoftTeamsGroupsFromWorkspace(workspace)

	for _, teamGroup := range teamsGroups {
		channels, err := GetMicrosoftTeamsChannels(*workspace.MicrosoftTeamsTenantId, teamGroup)
		if err != nil {
			return nil, err
		}
		allChannels = append(allChannels, channels...)
	}

	return allChannels, nil
}
```
I feel like there is a simpler way to do this but I'm unable to find one at the moment.
You can rewrite this or suggest a "better" way of going about this.

jemiluv8:
I've introduced support for fetching the ms teams channels directly from the microsoft graph api.

mayberryzane:
thanks @jemiluv8 , testing this today

mayberryzane:
@jemiluv8 just made a couple more changes - I stopped saving the groups with the workspace and instead just save tenant id, then use the `/teams` endpoint to load all teams. Tested a concurrent / non-concurrent version on a teams instance with 6 teams, took 1.6s for the concurrent version and 3.5s for the non-concurrent one so decided to keep it concurrent (changed it to use the errgroup library). For permissions, I think it just needs these:
Screen Shot 2024-01-29 at 3 34 20 PM

Should be good to merge in now, I'll make sure it gets merged in / passes checks. Thanks for all your work here!

mayberryzane (approved):

jemiluv8:
### Just to clarify
The changes look good to me. 

One issue with fetching all teams(under the tenant) and then all channels is that we cannot send to a channel until the bot is installed on the team. At the moment, we don't "default install" the bot on all teams under the tenant.

### This means
So we can setup an alert to be sent to a teams channel that will not receive the message if the bot is not installed on the "team"

### Conclusion
That was the reason I was storing the list of teams the bot was installed on in the workspace. So I only fetch the channels for these teams. Also, when a bot is uninstalled from a team, we remove the team from the list of teams. That fixes the above issue to some extent.

### Options
One thing we could do is to "auto install" the teams bot on all teams in the channel.

mayberryzane:
@jemiluv8 thanks for the additional context, I just added another check to verify that our app is installed in the team by using [this api](https://learn.microsoft.com/en-us/graph/api/team-list-installedapps?view=graph-rest-1.0&tabs=http). Not great because it's a 2n+1 query, but in testing this took about 2s which should be fine as it's only when loading the alerts page for users with many team. We can look at optimizing in the future if this becomes a user pain point, but it keeps all the state of which teams our bot is installed on on the MS Teams side so there's no issue of keeping our state in sync. Will merge this in today

jemiluv8:
Your update works just fine. I think I finally understand our differing “points of view” about “state”

Your preference for not having some state to keep in sync helps avoid a lot of problems 
with managing state even though it may not be the most performant.

At the same time too doing all this exxtra work to sync state over time can lead to bugs that we may have a hard time
keeping track of. The trade-off in this case makes sense because performance is not the priority right now. And the expected
number of users using this feature is “expected to be low”.

I’ll keep this in mind for future work on integrations. Its been cool collaborating on this.

PR #7160, opened 2023-11-16
Open session in right panel

ccschmitz:
## Summary

Opens a session in a flyout panel rather than redirecting to the session player.

### What's Left

- [ ] Don't load things that aren't needed for the streamlined player


ccschmitz:
This was a hackathon branch. Working on this in the next cycle.

PR #7161, opened 2023-11-17
swap for cbrotli libary to test performance

mayberryzane:
## Summary
- profiling shows 68% of the worker CPU time spent compressing brotli payloads
Screen Shot 2023-11-16 at 4 21 45 PM

- try replacing with the cbrotli library: https://pkg.go.dev/github.com/google/brotli@v1.0.7/go/cbrotli

- click tested locally to make sure sessions were being brotli encoded after processing
- ran `./run-backend.sh` to make sure brotli was being installed locally
- ran `docker buildx build --build-arg REACT_APP_COMMIT_SHA=zane --platform linux/arm64 -f deploy/Dockerfile.prod.arm64 .` to make sure the arm64 container build was working

- will monitor after deployment to check if things are actually faster. will revert if there's not a substantial difference because this adds a bunch of complexity to the build

- no


backend/main.go
```diff
@@ -359,7 +359,9 @@ func main() {
 	r.Use(httplog.RequestLogger(errorLogger))
 	compressor := middleware.NewCompressor(5, "application/json")
 	compressor.SetEncoder("br", func(w io.Writer, level int) io.Writer {
-		return brotli.NewWriterLevel(w, level)
+		return cbrotli.NewWriter(w, cbrotli.WriterOptions{
+			Quality: level,
```
Vadman97, line 363: 5 seems reasonable, though in the future could look at lowering the level further 
Screenshot 2023-11-17 at 9 43 50 AM

docker/run-backend.sh
```diff
@@ -1,11 +1,27 @@
...
 ./telemetry.sh
 source env.sh
 
 # start backend
 pushd ../backend
 go install github.com/cosmtrek/air@latest
 go install github.com/go-delve/delve/cmd/dlv@latest
+
+
+if [ ! /usr/local/bin/brotli ]; then
```
Vadman97, line 14: weird, this check didn't enter but i don't have the file
Screenshot 2023-11-17 at 9 52 09 AM
Screenshot 2023-11-17 at 9 52 26 AM

Vadman97 (approved):

mayberryzane:
from the benchmarks here, the gains for switching to cbrotli would be pretty marginal for the extra complexity this adds: https://groups.google.com/g/golang-nuts/c/J5MwJnmDpOw
going to close this in favor of #7164

backend/main.go
mayberryzane: 👍 agree that level 5 is probably overkill
I don't think this is a big deal atm because private graph uses a fraction of the CPU / mem out of all our services, but if this shows up in profiling would be good to do
Incidentally was looking at similar performance benchmarks and I think it makes sense to reduce the brotli level from 9 to 5 for our long-term session storage #7164

PR #7167, opened 2023-11-18
track sessions to process in redis instead of postgres

mayberryzane:
## Summary
- during periods of high DB load, the worker "get sessions to process" query is consistently the most expensive query. redis makes more sense for frequent polling
Screen Shot 2023-11-17 at 1 41 33 PM
- add 3 methods to add, remove, and get sessions to process from redis
- can remove `payload_updated_at` in the future to reduce some more DB load
- interacted w/ Highlight locally while observing the state in redis
Screen Shot 2023-11-17 at 5 04 30 PM

- will create a `sessions-to-process` key in redis and add all current unprocessed sessions to it
- can revert if there are issues

- no

### DIFF
--- a/backend/redis/utils.go
+++ b/backend/redis/utils.go
@@ -22,6 +22,7 @@ import (
 )
 
 const CacheKeyHubspotCompanies = "hubspot-companies"
+const CacheKeySessionsToProcess = "sessions-to-process"
 
 type Client struct {
 	Client  redis.Cmdable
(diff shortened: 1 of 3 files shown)


backend/redis/utils.go
```diff
@@ -344,6 +345,89 @@ func (r *Client) GetResources(ctx context.Context, s *model.Session, resources m
...
+
+	return nil
+}
+
+// Retrieves up to `limit` sessions to process. Sets the processing time
+// for each to `lockPeriod` minutes after the current time, so they
+// can be retried in case processing fails.
+func (r *Client) GetSessionsToProcess(ctx context.Context, lockPeriod int, limit int) ([]int64, error) {
+	now := time.Now().Unix()
+	timeAfterLock := float64(now+int64(60*lockPeriod)) + .5
```
Vadman97, line 395: what's the `.5` for?

backend/redis/utils.go
```diff
@@ -344,6 +345,89 @@ func (r *Client) GetResources(ctx context.Context, s *model.Session, resources m
...
+			table.insert(input, timeAfterLock)
+			table.insert(input, item)
+			count = count + 1
+		end
+
+		if count == 0 then
+			return {}
+		end
+
+		redis.call("ZADD", key, unpack(input))
```
Vadman97, line 415: what's the point of reinserting the values? does `zrangebyscore` pop off things that it returns?

Vadman97 (approved):

backend/redis/utils.go
mayberryzane: a bit hacky, but I wanted a way to encode that this was updated by the worker and not the public-worker. when removing processed sessions, I'm checking that the keys to remove have this non-integer score before removing, in case a session had new events added to it while it was being processed

backend/redis/utils.go
mayberryzane: the values were not removed, `ZADD` is just updating their scores here (with the new time)

PR #7173, opened 2023-11-20
Tune logs virtual table

SpennyNDaJets:
## Summary
Reports of the logs table flickering which is making it hard to use. Make the following updates:
- load 50 extra rows (i.e. 1 page) into the virtual table
- do not use debug mode
- add `data-index` to resolve warning
- memoize rows to allow expanding data objects

1) Visit the logs table and confirm better UX:
- scroll down to load more
- scroll up
- open rows
- expand objects

N/A

N/A


ccschmitz (approved):
😻

PR #7179, opened 2023-11-20
capture exceptions reported on logs

Vadman97:
## Summary

Python `logging` can report `exc_info` on the exception which should
be prioritized over the message, since the message will be a non-machine-readable
format of the stacktrace.

This supports loguru logging as well as custom usage of decorators to catch exceptions
with tools like loguru.

Updated e2e test, new unit test.

Screenshot 2023-11-20 at 11 16 49 AM

New python SDK version

No


SpennyNDaJets (approved):
LGTM

PR #7181, opened 2023-11-20
Refactor Docs titles

deltaepsilon:
## Summary

We've long suffered from duplicate page titles in our docs. It shows up in our Algolia crawler results. Searches sometimes return stacks of similarly-named pages. For instance, search for "overview" and you'll see over a dozen identical results.

I created a new `toc` frontmatter field that we can use to force the Table of Contents title to stay the same, while modifying the titles for each file.

I wrote a script to report on duplicate titles and to dump frontmatter to a file. Fire off `bun bin/generate-content-report.ts`.

image

image

### DIFF
--- a/bin/.gitignore
+++ b/bin/.gitignore
@@ -1,0 +1,1 @@
+reports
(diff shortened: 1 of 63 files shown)


SpennyNDaJets (approved):
LGTM

PR #7191, opened 2023-11-21
s3 script delete instead of tag

mayberryzane:
## Summary
- upgraded `github.com/golang-migrate/migrate/v4` due to missing dependency on the old version
- change script to delete instead of tag, because the tag endpoint's rate limits are too low

- tested locally against dev buckets

- will run in ec2 after merging
- will have to run again in the future to handle any current untagged objects

- no


SpennyNDaJets (approved):
Code makes sense to me

PR #7192, opened 2023-11-21
support additional fastapi attributes on http errors

Vadman97:
## Summary

A customer brought up that our [FastAPI integration](https://app.intercom.com/a/inbox/gm6369ty/inbox/shared/all/conversation/41483?view=List) does not report
much detail on HTTP errors, with a stacktrace that is not useful since
it will be pointing to FastAPI internal code.

This change introduces additional attributes to the FastAPI integration.

Unit test

Screenshot 2023-11-22 at 12 01 12 PM

No

No

### DIFF
--- a/sdk/highlight-py/highlight_io/integrations/fastapi.py
+++ b/sdk/highlight-py/highlight_io/integrations/fastapi.py
@@ -35,6 +35,10 @@ class FastAPIMiddleware(BaseHTTPMiddleware):
                     status_code=resp.status_code,
                     headers=resp.headers.__dict__,
                     detail=body.decode(),
+                    attributes={
+                        "http.method": request.method,
+                        "http.url": request.url,
+                    },
                 )
                 return Response(
                     content=body,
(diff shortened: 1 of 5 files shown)


sdk/highlight-py/highlight_io/integrations/fastapi.py
```diff
@@ -35,6 +35,10 @@ async def dispatch(
                     status_code=resp.status_code,
                     headers=resp.headers.__dict__,
                     detail=body.decode(),
```
elithe1, line 37: Wondering wether its a hustle to also record the request body here and not only the response body?
Often its useful and helpful for error reproduction to understand whats the request payload.

sdk/highlight-py/highlight_io/integrations/fastapi.py
```diff
@@ -35,6 +35,10 @@ async def dispatch(
...
+                    attributes={
+                        "http.method": request.method,
```
elithe1, line 39: 🙏🙏🙏

sdk/highlight-py/highlight_io/integrations/fastapi.py
Vadman97: updated to capture the request body when possible!

deltaepsilon (approved):

PR #7194, opened 2023-11-22
Don't use time fields in segments

ccschmitz:
## Summary

Segments currently persist the date filters that are being used when they are created/updated. This isn't the expected behavior and it's even more problematic because after 30 days there's a good chance a segment will show empty search results because there isn't any data to show anymore.

This PR prevents us from persisting the time range for the query and also filters it out when applying a segment so that existing segments will work as expected.

I also uncovered a bug where the query would be reset to the default when toggling the segment create/edit modal - that is fixed as well.

Ensure the date field isn't applied in the following scenarios:

- [ ] While on `main`, create a segment. This should persist with the date range field.
- [ ] Check out this branch and then apply that segment. The time range should not be applied.
- [ ] Create a new segment and ensure the time range is not persisted.
- [ ] Edit a segment and ensure the time range is not persisted.

N/A - all client-side changes

Nope!

### DIFF
--- a/frontend/src/__generated/index.css
+++ b/frontend/src/__generated/index.css
@@ -5613,6 +5613,24 @@ body {
 ._1vb6x0p0:hover {
   color: var(--_1pyqka9y);
 }
+._1ukfp2a0 {
+  border-radius: 0;
+  border: 0 !important;
+}
+._1ukfp2a0:focus,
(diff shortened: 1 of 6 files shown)


Vadman97 (approved):
click test with creating a segment in this PR and loading an existing segment looks great! :tada:

PR #7198, opened 2023-11-22
Loading style tweaks to logs and traces stats

ccschmitz:
## Summary

Cleans up the loading styles for stats on logs and traces and makes the styles consistent across both pages.

Local click test. Check out the Reflame preview + production and toggle between logs/traces pages to view the differences.

N/A - minor client-side tweaks.

N/A - these changes were requested by Julian.


frontend/src/pages/LogsPage/LogsCount/LogsCount.css.ts
SpennyNDaJets, line 1: Noticed that reflame won't go in and delete files, so wonder if we should also delete `frontend/src/__generated/ve/pages/LogsPage/LogsCount/LogsCount.css.js`?

SpennyNDaJets (approved):
Looks good! Left a question about the generated reflame files, but not blocking

frontend/src/pages/LogsPage/LogsCount/LogsCount.css.ts
ccschmitz: Thanks for the note!

PR #7201, opened 2023-11-22
optimize pushpayload data transfer

Vadman97:
## Summary

- use compression for pushpayload uploads in client. possibly use snappy instead of brotli since it's less cpu resource intensive.
- bring limit down xhr-listener for network requests (down from 64MB)
- add support to ignore certain graphql requests (we can use this to filter GetSessionPayload for live sessions)
- css animations cause a lot of events to push payload. we should consider debouncing these.
- sending requests despite disableSessionRecording

Screenshot 2024-01-10 at 6 16 02 PM

compressed session being recorded
Screenshot 2024-01-10 at 6 17 19 PM

Changeset

No


sdk/client/src/workers/highlight-client-worker.ts
```diff
@@ -143,37 +155,46 @@ function stringifyProperties(
...
 		if (highlightLogs) {
 			payload.highlight_logs = highlightLogs
 		}
 
-		const eventsSize = graphqlSDK
-			.PushPayload(payload)
-			.then((res) => res.pushPayload ?? 0)
+		const buf = strToU8(JSON.stringify(payload))
+		const compressed = compressSync(buf, {
+			level: 9,
```
mayberryzane, line 173: should we use the default level / mem here? I think 9 and 12 are the highest values respectively, but in this case because it's only compressed to limit data transmitted I would think this uses too much cpu / mem resources on the client. also curious how this library compares to the browser native https://developer.mozilla.org/en-US/docs/Web/API/CompressionStream

mayberryzane (approved):

sdk/client/src/workers/highlight-client-worker.ts
Vadman97: my thought here was this is running on a web worker so even if we have higher cpu, we run the risk of blocking the main thread. i'll stick to the defaults tho since there is probably little return for higher background cpu usage.

re. compressionstream - it sounds like [fflate](https://github.com/101arrowz/fflate?tab=readme-ov-file#what-about-compressionstream) thinks that it is better, but i will try some benchmarks and report back

Vadman97: tried compressionstream but ran into some issues with getting it to work in the web worker. looks like compressionstream is only supported by browsers of 90% of all devices so i'll stick to fflate for now with the defaults.

PR #7207, opened 2023-11-27
Improve logs table scrolling 

SpennyNDaJets:
## Summary
When viewing on large screens (i.e. >2000px in height) the logs table will only fill part of the page, and not let the user scroll for more logs.

- Fix the table to fetch logs until the page is full
- Improve heights of other logs tables
- Reduce width of timestamp column 

Fetching logs on large screens
1. While viewing Highlight, inspect the page and set the view to be responsive with a width of 4000px
2. Load the logs page
- [x] Logs are loaded until the page is full (3 or 4 requests)
- [x] Able to scroll to fetch more logs
- [x] No excess requests sent
3. View the logs table on the network resources
- [x] Height of logs table fits the space
- [x] No other changes to the other tables
4. View the logs table on the traces page
- [x] Height of logs table fits the space
- [x] No other changes to the other tables

https://www.loom.com/share/6de9eec69fdf44cfa335e77b26b83ea2?sid=37385118-ee70-4d3b-8bd2-88d1d55ecc40

N/A

N/A


frontend/src/pages/Player/RightPlayerPanel/components/NetworkResourcePanel/NetworkResourceLogs.tsx
```diff
@@ -146,6 +146,7 @@ export const NetworkResourceLogs: React.FC<{
 								query={query}
 								selectedCursor={undefined}
 								fetchMoreWhenScrolled={fetchMoreWhenScrolled}
+								bodyHeight="calc(100% - 28px)"
```
ccschmitz, line 149: Where does `28px` come from? Maybe worth adding a comment or assigning to a variable with a meaningful name?

Same goes for the change in `TraceLogs.tsx`.

ccschmitz (approved):
Looks great, but I did notice a couple small regressions. Let me know if you think they are worth following up on here before merging!

1. It looks like there is some scrolling now on the Logs tab on network resources and traces, and there didn't used to be.

https://www.loom.com/share/79b5ad313b3a436fb371acfa638dc0b7

2. I'm seeing we always are fetching when scrolling to the bottom now, even if we should know that there are no more results to load.

https://www.loom.com/share/868ba99d050f46c99bcd946840aa8922

frontend/src/pages/LogsPage/LogsTable/LogsTable.tsx
```diff
@@ -273,6 +278,23 @@ const LogsTableInner = ({
...
+	useEffect(() => {
+		setTimeout(() => {
+			if (!loadingAfter && bodyRef?.current) {
+				fetchMoreWhenScrolled(bodyRef.current, true)
+			}
+		}, 0)
+		// eslint-disable-next-line react-hooks/exhaustive-deps
+	}, [loadingAfter])
+
+	const handleFetchMoreWhenScrolled = (
```
SpennyNDaJets, line 290: Couldn't use `useRef` here because it wouldn't respect state changes when a new query was fetched (i.e. `hasNextPage` was always true since it was initialized this way)

frontend/src/pages/Traces/TraceLogs.tsx
```diff
@@ -130,6 +132,7 @@ export const TraceLogs: React.FC = () => {
 								query={query}
 								selectedCursor={undefined}
 								fetchMoreWhenScrolled={fetchMoreWhenScrolled}
+								bodyHeight="400px"
```
SpennyNDaJets, line 135: Can't use a percentage here since the panel is scrollable. We could look into making this panel not scrollable like the sessions' network tab, but will be a larger effort. I believe this is a good compromise/improvement over the previous state

ccschmitz (approved):

PR #7214, opened 2023-11-27
Using session_device_id for user filter if unidentified

### LINKED ISSUE #7183: clicking the 'search for other sessions' icon for a non-identified session populates an invalid filter
Screenshot 2023-11-20 at 1 28 22 PM

https://discord.com/channels/1026884757667188757/1174463824288567308

clicking the button next to the identifier on the top right for a session that is not identified populates the `identifier` query with something that doesn't match. instead, we should use the `clientID` query

image

deltaepsilon:
## Summary

Closes #7183 

image

Click test

nope!

nope!

### DIFF
--- a/frontend/src/pages/Player/MetadataBox/MetadataBox.tsx
+++ b/frontend/src/pages/Player/MetadataBox/MetadataBox.tsx
@@ -189,12 +189,16 @@ export const MetadataBox = React.memo(() => {
 		const displayName = getDisplayName(session)
 		const userParam = validateEmail(displayName) ? 'email' : 'identifier'
 
-		setSearchQuery((query) =>
-			buildQueryStateString({
+		setSearchQuery((query) => {
+			const params = session.identified
+				? { [`user_${userParam}`]: displayName }
+				: { session_device_id: String(session.fingerprint) }
+
+			return buildQueryStateString({
 				query,
-				[`user_${userParam}`]: displayName,
-			}),
-		)
+				...params,
+			})
+		})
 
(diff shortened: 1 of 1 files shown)


frontend/src/pages/Player/MetadataBox/MetadataBox.tsx
```diff
@@ -189,12 +189,16 @@ export const MetadataBox = React.memo(() => {
 		const displayName = getDisplayName(session)
 		const userParam = validateEmail(displayName) ? 'email' : 'identifier'
 
-		setSearchQuery((query) =>
-			buildQueryStateString({
+		setSearchQuery((query) => {
+			const params = session.identified
+				? { [`user_${userParam}`]: displayName }
```
SpennyNDaJets, line 194: Just wanted to check if you knew of the `getDisplayNameAndField` method? Seems to have the different cases listed with the value and identifier.

SpennyNDaJets (approved):
Click test looks good to me! Just one non-blocking question below to make sure we are handling any edge cases

frontend/src/pages/Player/MetadataBox/MetadataBox.tsx
deltaepsilon: Thanks for the heads up!

I've reviewed the function. It looks like it's doing a fall-through where it looks for name, email, identifier or fingerprint. I think this change still checks out, because the fingerprint should be universally available and maps to the `session_device_id`.

PR #7220, opened 2023-11-28
support selecting a random subset of sessions

Vadman97:
## Summary

Adds a new `sample` session filter that allows picking a random subset of sessions
based on a randomly generated (or set) hex string.
The subset does not have a given count limit but sets the order of
sessions that are returned, so that a given page number identifies the sample.

v0 https://www.loom.com/share/289a58d71f664e329dd73509887df035

No

No


SpennyNDaJets (approved):
Really cool! Code and click test look good to me!

PR #7226, opened 2023-11-28
Build antd less stylesheet

lewisl9029:
## Summary

Reflame doesn't have native support for less yet, so to customize the antd less stylesheet, we'll need a local build script to compile it into css.

This PR adds the build script and compiled css. It should only need to be reran when updating antd or its overrides, which should be rare, so I have excluded this script from the main reflame.mjs build script entry point so it can be ran only when necessary.

Reflame preview reflects style overrides now:

Screenshot 2023-11-28 at 2 54 04 PM


frontend/vite.config.ts
```diff
@@ -112,17 +112,6 @@ export default defineConfig(({ mode }) => {
 		},
 		css: {
 			devSourcemap: true,
-			preprocessorOptions: {
```
lewisl9029, line 115: No longer need this now that both reflame and vite are referring to the pre-built css.

frontend/src/pages/Sessions/SessionsFeedV3/SessionQueryBuilder/components/LengthInput/LengthInput.tsx
```diff
@@ -1,5 +1,3 @@
-import 'antd/dist/antd.css'
-
```
lewisl9029, line 2: Looks like we were importing the same css twice?

Vadman97 (approved):
looks great! there is something slightly weird happening with the antd-tabs margins but now i have a clue how to look more into this (now it might due to a different override we're setting). 

Screenshot from 2023-11-28 15-08-55

Vadman97:
@lewisl9029 i think something in the vite config has to be fixed to stop trying to bundle the antd less?
```
✓ built in 4.61s
  [vite:css] [less] Inline JavaScript is not enabled. Is it set in your options?
  file: /home/runner/actions-runner/_work/highlight/highlight/node_modules/antd/es/spin/style/index.less:110:0
  error during build:
  Error: [less] Inline JavaScript is not enabled. Is it set in your options?
      at less (file:///home/runner/actions-runner/_work/highlight/highlight/node_modules/vite/dist/node/chunks/dep-bb8a8339.js:39494:33)

```

lewisl9029:
Ah I didn't realize Vite was processing less in antd :thinking: 

Will dig further.

lewisl9029:
@Vadman97 I think https://github.com/highlight/highlight/pull/7226/commits/1d5c6ca9cd04512afd83deb91e67a472fd605a61 should do it.

This is what pulled less files into the dependency graph:

Screenshot 2023-11-29 at 1 12 09 PM

We should no longer need this now that we're using the prebuilt css stylesheet.

PR #7245, opened 2023-11-30
View logs by environment

SpennyNDaJets:
## Summary
Add environment filters for logs

1) Navigate to the logs page, and open a log
- [ ] Log displays the correct environment
- [ ] Logs can be filtered by env

https://www.loom.com/share/26206bcd30b143aaa2909a12af845575?sid=78e0b27d-f17b-482a-b52f-b5f8b8397edb

Make sure backwards compatible for sdks not passing in env

N/A


backend/private-graph/graph/schema.graphqls
```diff
@@ -916,6 +918,7 @@ enum ReservedTraceKey {
 }
 
 enum ReservedErrorObjectKey {
+	environment
```
SpennyNDaJets, line 921: Added this because believe it was missing?

PR #7251, opened 2023-12-01
Add environment clickhouse migrations

SpennyNDaJets:
## Summary
We want to apply the environment from the backend sdk to the logs and traces information. Before we do so, we must migrate the clickhouse house tables to accept the environment column

Specs pass

Make sure backwards compatible for SDKs that do not report environment

N/A


frontend/src/pages/LogsPage/LogsTable/LogDetails.tsx
```diff
@@ -106,6 +106,7 @@ export const LogDetails: React.FC<Props> = ({
 		source,
 		service_name: serviceName,
 		service_version: serviceVersion,
+		environment: null, // TODO: replace this when implementing for logs
```
Vadman97, line 109: just confirming this todo is for the follow up PR which will add querying for the environment value?

Vadman97 (approved):

PR #7254, opened 2023-12-01
fix reflame antd style discrepancies

Vadman97:
## Summary

Fixes additional antd override discrepancies in reflame deploys.

before
Screenshot from 2023-11-30 18-21-16

after
Screenshot from 2023-11-30 18-21-21

before
Screenshot from 2023-11-30 18-20-44

after
Screenshot from 2023-11-30 18-20-49

No

No

### DIFF
--- a/frontend/src/index.tsx
+++ b/frontend/src/index.tsx
@@ -1,8 +1,8 @@
-import './__generated/antd.css'
 import '@highlight-run/rrweb/dist/rrweb.min.css'
 import '@fontsource/poppins'
 import '@highlight-run/ui/styles.css'
 import './index.css'
+import './__generated/antd.css'
 import './style/tailwind.css'
 
 import { ApolloError, ApolloProvider } from '@apollo/client'
--- a/frontend/src/index.tsx
+++ b/frontend/src/index.tsx
@@ -1,8 +1,8 @@
+import './__generated/antd.css'
 import '@highlight-run/rrweb/dist/rrweb.min.css'
 import '@fontsource/poppins'
 import '@highlight-run/ui/styles.css'
 import './index.css'
-import './__generated/antd.css'
 import './style/tailwind.css'
 
 import { ApolloError, ApolloProvider } from '@apollo/client'
--- a/frontend/vite.config.ts
+++ b/frontend/vite.config.ts
@@ -4,7 +4,6 @@ import react from '@vitejs/plugin-react-swc'
 import { dirname, join } from 'path'
 import { fileURLToPath } from 'url'
 import { defineConfig, loadEnv } from 'vite'
-import vitePluginImp from 'vite-plugin-imp'
 import svgr from 'vite-plugin-svgr'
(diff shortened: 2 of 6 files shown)


frontend/src/index.tsx
```diff
@@ -1,8 +1,8 @@
-import './__generated/antd.css'
 import '@highlight-run/rrweb/dist/rrweb.min.css'
 import '@fontsource/poppins'
 import '@highlight-run/ui/styles.css'
 import './index.css'
+import './__generated/antd.css'
```
lewisl9029, line 5: There might not be any practical implications looking at the preview, but I think it makes more sense for our custom css to go after the generated antd.css so they can override them when necessary.

Vadman97: oops yeah, was trying that - will revert!

frontend/vite.config.ts
```diff
@@ -108,6 +112,17 @@ export default defineConfig(({ mode }) => {
...
+				less: {
+					javascriptEnabled: true,
+					modifyVars: {
+						hack: `true; @import "${join(
+							__dirname,
+							'src/style/AntDesign/antd.overrides.less',
+						)}";`,
+					},
+				},
+			},
```
Vadman97, line 125: @lewisl9029 i put this all back because vite dev locally would not have antd styling otherwise. should be ok, right?

Vadman97: yeah, reflame preview is good without this change (isn't affected by vite.config.ts), but `vite dev` didn't have the antd styling without both of these changes in the vite.config.ts

Vadman97: Screenshot from 2023-11-30 19-07-00

hmmm

Vadman97: this is actually how main looks right now for me, running with `./run-frontend.sh`

lewisl9029: Sorry took a while, but finally managed to repro. Had to comment out [these lines](https://github.com/lewisl9029/highlight/blob/ab02ba0554ffa62385cec6171d9b9aca82a087e6/docker/env.sh#L41-L48) for run-backend.sh to work properly after running start-infra.sh

Looking into it now.

lewisl9029: Wow, turns out, vite-plugin-imp has a preset option for antd to add imports to their css entry points:
Screenshot 2023-11-30 at 8 31 22 PM

And these overwrote classes from our prebuilt css. 🤦 

Looking at the list of libraries with built-in support, I don't think we needed it for anything other than antd previously, and we no longer need it for antd thanks to the prebuilt stylesheet, so I ended up just removing it altogether in https://github.com/highlight/highlight/pull/7255/commits/ebb92c7f65db0182958e2d7e289de40f13c2790b. 

Longer term we might add something similar natively in Reflame to help users map named imports to specific entry points automatically to better optimize their prod bundles.

Vadman97: wow, interesting! thanks for getting to the bottom of this. looks good with your changes locally, merging!

PR #7256, opened 2023-12-01
switch frontend vite to lightningcss

Vadman97:
## Summary

Update frontend CSS vite settings to match reflame more closely per recommendation from @lewisl9029 

## Testing

Reflame preview and local vite dev look the same
Screenshot from 2023-12-06 19-06-06
Screenshot from 2023-12-06 19-08-31


lewisl9029:
Made [this PR](https://github.com/highlight/highlight/pull/7257) to fix the error in the prod build (as seen in the monorepo workflow here). Didn't actually end up running into any errors in dev. Hopefully that's the only issue.

Might be worth trying a render preview to make sure prod vite build still works as expected, to make sure that backup is available if we ever need it (🤞  that we won't ofc).

ccschmitz (approved):
Nice!

PR #7259, opened 2023-12-01
Update insights lambda to handle multiple requested emails

SpennyNDaJets:
## Summary
The previously created lambda was handling one email format, but will need to respond to multiple email template requests. Allow a parameter to be passed to determine which template should be used.

Also updates the session alerts to not show unavailable data (i.e. these alerts are sent out for live sessions)

Step function test on AWS

Breaking change to previous lambda function

N/A

### DIFF
--- a/packages/react-email-templates/src/index.ts
+++ b/packages/react-email-templates/src/index.ts
@@ -1,13 +1,28 @@
 import { render } from '@react-email/render'
 import { APIGatewayEvent } from 'aws-lambda'
 import {
+	ErrorAlertEmail,
+	LogAlertEmail,
+	NewSessionAlertEmail,
+	NewUserAlertEmail,
+	RageClickAlertEmail,
 	SessionInsightsEmail,
-	SessionInsightsEmailProps,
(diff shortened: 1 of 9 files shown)


packages/react-email-templates/src/index.ts
```diff
@@ -1,13 +1,28 @@
...
+
+	const EmailTemplate = getEmailTemplate(args.template)
+	if (!EmailTemplate) {
+		return {
+			statusCode: 400,
+			body: 'Request is missing a valid "template" property',
+		}
+	}
+
+	const html = render(EmailTemplate(args.data))
```
SpennyNDaJets, line 25: Breaking change - wondering if we should redeploy this as an entirely different lambda function since it is not longer specific to the insights email?

mayberryzane (approved):

packages/react-email-templates/src/index.ts
SpennyNDaJets: Synced with Zane, and this shouldn't be an issue

PR #7276, opened 2023-12-04
Tracing support for Node and Next SDKs

deltaepsilon:
## Summary

`BatchSpanProcessorBase` doesn't set its own exporter, which makes zero sense... but whatever. We can force the issue and set `this._exporter` on our own.

A little `asyncLocalStorage` cleanup and we can reliably add `highlight.session_id` and `highlight.request_id` headers to each span's `Resource` whether OTEL wants us to or not.

Deployed to Vercel, clicked the buttons, watched the traces land on the Traces Dashboard.

n/a

n/a


sdk/highlight-next/src/util/with-highlight-nodejs-app-router.ts
```diff
@@ -1,27 +1,56 @@
...
+
+										return resolve(result)
+									} catch (error) {
+										reject(error)
+									}
+								},
+							)
+						})
+					},
+				)
```
Vadman97, line 52: as discussed offline, let's switch this to the approach of populating the trace id -> highlight ctx map and applying the highlight ctx on end of spans.

Vadman97 (changes requested):

sdk/highlight-node/src/client.ts
```diff
@@ -58,9 +105,8 @@ export class Highlight {
 	_projectID: string
 	_debug: boolean
 	otel: NodeSDK
-	private tracer: Tracer
+	tracer: Tracer
```
Vadman97, line 108: rather than exposing the underlying otel tracer and having folks call `H.tracer.startActiveSpan`, we should have a helper `H.startSpan` like we do in other languages that calls `H.tracer.startActiveSpan` with the right args

docs-content/getting-started/fullstack-frameworks/next-js/7_advanced-config.md
```diff
@@ -33,6 +33,38 @@ all-in-one.
...
+	return new Promise<void>((resolve) => {
+		highlight?.tracer.startActiveSpan('page-router-span', async (span) => {
+			console.info('Here: /pages/api/page-router-trace.ts ⌚⌚⌚')
+
+			res.send(
+				`Trace sent! Check out this random number: ${Math.random()}`,
+			)
+			span.end()
+			resolve()
+		})
```
Vadman97, line 62: update per comment below, unless you think this is the right interface for creating a span to expose to our customers

e2e/nextjs/.env
```diff
@@ -1,3 +1,3 @@
-NEXT_PUBLIC_HIGHLIGHT_PROJECT_ID='1jdkoe52'
-NEXT_PUBLIC_HIGHLIGHT_OTLP_ENDPOINT='http://localhost:4318'
-NEXT_PUBLIC_HIGHLIGHT_BACKEND_URL='https://localhost:8082/public'
+NEXT_PUBLIC_HIGHLIGHT_PROJECT_ID='1383'
+# NEXT_PUBLIC_HIGHLIGHT_OTLP_ENDPOINT='http://localhost:4318'
+# NEXT_PUBLIC_HIGHLIGHT_BACKEND_URL='https://localhost:8082/public'
```
Vadman97, line 3: this is probably what's failing the e2e test - we expect all the e2e tests to route to the local deploy

docs-content/getting-started/fullstack-frameworks/next-js/7_advanced-config.md
```diff
@@ -33,6 +33,38 @@ all-in-one.
...
+
+import { withPageRouterHighlight } from '@/app/_utils/page-router-highlight.config'
+
+export default withPageRouterHighlight(async function handler(
+	req: NextApiRequest,
+	res: NextApiResponse,
+	highlight,
+) {
+	return new Promise<void>((resolve) => {
+		highlight?.tracer.startActiveSpan('page-router-span', async (span) => {
```
deltaepsilon, line 54: Update these docs

PR #7283, opened 2023-12-05
Spenny/fix log key values ux

SpennyNDaJets:

### DIFF
--- a/backend/clickhouse/query.go
+++ b/backend/clickhouse/query.go
@@ -353,7 +353,7 @@ func expandJSON(logAttributes map[string]string) map[string]interface{} {
 	return out
 }
 
-func KeysAggregated(ctx context.Context, client *Client, tableName string, projectID int, startDate time.Time, endDate time.Time) ([]*modelInputs.QueryKey, error) {
+func KeysAggregated(ctx context.Context, client *Client, tableName string, projectID int, startDate time.Time, endDate time.Time, query *string) ([]*modelInputs.QueryKey, error) {
 	chCtx := clickhouse.Context(ctx, clickhouse.WithSettings(clickhouse.Settings{
 		"max_rows_to_read": KeysMaxRows,
 	}))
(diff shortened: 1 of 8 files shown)


frontend/src/components/Search/SearchForm/SearchForm.tsx
```diff
@@ -252,18 +237,21 @@ export const Search: React.FC<{
 	const cursorIndex = inputRef.current?.selectionStart || 0
 	const activeTermIndex = getActiveTermIndex(cursorIndex, queryTerms)
 	const activeTerm = queryTerms[activeTermIndex]
+	const debouncedKeyValue = useDebouncedValue<string>(activeTerm.value)
 
+	// TODO: code smell, user is not able to use "message" as a search key
+	// because we are reserving it for the body implicitly
```
Vadman97, line 243: hmm, is there some cleaner thing we can do about this?

backend/clickhouse/query.go
```diff
@@ -363,10 +363,15 @@ func KeysAggregated(ctx context.Context, client *Client, tableName string, proje
...
+		Where(fmt.Sprintf("Day <= toStartOfDay(%s)", sb.Var(endDate)))
+
+	if query != nil && *query != "" {
+		sb.Where(fmt.Sprintf("Key LIKE %s", sb.Var("%"+*query+"%")))
+	}
+
+	sb.GroupBy("1").
 		OrderBy("2 DESC, 1").
-		Limit(500)
+		Limit(10)
```
Vadman97, line 374: does the limit actually affect the max number of rows clickhouse scans when building the result set? limiting to the query should help with that tho

Vadman97 (approved):

frontend/src/components/Search/SearchForm/SearchForm.tsx
SpennyNDaJets: Went back and forth on this with @ccschmitz - I think the proposed solution is outlined in [this slack message](https://highlightcorp.slack.com/archives/C0454PWLLLW/p1699401258239049?thread_ts=1699390242.808739&cid=C0454PWLLLW)

frontend/src/components/Search/SearchForm/SearchForm.tsx
```diff
@@ -252,18 +237,21 @@ export const Search: React.FC<{
...
 	const showValues =
-		activeTerm.key !== BODY_KEY ||
-		!!keys?.find((k) => k.name === activeTerm.key)
+		activeTerm.key !== BODY_KEY &&
+		!!keysData?.keys?.find((k) => k.name === activeTerm.key)
```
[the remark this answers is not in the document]
SpennyNDaJets: I believe they are being logged as any other key/value pair, just under the message key. We could try an add that as a reserved column, but not clear how that works as it seems other reserved columns still have entries in the table. It does seem the body values are causing the fetching of key values to return an error, `"code: 158, message: Limit for rows (controlled by 'max_rows_to_read' setting) exceeded, max rows: 1.00 million, current rows: 2.79 million"`.

With this solution, we would no longer fetch the message key, even if users were using a custom value. I know we are addressing that as part of [this](https://highlightcorp.slack.com/archives/C0454PWLLLW/p1699401258239049?thread_ts=1699390242.808739&cid=C0454PWLLLW), but could be useful for a sync at Lacroix on the expected behavior and future behavior here

backend/clickhouse/query.go
SpennyNDaJets: Gotcha! I 'm seeing that the error is coming from this setting
https://github.com/highlight/highlight/blob/main/backend/clickhouse/query.go#L358

Seems like we wouldn't want it to error here if their were too many rows to scan?

PR #7287, opened 2023-12-06
Improve syntax highlight on query builder

ccschmitz:
## Summary

A few updates to the query builder for logs and traces:

* Fixes an issue where tags were being rendered in the wrong position
* Extends syntax highlight to highlight special characters like `>`, `<`, `>=`, `=`, and `!=` (not supported on the backend yet though)
* Updates the tag styles to match [the new mocks](https://www.figma.com/file/rdBPNzn7Klk6RG1LHUCE5B/Screen-Designs?type=design&node-id=13331%3A41232&mode=design&t=LjbC92IrAhQyaYfQ-1)

https://www.loom.com/share/567217370e064d3e9fd2c32ce5609d08

Also looked at making the tags more interactive (e.g. hover the tag to make it change color and show a large "X" to remove it) but it was [getting somewhat complicated](https://github.com/highlight/highlight/pull/7287/commits/3f887799603784d07fa043c2f8b86f151a051d47#diff-d17856357f5ac30cbd28cd4b82e9ceff41c043e4f5791bbf7b763c81d67cbf14R408-R440) and not working perfectly, so decided to ship these changes on their own for now.

I also did [some explorations](https://stackblitz.com/edit/stackblitz-starters-tn8agv?description=React%20%20%20TypeScript%20starter%20project&file=src%2FApp.tsx&title=React%20Starter) with building a new version of the search form that uses `contenteditable` so you can render anything you want to represent a tag and keep things interactive. However, there were a lot of edge cases with this that weren't immediately obvious to me how to work around, especially around managing cursor position and text selection.

N/A - all client side changes.

Will have Julian take a look at these changes in the PR preview and get feedback.


frontend/src/components/Search/SearchForm/SearchForm.tsx
Vadman97, line 1: pasting `STRIPE level:error ` and hitting enter seems to create a gap in the tags
Screenshot from 2023-12-06 13-29-36

Vadman97, line 1: should this be valid syntax? actually perhaps not, since we correctly highlight `level:!=error` and `level:!=error`
Screenshot from 2023-12-06 13-31-17

Vadman97, line 1: a thought for a future pr - do we plan to have some error states for invalid queries, like if you provide a key with no value?

Screenshot from 2023-12-06 13-32-56

Vadman97 (approved):

frontend/src/components/Search/SearchForm/SearchForm.tsx
ccschmitz: Yeah, definitely plan to show some error states in the future! Will create some cards to follow up on additional functionality like this.

ccschmitz: I think this was is resolved now! Would be curious if this works for you if you try pasting in again. Though I did notice one issue with pasting if you copy a value from the DOM and it has some rich text formatting. This could produce an extra space when pasted, which I confirmed is the default behavior of an `input`. We could add some custom handling of pasted values in the future if we decide this is an issue.

ccschmitz: Tracking in: https://github.com/highlight/highlight/issues/7298

PR #7289, opened 2023-12-06
Add trace_id and span_id as trace key suggestions

ccschmitz:
## Summary

We don't populate suggestions for `trace_id` and `span_id` in the `trace_keys` table because the results are so unique. However, this means these keys also don't show up in the typeahead for suggestions when creating filters on the traces search.

This PR manually adds `trace_id` and `span_id` to the suggestions for trace keys.

Screenshot 2023-12-07 at 11 22 37 AM

Local click test. Couldn't test in the PR preview because the backend changes aren't deployed.

Should verify functionality in prod.

N/A

### DIFF
--- a/backend/clickhouse/traces.go
+++ b/backend/clickhouse/traces.go
@@ -61,6 +61,11 @@ var traceColumns = []string{
 	"StatusMessage",
 }
 
+var defaultTraceKeys = []*modelInputs.QueryKey{
+	{Name: "trace_id", Type: modelInputs.KeyTypeString},
+	{Name: "span_id", Type: modelInputs.KeyTypeString},
+}
+
 var tracesTableConfig = tableConfig[modelInputs.ReservedTraceKey]{
 	tableName:        TracesTable,
 	keysToColumns:    traceKeysToColumns,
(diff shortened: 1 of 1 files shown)


backend/clickhouse/traces.go
```diff
@@ -471,7 +471,24 @@ func (client *Client) ReadTracesMetrics(ctx context.Context, projectID int, para
...
+		Type: modelInputs.KeyTypeString,
+	})
+	defaultTraceKeys = append(defaultTraceKeys, &modelInputs.QueryKey{
+		Name: "span_id",
+		Type: modelInputs.KeyTypeString,
+	})
+
+	traceKeys = append(traceKeys, defaultTraceKeys...)
+
+	return traceKeys, nil
```
Vadman97, line 491: just checking that this doesn't break the value recommendation when you input `trace_id:` on the frontend

backend/clickhouse/traces.go
```diff
@@ -471,7 +471,24 @@ func (client *Client) ReadTracesMetrics(ctx context.Context, projectID int, para
...
 func (client *Client) TracesKeys(ctx context.Context, projectID int, startDate time.Time, endDate time.Time) ([]*modelInputs.QueryKey, error) {
-	return KeysAggregated(ctx, client, TraceKeysTable, projectID, startDate, endDate)
+	traceKeys, err := KeysAggregated(ctx, client, TraceKeysTable, projectID, startDate, endDate)
+	if err != nil {
+		return nil, err
+	}
+
+	defaultTraceKeys := []*modelInputs.QueryKey{}
+	defaultTraceKeys = append(defaultTraceKeys, &modelInputs.QueryKey{
+		Name: "trace_id",
```
Vadman97, line 481: are there any other keys we may add here? perhaps this `defaultTraceKeys` could be populated in a loop here from some 'defaultTraceKeys' top-level var, like line 45

Vadman97 (approved):

backend/clickhouse/traces.go
ccschmitz: Cleaned up in https://github.com/highlight/highlight/pull/7289/commits/d69c5aa8940709aabecd9e515d67fb4d2b8c7cc9 - good call!

PR #7290, opened 2023-12-06
session feed report

Vadman97:
## Summary

Add feature flagged ability to export a session report in the following format:

filters | num_results |
-- | -- |
custom_created_at:between_date:2023-11-07T01:58:22.327Z_2023-12-07T01:58:22.328Z\|session_country:is:United States\|custom_sample:is_editable:496974bfe286b849 | 15 |

key | user_properties | num_sessions | num_days_visited | num_months_visited | avg_active_length_mins | max_active_length_mins | total_active_length_mins | avg_length_mins | max_length_mins | total_length_mins | location
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
vadim@highlight.run |   | 14 | 5 | 1 | 15.3720535714286 | 76 | 215.20875 | 1702.18174285714 | 13089 | 23830.5444 | California
[::1]:59662 |   | 1 | 1 | 1 | 3.40421666666667 | 3 | 3.40421666666667 | 9816.60625 | 9816 | 9816.60625 | California

secure_id | identifier | active_length | avatar | domain | email | identified_email | name |   |  
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
0K6joaR8LzamQI91t3T4n6yzjLg2 | 5 | 34998 | https://lh3.googleusercontent.com/a-/x-x=s96-c | highlight.run | vadim@example.com | true | vadim |   |  

https://www.loom.com/share/4c778566d7f74a1496ad13199a840642

No

Yes @julian-highlight let me know if you have designs for a better button
Screenshot from 2023-12-06 18-47-30

### DIFF
--- a/backend/private-graph/graph/schema.resolvers.go
+++ b/backend/private-graph/graph/schema.resolvers.go
@@ -5500,6 +5500,57 @@ func (r *queryResolver) SessionsHistogramClickhouse(ctx context.Context, project
 	}, nil
 }
 
+// SessionsReport is the resolver for the sessions_report field.
+func (r *queryResolver) SessionsReport(ctx context.Context, projectID int, query modelInputs.ClickhouseQuery) ([]*modelInputs.SessionsReportRow, error) {
+	project, err := r.isAdminInProjectOrDemoProject(ctx, projectID)
+	if err != nil {
+		return nil, err
+	}
+	workspace, err := r.GetWorkspace(project.WorkspaceID)
+	if err != nil {
+		return nil, err
+	}
+	retentionDate := GetRetentionDate(workspace.RetentionPeriod)
+
+	// If there's no admin for the context, use `admin=nil`
+	// (admin is used by the "viewed by me" filter)
+	admin, err := r.getCurrentAdmin(ctx)
+	if errors.Is(err, AuthenticationError) {
+		admin = nil
+	} else if err != nil {
+		return nil, err
+	}
+
+	ids, _, _, err := r.ClickhouseClient.QuerySessionIds(ctx, admin, projectID, 1_000_000_000, query, "ID", nil, retentionDate)
+	if err != nil {
+		return nil, err
+	}
+
+	var results []*modelInputs.SessionsReportRow
+	if err := r.DB.Raw(`
+select coalesce(email, ip, client_id, identifier)                      as key,
+       max(user_properties::text)                                      as user_properties,
+       count(*)                                                        as num_sessions,
+       count(distinct date_trunc('day', created_at))                   as num_days_visited,
+       count(distinct date_trunc('month', created_at))                 as num_months_visited,
+       avg(active_length) / 1000 / 60                                  as avg_active_length_mins,
+       max(active_length) / 1000 / 60                                  as max_active_length_mins,
+       sum(active_length) / 1000 / 60                                  as total_active_length_mins,
+       avg(length) / 1000 / 60                                         as avg_length_mins,
+       max(length) / 1000 / 60                                         as max_length_mins,
+       sum(length) / 1000 / 60                                         as total_length_mins,
+       max(case when state is not null then state || ', ' || city end) as location
+from sessions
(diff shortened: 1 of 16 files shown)


frontend/src/pages/Sessions/SessionsFeedV3/SessionQueryBuilder/components/SessionFeedConfigurationV2/SessionFeedConfigurationV2.tsx
```diff
@@ -46,10 +47,15 @@ export const DropdownMenu = function ({
...
 	})
+	const { generateSessionsReportCSV, loading: reportLoading } =
+		useGenerateSessionsReportCSV()
 
 	const showDeleteButton =
 		canDelete &&
 		workspaceSettingsData?.workspaceSettings?.enable_data_deletion !== false
+	const showReportButton =
+		workspaceSettingsData?.workspaceSettings?.enable_session_export !==
+		false
```
SpennyNDaJets, line 58: Just want to make sure that if `workspaceSettingsData` was `{}`, then we should want to show the report button?

backend/private-graph/graph/schema.resolvers.go
```diff
@@ -5500,6 +5500,57 @@ func (r *queryResolver) SessionsHistogramClickhouse(ctx context.Context, project
...
+	// If there's no admin for the context, use `admin=nil`
+	// (admin is used by the "viewed by me" filter)
+	admin, err := r.getCurrentAdmin(ctx)
+	if errors.Is(err, AuthenticationError) {
+		admin = nil
+	} else if err != nil {
+		return nil, err
+	}
+
+	ids, _, _, err := r.ClickhouseClient.QuerySessionIds(ctx, admin, projectID, 1_000_000_000, query, "ID", nil, retentionDate)
```
SpennyNDaJets, line 5524: Definite edge case, but wonder if we should limit the count to less than 1,000,000,000? I think clickhouse will only return a max of 1,000,000 rows before erroring and not sure if most programs can open more than a few hundred thousand rows

frontend/src/util/session/report.ts
```diff
@@ -0,0 +1,125 @@
...
+			return Object.values(report).filter(
+				(value) => value !== 'SessionsReportRow',
+			)
+		}),
+	]
+}
+
+const getSessionRows = (sessions: Session[]) => {
+	const rows = []
+	for (const session of sessions) {
```
SpennyNDaJets, line 29: Maybe an improvement for the future, but think it would be nice to only have the headers once instead of per row

SpennyNDaJets (approved):
Click test looks good to me

frontend/src/pages/Sessions/SessionsFeedV3/SessionQueryBuilder/components/SessionFeedConfigurationV2/SessionFeedConfigurationV2.tsx
Vadman97: good point - i think originally i thought to make this more lenient in case the workspace settings were cached and had an undefined value, but it still makes sense to be strict in that point.

frontend/src/util/session/report.ts
Vadman97: yeah, the customer actually brought this up (i didn't implement it that way since rows can have different keys), so i'll be aggregating keys for one header and leaving unset ones blank

backend/private-graph/graph/schema.resolvers.go
Vadman97: good point... perhaps we can just error if there are more results? on the backend, it's kind of an edge case since the data is aggregated by user via the postgres query, so the large number of ids is just possible in between (with too many id arguments, the postgres query won't run anyway because of the argument limits). will make this 1M and throw an error if more results are returned by the query.

Issue #7298, opened 2023-12-07
Add client-side query validation


ccschmitz:
Translate the grammar developed in #7297 so we can get client-side validation on queries, or return the validation errors from the perser on the server and render them on the client.

We should also clean up the old logic for parsing a search query and converting query terms to a search string. This logic is used in many places and the data structure for a search expression is more verbose than what we had for query terms, so it might be helpful to create some helpers to make this easier.

Issue #7297, opened 2023-12-07
Build grammar for parsing search queries


ccschmitz:
Develop a grammar for parsing search queries. This grammar will be used parsing a search query and:

1. Building a SQL statement to query for sessions, errors, logs, and traces.
2. Validating queries are formatted correctly and displaying error feedback in the client.

We will start with implementing the below grammar in our backend to parse query strings from the client and build SQL statements. This grammar was built by @mayberryzane during a pairing session with @ccschmitz.

```
lexer grammar ExprLexer;
AND : 'AND' ;
OR : 'OR' ;
NOT : 'NOT' ;
EQ : '=' ;
NEQ : '!=' ;
LT : '<' ;
LTE : '<=' ;
GT : '>' ;
GTE : '>=' ;
COMMA : ',' ;
SEMI : ';' ;
QUOT : '"' ;
LPAREN : '(' ;
RPAREN : ')' ;
LCURLY : '{' ;
RCURLY : '}' ;
ID: [a-zA-Z_0-9\-]+ ;
STRING: '"'.*?'"' ;
WS: [ \t\n\r\f]+ -> skip ;
parser grammar ExprParser;
options { tokenVocab=ExprLexer; }
search_query
  : EOF
  | search_expr EOF
  ;
col_expr
  : LPAREN col_expr RPAREN
  | col_expr AND col_expr
  | col_expr col_expr
  | col_expr OR col_expr
  | NOT col_expr
  | STRING
  | ID
  ;
search_expr 
  : LPAREN search_expr RPAREN
  | search_expr AND search_expr
  | search_expr search_expr
  | search_expr OR search_expr
  | NOT search_expr
  | ID bin_op col_expr
  ;
bin_op
  : EQ
  | NEQ
  | GT
  | GTE
  | LT
  | LTE
  ;
  
service_name=(private-graph OR public-graph) OR (status>=400 AND span_kind!=internal) name!="Zane Mayberry"
```

You can test it live at http://lab.antlr.org with a query like `service_name=(private-graph OR public-graph) OR (status>=400 AND span_kind!=internal) name!="Zane Mayberry"`.

![Image](https://github.com/highlight/highlight/assets/308182/5689d901-6ae7-42c6-83a0-70a84fe0295c)

PR #7299, opened 2023-12-07
Fix query parsing with quotes and whitespace in tags

ccschmitz:
## Summary

There's a bug in our current query parsing logic with quotes that messes up formatting unless you have matching opening/closing quotes. There's also a bug with how whitespace is rendered. This PR still isn't perfect, but the behavior is better. Also note that we are iterating on this area of the product this week to improve query parsing further as we develop our own grammar for search queries and parsers for client and server to provide a structured representation of a query (see #7297, #7298, and #7309 for more details).

### Before
https://www.loom.com/share/1062c60e94a14460b44aca8f46ee8b86

### After
https://www.loom.com/share/04fda202cd8d4dc3b5c2ca82d4c91d60

N/A

N/A

### DIFF
--- a/frontend/src/components/Search/SearchForm/utils.ts
+++ b/frontend/src/components/Search/SearchForm/utils.ts
@@ -15,7 +15,8 @@ export const BODY_KEY = 'message'
 const PARSE_REGEX =
 	/(\S+:'(?:[^'\\]|\\.)*')|(\S+:"(?:[^"\\]|\\.)*")|(-?"(?:[^"\\]|\\.)*")|(-?'(?:[^'\\]|\\.)*')|(\S+:\((?:[^\)\\]|\\.)*\))|\S+|\S+:\S+|\s$/g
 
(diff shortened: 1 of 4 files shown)


frontend/src/components/Search/SearchForm/utils.ts
```diff
@@ -15,7 +15,8 @@ export const BODY_KEY = 'message'
 const PARSE_REGEX =
 	/(\S+:'(?:[^'\\]|\\.)*')|(\S+:"(?:[^"\\]|\\.)*")|(-?"(?:[^"\\]|\\.)*")|(-?'(?:[^'\\]|\\.)*')|(\S+:\((?:[^\)\\]|\\.)*\))|\S+|\S+:\S+|\s$/g
 
-const WHITESPACE_REGEX = /\s+(?=(?:[^"]*"[^"]*")*[^"]*$)(?![^\(]*\))/g
+const WHITESPACE_REGEX =
```
mayberryzane, line 18: 🤯

mayberryzane (approved):

Issue #7309, opened 2023-12-11
Backend query parsing


ccschmitz:
Build a parser in Go for the grammar developed in #7314 to parse a query string and build a query on the backend. Use [antlr4 CLI to generate a lexer and parser in Go](https://blog.gopheracademy.com/advent-2017/parsing-with-antlr4-and-go/).

PR #7314, opened 2023-12-11
Add antlr grammar + JS parser

ccschmitz:
## Summary

Adds an [ANTLR](https://www.antlr.org/) grammar (`.g4`) file, generates some TypeScript boilerplate, and adds a custom listener + utilities for parsing a search query string and returning a structure representation of the query that can be used for displaying a query with syntax highlighting.

The functionality should actually be no different than what we currently have in production, but this paves the way for new features which will be added in #7298 and #7309.

Local + PR preview click test. Also have some unit tests.

This is all just updating the syntax highlight on searches right now, and it's only the pill rendering, nothing to do with the actual query string building logic yet, so it should be fairly low risk.

N/A - will coordinate with Julian on https://github.com/highlight/highlight/issues/7298.


frontend/.eslintrc
```diff
@@ -30,7 +30,11 @@
 			"jsx": true
 		}
 	},
-	"ignorePatterns": ["src/graph/generated/*", "src/__generated/*"],
+	"ignorePatterns": [
+		"src/graph/generated/*",
+		"src/__generated/*",
+		"src/components/Search/Parser/*"
```
ccschmitz, line 36: Ignored these files from linting because they are (mostly) generated.

frontend/src/components/Search/utils.test.ts
```diff
@@ -0,0 +1,177 @@
+import SearchGrammarParser from '@/components/Search/Parser/SearchGrammarParser'
+import { parseSearch, STRING_TYPE } from '@/components/Search/utils'
+
+describe('parseSearch', () => {
+	it('should parse a string correctly', () => {
```
SpennyNDaJets, line 5: There seems to be a weird case where certain characters do not show up in the input: `@$%&` and maybe a few others

SpennyNDaJets (approved):
Looks good, but maybe a small regression with some characters being hidden? (see comment below)

frontend/src/components/Search/utils.test.ts
ccschmitz: Addressed that in 6c6eb6a2d7c7564c8e4b90cee486bb94590718ee - thanks for catching!

PR #7319, opened 2023-12-12
Support python requests autoinstrumentation

SpennyNDaJets:
## Summary
Add tracing auto-instrumentation for the requests python library

1) Start the e2e flask app
```
cd sdk/highlight-py/e2e/highlight_flask
poetry build
poetry run flask run
```
2) Hit the app via curl
```
curl -H "x-highlight-request: bs1/br1" http://localhost:5000
```
3) Confirm that the `GET` trace for the requests library is created and linked to the same trace as the parent span

Screenshot 2023-12-12 at 1 46 17 PM

N/A

N/A


sdk/highlight-py/e2e/highlight_flask/app.py
```diff
@@ -1,29 +1,37 @@
...
 H = highlight_io.H(
     "1",
-    integrations=[FlaskIntegration()],
+    integrations=[FlaskIntegration(), RequestsIntegration()],
     instrument_logging=True,
     otlp_endpoint="http://localhost:4318",
     service_name="my-flask-app",
     service_version="1.0.0",
     environment="e2e-test",
+    tracing_origins=["localhost", "highlight.io"],
```
Vadman97, line 21: does this option fit better in the `RequestsIntegration()` itself? that makes more sense to me since that's what is actually able to trace a request (outgoing http request), and if there is a future other external service integration (like `GRPCIntegration()`) that may have a different set of tracing origins

i propose moving the option into the ingration itself like so: `RequestsIntegration(tracing_origins=[...])`

sdk/highlight-py/highlight_io/sdk.py
```diff
@@ -44,6 +46,7 @@ class H(object):
     OTLP_HTTP = "https://otel.highlight.io:4318"
     _instance: "H" = None
     _logging_instrumented = False
+    _context_map = LRUCache(1000)
```
Vadman97, line 49: a comment here would be useful for the future why we use an LRU-expiring map here

sdk/highlight-py/highlight_io/integrations/requests.py
Vadman97, line 1: could we add a test for this file?

sdk/highlight-py/e2e/highlight_flask/app.py
SpennyNDaJets: That makes sense to me. I was originally thinking that these tracing origins would be across multiple integrations to avoid duplicative code if you were using multiple libraries to make requests, but could see why you may want a different list for grpc vs http.

r.e. out conversation of having these library auto-instrumentations on by default - my thought is:
1. If not included in `integrations` list -> RequestsInstrumentation is `on` with no tracing_origins
2. If added to exclusions list -> RequestsInstrumentation is `off`
3. If added to `integrations` list with tracing origins -> use those integration parameters and leave out the default integration

sdk/highlight-py/highlight_io/sdk.py
```diff
@@ -126,6 +137,15 @@ def __init__(
...
             integration.enable()
+            try:
+                key = integration.INTEGRATION_KEY
+                self._disabled_integrations.append(key)
+            except AttributeError:
+                pass
+
+        for [key, integration] in DEFAULT_INTEGRATIONS:
+            if key not in self._disabled_integrations:
+                integration().enable()
```
Vadman97, line 148:
```suggestion
        for integration in DEFAULT_INTEGRATIONS:
                    if integration.INTEGRATION_KEY not in self._disabled_integrations:
                        integration().enable()
```

sdk/highlight-py/highlight_io/sdk.py
```diff
@@ -21,6 +21,11 @@
 from opentelemetry.trace import INVALID_SPAN
 
 from highlight_io.integrations import Integration
+from highlight_io.utils.lru_cache import LRUCache
+from highlight_io.integrations.requests import RequestsIntegration
+
+
+DEFAULT_INTEGRATIONS = [[RequestsIntegration.INTEGRATION_KEY, RequestsIntegration]]
```
Vadman97, line 28: make INTEGRATION_KEY an attribute of the `Integration` class and have every integration override it

sdk/highlight-py/pyproject.toml
```diff
@@ -31,13 +31,14 @@ django = { version = ">=3", optional = true }
...
-opentelemetry-proto = "1.20.0"
-opentelemetry-sdk = "1.20.0"
+opentelemetry-api = "1.21.0"
+opentelemetry-distro = { extras = ["otlp"], version = "0.42b0" }
+opentelemetry-exporter-otlp-proto-http = "1.21.0"
+opentelemetry-instrumentation = "0.42b0"
+opentelemetry-instrumentation-logging = "0.42b0"
+opentelemetry-proto = "1.21.0"
+opentelemetry-sdk = "1.21.0"
+opentelemetry-instrumentation-requests = "0.42b0"
```
Vadman97, line 41: in a future pr, we can break these auto instrumentations into their own extras that determine whether the instrumentation is auto enabled. no need to do that here tho

Vadman97 (approved):

sdk/highlight-py/highlight_io/integrations/requests.py
```diff
@@ -0,0 +1,39 @@
+import typing as t
+
+import highlight_io
+from highlight_io.integrations import Integration
+from highlight_io.utils import utils
+
+try:
+    import requests
```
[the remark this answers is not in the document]
Vadman97: should be fine, might behave the same as importing `opentelemetry.instrumentation.requests` but good to be explicit here

Vadman97 (approved):

PR #7326, opened 2023-12-13
ensure batched workers flush correctly if no new messages are arriving

Vadman97:
## Summary

A bug in the batched queue flush logic is causing the last message to never get flushed.

- datasync batch worker receives the error object update
- datasync batch worker sets the last message timestamp and waits for more messages to flush
- on the next kafka receive(), there are no more messages so the call returns nil
- we return without ever calling flush()

The change ensures we keep trying to flush (to purge previously received messages) even if we stop receiving a message.
The change also sets a receive timeout in the batched worker to wait for no more than the flush timeout
for a new message, to avoid delaying flushes.

e2e test

No

No


SpennyNDaJets (approved):
LGTM

PR #7331, opened 2023-12-14
add jira issue type selection back to modal

### LINKED ISSUE #7258: Jira integration bug
**Describe the bug**
The Jira integration is not working. It outputs this error message in console:

> Error: error creating Jira task: Jira API responded with error; status_code=400 Bad Request; body={"errorMessages":[],"errors":{"issuetype":"The issue type selected is invalid."}}

**To Reproduce**

1. Go to 'Create issue'
2. Click on 'jira issue'
3. Click submit

**Expected behavior**
It should send the ticiket to Jira.

I think there is a misconfiguration or your end, about the issueId.

mayberryzane:
## Summary
- issue type selection was removed here: https://github.com/highlight/highlight/pull/6754/commits/73f443a874997d1bbf9dc6df2839c20971fb51c1
- adding it back because the assumption that all Jira projects would have a default "task" type was incorrect
- fixes #7258 

- clicktested locally
- https://www.loom.com/share/afb1196901bb476993a660c763d9d95a

- no

- no


backend/private-graph/graph/schema.resolvers.go
```diff
@@ -1990,7 +1999,11 @@ func (r *mutationResolver) CreateErrorComment(ctx context.Context, projectID int
 
 			errorComment.Attachments = append(errorComment.Attachments, attachment)
 		} else if *s == modelInputs.IntegrationTypeJira {
-			if err := r.CreateJiraTaskAndAttachment(ctx, workspace, attachment, title, desc, *issueTeamID); err != nil {
+			if issueTeamID == nil || issueTypeID == nil {
```
SpennyNDaJets, line 2002: I noticed that `issueTeamID` is getting defaulted to an empty string in the request, whereas `issueTypeID` is defaulted to undefined. Is the case where it is an empty string a relevant edgecase?

Seems like these two variables should behavior similarly in the ts files, but not blocking

SpennyNDaJets (approved):
LGTM - left one question about the code, then think we just need to merge/rebase main with the flaky spec fix

backend/private-graph/graph/schema.resolvers.go
mayberryzane: I think changing `issueTeamID` to default to undefined makes sense since the backend has some logic to default in case it's `nil` but doesn't do anything to check the empty string, but it does look like there's some possible panics where we dereference it. Will take a look at this

PR #7341, opened 2023-12-15
Infinite scrolling for traces

SpennyNDaJets:
## Summary
Update the traces table to have an infinite scroll

Note: I also removed the "backwards" infinite scroll on the logs page since this it is not really used in favor of the "Load more" option

1. View the traces page
2. Scroll down to the bottom of the table
- [ ] More traces loaded
- [ ] Can select a span/trace
- [ ] Can load more when polled

https://www.loom.com/share/ff7c9ac5b00e4f468289f154df0efba7?sid=f4efff9d-c624-463d-9d5b-d9757b72b1d3

N/A

N/A

### DIFF
--- a/frontend/src/pages/Traces/TraceLogs.tsx
+++ b/frontend/src/pages/Traces/TraceLogs.tsx
@@ -35,38 +35,27 @@ export const TraceLogs: React.FC = () => {
 	const { projectId } = useProjectId()
 	const [query, setQuery] = useState('')
 
-	const {
-		logEdges,
-		loading,
-		error,
-		loadingAfter,
-		fetchMoreForward,
-		fetchMoreBackward,
-	} = useGetLogs({
-		query,
-		project_id: projectId,
-		logCursor: undefined,
-		startDate,
-		endDate,
-	})
+	const { logEdges, loading, error, loadingAfter, fetchMoreForward } =
+		useGetLogs({
+			query,
+			project_id: projectId,
+			logCursor: undefined,
+			startDate,
+			endDate,
+		})
 
 	const fetchMoreWhenScrolled = React.useCallback(
-		(
-			containerRefElement?: HTMLDivElement | null,
-			disableBackwards?: boolean,
-		) => {
+		(containerRefElement?: HTMLDivElement | null) => {
 			if (containerRefElement) {
 				const { scrollHeight, scrollTop, clientHeight } =
 					containerRefElement
 
 				if (scrollHeight - scrollTop - clientHeight < 100) {
 					fetchMoreForward()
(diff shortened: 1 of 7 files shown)


frontend/src/pages/Traces/TraceLogs.tsx
```diff
@@ -35,38 +35,27 @@ export const TraceLogs: React.FC = () => {
...
-			disableBackwards?: boolean,
-		) => {
+		(containerRefElement?: HTMLDivElement | null) => {
 			if (containerRefElement) {
 				const { scrollHeight, scrollTop, clientHeight } =
 					containerRefElement
 
 				if (scrollHeight - scrollTop - clientHeight < 100) {
 					fetchMoreForward()
-				} else if (!disableBackwards && scrollTop === 0) {
```
SpennyNDaJets, line 64: Removed this behavior for logs because it resulted in extra unused requests, and we favor the load more via the "X new logs" button

frontend/src/pages/Traces/TracesPage.tsx
```diff
@@ -109,37 +107,19 @@ export const TracesPage: React.FC = () => {
 			fetchPolicy: 'cache-and-network',
 		})
 
-	const [moreDataQuery] = useGetTracesLazyQuery({
-		fetchPolicy: 'network-only',
-	})
-
-	const { numMore: numMoreTraces, reset: resetMoreTraces } = usePollQuery<
```
SpennyNDaJets, line 116: Removed and moved into the `useGetTraces`

frontend/src/pages/Traces/useGetTraces.ts
SpennyNDaJets, line 1: Largely copied over from the `useGetLogs`

mayberryzane (approved):

frontend/src/pages/Traces/TraceLogs.tsx
ccschmitz: Maybe worth discussing if/when we would want to have this UX in the future. If we are going to rely on the "X more traces/logs" feature instead of this, maybe we want to just remove the `fetchMoreBackward` methods (and not add it in `useGetTraces`)?

ccschmitz (commented):
Saw @mayberryzane already reviewed, but wanted to leave a thought re background infinite scroll - maybe something we can clean up altogether?

Looks great otherwise!

frontend/src/pages/Traces/TraceLogs.tsx
SpennyNDaJets: I agree, right now this doesn't seem to be behaving as expected and was creating tech debt on some of the infinite loading side. So definitely worth a discussion on the behavior here

PR #7344, opened 2023-12-16
add extra whitespace to fix node.js code highlighting

mayberryzane:
## Summary
- these specific scripts are being passed to the `HighlightCodeBlock` as a react component rather than text, not sure why this is but it's rendered as `[object Object]`
- adding extra whitespace is a workaround 

- local docs build

- no

- no


Vadman97 (commented):
🙇

Vadman97 (approved):

PR #7349, opened 2023-12-18
chunk push payload messages

Vadman97:
## Summary

Chunk data for kafka to reduce message size (and reduce the fetch size of a consumer).

Local deploy.

Monitoring kafka.

No

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -3,7 +3,6 @@ on:
     push:
         branches:
             - 'main'
-            - 'vadim/chunk-push-payload'
         paths:
             - 'go.work'
             - 'go.work.sum'
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -3,6 +3,7 @@ on:
     push:
         branches:
             - 'main'
+            - 'vadim/chunk-push-payload'
         paths:
             - 'go.work'
             - 'go.work.sum'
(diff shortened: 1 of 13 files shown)


.github/workflows/deploy.yml
```diff
@@ -3,6 +3,7 @@ on:
     push:
         branches:
             - 'main'
+            - 'vadim/chunk-push-payload'
```
mayberryzane, line 6: should delete?

backend/public-graph/graph/resolver.go
```diff
@@ -2494,8 +2514,9 @@ func (r *Resolver) ProcessPayload(ctx context.Context, sessionSecureID string, e
 	if sessionSecureID == "" {
 		return e.New("ProcessPayload called without secureID")
 	}
-	sessionObj := &model.Session{}
-	if err := r.DB.WithContext(ctx).Where(&model.Session{SecureID: sessionSecureID}).Limit(1).Take(&sessionObj).Error; err != nil {
+	var err error
+	var sessionObj *model.Session
+	if sessionObj, err = r.Store.GetSessionFromSecureID(ctx, sessionSecureID); err != nil {
```
mayberryzane, line 2519: is it ok to use the cached version here? I would have thought there were some session attributes that could have changed if we're processing many payloads for the same session

mayberryzane (approved):

backend/public-graph/graph/resolver.go
Vadman97: turned out is was not - though i didn't get to the bottom of what was important to not be cached (perhaps this would sometimes return `nil` after an initialize session?). reverted.

PR #7351, opened 2023-12-18
Allow autojoin email input

SpennyNDaJets:
## Summary
Update the workspace settings to allow users to input email domains they want to auto-approve for their workspace

1) Try to create a new workspace
- [x] Checkbox is displayed to allow autojoin emails from the admin's domain
- [x] Workspace is created correctly with the option selected
2) Go to the workspace settings
- [x] Checkbox displayed to allow auto-approve email domains
- [x] Select displayed to change the auto-approved email domains
- [x] Can select domains
- [x] Can create new domain options
- [x] Allowing auto-approved email domains will default to admin's email domain
- [x] Consistent across reloads and pages (i.e. "Members" and "properties"

https://www.loom.com/share/5948481d3e4241c4bbce61587d0e3a1a?sid=04b1be01-2da0-4601-be66-cd6bfc2f8817

N/A

Sure

### DIFF
--- a/frontend/src/pages/NewProject/AutoJoinInput.module.css
+++ b/frontend/src/pages/NewProject/AutoJoinInput.module.css
@@ -1,0 +1,21 @@
+.container {
+	align-items: flex-start;
+	row-gap: var(--size-xxSmall);
+	display: flex;
+	flex-direction: column;
+	justify-content: space-between;
+	max-height: fit-content;
+	min-height: var(--size-xxLarge);
+}
+
+.select {
+	&:global(.ant-select-disabled .ant-select-selector) {
+		background-color: transparent !important;
+	}
+
+	:global(.ant-select-selector) {
+		padding: 0 !important;
+		border: 0px solid var(--color-gray-300) !important;
+		background-color: transparent !important;
+	}
+}
--- a/frontend/src/pages/WorkspaceSettings/WorkspaceSettings.tsx
+++ b/frontend/src/pages/WorkspaceSettings/WorkspaceSettings.tsx
@@ -2,7 +2,7 @@ import Alert from '@components/Alert/Alert'
 import { FieldsBox } from '@components/FieldsBox/FieldsBox'
 import { AdminRole } from '@graph/schemas'
 import { Box } from '@highlight-run/ui/components'
-import AutoJoinForm from '@pages/WorkspaceTeam/components/AutoJoinForm'
+import { AutoJoinForm } from '@pages/WorkspaceTeam/components/AutoJoinForm'
 import { Authorization } from '@util/authorization/authorization'
 import React from 'react'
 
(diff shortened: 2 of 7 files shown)


frontend/src/pages/NewProject/AutoJoinInput.module.css
```diff
@@ -0,0 +1,21 @@
...
+	&:global(.ant-select-disabled .ant-select-selector) {
+		background-color: transparent !important;
+	}
+
+	:global(.ant-select-selector) {
+		padding: 0 !important;
+		border: 0px solid var(--color-gray-300) !important;
+		background-color: transparent !important;
+	}
+}
```
Vadman97, line 21: i'm guessing this syntax might be causing the discrepancy with reflame? see https://github.com/highlight/highlight/pull/7254

frontend/src/pages/WorkspaceSettings/WorkspaceSettings.tsx
```diff
@@ -2,7 +2,7 @@ import Alert from '@components/Alert/Alert'
 import { FieldsBox } from '@components/FieldsBox/FieldsBox'
 import { AdminRole } from '@graph/schemas'
 import { Box } from '@highlight-run/ui/components'
-import AutoJoinForm from '@pages/WorkspaceTeam/components/AutoJoinForm'
+import { AutoJoinForm } from '@pages/WorkspaceTeam/components/AutoJoinForm'
```
Vadman97, line 5: does it make sense to have the auto join checkbox on the settings page? seems redundant to me with having the email inputs, since if i want to turn it off, i can just remove all the email domains from the list.

Vadman97 (approved):
looks great! just thinking the checkbox on the settings page is a bit redundant (we can just use the email input deletion to turn off auto join domains). @julian-highlight what do you think of the UX?

frontend/src/pages/WorkspaceSettings/WorkspaceSettings.tsx
SpennyNDaJets: Synced with Julian here:

> The reason we put the check-box there is to easily opt-out of this. I believe that this won’t be unchecked/checked quite frequently, therefor I don’t think it will be confusing. In the unchecked state we could maybe show the text that we have in the callout on the sign up process. The checked state could show the input instead

Personally, I like always having the select showing, but don't feel too strongly about the checkbox if we have an easy way to clear the select. The checkbox also adds some context, so want to callout that if its removed we should replace with some text describing what the select input controls.

frontend/src/pages/NewProject/AutoJoinInput.module.css
SpennyNDaJets: I think the select looks inline with vite and reflame 🤔 but can take another look.

Lewis did some research on some the discrepancies [here](https://highlightcorp.slack.com/archives/C04DVJJ0ZJ6/p1703025239265259), and agree that I think this is a regression on the vite side of things, and reflame is handling our css better (correctly?)

PR #7352, opened 2023-12-18
Fix paginated Clickhouse queries

ccschmitz:
## Summary

Fixes an issue where paginated Clickhouse queries were returning the wrong value for `hasNextPage`, which was preventing users from being able to scroll and view more logs/traces.

The problem is that our logic for determining if there is more than one page was to see if there was 1 log more than the number we return for the query (51 total). However, this would often break because our paginated queries would return _more_ than 51 and our check was that we return _exactly_ 51. In this case `hasNextPage` should be `true`, which was fixed by changing that check.

The reason we are returning more than 51 items is because we have a number of instances where we have more than one record with the same timestamp + UUID. This may be a separate issue we need to dig into, but the only `LIMIT` we were setting on the query was on an inner query which was being used on an `IN` clause. This PR also sets the `LIMIT` on the outer `SELECT` to ensure we aren't over fetching if we have >51 items that match the inner query.

I was never able to repro the issue locally, but I confirmed that logs and traces still worked as expected in my local click tests and added a failing test that I was able to get to pass with the new logic.

Need to verify the bug is fixed in prod and notify users who reported being impacted by the bug.

Nope!

### DIFF
--- a/backend/clickhouse/cursor.go
+++ b/backend/clickhouse/cursor.go
@@ -47,18 +47,18 @@ func getConnection[T interface{}](edges []*Edge[T], pagination Pagination) *Conn
 		}
 	} else if pagination.After != nil && len(*pagination.After) > 1 {
 		hasPreviousPage = true // implicitly true because the passed in cursor should match
-		if len(edges) == LogsLimit+1 {
+		if len(edges) >= LogsLimit+1 {
 			hasNextPage = true
 			edges = edges[:len(edges)-1]
 		}
 	} else if pagination.Before != nil && len(*pagination.Before) > 1 {
 		hasNextPage = true // implicitly true because the passed in cursor should match
-		if len(edges) == LogsLimit+1 {
+		if len(edges) >= LogsLimit+1 {
 			hasPreviousPage = true
 			edges = edges[1 : len(edges)-1]
 		}
 	} else {
-		if len(edges) == LogsLimit+1 { // has forward page
+		if len(edges) >= LogsLimit+1 { // has forward page
 			hasNextPage = len(edges) == LogsLimit+1
 			edges = edges[:LogsLimit]
 		}
--- a/backend/clickhouse/logs_test.go
+++ b/backend/clickhouse/logs_test.go
@@ -327,7 +327,7 @@ func TestReadLogsHasNextPage(t *testing.T) {
 	assert.Len(t, payload.Edges, 50)
 	assert.False(t, payload.PageInfo.HasNextPage)
 
-	// Add another row to have >50 rows + duplicates
+	// Add another row to have >50 rows
 	assert.NoError(t, client.BatchWriteLogRows(ctx, []*LogRow{NewLogRow(now, 1)}))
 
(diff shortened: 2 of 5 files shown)


backend/clickhouse/cursor.go
```diff
@@ -47,18 +47,18 @@ func getConnection[T interface{}](edges []*Edge[T], pagination Pagination) *Conn
...
+		if len(edges) >= LogsLimit+1 {
 			hasNextPage = true
 			edges = edges[:len(edges)-1]
 		}
 	} else if pagination.Before != nil && len(*pagination.Before) > 1 {
 		hasNextPage = true // implicitly true because the passed in cursor should match
-		if len(edges) == LogsLimit+1 {
+		if len(edges) >= LogsLimit+1 {
 			hasPreviousPage = true
 			edges = edges[1 : len(edges)-1]
```
ccschmitz, line 58: Was also wondering about this line and if we shouldn't be trimming on the front and end of the `edges` 🤔 

```suggestion
			edges = edges[1:]
```

backend/clickhouse/cursor.go
```diff
@@ -47,18 +47,18 @@ func getConnection[T interface{}](edges []*Edge[T], pagination Pagination) *Conn
 		}
 	} else if pagination.After != nil && len(*pagination.After) > 1 {
 		hasPreviousPage = true // implicitly true because the passed in cursor should match
```
ccschmitz, line 49: Technically, these changes to the conditional logic here shouldn't be necessary with the `LIMIT` on the outer query, but I didn't see how they hurt. @Vadman97 let me know if you think we should keep them around or revert.

ccschmitz:
Here is an example where I was seeing duplicate UUIDs on logs:

Screenshot 2023-12-19 at 3 55 58 PM

You can play with it in Clickhouse here: https://console.clickhouse.cloud/services/838a9788-b028-4494-90a1-78b281f531fc/query/b777e0f2-7805-430a-a9c3-f58fd570917a

@Vadman97 I assume we should never have logs with duplicate UUIDs, correct? Think we should create another ticket to investigate that?

backend/clickhouse/cursor.go
Vadman97: sounds good to keep them!

Vadman97 (approved):
sounds good! will try to repro this in prod with the example i had before a well once this is merged

Vadman97:
@ccschmitz re the duplicates, yeah - i believe this is from reingesting logs from kafka and writing them again to clickhouse, since we don't deduplicate by the UUID with the mergetree. mind creating a ticket for that ( actually thought I already had one, but I guess not ).

ccschmitz:
@Vadman97 just created a ticket to track that: https://github.com/highlight/highlight/issues/7366 - thanks for the review!

backend/clickhouse/cursor.go
ccschmitz: Not going to change this for now. Doesn't actually seem to be causing issues in the tests I was looking at.

PR #7357, opened 2023-12-19
add errors and sessions to grafana plugin

### LINKED ISSUE #7310: support logs, errors, sessions in the grafana plugin
support for:

1. [x] logs
2. [ ] errors count, group by error group, number of unique browsers, environment
3. [ ] sessions

mayberryzane:
## Summary
- add support for error and session metrics by adding two new views which join errors/sessions with their fields to mirror the format for logs / traces
  - downside to this is worse performance for projects with large amounts of session fields since all fields need to be loaded
- for errors, only support count since none of the other error columns are numeric
- closes #7310 
https://www.loom.com/share/a668cdcc044a40b7bfb9b4a39bb31ab2

- click tested locally

- devs will need to run `docker pull clickhouse/clickhouse-server` locally, else the migration will fail due to the `splitby_max_substrings_includes_remaining_string` setting

- no


.github/workflows/backend.yml
```diff
@@ -104,7 +104,7 @@ jobs:
                 ports:
                     - 5432:5432
             clickhouse:
-                image: clickhouse/clickhouse-server:23.5.1.3174
+                image: clickhouse/clickhouse-server:23.11.2.11
```
Vadman97, line 107: should we just use latest as we do with all other images? our clickhouse cloud is always upgraded to latest anyways

Vadman97 (approved):
nice!

.github/workflows/backend.yml
mayberryzane: sounds good, especially since this is just for testing

PR #7360, opened 2023-12-19
Sitemap fix

Abhishek-More:
## Summary

Sitemap is currently returning a server error. On localhost, the error message states
```
Error: unable to determine transport target for "@highlight-run/pino"
```

Inspecting localhost:4000/sitemap.xml

No

No


Vadman97:
i made some changes to the pino transport that may fix it - @Abhishek-More can you try rebase your branch on main and see if there is something else in the sitemap fetching that causes the request to time out?

Abhishek-More:
Will do

PR #7371, opened 2023-12-20
WIP: use relative times in search features

SpennyNDaJets:


frontend/src/pages/Player/RightPlayerPanel/components/NetworkResourcePanel/NetworkResourceLogs.tsx
```diff
@@ -67,6 +67,7 @@ export const NetworkResourceLogs: React.FC<{
 		logCursor: undefined,
 		startDate,
 		endDate,
+		disablePolling: true,
```
SpennyNDaJets, line 70: Turned off polling here - I don't think we would ever want it with the set dates?

PR #7372, opened 2023-12-20
Use grammar for query parsing on server

ccschmitz:
## Summary

Uses the ANTLR grammar defined in #7314 to build a parser in Go and uses the results from the parser to build the SQL query for Clickhouse queries. Also adds support for additional numeric comparison operators like `>`, `<=`, etc.

The grammar was also extended so we could get more information about what the different parts of a `col_expr` are.

Since we are using the ANTLR grammar for more than just client code, I also moved it to a top-level `antlr` directory and added scripts in `./package.json` for generating all the language-specific code.

This also rewrites the display logic for rendering the query in the client and breaking up the tokens into filter groups. It also adds some basic client-side validation errors when queries are formatted incorrectly and we encounter a parsing error. There will be more follow ups on these validation errors in #7442.

### Breaking Changes

* If there were multiple filters applied on the same key these would be merged together into an `OR` filter. This is now an `AND`, and the new `(... OR ...)` syntax is required. e.g. `level:warn level:error` now needs to be `level:(warn OR error)`.
* If you use a comparison operator in the value of your query you will need to wrap it in quotes. e.g. `service:foo:bar` used be interpreted as `service = "foo:bar"`, but now you need to explicitly wrap `foo:bar` in quotes or it can't be parsed correctly.
* The syntax for applying a `NOT` clause has changed from adding a `-` prefix (`service:-frontend`) to using `!=` (`service!=frontend`) as the comparison operator.

Local click test. Can't be tested in a PR preview because of the backend changes. Also analyzed the SQL produced by this new logic vs the old to ensure results were the same, aside from the new syntax that is supported.

Being careful to ensure the SQL produced matches what we already had, aside from supporting the new syntax, but should still monitor query performance and confirm behavior the best we can in production via click testing.

Nope!


backend/clickhouse/logs_test.go
```diff
@@ -656,14 +659,16 @@ func TestReadLogsWithKeyFilter(t *testing.T) {
...
 	payload, err = client.ReadLogs(ctx, 1, modelInputs.QueryInput{
 		DateRange: makeDateWithinRange(now),
-		Query:     `service:"image processor" service:"different processor"`,
+		Query:     `service:("image processor" OR "different processor")`,
 	}, Pagination{})
 	assert.NoError(t, err)
 	assert.Len(t, payload.Edges, 2)
 
+	// TODO: Do we want to support this syntax still, or should we force people to
+	// wrap this in a string? `service:"foo:bar"`` if contains a bin_op separator?
```
ccschmitz, line 668: @Vadman97 what do you think about this change? I can probably do something with an error message saying that comparison operators need to be wrapped in quotes or something. Or maybe @mayberryzane has an idea on how to evolve the search grammar to preserve the existing behavior 🤔

backend/parser/listener/listener.go
```diff
@@ -0,0 +1,211 @@
...
+	s.currentOp = "="
+}
+func (s *searchListener) ExitBody_search_expr(ctx *parser.Body_search_exprContext) {}
+
+// There is an issue here when grouping clauses together that contain a LIKE.
+// e.g. a:b* a!=bc a!=bb will result in a query like:
+// SELECT * FROM t WHERE ((a LIKE 'b%' AND a <> 'bc') AND a <> 'bb')
+// Commented out for now, but we will probably want to support something like
+// this in the future.
+// TODO: Review w/ Zane.
```
ccschmitz, line 96: @mayberryzane wanted to look at this one with you. Seems like we want this grouping from `ExitAnd_search_expr`, but only when we encounter an explicit `AND` clause - would you agree?

ccschmitz: I ended up implementing this in e2add65322c380ebfef321bc089a34d319f2bf3e - let me know if you think that's not the right approach!

backend/clickhouse/logs_test.go
Vadman97: should be fine as long as the input `colon_delimited:foo:bar` shows an error to the user. perhaps the error / ui could make it clear that wrapping the value in `"` would escape any special characters?

backend/clickhouse/query.go
```diff
@@ -238,121 +248,17 @@ func makeSelectBuilder[T ~string](config tableConfig[T], selectStr string, selec
 		}
 	}
 
-	appendWhereConditions(sb, config, params.Query)
-
-	return sb, nil
-}
-
-func appendWhereConditions[T ~string](sb *sqlbuilder.SelectBuilder, config tableConfig[T], query string) {
```
Vadman97, line 246: make `AssignSearchFilters` accept a template `[T ~string]` and the `tableConfig` type a public `TableConfig`

Vadman97 (commented):
few more bugs i've spotted in a clicktest, tho this is working great overall for me
https://www.loom.com/share/fe4bf3274e6d46f7bd17e3114053ba21

backend/parser/listener/listener.go
```diff
@@ -0,0 +1,209 @@
...
+				s.rules = append(s.rules, "hasTokenCaseInsensitive("+s.tableConfig.BodyColumn+", "+s.sb.Var(v)+")")
+			}
+		}
+
+		return
+	}
+
+	filterKey, ok := s.tableConfig.KeysToColumns[T(s.currentKey)]
+	if !ok {
+		filterKey = fmt.Sprintf("%s['%s']", s.attributesColumn, s.currentKey)
```
Vadman97, line 169: is `currentKey` is user provided, does this mean that we are potentially injecting the raw user input into the SQL without escaping it?

Vadman97 (approved):
🎉

backend/parser/listener/listener.go
ccschmitz: Ah, yeah, now I understand why we needed some of our existing logic for querying on custom attribute keys. Resolved in https://github.com/highlight/highlight/pull/7372/commits/55d0d8cb276d797229e15d2319e394dbdf2673bc.

frontend/src/components/Search/Parser/listener.ts
```diff
@@ -1,38 +1,98 @@
...
+	}
+
+	exitBody_search_expr = (_ctx: Body_search_exprContext) => {
+		this.currentExpression.value = this.currentExpression.text
+		this.expressions.push(this.currentExpression)
+		this.currentExpression = { ...DEFAULT_EXPRESSION }
+	}
+
+	enterEveryRule(ctx: ParserRuleContext): void {
+		console.log('::: enterEveryRule', ctx.getText(), ctx)
```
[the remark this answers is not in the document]
ccschmitz: Thanks - cleaned up now!

ccschmitz:
> few more bugs i've spotted in a clicktest, tho this is working great overall for me

Thanks for sharing, Vadim! I think I addressed everything you ran into + your other comments. Should be ready for another pass if you want to give a final click test, otherwise I'll plan on merging this in tomorrow.

Vadman97 (commented):
got this error locally typing in `level:info OR level:error`

Screenshot 2024-01-17 at 9 42 49 AM

https://localhost:3000/1/logs?query=level%3Ainfo%20OR%20level%3Aerror%20 consistently crashes

Vadman97:
managed to get into a broken state with a trailing AND with the same traceback as above
https://localhost:3000/1/logs?query=level%3Ainfo%20container_name%3D%28%22%2Fcollector%22%20OR%20%22foo%22%29%20AND

Vadman97 (approved):
looks good! feel free to ship this. outlining the last bugs i've noticed
https://www.loom.com/share/19857dc51300453898b55274df47b81f?focus_title=1&muted=1&from_recorder=1

ccschmitz:
@Vadman97 addressed the issues you came across - thank you for click testing and reporting! Will merge once I get a green build ✅

PR #7383, opened 2023-12-21
Changelog 27 first draft

deltaepsilon:
## Summary

image


Vadman97 (approved):
nice! could also add the next tracing upgrades to be released with https://github.com/highlight/highlight/pull/7380

I'm still finalizing those docs, so I figured we'd save the Tracing stuff until after Christmas?

PR #7387, opened 2023-12-21
update dependencies

Vadman97:
## Summary

Update dependencies based on dependabot alerts.

E2E test

No (new SDKs affected will be released separately)

No


SpennyNDaJets (changes requested):
I think there may be some breaking changes with antd? Looks like changing tabs is causing some weird behavior (See https://preview.highlight.io/1/settings/general) - happy to take a look/create a card/leave out this upgrade

Edit: Possibly in [4.23.0](https://github.com/ant-design/ant-design/blob/4.x-stable/CHANGELOG.en-US.md#4230)

Screenshot 2023-12-22 at 10 36 14 AM

Vadman97:
thanks for catching that @SpennyNDaJets ! couldn't figure out how to fix it even with the new tabs `items=` prop, so i've gone down to the previous minor version.

SpennyNDaJets (approved):
Click test looks good to me

PR #7388, opened 2023-12-22
get Grafana plugin ready for publishing

mayberryzane:
## Summary
- rename grafana plugin's folder per convention
- support querying demo project w/o oauth credentials
- create a provisioning datasource.yaml config to set up data source for the demo project for testing
- fill out README

- built the plugin and ran the grafana plugin validator https://github.com/grafana/plugin-validator. The only remaining issues are that the plugin is not signed (expected until later in the plugin submission process) and some of the README image links will be broken until this is merged.

- no, will submit plugin after merging

- no


sdk/highlightinc-highlight-datasource/provisioning/datasources/datasource.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: 1
+
+datasources:
+  - name: Highlight.io
+    type: highlightinc-highlight-datasource
+    jsonData:
+      clientID: ''
+      projectID: 1344
+      tokenURL: 'https://pri.highlight.io/oauth/token'
+      backendURL: 'https://pri.highlight.io'
```
Vadman97, line 10: does our api just work for the demo project, or do we need to add some special logic to support it?

Vadman97 (approved):
very cool! 🚢

sdk/highlightinc-highlight-datasource/provisioning/datasources/datasource.yaml
mayberryzane: it just works! if no oauth credentials are entered into the data source, i'm using the default http client rather than the oauth client, and the auth in the resolvers lets you query for demo project data

PR #7402, opened 2023-12-28
Use Reflame exports detection

lewisl9029:
## Summary

Removes some redundant options that have been made default, and migrates to our new package.json exports compat where possible so there's a single source of truth for import resolution.

Left comments on each change for more context.

Reflame preview & tests still work.

### DIFF
--- a/.reflame.config.jsonc
+++ b/.reflame.config.jsonc
@@ -3,9 +3,9 @@
   // It's basically JSON, with the addition of comments, and looser syntax
   // (trailing commas!).
   // Reflame uses this to identify your app.
-  // "appId": "01GVH35S71FJP26W72V3K7CB2N", // prod
+  "appId": "01GVH35S71FJP26W72V3K7CB2N", // prod
   // "appId": "01GSXZYQT1W0TZDJAG94P9S4BZ", // prod fork
-  "appId": "01GR550WASJY2G2NSYZ4P54JJV", // dev
+  // "appId": "01GR550WASJY2G2NSYZ4P54JJV", // dev
   // This is what shows up in the browser's tab bar.
   "title": "highlight.io",
   // This is the description that shows up in Google search.
--- a/.reflame.config.jsonc
+++ b/.reflame.config.jsonc
@@ -3,9 +3,9 @@
   // It's basically JSON, with the addition of comments, and looser syntax
   // (trailing commas!).
   // Reflame uses this to identify your app.
-  "appId": "01GVH35S71FJP26W72V3K7CB2N", // prod
+  // "appId": "01GVH35S71FJP26W72V3K7CB2N", // prod
   // "appId": "01GSXZYQT1W0TZDJAG94P9S4BZ", // prod fork
-  // "appId": "01GR550WASJY2G2NSYZ4P54JJV", // dev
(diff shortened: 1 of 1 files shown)


.reflame.config.jsonc
```diff
@@ -167,12 +167,7 @@
     "testStories": true,
   },
   "nodejsCompatibility": {
-    "omitModuleExtension": true,
-    "mapIndexToDirectory": true,
```
lewisl9029, line 171: These compat options are now enabled by default, so we can get rid of them.

.reflame.config.jsonc
```diff
@@ -306,10 +294,10 @@
         },
       },
       "nodejsCompatibility": {
-        "omitModuleExtension": true,
-        "mapIndexToDirectory": true,
         "packageJson": {
-          "dependenciesSynced": true,
+          "exportsDetection": {
+            "ignoreTopLevelDirectory": true
```
lewisl9029, line 299: This is a bit of a hack to map the build artifact paths (`./dist/*`, which don't exist in our source code) in our package.json exports to source paths relative to `sourceDirectory`.

.reflame.config.jsonc
```diff
@@ -273,16 +268,9 @@
...
-        "/keyframes": "/keyframes.ts",
-        "/sprinkles": "/sprinkles.ts",
-        "/vars": "/vars.ts",
-        "/theme": "/theme.ts",
-        "/colors": "/colors.ts",
-        "/borders": "/borders.ts",
-        "/styles.css": "/noop.ts",
+      "exports": {
+        "./components": "./components/index.ts",
+        "./styles.css": "./noop.ts",
```
lewisl9029, line 273: We need to override components because the dist path points to `dist/components.js` which is not a 1:1 mapping to the source modules, and styles.css because we're using our custom css bundle.

.reflame.config.jsonc
```diff
@@ -346,24 +331,20 @@
...
+        "./src/listeners/first-load-listeners": "./listeners/first-load-listeners.tsx",
+        "./src/listeners/first-load-listeners.js": "./listeners/first-load-listeners.tsx",
+        "./src/listeners/network-listener/utils/utils": "./listeners/network-listener/utils/utils.ts",
+        "./src/utils/secure-id": "./utils/secure-id.ts",
+        "./src/utils/secure-id.js": "./utils/secure-id.ts",
+        "./src/utils/sessionStorage/highlightSession": "./utils/sessionStorage/highlightSession.ts",
+        "./src/utils/sessionStorage/highlightSession.js": "./utils/sessionStorage/highlightSession.ts",
+        "./src/utils/sessionStorage/sessionStorageKeys": "./utils/sessionStorage/sessionStorageKeys.ts",
+        "./src/utils/sessionStorage/sessionStorageKeys.js": "./utils/sessionStorage/sessionStorageKeys.ts",
+        "./src/utils/storage.js": "./utils/storage.ts",
```
lewisl9029, line 344: There's no exports map in the packageJson for this lib, so we still have to specify every entry point. Having an explicit allowlist allows us to take a bunch of shortcuts when resolving imports from libs for performance, but it should be pretty straightforward to allow importing from every file if this ever becomes a pain. Just let me know!

Vadman97 (approved):

PR #7409, opened 2024-01-02
fixed white line on navbar

Abhishek-More:
## Summary

See attached issue. Flashing white line under navbar.

Scroll testing

No

No


Vadman97 (approved):
nice! 

https://www.loom.com/share/cfa486a3a02948ec8efb71ee610d6ac0
maybe for `window.scrollY > 60` we should have a bit of a larger buffer than 60px to make it so that the line doesn't show until you scroll more?

Abhishek-More:
Increased the distance from 60 to 120px. Looks better now

PR #7408, opened 2024-01-02
Support saved segments (traces/logs)

SpennyNDaJets:
## Summary
Support saved segments on traces and logs

## How did you test this change?
1) View the logs page
- [ ] Saved segment buttons are displayed
- [ ] Save button is disabled
2) Add a search criteria
- [ ] Able to save the segment
- [ ] Able to rename the saved segment
- [ ] Able to duplicate the saved segment
- [ ] Able to edit the query of the saved segment
- [ ] Able to switch between saved segments
- [ ] Able to delete a saved segment
- [ ] Able to restore defaults of a saved segment
- [ ] Able to restore defaults of the search
3) Repeat for traces

https://www.loom.com/share/a81a439bb9884a5f9a73919844cc37cf?sid=c70de6d1-9c96-4d4d-afd6-0138eb2b3b24

## Are there any deployment considerations?
N/A
Closes #7030

## Does this work require review from our design team?
Sure

PR #7410, opened 2024-01-02
Support saved segments (backend)

SpennyNDaJets:
## Summary
Create the necessary tables and graphql mutations/queries needed to support saved searches for the traces and logs table.

- Check code is valid.
- Click test in https://github.com/highlight/highlight/pull/7408

N/A

N/A


backend/model/model.go
```diff
@@ -2109,6 +2110,21 @@ type LogAlertEvent struct {
...
+	SavedSegmentEntityTypeLog   SavedSegmentEntityType = "Log"
+	SavedSegmentEntityTypeTrace SavedSegmentEntityType = "Trace"
+)
+
+type SavedSegment struct {
+	Model
+	Name       string
+	EntityType SavedSegmentEntityType `gorm:"index:idx_saved_segment"`
+	Params     string                 `json:"params"`
+	ProjectID  int                    `gorm:"index:idx_saved_segment" json:"project_id"`
```
SpennyNDaJets, line 2125: Add an index here since most queries will be by `project_id` and `entity_type` (e.g. "Log saved segments for project 1", "Trace saved segments for Project 23")

backend/private-graph/graph/schema.graphqls
```diff
@@ -1137,27 +1154,6 @@ input NetworkHistogramParamsInput {
 	attribute: NetworkRequestAttribute
 }
 
-type SearchParams {
```
SpennyNDaJets, line 1140: Not used anymore - just all included in the `query` param

backend/model/model.go
```diff
@@ -2109,6 +2110,14 @@ type LogAlertEvent struct {
 	SentAt     time.Time
 }
 
+type SavedSegment struct {
+	Model
+	Name       string
+	EntityType modelInputs.SavedSegmentEntityType `gorm:"index:idx_saved_segment"`
```
SpennyNDaJets, line 2116: Add an index here since most queries will be by project_id and entity_type (e.g. "Log saved segments for project 1", "Trace saved segments for Project 23")

mayberryzane: nit: can use the `priority` field in gorm so that project id is ordered before entity type - assuming sometime in the future we might find it useful to query all segments for all entity types in a project, but would never want to query segments for a single entity type across all projects https://gorm.io/docs/indexes.html#:~:text=You%20can%20specify%20the%20order%20with%20the%20priority%20option%2C%20the%20default%20priority%20value%20is%2010%2C%20if%20priority%20value%20is%20the%20same%2C%20the%20order%20will%20be%20based%20on%20model%20struct%E2%80%99s%20field%20index

backend/private-graph/graph/resolver.go
```diff
@@ -703,6 +703,20 @@ func (r *Resolver) isAdminErrorSegmentOwner(ctx context.Context, error_segment_i
 	return segment, nil
 }
 
+func (r *Resolver) isAdminSavedSegmentOwner(ctx context.Context, segment_id int) (*model.SavedSegment, error) {
+	authSpan, ctx := util.StartSpanFromContext(ctx, "isAdminSavedSegmentOwner", util.ResourceName("resolver.internal.auth"))
+	defer authSpan.Finish()
+	segment := &model.SavedSegment{}
+	if err := r.DB.WithContext(ctx).Where(&model.SavedSegment{Model: model.Model{ID: segment_id}}).Take(&segment).Error; err != nil {
```
mayberryzane, line 710: should replace with `.Where("id = ?", segment_id)` - if segment_id is zero, the where clause is not applied and a segment is picked arbitrarily

backend/private-graph/graph/schema.resolvers.go
```diff
@@ -1189,6 +1189,82 @@ func (r *mutationResolver) DeleteErrorSegment(ctx context.Context, segmentID int
...
+
+	var count int64
+	if err := r.DB.WithContext(ctx).Model(&model.SavedSegment{}).Where("project_id = ? AND name = ? AND entity_type = ? AND id <> ?", projectID, name, entityType, id).Count(&count).Error; err != nil {
+		return nil, err
+	}
+	if count > 0 {
+		return nil, e.New("saved segment with this name already exists")
+	}
+
+	if err := r.DB.WithContext(ctx).Model(&model.SavedSegment{Model: model.Model{ID: id}}).Updates(&model.SavedSegment{
```
mayberryzane, line 1247: same .where feedback here

backend/private-graph/graph/schema.resolvers.go
```diff
@@ -1189,6 +1189,82 @@ func (r *mutationResolver) DeleteErrorSegment(ctx context.Context, segmentID int
...
+	return &model.T, nil
+}
+
+// DeleteSavedSegment is the resolver for the deleteSavedSegment field.
+func (r *mutationResolver) DeleteSavedSegment(ctx context.Context, segmentID int) (*bool, error) {
+	_, err := r.isAdminSavedSegmentOwner(ctx, segmentID)
+	if err != nil {
+		return nil, err
+	}
+	if err := r.DB.Delete(&model.SavedSegment{Model: model.Model{ID: segmentID}}).Error; err != nil {
```
mayberryzane, line 1262: same .where feedback here

mayberryzane (approved):

PR #7411, opened 2024-01-03
add an enterprise plan comparison page

Vadman97:
## Summary

Building a new checkout flow that provides an enterprise pricing comparison.

https://www.loom.com/share/186d12127ff74fd29b4cc170b3c0892a

All frontend.

No


frontend/src/pages/Billing/PlanComparisonPage.tsx
```diff
@@ -0,0 +1,235 @@
...
+				per month
+			</Text>
+			<Button
+				trackingId={`planSelect-${plan.name}`}
+				kind="secondary"
+				size="small"
+				emphasis={enterprise || current ? 'medium' : 'high'}
+				disabled={current}
+				onClick={() => {
+					if (enterprise) {
```
SpennyNDaJets, line 123: Opinion: should we disable these buttons when the user is on an enterprise account? Seems weird to leave the button enabled with no onClick, unless we were going to have another action

SpennyNDaJets (approved):
LGTM

frontend/src/pages/Billing/PlanComparisonPage.tsx
Vadman97: my bad - missed the desired action here. should be fixed now.

PR #7415, opened 2024-01-03
create hnsw indexes for all partitions of the error_object_embeddings_partitioned table

mayberryzane:
## Summary
- hnsw is a new vector index type introduced in pgvector v0.5.0 with better performance than ivfflat
- this migration adds hnsw indexes for all partitions of error_object_embeddings_partitioned

- ran script locally, currently running in prod

- new partitions are already using hnsw indexes
- will remove the existing ivfflat indexes after confirming these new indexes are created for all partitions and working as expected

- no


Vadman97 (approved):
🎉

PR #7427, opened 2024-01-05
ensure pino transport works correctly in non-node environments  @Vadman97

Vadman97:
## Summary

When using the pino transport in next-js, the import would cause errors in the browser
because client code would try to use the transport.
Detect next runtimes and avoid importing the node-specific SDK.

Testing on highlight.io and in the e2e app.
https://www.loom.com/share/bdcc5faf75db45c8bbf0e8d12c3d9981

Changeset

No

### DIFF
--- a/.changeset/pink-baboons-look.md
+++ b/.changeset/pink-baboons-look.md
@@ -1,0 +1,5 @@
+---
+'@highlight-run/pino': patch
+---
+
+ensure pino transport works correctly in non-node environments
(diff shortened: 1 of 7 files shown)


Vadman97:
cc @Abhishek-More the pino changes seem to fix highlight.io/sitemap.xml locally for me, but the vercel preview is still timing out (maybe in production some of the sitemap logic now just takes too long, longer than 30 seconds?)

mayberryzane (approved):

PR #7432, opened 2024-01-05
Use relative times in search

SpennyNDaJets:
## Summary
Use relative times in the logs and traces search product

Other product changes
- Logs on sessions network requests no long polls
- Logs on traces page no longer polls
- Project filters use new component, but handle state within the component
- Change all date range pickers to support same day selection

1) Load the logs/traces page
- [ ] Time frame defaults to last 15 minutes in search
- [ ] Polling for future traces/logs
2) Select a relative time (e.g. last 4 hours)
- [ ] Time frame is correct in search
- [ ] Polling for future traces/logs
3) Refresh the page
- [ ] Time frame in search is most recent to last X, and not hardcoded to previous times
4) Select a hardcoded time
- [ ] Can select the same day
- [ ] Time frame is correct in search
- [ ] Polling for future logs/traces is disabled
- [ ] Refreshing the page uses the same time frame
- [ ] Can switch back to relative time

https://www.loom.com/share/c209275f22dd4e90a92bc4ad100a902f?sid=ffefdb9d-7710-459a-a94e-870e7f9ee191

N/A

N/A


frontend/src/pages/Player/RightPlayerPanel/components/NetworkResourcePanel/NetworkResourceLogs.tsx
```diff
@@ -67,6 +67,7 @@ export const NetworkResourceLogs: React.FC<{
 		logCursor: undefined,
 		startDate,
 		endDate,
+		disablePolling: true,
```
SpennyNDaJets, line 70: Disable polling for network resource panel since this is locked into the time a session starts and ends

frontend/src/pages/ProjectSettings/ProjectFilters/ProjectFilters.tsx
```diff
@@ -141,9 +146,12 @@ export const ProjectProductFilters: React.FC<{
...
 	const navigate = useNavigate()
 	const { currentWorkspace } = useApplicationContext()
+	const [selectedPreset, setSelectedPreset] =
+		React.useState<DateRangePreset>(DEFAULT_PRESET)
+
 	const [dateRange, setDateRange] = React.useState<DateRange>({
-		start: defaultPresets[1].startDate,
-		end: getNow().toDate(),
+		start: presetStartDate(DEFAULT_PRESET),
+		end: moment().toDate(),
```
SpennyNDaJets, line 154: Manage the start within the component instead of url since multiple of these will be rendered on 1 page and should behave independently

frontend/src/pages/Traces/TraceLogs.tsx
```diff
@@ -42,6 +42,7 @@ export const TraceLogs: React.FC = () => {
 			logCursor: undefined,
 			startDate,
 			endDate,
+			disablePolling: true,
```
SpennyNDaJets, line 45: Disable polling on this logs search since fixed to the time frame of a trace

packages/ui/src/components/DatePicker/PreviousDateRangePicker.tsx
```diff
@@ -398,12 +398,35 @@ const PreviousDateRangePickerImpl = ({
...
 				selectedDates,
-				onDatesChange: handleDatesChange,
-				dates: { mode: 'range', minDate, maxDate: new Date() },
+				onDatesChange: handleAbsoluteDateChange,
+				dates: {
+					mode: 'range',
+					minDate,
+					maxDate: new Date(),
+					selectSameDate: true,
+				},
```
SpennyNDaJets, line 429: Kept this consistent across all the products to avoid confusion (i.e. sessions vs logs)

frontend/src/pages/LogsPage/useGetLogs.ts
```diff
@@ -85,6 +87,7 @@ export const useGetLogs = ({
 		GetLogsQuery,
 		GetLogsQueryVariables
 	>({
+		skip: disablePolling,
```
Vadman97, line 90: one thing i noticed from click testing is that the polling query still uses a fixed time range (that looks 1 hour into the future, so if the page is open for more than 1 hour, it won't query new data). might be something we should change, though don't necessarily need to do this in this pr

Vadman97 (approved):
clicktest looks great, big improvement :tada:

frontend/src/pages/LogsPage/useGetLogs.ts
SpennyNDaJets: Yes, I saw this. I think ideally it would use the current time as the end date, which should hopefully be a quick fix.

This would be an issue if the user did not receive any logs/traces for an hour, but we would show the "X more logs" that were fetched in the time frame. Then when the user clicks it, we would fetch the logs with the "Last Y time" using the current time as the end date.

frontend/src/pages/ProjectSettings/ProjectFilters/ProjectFilters.tsx
```diff
@@ -568,17 +579,29 @@ export const ProjectProductFilters: React.FC<{
 					<Text weight="medium" size="xSmall" color="weak">
 						{label}s
 					</Text>
-					<PreviousDateRangePicker
-						selectedDates={[dateRange.start, dateRange.end]}
-						onDatesChange={(dates) =>
+					{/* TODO(spenny): figure this out */}
```
ccschmitz, line 582: Is there still a follow up here?

ccschmitz (approved):
Code + click test LGTM 😻

frontend/src/pages/ProjectSettings/ProjectFilters/ProjectFilters.tsx
SpennyNDaJets: Nope! 😄

PR #7433, opened 2024-01-05
fix more grafana bugs

mayberryzane:
## Summary
- fixes bug where numeric keys would not show up as group by options (e.g. enum types), should handle those as both potential metric types and group by types
- applies `where` conditions before limiting the number of groups
- removes prewhere logic as clickhouse should handle this automatically, did not see any meaningful performance difference between the two

- validated that the correct data was shown in the grafana plugin after applying a filter and limit = 1 

- no

- no


Vadman97 (approved):
🙇

PR #7434, opened 2024-01-05
remove message from large kafka message logs

mayberryzane:
## Summary
- these messages are pretty big and could be a cause for elevated cloudwatch ingest, removing them 

- will monitor cloudwatch ingest bytes for public graph after deploying

- no

- no


Vadman97 (approved):
🙏

PR #7445, opened 2024-01-08
Customer Testimonial OG image

Abhishek-More:
## Summary

See attached issue.
We want an OG image for customer testimonials, similar to blog posts.

Checking opengraph.xyz

https://www.opengraph.xyz/url/https%3A%2F%2Fhighlight-landing-g0mpgbx8k-highlight-run.vercel.app%2Fcustomers%2Fmotion

image

No

No


highlight.io/pages/docs/[[...doc]].tsx
Comment (Vadman97) [inline highlight.io/pages/docs/[[...doc]].tsx:1]: was it necessary to remove the logger here? my guess is that it is only necessary to remove in `github.ts` since that is a file that is imported by the `og` edge runtime paths.

Vadman97 (approved):

highlight.io/pages/docs/[[...doc]].tsx
Comment (Abhishek-More) [reply highlight.io/pages/docs/[[...doc]].tsx:1]: Yeah, that part was also causing issues. Adding it back breaks OG previews for blog too

PR #7458, opened 2024-01-09
use clickhouse for session metrics

mayberryzane:
## Summary
- device memory and web vitals are the last two types of metrics being read from postgres
- add a clickhouse materialized view indexed on session secure id to look up these metrics efficiently from clickhouse

- tested locally to make sure migration created the table/mv and the resolver was behaving correctly
Screen Shot 2024-01-09 at 3 39 41 PM

- table / mv already created in prod, will remove the code for saving to the metrics table once this is looking good

- no


Vadman97 (approved):
neat that the MV will capture other metrics (any metrics that report a session ID). :ship:

PR #7464, opened 2024-01-10
Fixed "Suggested Edits" Docs link

Abhishek-More:
## Summary

Fix for "Suggest Edits" docs button not linking properly in prod. When merging the github docs link with the relative docs path, precautions were taken to prevent double slashes. The links would replace every '//' with '/'. However, this also applied to the '//' in 'https://', which broke the link.

Click-testing

No.

No.


Vadman97 (approved):
preview looks good!

PR #7465, opened 2024-01-10
Use relative times on the sessions and errors page

SpennyNDaJets:
## Summary
Apply the new relative time range to the query builder (sessions/errors) page.

Updates to logic:
- session and error search no longer use `TimeRangeField` on frontend, and all logic is stripped away
- Remove trying to build the backend query on the frontend (was necessary for opensearch)
- Update the time range component to always send a start and end date, even when a preset is selected
- Update the `useSearchTime` hook to set the state first, with a `useEffect` to set the url params (was switched previously)

https://www.loom.com/share/a9a8f467052a44babd927af43356baaf?sid=7b32906e-f324-4dcd-b773-9769e094cc92

1) Load the sessions page
- [ ] Correct default time of sessions fetched
2) Change the time range to a different relative time
- [ ] Correct time of sessions fetched
3) Reload to page
- [ ] Correct time of sessions fetched
4) Change to a absolute time
- [ ] Correct time of sessions fetched
5) Refresh the page
- [ ] Correct time of sessions fetched
7) Save a query
- [ ] Query saved without time
8) Change the time by dragging the histogram
- [ ] Correct time of sessions fetched
9) Load the errors page via the topbar
- [ ] Correct default time of errors fetched
10) Change the time range to a different relative time
- [ ] Correct time of errors fetched
11) Reload to page
- [ ] Correct time of errors fetched
12) Change to a absolute time
- [ ] Correct time of errors fetched
13) Refresh the page
- [ ] Correct time of errors fetched
14) Save a query
- [ ] Query saved without time
15) Change the time by dragging the histogram
- [ ] Correct time of errors fetched
16) Test searching errors with Command-K bar
- [ ] Correctly fetches errors
- [ ] Correctly fetches sessions

N/A

N/A


SpennyNDaJets:
> looking good! issue i noticed locally is that the relative time choice gets reset on tab refresh (different behavior from logs / traces) https://www.loom.com/share/0cd205fcb42b4a13958240e8282a2d1b

Thanks for raising! I found two separate issues here:
1) The time and segment url params are clearing on refresh
  a) This is an issue locally due to this [useEffect](https://github.com/highlight/highlight/pull/7465/files#diff-d54385688568eb9bced4be7622e705e692296e25dab2778da5c1b82553d0d2bfR1555) in `<React.StrictMode>` we may want to look into how we can solve this, but removing the strict mode causes it to work properly. Without this use effect we see the times and selected segments bleeding into the errors and segments search since the context is always rendered. Ideally they searches would be completely separate.

2) You cannot "Save as a new segment" on Sessions and Errors.
  a) Taking a look at this!

PR #7470, opened 2024-01-11
enhance node.js errors with sourcemaps

Vadman97:


e2e/express-ts/package.json
```diff
@@ -8,7 +8,8 @@
 		"build": "tsc",
 		"dev": "ts-node src/index.ts",
 		"start": "node dist/index.js",
-		"test:dev": "ts-node src/test.ts"
+		"test:dev": "ts-node src/test.ts",
+		"sourcemaps": "yarn exec ../../sourcemap-uploader/dist/index.js upload --apiKey \"${HIGHLIGHT_API_KEY}\" --appVersion \"${APP_VERSION}\" --path ./dist --basePath /Users/vkorolik/work/highlight/e2e/express-ts/dist/"
```
SpennyNDaJets, line 12: Not a blocker, but this will only work for your laptop with path `/Users/vkorolik/work`?

backend/public-graph/graph/resolver.go
```diff
@@ -2213,14 +2212,30 @@ func (r *Resolver) ProcessBackendPayloadImpl(ctx context.Context, sessionSecureI
 			ServiceVersion: v.Service.Version,
 		}
 
-		var mappedStackTrace *string
 		var structuredStackTrace []*privateModel.ErrorTrace
-		mappedStackTrace, structuredStackTrace, err = r.Store.EnhancedStackTrace(ctx, v.StackTrace, workspace, &project, errorToInsert, nil)
+		var stackFrameInput []*publicModel.StackFrameInput
 
+		if err := json.Unmarshal([]byte(v.StackTrace), &stackFrameInput); err == nil {
+			mapped, structured, err := r.getMappedStackTraceString(ctx, stackFrameInput, projectID, errorToInsert, pointy.String(fmt.Sprintf("%s-%s", v.Service.Name, v.Service.Version)))
```
SpennyNDaJets, line 2219: Just want to make this edge case won't cause issues - a lot of users leave this hardcoded to "git-sha" or an empty string when it isn't provided to the Highlight instance.

This could help promote users supplying their correct service version to the instance, but want to call out incase any one may be having issues with it working

Seems like worst case we would log the error and continue

SpennyNDaJets (approved):
Code looks great to me! Left some considerations but nothing blocking

e2e/express-ts/package.json
Vadman97: good point, making this the absolute working dir to make this work for others

backend/public-graph/graph/resolver.go
Vadman97: seems to me that now for accounts with an invalid git sha, we may get log spam from this happening and then the project's service would become unhealthy, but both seem fine. do you think there might be other issues with having this run by default?

PR #7479, opened 2024-01-12
fix promise patch for angular.js

Vadman97:


sdk/highlight-node/src/client.ts
```diff
@@ -332,10 +332,22 @@ export class Highlight {
 			span.setAttributes(metadata)
 		}
 		if (secureSessionId) {
-			span.setAttributes({ ['highlight.session_id']: secureSessionId })
+			span.setAttribute('highlight.session_id', secureSessionId)
```
deltaepsilon, line 335: Well THAT'S a good catch.

deltaepsilon (approved):

sdk/highlight-node/src/client.ts
Vadman97: haha i think it's equivalent; just thought i'd simplify this

PR #7503, opened 2024-01-16
fix potential bugs in otel.go, better logging in resolver.go

mayberryzane:
## Summary
- pointer reference in loop points to last project id
- declare `messages` in loop to prevent duplicate metrics

- will monitor logs in prod after deploying

- no

- no

### DIFF
--- a/backend/otel/otel.go
+++ b/backend/otel/otel.go
@@ -324,7 +324,7 @@ func (o *Handler) HandleTrace(w http.ResponseWriter, r *http.Request) {
 				keyedErrorMessages[sessionID] = append(keyedErrorMessages[sessionID], &kafkaqueue.Message{
 					Type: kafkaqueue.PushBackendPayload,
 					PushBackendPayload: &kafkaqueue.PushBackendPayloadArgs{
-						ProjectVerboseID: &projectID,
+						ProjectVerboseID: pointy.String(projectID),
 						Errors:           []*model.BackendErrorObjectInput{errorObject},
 					}})
 			}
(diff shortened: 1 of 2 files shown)


backend/otel/otel.go
```diff
@@ -324,7 +324,7 @@ func (o *Handler) HandleTrace(w http.ResponseWriter, r *http.Request) {
 				keyedErrorMessages[sessionID] = append(keyedErrorMessages[sessionID], &kafkaqueue.Message{
 					Type: kafkaqueue.PushBackendPayload,
 					PushBackendPayload: &kafkaqueue.PushBackendPayloadArgs{
-						ProjectVerboseID: &projectID,
```
Vadman97, line 327: just making sure i understand this bug:

- `projectID` on 314 is an integer initialized once for the whole loop
- `projectID` set to first value in the map
- this line puts the memory address of the variable in the message
- next iteration of the loop updates `projectID` to be second value in the map
1st message and 2nd message have the same pointer here with the same value, rather than the two different project ids

backend/otel/otel.go
```diff
@@ -339,9 +339,9 @@ func (o *Handler) HandleTrace(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
-	var messages []*kafkaqueue.Message
 	for projectID, traceMetrics := range projectTraceMetrics {
 		for sessionID, metrics := range traceMetrics {
```
Vadman97, line 343: same bug here?

Vadman97 (approved):

backend/otel/otel.go
mayberryzane: this was different but could cause duplicate metrics as the `messages` list could be submitted to kafka multiple times

Vadman97: what about the `&projectID` reference on line 349?

PR #7506, opened 2024-01-17
improve public worker compressed payload processing

Vadman97:
## Summary

* write compressed push payload messages and decompress in the worker
* switch to [segmentio](https://github.com/segmentio/encoding) json encoding / decoding for which boasts better performance and lower cpu consumption

E2E ci recording session. Existing `PushPayload` messages from old clients will not be affected.

local deploy
Screenshot 2024-01-17 at 4 33 59 PM

Will monitor worker CPU impact.

No


SpennyNDaJets (approved):
image

PR #7516, opened 2024-01-19
Don't set auth redirect if on homepage

ccschmitz:
## Summary

Fixes a bug where users could get stuck in the loading state of the app permanently. The problem was that we were storing the initial URL the user was supposed to be redirected to which included a search string in the URL. We would try to redirect back to this from the `ProjectRedirectionRouter` which would give precedence to any URL stored in the auth redirect cookie vs falling back to the default project.

This PR ensures we don't store a redirect URL for the root path (`/`) even if there is a search string after it.

In the PR preview, I dropped a debugger in `SignInRedirect` to see if `authRedirect.set` was called. I confirmed that the new logic was avoiding storing the redirect URL and that I was able to get into the app after coming to it from the homepage.
 
N/A

N/A

### DIFF
--- a/frontend/src/pages/Auth/SignInRedirect.tsx
+++ b/frontend/src/pages/Auth/SignInRedirect.tsx
@@ -8,8 +8,8 @@ export const SignInRedirect: React.FC = () => {
 		// Store the original path so we can redirect back to it later.
 		const dest = window.location.href.replace(window.location.origin, '')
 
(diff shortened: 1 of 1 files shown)


SpennyNDaJets (approved):

Vadman97 (approved):
ahh, so having a query string would prevent the redirect from happening. nice find!

PR #7519, opened 2024-01-19
Update exist and not exists documentation

SpennyNDaJets:
## Summary
Instead of adding a default filter to only show root level traces, we want to update the documentation to show users how to create these filters. This is in part because the filter is not clear what it is doing, and may not be the best default experience

View docs

N/A

N/A


ccschmitz (commented):
Works for what we're trying to accomplish, but I'm concerned making this a default filter will be confusing. I know we've gone back and forth on adding an actual control in the UI to apply this filter, but I'm thinking that would be much more clear. If I came to the traces page and saw that default filter I think I would be confused. Definitely subjective, so curious what others think.

Vadman97 (approved):
could we add docs to https://www.highlight.io/docs/general/product-features/tracing/trace-search to describe how one would search for root / non-root traces with the `parent_span_id` filter, and how one could use the `=""` or `!=""` to look for keys being set vs empty?

Vadman97:
don't think that the `parent_span_id` key shows up for me here

Screenshot 2024-01-22 at 11 19 10 AM

SpennyNDaJets:
> don't think that the `parent_span_id` key shows up for me here
> 
> Screenshot 2024-01-22 at 11 19 10 AM

@Vadman97 just want to make sure this was intended for this PR? https://github.com/highlight/highlight/pull/7529

mayberryzane:
@SpennyNDaJets if we're still applying a default filter, should we switch this over to the `exists` syntax here once this gets approved: https://github.com/highlight/highlight/pull/7548 ?

SpennyNDaJets:
> @SpennyNDaJets if we're still planning to have a default filter, should we switch this over to the `exists` syntax here once this gets approved: #7548 ?

@mayberryzane that sounds great! I can handle that and update the documentation

PR #7529, opened 2024-01-22
Add clickhouse materialized views to add missing keys

SpennyNDaJets:
## Summary
A few keys were missing from the traces and logs search since they are not added via attributes, and there were no materialized views to add them to the correct tables.

Add materialized views to the following attributes so they show up in search:
- Logs
  - environment
- Traces
  - environment
  - parent_span_id (add to default with no values)
  - secure_session_id (add to default with no values)

Remove high cardinality attributes from logs - span_id, trace_id, secure_session_id, body (not used)

1) View the logs and traces page
2) try searching by the keys above
- [ ] Keys are correctly suggested
- [ ] Values are correctly suggested

Screenshot 2024-01-22 at 2 29 34 PM

Screenshot 2024-01-22 at 2 47 06 PM

Watch out for too many table row scan errors

N/A


backend/clickhouse/migrations/000073_create_trace_secure_session_id_mv.up.sql
```diff
@@ -0,0 +1,14 @@
...
+    `Key` String,
+    `TraceTimestamp` DateTime,
+    `TraceUUID` UUID,
+    `Value` String
+) AS
+SELECT ProjectId AS ProjectId,
+    'secure_session_id' AS Key,
+    Timestamp AS TraceTimestamp,
+    UUID AS TraceUUID,
+    SecureSessionId AS Value
```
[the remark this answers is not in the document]
Vadman97: cc @mayberryzane in case you have oppinions here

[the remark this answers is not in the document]
SpennyNDaJets: I can do that instead. I noticed this was already added to the logs table, so curious if we should synchronize how these session ids are handled? https://github.com/highlight/highlight/blob/main/backend/clickhouse/migrations/000032_create_log_secure_session_id_mv.up.sql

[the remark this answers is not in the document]
Vadman97: yeah, good point. on logs right now it looks like both `session_secure_id` and `secure_session_id` recommend values. perhaps we can drop both there as well as on traces?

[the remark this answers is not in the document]
mayberryzane: yep, for performance I agree it would be better to hardcode any high cardinality keys and omit them from the keys mv

backend/clickhouse/traces.go
```diff
@@ -349,7 +352,7 @@ func (client *Client) TracesKeys(ctx context.Context, projectID int, startDate t
 		traceKeys = append(traceKeys, defaultTraceKeys...)
 	} else {
 		for _, key := range defaultTraceKeys {
-			if !strings.Contains(key.Name, *query) {
+			if strings.Contains(key.Name, *query) {
```
SpennyNDaJets, line 355: Believe this was flipped?

Vadman97: makes sense, also shouldnt this be `if strings.Contains(*query, key.Name) {`
to look for the default key in the query?

backend/clickhouse/logs.go
```diff
@@ -36,6 +37,13 @@ var logKeysToColumns = map[modelInputs.ReservedLogKey]string{
 	modelInputs.ReservedLogKeyEnvironment:     "Environment",
 }
 
+// These keys show up as recommendations, but with no recommended values due to high cardinality
+var defaultLogKeys = []*modelInputs.QueryKey{
```
Vadman97, line 41: should `body` be part of this since we are dropping the mv?

Vadman97 (approved):
makes sense to me!

backend/clickhouse/logs.go
SpennyNDaJets: Currently the body key is never being used since it is the default key and being filtered out on the frontend. However, I'll probably add this back in another PR for customizing columns on the logs table

backend/clickhouse/traces.go
SpennyNDaJets: Wait actually I think its currently right - for example, if the user types in "env" for `*query`, we would want to recommend the key, `environment`. So searching for the `*query` substring within the `key.Name`. Let me know if there is another case for this that I am missing

PR #7540, opened 2024-01-23
Show shorter sessions

SpennyNDaJets:
## Summary
Sessions that are less than a second but not excluded are not showing up on session search, but can be navigated to via the url.

N/A

N/A

N/A


Vadman97 (approved):
🎉

PR #7545, opened 2024-01-23
Search typeahead updates

ccschmitz:
## Summary

* Updates the number of typeahead results to show to 25
* Displays search operators as typeahead results when there is an exact match on an attribute key
* Removes the no results text
* Gives all typeahead items a `type` attribute
* Ensures all selected values are wrapped in quotes when necessary
* Fixes an issue where the typeahead results would display all results when deleting a character
* Cache results from keys and values queries
* Eliminate the delay in fetching new keys when starting a new filter

Local + PR preview click test. All changes except upping the number of results to display should be testable in the PR preview.

N/A

Will share the PR preview with Julian and get his feedback on the new UX since I've made a bunch of assumptions about how things should work. I'm sure he'll have some ideas for impovements 😄


Vadman97:
unrelated to this PR, but another bug i recently found - made a logs query for "running api docs search query" and it's no longer an exact string match
https://app.highlight.io/demo/logs?query=%22running%20api%20docs%20search%20query%22&relative_time=last_4_hours

Screenshot 2024-01-23 at 3 37 40 PM

ccschmitz:
@Vadman97 thanks for reporting that! Just filed https://github.com/highlight/highlight/issues/7581 to follow up on it.

frontend/src/components/Search/SearchForm/SearchForm.tsx
```diff
@@ -240,46 +243,65 @@ export const Search: React.FC<{
...
+	const [getKeys, { loading: keysLoading }] = fetchKeysLazyQuery()
+	const [getKeyValues, { loading: valuesLoading }] = fetchValuesLazyQuery()
 	const [cursorIndex, setCursorIndex] = useState(0)
+	const [isPending, startTransition] = React.useTransition()
 
 	const { queryParts, tokens } = parseSearch(query)
 	const tokenGroups = buildTokenGroups(tokens, queryParts, query)
 	const activePart = getActivePart(cursorIndex, queryParts)
-	const debouncedKeyValue = useDebouncedValue<string>(activePart.value)
+	const { debouncedValue, setDebouncedValue } = useDebounce<string>(
```
SpennyNDaJets, line 259: Not in this PR, but should we be using `useDebounce` > `useDebouncedValue` in all places?

SpennyNDaJets (commented):
Click test and code look good to me!

Just want to double check that this works well with the custom column updates that was recently merged in. Seems like we can use this type to help us figure out how to best format in the table as well 🙌

frontend/src/components/Search/SearchForm/SearchForm.tsx
ccschmitz: Not something we need to do everywhere, I just wanted to be able to update the debounced value independently in that one case where someone is transitioning to a new query filter. `useDebouncedValue` didn't return a setter to do that, so I created my own hook here. Sorry if you think that will be a source of confusion.

SpennyNDaJets: That make sense to me! Feels like we should move towards one, but I don't feel strongly about the urgency

SpennyNDaJets (approved):
Oops - approved

PR #7548, opened 2024-01-24
add exists / not exists operators

mayberryzane:
## Summary
- adds exists/not exists operators to the antlr grammar
- implemented as `= ""` and `!= ""` in sql
- some small frontend changes to get syntax highlighting working
- makes the grammar case insensitive (@ccschmitz is this ok?)

- clicktested locally and in grafana plugin
https://www.loom.com/share/ace9726d178e4a16a9f5af6997c303c2

- no

- no


backend/parser/listener/listener.go
```diff
@@ -134,6 +137,21 @@ func (s *searchListener[T]) EnterBin_op(ctx *parser.Bin_opContext) {
...
+	switch op {
+	case "EXISTS":
+		s.currentOp = "!="
+	case "NOTEXISTS":
+		s.currentOp = "="
+	default:
+		fmt.Printf("Unknown exists operator: %s\n", op)
+	}
+
+	s.appendRules("")
```
Vadman97, line 152: just checking that this will work for numeric attributes too

Vadman97 (approved):

backend/parser/listener/listener.go
mayberryzane: just added that as a special case below - casting to a string before comparing against `''`, I think the only column this matters for on traces/logs is `Duration` atm, not sure how we want to handle `Duration = 0` but this treats it as exists

PR #7549, opened 2024-01-24
switch ingest parsing to antlr

### LINKED ISSUE #7269: Ingest filters should support non-is operators and be more clear what is excluded
Received feedback from Clearspace that ingest filters were missing some functionality for filtering and they would like it to be more clear what data is getting filtered.

Vadman97:
## Summary

Our ingest filters would not work with the new traces / logs syntax
as they were built for parsing the older syntax used for sessions and errors. 
Switches the ingestion filters to use the new syntax.
Closes #7269

Unit testing / local deploy
https://www.loom.com/share/0fc5aad6a04842f6b3b562a804b2381b

We may need to convert old exclusion queries to the new syntax, but there are only a handful configured right now so I'll do it manually once this is deployed.

No

### DIFF
--- a/backend/clickhouse/query.go
+++ b/backend/clickhouse/query.go
@@ -344,7 +344,7 @@ func KeyValuesAggregated(ctx context.Context, client *Client, tableName string, 
 }
 
 // clickhouse token - https://clickhouse.com/docs/en/sql-reference/functions/splitting-merging-functions#tokens
-var nonAlphaNumericChars = regexp.MustCompile(`\s`)
+var nonAlphaNumericChars = regexp.MustCompile(`[^\w:*]`)
 
 func matchFilter[TObj interface{}, TReservedKey ~string](row *TObj, config model.TableConfig[TReservedKey], filter *listener.FilterOperation) bool {
 	bodyFilter := config.BodyColumn != "" && filter.Column == "" && filter.Key == config.BodyColumn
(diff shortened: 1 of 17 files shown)


backend/parser/listener/listener.go
```diff
@@ -62,13 +93,27 @@ func (s *searchListener[T]) ExitAnd_col_expr(ctx *parser.And_col_exprContext) {
...
 
 func (s *searchListener[T]) EnterOr_col_expr(ctx *parser.Or_col_exprContext) {}
 func (s *searchListener[T]) ExitOr_col_expr(ctx *parser.Or_col_exprContext) {
 	rules := s.rules[len(s.rules)-2:]
 	s.rules = s.rules[:len(s.rules)-2]
 	s.rules = append(s.rules, s.sb.Or(rules...))
+
+	ops := s.ops[len(s.ops)-2:]
+	s.ops = s.ops[:len(s.ops)-2]
+	s.ops = append(s.ops, &FilterOperation{
```
mayberryzane, line 113: do we have to add a rule for `OperatorNot` too? (`ExitNegated_search_expr` below)

backend/clickhouse/query.go
```diff
@@ -371,79 +371,112 @@ func KeyValuesAggregated(ctx context.Context, client *Client, tableName string,
...
+			for _, childFilter := range filter.Filters {
+				if matchesQuery(row, config, listener.Filters{childFilter}) {
+					anyMatch = true
+					break
+				}
+			}
+			if !anyMatch {
+				return false
+			}
+		default:
```
mayberryzane, line 476: and handle `OperatorNot` here?

mayberryzane (approved):

backend/parser/listener/listener.go
Vadman97: whoops, not sure how i missed this one. good catch!

PR #7554, opened 2024-01-24
Fix link back to project from settings pages

ccschmitz:
## Summary

The settings pages have a button that says "Back to Project" in the top left. This actually opens a dropdown of projects to pick from rather than taking you back to the current project. This is also confusing because there is already a list of projects in the sidebar to choose from.

N/A - client only change.

N/A

### DIFF
--- a/frontend/src/components/Header/Header.tsx
+++ b/frontend/src/components/Header/Header.tsx
@@ -185,7 +185,7 @@ export const Header: React.FC<Props> = ({ fullyIntegrated }) => {
 					py="8"
 					justifyContent="space-between"
 				>
-					{isSetup ? (
+					{isSetup || isSettings ? (
 						<LinkButton
 							to={goBackPath}
 							kind="secondary"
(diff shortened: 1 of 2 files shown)


SpennyNDaJets (commented):
Looking good! I noticed a case when the user is on the workspace settings (i.e. no project selected) that the link has an undefined project id -> signed in as me, this causes me to get redirected to `/1/sessions`, but not sure the behavior as someone not in the Highlight project

test

ccschmitz:
Thanks for catching that, @SpennyNDaJets! I think that link was working as anticipated for me and I missed that it was actually the wrong path. I just pushed a couple changes.

1. Removes some dead code I noticed for what I assume were some old header components.
2. I started storing the project ID in local storage so we can reference it later from pages like the workspace settings to redirect back correctly.

Please take a look and let me know what you think!

SpennyNDaJets (approved):
Click test looks great!

PR #7580, opened 2024-01-25
support aws marketplace subscriptions

Vadman97:
## Summary

Support for setting up a self-serve AWS Marketplace subscription.
Applies usage-based pricing for the 4 products.

High level flow:
* Upon sign-up, AWS Marketplace initates a callback to https://pri.highlight.io/callback/aws-mp
* Handler resolves the AWS MP token to customer metadata. Handler doesn't have app.highlight.io auth context, so we store the metadata in a temporary cache under a random generated code. Handler redirects to app.highlight.io with code in query string.
* Frontend receives backend code and fires an authenticated mutation with the code and workspace id.
* Backend (private graph) uses code to look up aws customer metadata and associates it with the workspace.

AWS Lambda:
* We listen for SQS notifications per the [AWS docs](https://docs.aws.amazon.com/marketplace/latest/userguide/saas-integrate-contract-with-pay.html#saas-contract-with-pay-monitor-changes) to the subscription updates SNS topic to detect subscription changes.

Background metering:
* We now report usage as part of our background stripe metering job to AWS,

Some testing locally, but full integration testing to happen once this is released in prod per https://docs.aws.amazon.com/marketplace/latest/userguide/saas-integrate-contract-with-pay.html 

Deploying new lambda.

No


mayberryzane (approved):
🆒

PR #7583, opened 2024-01-25
Fix log attribute matching on body queries

ccschmitz:
## Summary

Matching attribute highlighting is broken on logs, but this fixes it! It improves the backend logic for querying matching logs to always use an `ILIKE` query when we have special characters in the body query and also ensures we use an `ILIKE` rather than a `LIKE` for other wildcard matches. Finally, it fixes some client logic for displaying the matching attributes on body queries since we now could have multiple body query filters rather than merging them all together.

**Before**
Screenshot 2024-01-25 at 3 16 59 PM

**After**
Screenshot 2024-01-25 at 3 16 16 PM

Updated tests to ensure backend logic is working as expected. Should keep an eye out for any body query behavior not working as expected.

N/A


SpennyNDaJets (approved):
Click test looks good to me!

PR #7584, opened 2024-01-25
Fix Github enhanced stacktraces for backend errors

SpennyNDaJets:
## Summary
Backend stacktraces were not being enhanced because the mapped stacktrace was `"null"`. Add a check to make sure we don't overwrite any stacktrace in this condition.

1) Integrate with Github (make sure not to connect to the prod app, reachout if you need to use my personal GitHub app)
2) Cause a backend error
- [ ] Error was enhanced successfully

https://www.loom.com/share/e9cc85b49734403988efd6b5e48d965b

N/A

N/A


Vadman97 (approved):
thanks for catching this!

PR #7588, opened 2024-01-26
fix discord notifications by adding emoji buttons

Vadman97:
## Summary

Discord recently changed the API to cause our go client to get an error when sending a button component with no emoji.
This must be because our client does not have an `omitEmpty` on the json serialization for the `emoji` key, but discord
must have made the value required when the key is present.

```json
{"message": "Invalid Form Body", "code": 50035, "errors": {"components": {"0": {"components": {"0": {"emoji": {"name": {"_errors": [{"code": "BUTTON_COMPONENT_INVALID_EMOJI", "message": "Invalid emoji"}]}}}}}}}}
```

Fixes the error by adding a highlight emoji to the button.

local testing
Screenshot 2024-01-25 at 4 46 02 PM

no

no

### DIFF
--- a/backend/alerts/integrations/discord/messages.go
+++ b/backend/alerts/integrations/discord/messages.go
@@ -18,6 +18,12 @@ func newMessageEmbed() *discordgo.MessageEmbed {
 var RED_ALERT = 0x961e13
 var YELLOW_ALERT = 0xf2c94c
 
+var highlightEmoji = discordgo.ComponentEmoji{
+	Name:     "highlight",
+	ID:       "1094349627953778788",
+	Animated: false,
+}
+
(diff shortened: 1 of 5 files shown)


SpennyNDaJets (approved):
Alerts are working! 🙌 

Should we create a card to clean up these alerts (i.e. the errors) since it looks a little overwhelming with all the emojis?
Screenshot 2024-01-26 at 10 07 26 AM

Vadman97:
yeah good point @SpennyNDaJets , though i don't think there is a fix without using some emoji right now (unless we fork the discord lib we use to fix the json serialization issue). https://github.com/highlight/highlight/issues/7590

PR #7594, opened 2024-01-26
Fix the width of the traces column popover

SpennyNDaJets:
## Summary
The traces popover changes width based on the longest attribute name.
- fix the width to 300px
- hide any text that is more than the 300px width
- add a popover to show the full text

Screenshot 2024-01-26 at 1 48 57 PM
Screenshot 2024-01-26 at 1 48 20 PM

N/A

Sure


Vadman97 (approved):
looks great in preview

PR #7617, opened 2024-01-30
Link repair

deltaepsilon:
We've got a bad link


Abhishek-More (changes requested):
I think you changed the DayOne link instead of the DayTwo link

Abhishek-More (approved):
Looks good!

PR #7629, opened 2024-01-31
Search traces by has errors

SpennyNDaJets:
## Summary
Allow users to search traces with errors

0) Create traces with errors

1) Load the traces page
2) Add the `has_errors` column
3) Search by `has_errors:true`
- [ ] Only traces with errors are displayed
4) Click into a trace
- [ ] Flame graph is highlighting the trace with the error

https://www.loom.com/share/2f0d4067a9874cb08bc1c7c2f15f04cb

N/A

N/A


frontend/src/pages/Traces/CustomColumns/columns.ts
```diff
@@ -56,14 +64,6 @@ const TIMESTAMP_COLUMN: TraceCustomColumn = {
 	accessKey: 'timestamp',
 }
 
-const START_TIME_COLUMN: TraceCustomColumn = {
```
SpennyNDaJets, line 59: Not used - just relevant in the context of sessions

frontend/src/pages/Traces/CustomColumns/columns.ts
```diff
@@ -48,6 +48,14 @@ const SECURE_SESSION_ID_COLUMN: TraceCustomColumn = {
 	accessKey: 'secureSessionID',
 }
 
+const HAS_ERRORS_COLUMN: TraceCustomColumn = {
+	id: 'has_errors',
+	label: 'Has Errors',
+	type: 'boolean',
+	size: '1fr',
+	accessKey: 'hasErrors',
+}
```
ccschmitz, line 57: Agree with your comment about displaying this in a better way in the future. I think we could show whether a trace has errors in a subtle but obvious way rather than giving it an entire column. Do we have a ticket for that?

ccschmitz (approved):
Code + local click test look great! ✨

PR #7633, opened 2024-02-01
allow loading traces for publicly-shared sessions

Vadman97:
## Summary

Our ShowHN links a [session](https://app.highlight.io/1383/sessions/COu90Th4Qc3PVYTXbx9Xee7xbQqN?network-resource-id=25&page=1&query=and%7C%7Ccustom_processed%2Cis%2Ctrue&ts=5.243) that has a flame graph,
but the flame graph does not load despite the session being public.
Update the Trace resolver to consider whether the session is public when loading data.

Local deploy

No

No


ccschmitz (approved):
Code + local click test look good!

PR #7634, opened 2024-02-01
fix ingest graphs, ignore column if count metric is used

mayberryzane:
## Summary
- fixes graphs shown on ingest filters page
- fixes "count" query if a metric is provided by ignoring that metric

- clicktested ingest page + grafana

- no

- no

### DIFF
--- a/backend/clickhouse/query.go
+++ b/backend/clickhouse/query.go
@@ -429,7 +429,7 @@ func getFnStr(aggregator modelInputs.MetricAggregator, column string) string {
 	case modelInputs.MetricAggregatorCount:
 		return "round(count() * any(_sample_factor))"
 	case modelInputs.MetricAggregatorCountDistinctKey:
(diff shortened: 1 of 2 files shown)


backend/clickhouse/query.go
```diff
@@ -516,6 +512,20 @@ func readMetrics[T ~string](ctx context.Context, client *Client, sampleableConfi
 		metricExpr = "1.0"
 	}
 
+	needsValue := false
+	for _, metricType := range metricTypes {
```
Vadman97, line 516: does the `tableConfig.defaultFilter` exclude the `IsIngestedBy` traces for sampling?

Vadman97 (approved):
thank you! was just looking at fixing this in https://github.com/highlight/highlight/pull/7549/commits/2386af17bdba6bd81c5dead4a7db9911414ba650

backend/clickhouse/query.go
mayberryzane: yep, this is a little hacky but this skips applying the default `<> Internal` filter

PR #7641, opened 2024-02-02
Show trace even if no root span

ccschmitz:
## Summary

Falls back to a span in a trace even if there is no root span. This makes it possible to render a trace even if the root span isn't available, which we've noticed happening because of issues with our sampling logic.

Local + PR preview click test. I viewed [a problematic trace](https://app.highlight.io/1/traces/84964c51a1df8e9ae344b60dbd5ddeeb/c78b971eb0bcb4ac) in prod and in the PR preview to ensure it was working as expected.

Screenshot 2024-02-02 at 2 06 49 PM

N/A

N/A


SpennyNDaJets (approved):
image

PR #7645, opened 2024-02-05
use ReplacingMergeTree for logs engine

mayberryzane:
## Summary
- creates a separate logs_new table using a ReplacingMergeTree for deduplication
- attaches all partitions of logs to logs_new
- exchanges the table names

- ran migration locally
- made sure dropping the old table after exchanging doesn't affect the new table
- confirmed queries against the new table were being deduplicated with FINAL
- confirmed data was being ingested and dependent materialized views were correctly updated

- once this is approved, will temporarily disable the public workers and run these scripts manually against prod
- scripts should complete very quickly (< 5s), after which we can restart the public workers

- no

### DIFF
--- a/backend/clickhouse/migrations/000077_create_logs_replacing_merge_tree.down.sql
+++ b/backend/clickhouse/migrations/000077_create_logs_replacing_merge_tree.down.sql
@@ -1,0 +1,3 @@
+EXCHANGE TABLES logs
+AND logs_new;
+DROP TABLE IF EXISTS logs_new;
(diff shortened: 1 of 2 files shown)


backend/clickhouse/migrations/000077_create_logs_replacing_merge_tree.down.sql
```diff
@@ -0,0 +1,3 @@
+EXCHANGE TABLES logs
+AND logs_new;
+DROP TABLE IF EXISTS logs_new;
```
Vadman97, line 3: would going the other way technically require `attach partition` commands here? maybe we can have a comment to remember this for the future

Vadman97 (approved):

backend/clickhouse/migrations/000077_create_logs_replacing_merge_tree.down.sql
mayberryzane: the partitions stay attached to the original table too - essentially the pointer to the part gets copied over to the new table but stays on the old one as well

PR #7656, opened 2024-02-05
Logs: add link to related trace

SpennyNDaJets:
## Summary
Allow users to see the related trace from the logs

1) View a log
2) Expand the row
3) Click on related trace link
- [ ] Opens traces search page with relevant filters

https://www.loom.com/share/d6c0739dbf684aec8d6949eca7a3c291?sid=67b9efd5-c305-4a22-b158-101ed649b486

N/A

N/A

### DIFF
--- a/backend/clickhouse/traces.go
+++ b/backend/clickhouse/traces.go
@@ -3,9 +3,10 @@ package clickhouse
 import (
 	"context"
 	"fmt"
-	"github.com/highlight-run/highlight/backend/parser/listener"
 	"strings"
 	"time"
+
+	"github.com/highlight-run/highlight/backend/parser/listener"
 
 	"github.com/highlight/highlight/sdk/highlight-go"
 
@@ -70,6 +71,7 @@ var traceColumns = []string{
 var defaultTraceKeys = []*modelInputs.QueryKey{
 	{Name: string(modelInputs.ReservedTraceKeyTraceID), Type: modelInputs.KeyTypeString},
 	{Name: string(modelInputs.ReservedTraceKeySpanName), Type: modelInputs.KeyTypeString},
+	{Name: string(modelInputs.ReservedTraceKeySpanID), Type: modelInputs.KeyTypeString},
 	{Name: string(modelInputs.ReservedTraceKeyDuration), Type: modelInputs.KeyTypeNumeric},
 	{Name: string(modelInputs.ReservedTraceKeyParentSpanID), Type: modelInputs.KeyTypeString},
 	{Name: string(modelInputs.ReservedTraceKeySecureSessionID), Type: modelInputs.KeyTypeString},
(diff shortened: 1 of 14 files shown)


frontend/src/pages/LogsPage/LogsTable/LogDetails.tsx
```diff
@@ -299,6 +317,25 @@ export const LogDetails: React.FC<Props> = ({
 							</Box>
 						</LinkButton>
 					)}
+					{row.original.node.traceID &&
+						row.original.node.source === 'backend' && (
```
SpennyNDaJets, line 321: Added this check since we do not record frontend traces - there is a few cases that we do not write a [trace](https://github.com/highlight/highlight/blob/main/backend/otel/otel.go#L198)
- metric events
- logging events

I did notice that we still add a trace_id and span_id for logs [in handleTrace](https://github.com/highlight/highlight/blob/main/backend/otel/otel.go#L265-L266) and [in handleLog](https://github.com/highlight/highlight/blob/main/backend/otel/otel.go#L439-L440). I'm not sure which one, but should we remove this link from one since the trace may not exists? Or could there be a outer trace used?

SpennyNDaJets: It seems like the log may still be linked to a previously created trace, but we do not want to create a span for this log. This may result in some logs appearing to have a related trace, but none in fact, if it is the only span in the trace 🤔

Vadman97: believe it should be fine to stop writing a `trace id` or a `span id` for logs when they do not have one, rather than creating one. perhaps there is something else we can use to remember that this is a trace-less log record though?

Vadman97 (commented):
have a case where a lot shows no traces, is this expected?
https://app.highlight.io/1/logs?query=http%20level%3Derror%20 -> 

https://www.loom.com/share/c058099f57af4fc4a65e414b76775f30

backend/otel/otel.go
```diff
@@ -263,7 +263,6 @@ func (o *Handler) HandleTrace(w http.ResponseWriter, r *http.Request) {
 						logRow := clickhouse.NewLogRow(
 							fields.timestamp, uint32(fields.projectIDInt),
 							clickhouse.WithTraceID(traceID),
-							clickhouse.WithSpanID(spanID),
```
SpennyNDaJets, line 266: Seems like we should not include these since they will not have a related span in the traces search

frontend/src/pages/Traces/CustomColumns/columns.ts
```diff
@@ -103,6 +111,7 @@ export const HIGHLIGHT_STANDARD_COLUMNS: Record<string, TraceCustomColumn> = {
 	service_version: SERVICE_VERSION_COLUMN,
 	trace_id: TRACE_ID_COLUMN,
 	parent_span_id: PARENT_SPAN_ID_COLUMN,
+	span_id: SPAN_ID_COLUMN,
```
SpennyNDaJets, line 114: Support this column, since it is not supported in filtering keys

backend/clickhouse/traces.go
```diff
@@ -281,6 +283,41 @@ func (client *Client) ReadTraces(ctx context.Context, projectID int, params mode
 	}, nil
 }
 
+func (client *Client) ExistingTraceIds(ctx context.Context, projectID int, traceIDs []string) ([]string, error) {
```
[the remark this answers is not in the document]
SpennyNDaJets: c.c. @ccschmitz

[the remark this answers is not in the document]
SpennyNDaJets: Synced with Chris here and we think this is good for our use case and in-relation to the flyout (i.e. a trace could have hundreds of spans)

Vadman97, line 286: to check my understanding here, essentially this is a way to bulk query for traces by id, but only returning the ids (rather than the full trace rows)

Vadman97 (approved):
looks great locally! just running into some weird behavior with the span in the trace that's being loaded, but it might be unrelated to your actual changes
https://www.loom.com/share/94f076493e9b4d22ab62ec50feee9980

SpennyNDaJets:
@Vadman97 I'll take a look! So we just link to the trace without a span id, since the log's span id most likely does not exist. There may be an issue with finding the incorrect span (if clashing) or the span may fall outside of the time frame we are providing

PR #7659, opened 2024-02-05
Don't redirect to projects outside workspace

ccschmitz:
## Summary

Fixes a bug where we would redirect a user to a project from a different workspace after switching workspaces. The new logic ensures we wipe the project ID we store in local storage if it doesn't not exist in the current workspace. It also enhances the back button by showing the project name references in local storage even if you are on a workspace settings page.

https://www.loom.com/share/d91b88e3c9a943a39e0221513d930460

N/A

N/A


SpennyNDaJets (approved):
This looks great!

PR #7663, opened 2024-02-05
Fix zoom controls on trace flame graphs

ccschmitz:
## Summary

Fixes a couple issues with the zoom controls on trace flame graphs.

1. The `+` and `-` buttons were not working.
2. The scroll bar was jumping around a lot and causing some jank when zooming in/out using the `wheel` event or pinch to zoom.

https://www.loom.com/share/7da4ab626a0e4be7902847dd0b3b17cc

N/A

N/A


frontend/src/pages/Traces/TraceFlameGraph.tsx
```diff
@@ -124,10 +124,10 @@ export const TraceFlameGraph: React.FC = () => {
 					(svgContainerRef.current?.scrollLeft ?? 0) *
 						(newZoom / prevZoom)
 
-				setTimeout(() => {
+				requestAnimationFrame(() => {
```
Vadman97, line 127: seems like if i scroll with mouse wheel too quickly, the graph still jumps to a random spot
https://www.loom.com/share/7b19c395b1824d6d8fd850a5dcda8b86

Vadman97 (approved):
much better now! just finding that too fast of a mouse scroll can cause random jumps, but this is already much better than before

frontend/src/pages/Traces/TraceFlameGraph.tsx
ccschmitz: I think I got this sorted out in https://github.com/highlight/highlight/pull/7663/commits/4e9987d13d33cec0a7444bbd2d3bb934707b166d! Also added support for scrolling sideways while holding the ctrl key and improved alignment of the spans when using the buttons to zoom in and out. Will merge once it's green!

PR #7671, opened 2024-02-06
Custom columns: popover updates

SpennyNDaJets:
## Summary
Make the following updates to the custom column popover:
- add loading state
- add empty state
- remove tooltip and rely on title

1) Load the traces or logs page
2) Open the custom column selector on the far right of the table
3) Hover over a key
- [ ] More discrete tooltip is displayed
3) Type in a query
- [ ] Loading state shown
4) Type in a key that does not exist
- [ ] Empty state shown

General
- [ ] No changes to PopoverSelect on Query builder

N/A

Sure

### DIFF
--- a/frontend/src/components/CustomColumnPopover/index.tsx
+++ b/frontend/src/components/CustomColumnPopover/index.tsx
@@ -3,18 +3,17 @@ import {
 	useGetTracesKeysLazyQuery,
 } from '@graph/hooks'
 import {
-	Box,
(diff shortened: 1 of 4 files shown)


SpennyNDaJets:
> One quick thought - do you think we should have a line separating the text input and the empty/loading state content? Feels a little off without some divider there, but that's subjective, so don't hold up merging or anything unless you agree or want to as Julian.

Good catch! Adding this in

PR #7678, opened 2024-02-06
support conversion of capacitor assets

Vadman97:
## Summary

[Capacitor](https://capacitorjs.com/) references `capacitor://localhost` urls that cannot be referenced remotely
and in some cases cannot be inlined on the client.
Support fetching them from another static host where they are publicly available.
In the future, we can support replacing them with an s3 bucket upload, similar to our sourcemap uploader.

Unit tests
Only enabled with asset replacement on and for a specific project.

No

No


backend/event-parse/parse.go
```diff
@@ -459,16 +464,23 @@ func getOrCreateUrls(ctx context.Context, projectId int, originalUrls []string,
...
 		parsedUrl.Fragment = ""
-		urlMap[u] = parsedUrl.String()
+		assetKey := parsedUrl.String()
+		assetURL := u
+		if (projectId == 33914 || projectId == 33886) && parsedUrl.Scheme == "capacitor" {
+			parsedUrl.Scheme = "https"
+			parsedUrl.Host = "app.priceworx.co.uk"
+			assetURL = parsedUrl.String()
+		}
+		urlMap[u] = AssetValue{assetKey, assetURL}
```
SpennyNDaJets, line 474: Wanted to sum up my understanding:

1. Store all the new asset value under the key `u` so that we can look it up by its original url
2. The old value of the dictionary is moved to `assetKey`, but overwritten for capacitor urls. This `u.Value.AssetKey` is used in place of `u.Value`.
3. The key of the dictionary key is reiterated in `assetURL`, but overwritten for capacitor urls. This `u.Value.URL` is used in place of `u.Key`.

Just had to wrap my head around this logic cause the flipping of the naming was a little confusing, but makes sense and looks equivalent for the old case.

SpennyNDaJets (approved):
LGTM!

PR #7685, opened 2024-02-07
new flattened format for log messages

mayberryzane:
## Summary
- creates a new `RetryableMessage` interface, implemented by `Message` and `LogRowMessage`
- `LogRowMessage` has a flattened format to make it easier to use with Kafka Connect
- includes consumer code to handle both new and old formats

- clicktested locally to make sure logs and other kafka messages were being produced/consumed correctly
- ensured `LogRowMessage` format matched what was expected

- new format messages will be dropped in case of revert, will quickly verify that logs are still being ingested properly after deployment

- no

### DIFF
--- a/backend/kafka-queue/kafkaqueue.go
+++ b/backend/kafka-queue/kafkaqueue.go
@@ -60,8 +60,8 @@ type Queue struct {
 
 type MessageQueue interface {
 	Stop(context.Context)
-	Receive(context.Context) *Message
-	Submit(context.Context, string, ...*Message) error
+	Receive(context.Context) RetryableMessage
+	Submit(context.Context, string, ...RetryableMessage) error
 	LogStats()
 }
 
@@ -303,7 +303,7 @@ func (p *Queue) Stop(ctx context.Context) {
 	}
 }
 
-func (p *Queue) Submit(ctx context.Context, partitionKey string, messages ...*Message) error {
+func (p *Queue) Submit(ctx context.Context, partitionKey string, messages ...RetryableMessage) error {
 	if len(messages) == 0 {
 		return nil
 	}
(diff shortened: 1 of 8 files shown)


backend/kafka-queue/kafkaqueue.go
```diff
@@ -315,7 +315,8 @@ func (p *Queue) Submit(ctx context.Context, partitionKey string, messages ...*Me
 
 	var kMessages []kafka.Message
 	for _, msg := range messages {
-		msg.MaxRetries = TaskRetries
+		msg.SetMaxRetries(TaskRetries)
+		// msg.MaxRetries = TaskRetries
```
Vadman97, line 319: old?

backend/kafka-queue/kafkaqueue.go
```diff
@@ -452,22 +453,37 @@ func (p *Queue) LogStats() {
...
 
-func (p *Queue) deserializeMessage(compressed []byte) (msg *Message, error error) {
+func (p *Queue) deserializeMessage(compressed []byte) (RetryableMessage, error) {
 	if int64(len(compressed)) >= p.MessageSizeBytes {
 		return nil, errors.New("message too large")
 	}
+	var msgType struct {
+		Type PayloadType
+	}
+	if err := json.Unmarshal(compressed, &msgType); err != nil {
```
Vadman97, line 471: just curious, what's the performance impact of this extra unmarshal? is it as slow as the second unmarshal on line 482?

Vadman97 (approved):
:ship: very cool!

backend/kafka-queue/kafkaqueue.go
mayberryzane: there is probably some performance impact, could also unmarshal to the default message type and then only marshal to the new type in case it's a flattened message.

I think ideally this type information would be encoded in the key. Do we rely on the key for anything at the moment? Or can we manually choose a partition to use regardless of the key?

PR #7690, opened 2024-02-07
Search errors by trace and session ids

SpennyNDaJets:
## Summary
Add two new error instance attributes to search by:
- `trace_id`
- `secure_session_id`

These will be needed in the new search experience, as well as to better link to the errors page from a session or trace.

1) View the errors
2) Filter out errors by trace id
- [ ] "Trace ID" key is displayed
- [ ] `trace_id` options can be chosen
- [ ] Correct options to filter by
- [ ] Correct filters
3) Filter out errors by secure session id
- [ ] "Secure Session ID" key is displayed
- [ ] `secure_session_id` options can be chosen
- [ ] Correct options to filter by
- [ ] Correct filters

https://www.loom.com/share/049e78dff7d04f7080659ee6da9c48b2?sid=91bf5ccc-ba0b-45eb-961d-bfc6a99bbc0f

Confirm switch of key name will not break anything in grafana

N/A


backend/private-graph/graph/schema.graphqls
```diff
@@ -980,9 +980,9 @@ enum ReservedErrorObjectKey {
 	os
 	payload
 	request_id
+	secure_session_id
```
SpennyNDaJets, line 983: I wanted to keep this consistent with logs and traces to not confuse the user with whether they should use `secure_session_id` or `session_secure_id`. This is not being used in the product, but wanted to confirm that this will not break anything in Grafana? c.c. @mayberryzane

mayberryzane (approved):

PR #7702, opened 2024-02-08
Tracking updates for related resources

ccschmitz:
## Summary

Adds some tracking for page views and key interactions on sessions, errors, logs, and traces. These changes will allow us to see how users are navigating to related resources, and glean other insights in the future about how different parts of the product are being used.

This also adds a required argument to `analytics.page` for the name of the page and introduces some new tracking conventions for analytics events. Very open to feedback on these conventions.

### Page View Tracking

All calls to `analytics.page` now require a `name` argument. The name should be title cased and just the name of the object being referenced. e.g. 'Traces', 'Logs', 'Sessions', 'Session', etc. Note that I do consider 'Sessions' and 'Session' a page view, but we don't consider switching to the 'Errors' tab on a trace a view, that is tracked as an `analytics.track` event.

The reason I didn't follow the same convention as with event tracking (see below) is because the `name` argument just controls the `title` property on a standard track event. Page views don't require the `context` or `action` since they are higher level and `view` is the assumed action.

### Event Tracking

Event tracking follows the `Object` => `Context` => `Action` pattern for naming events. Each section name is kebab cased and separated with an underscore (`_`). e.g. `session_console-logs_view`, `error_related-session_click`, etc.

I click around the following areas of the app ensuring I saw network requests for tracking as I expected.

### Sessions

* [ ] Errors
* [ ] Logs
* [ ] Traces

### Errors

* [ ] Session
* [ ] Logs
* [ ] Traces

### Logs

* [ ] Session
* [ ] Errors
* [ ] Trace

### Traces

* [ ] Session
* [ ] Errors
* [ ] Logs

N/A - will finish building Mixpanel dashboard once this is deployed.

N/A


frontend/src/pages/Player/RightPlayerPanel/components/NetworkResourcePanel/NetworkResourcePanel.tsx
```diff
@@ -248,11 +248,12 @@ function NetworkResourceDetails({
 	}, [resource.id])
 
 	useEffect(() => {
-		analytics.page(
-			`/sessions/${sessionSecureId}/network-resource/${
+		analytics.page('Network Resource Details', {
```
Vadman97, line 251: as we discussed, does it make sense to update all the `.page()` calls with page names?

Vadman97 (approved):

frontend/src/pages/Player/RightPlayerPanel/components/NetworkResourcePanel/NetworkResourcePanel.tsx
ccschmitz: Updated this and added `page: string` as a required argument on `analytics.page` to ensure we always do this.

PR #7704, opened 2024-02-08
Create partitioned trace_by_id table

SpennyNDaJets:
## Summary
Search multiple traces by id is slow since it needs to search many parts in clickhouse. Create a new table to partition the data by the first character of the UUID

1) Load Clickhouse locally (localhost:8123/play)
2) Confirm the `traces_by_id` and `traces_by_id_new` table match

1) Deploy to create new table and materialized view
2) Find latest time stamp and copy over data from old table
3) Exchange tables

N/A


backend/clickhouse/migrations/000082_create_new_traces_by_id_mv.up.sql
```diff
@@ -0,0 +1,60 @@
...
+    `Events.Timestamp` Array(DateTime64(9)),
+    `Events.Name` Array(LowCardinality(String)),
+    `Events.Attributes` Array(Map(LowCardinality(String), String)),
+    `Links.TraceId` Array(String),
+    `Links.SpanId` Array(String),
+    `Links.TraceState` Array(String),
+    `Links.Attributes` Array(Map(LowCardinality(String), String))
+)
+    ENGINE = MergeTree
+        PARTITION BY substring(toString(UUID), 1, 1)
```
SpennyNDaJets, line 29: Does it make sense to partition by the UUID if then a single trace could have spans across different partitions?

mayberryzane: I think it should be partitioned by traceId instead. Because traceId can be either a UUID or one of our own requestIds, maybe using something like `farmHash64(TraceId) % 100` would be best? I think we saw that a reduction of about 100x would work well for our own project

SpennyNDaJets: @mayberryzane I like that, and makes sense based on the trace id! Quick question: this would essentially bring us from 1 -> 100 partitions. Curious if we have a method for deciding how many partitions is the right amount?

mayberryzane (approved):

backend/clickhouse/migrations/000082_create_new_traces_by_id_mv.up.sql
Vadman97: makes sense why we should use the hash fn. does it make more sense to use TraceId here because we would want to load multiple spans for a given trace id from the same part rather than across parts (which would be the case if we partition by UUID)?

SpennyNDaJets: That is my understanding, but will let Zane fill in

Resolving to merge

PR #7709, opened 2024-02-08
fix hobby deploy routing logs/traces to self

Vadman97:
## Summary

When running the hobby deploy with `./run-hobby.sh`, logs and traces would be sent to the stack itself
rather than to our cloud collector (when telemetry was enabled) because of the default value of the
`OTLP_ENDPOINT`. Correct this to allow collecting phonehome telemetry data.
Fixes recent regression that would prevent local dev from recording sessions.

before
Screenshot 2024-02-08 at 12 34 33 PM

after
Screenshot 2024-02-08 at 12 36 29 PM

Screenshot 2024-02-08 at 12 36 17 PM

https://www.loom.com/share/90b4ade1a6c749af93d10b8b54cbd35e

no

no


SpennyNDaJets (approved):
image

PR #7731, opened 2024-02-12
clean up error filtering logic

Vadman97:
## Summary

Remove manual error filter patterns to be replaced with system configuration `ErrorFilters` entries
as part of the sampling logic determining ingest.
Ensure frontend error ingestion also goes through the same ingest sampling logic as backend errors.

Local deploy ingesting errors when no filter is applied, and not ingesting when filters are applied
Screenshot from 2024-02-12 12-34-24

converting static filters to project / global ones in the db as needed

no

### DIFF
--- a/backend/public-graph/graph/resolver.go
+++ b/backend/public-graph/graph/resolver.go
@@ -2186,7 +2186,11 @@ func (r *Resolver) ProcessBackendPayloadImpl(ctx context.Context, sessionSecureI
 
 		group, err := r.HandleErrorAndGroup(ctx, errorToInsert, structuredStackTrace, extractErrorFields(session, errorToInsert), projectID, workspace)
 		if err != nil {
-			log.WithContext(ctx).WithError(err).Error("Error updating error group")
+			if e.Is(err, ErrQuotaExceeded) || e.Is(err, ErrUserFilteredError) {
+				log.WithContext(ctx).WithError(err).Info("Will not update error group")
+			} else {
+				log.WithContext(ctx).WithError(err).Error("Error updating error group")
+			}
 			continue
 		}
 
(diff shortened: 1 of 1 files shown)


backend/public-graph/graph/resolver.go
```diff
@@ -2781,12 +2723,29 @@ func (r *Resolver) ProcessPayload(ctx context.Context, sessionSecureID string, e
 				StackTrace:     &traceString,
 				Timestamp:      v.Timestamp,
 				Payload:        v.Payload,
-				RequestID:      nil,
 				IsBeacon:       isBeacon,
 				ServiceVersion: serviceVersion,
 				ServiceName:    sessionObj.ServiceName,
 			}
 
+			if !r.IsErrorIngested(ctx, projectID, &publicModel.BackendErrorObjectInput{
```
SpennyNDaJets, line 2731: I noticed that downstream we call `isIngestedByRateLimit` - will we be double counting errors toward this rate limit? Or were these previously not being counted?

SpennyNDaJets (approved):
Looks good to me! Do we need to translate these filtered out errors to their filters or give them an update incase they start noticing changes?

backend/public-graph/graph/resolver.go
```diff
@@ -2238,15 +2188,7 @@ func (r *Resolver) ProcessBackendPayloadImpl(ctx context.Context, sessionSecureI
...
-			if e.Is(err, ErrNoisyError) {
-				log.WithContext(ctx).Warn(e.Wrap(err, "Error updating error group"))
-			} else if e.Is(err, ErrQuotaExceeded) {
-				log.WithContext(ctx).Warn(e.Wrap(err, "Error updating error group"))
-			} else if e.Is(err, ErrUserFilteredError) {
-				log.WithContext(ctx).Info(e.Wrap(err, "Error updating error group"))
-			} else {
-				log.WithContext(ctx).WithError(err).Error(e.Wrap(err, "Error updating error group"))
-			}
+			log.WithContext(ctx).WithError(err).Error("Error updating error group")
```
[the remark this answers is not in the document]
Vadman97: good catch - cleaning this up

backend/public-graph/graph/resolver.go
Vadman97: these errors (frontend errors) were incorrectly not counted previously / not filtered out. this is the only place frontend errors would go thru the sampling check, so they won't be double counted.

Vadman97:
> Looks good to me! Do we need to translate these filtered out errors to their filters or give them an update incase they start noticing changes?

thanks! will be migrating these as needed to the ingestion filters

PR #7732, opened 2024-02-12
Display exists values in search

SpennyNDaJets:
## Summary
When a user is searching if attribute exists on a log row, there is no feedback to which rows are matching. Add support to display row when searching if an attribute exists.

1) Load the logs page
2) Search for an attribute existing
- [ ] Results display that attribute key/value under the row

Screenshot 2024-02-12 at 5 38 48 PM

N/A

N/A


ccschmitz (changes requested):
One thing I'm noticing right away is that we show the `message` attribute as always matching even when there are no filters.

Screenshot 2024-02-13 at 5 01 17 PM

Another problem I'm seeing is that if you use `!=` all those attributes will still show up as matching still.

Screenshot 2024-02-13 at 5 03 17 PM

I'm guessing there are still some edge cases we'll run into with the new operators like `!=`, but wondering if we want to keep the old logic for `=` and maybe create new code paths for each operator to determine if it's matching?

ccschmitz (approved):
Looks good to me! ✨

PR #7735, opened 2024-02-13
fix text obfuscation logic

Vadman97:
## Summary

innerHTML changes would not be obfuscated properly in default obfuscation mode.
https://github.com/highlight/rrweb/pull/116

fixes HIG-3970

before
Screenshot from 2024-02-13 09-49-45

after
Screenshot from 2024-02-13 09-54-40

new changeset

no

### DIFF
--- a/frontend/src/pages/Buttons/Buttons.tsx
+++ b/frontend/src/pages/Buttons/Buttons.tsx
@@ -26,6 +26,7 @@ import {
 export const Buttons = () => {
 	const [hasError, setHasError] = useState(false)
 	const [showWebSocket, setShowWebSocket] = useState(false)
+	const [email, setEmail] = useState('')
 	const [sendEmail, { loading }] = useSendEmailSignupMutation()
 	if (hasError) {
 		throw new Error('this is a buttons error', {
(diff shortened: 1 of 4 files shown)


frontend/src/pages/Buttons/Buttons.tsx
```diff
@@ -129,6 +130,48 @@ export const Buttons = () => {
...
+								.toString(36)
+								.substring(7)}`,
+						)
+					}}
+				>
+					set email
+				</button>
+				<div>
+					<b>hello, this is a b tag</b>
+					<b key="email-1">{email}</b>
```
SpennyNDaJets, line 160: @Vadman97 have you been able to reproduce the bug? I've been messing with some of the state, but the elements are being obfuscated through the `adds` nodes rather than the `texts`. Been messing with some css to see if I can reproduce better

SpennyNDaJets (approved):
Overall think this is the fix we want. Not sure if we want to include in this PR but from our conversation, we still need to:
- [ ] Confirm our edge case
  - incremental update through `texts`, only been able to update through `adds` which was working properly
  - see comment above
- [ ] Add end-to-end tests

frontend/src/pages/Buttons/Buttons.tsx
Vadman97: got it to repro not by using the react state, but by manually setting `innerText` see https://github.com/highlight/highlight/blob/669886629b7bfdefa079c5c097899a2e7ceabb6e/frontend/src/pages/Buttons/Buttons.tsx#L137-L141 and the screenshots above

Vadman97: i'm guessing the customer's app is using jquery or some other library to directly edit the DOM innerText, hence we are seeing this bug for them

Vadman97:
better e2e tests coming in https://linear.app/highlight/issue/HIG-4345/e2e-test-for-obfuscation

PR #7736, opened 2024-02-13
enforce +/- 2 hour difference for ingested logs and traces

mayberryzane:
## Summary
- this revives a PR from a few months ago that was applying this to logs
- changed limit to 2 hour to account for any potential daylight savings time weirdness
- expecting this to help clickhouse merge performance as effort won't have to be spent merging large parts in old partitions as new data won't be ingested for those partitions anymore

- unit test, clicktested locally to make sure new logs + traces were ingested

- no, can revert if issues

- no

### DIFF
--- a/backend/otel/extract.go
+++ b/backend/otel/extract.go
@@ -74,6 +74,7 @@ type extractFieldsParams struct {
 	event     *ptrace.SpanEvent
 	scopeLogs *plog.ScopeLogs
 	logRecord *plog.LogRecord
+	curTime   time.Time
 }
 
 func extractFields(ctx context.Context, params extractFieldsParams) (*extractedFields, error) {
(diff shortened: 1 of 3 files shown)


Vadman97 (approved):
do we have an idea what the out-of-bounds logs / traces look like? are they a potential problem because of some other time converstion, ie. with heroku / render.com timestamps not being parsed correctly?

mayberryzane:
> do we have an idea what the out-of-bounds logs / traces look like? are they a potential problem because of some other time converstion, ie. with heroku / render.com timestamps not being parsed correctly?

@Vadman97 
The vast majority of these are `Client` spans - assuming these are just browsers with incorrectly set clocks.
The rest are `Internal` or blank - not too sure what's going on with those but it's a tiny percentage of all spans

```
select SpanKind, count()
from traces
where Timestamp >= now() + interval '2 hours'
group by 1
order by 2 desc
limit 10

Client   | 110671510
Internal | 287988
         | 15030
```

PR #7737, opened 2024-02-13
skip trace_attributes table, write directly to trace_key_values

mayberryzane:
## Summary
- we don't use `trace_attributes` for anything and it uses a large number of parts, contributing to performance issues for trace ingest
- write key/value counts directly to `trace_key_values` (about 1000x fewer data)

- ran migration locally, ensured no new rows in `trace_attributes`, new rows for static and custom keys in `trace_key_values`

- no

- no

### DIFF
--- a/backend/clickhouse/migrations/000084_rewrite_trace_attributes_mv.up.sql
+++ b/backend/clickhouse/migrations/000084_rewrite_trace_attributes_mv.up.sql
@@ -14,7 +14,6 @@ SELECT ProjectId,
 FROM traces
 WHERE (
         Key NOT IN (
-            'level',
             'trace_state',
             'span_name',
             'span_kind',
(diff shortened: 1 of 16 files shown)


backend/clickhouse/migrations/000084_rewrite_trace_attributes_mv.up.sql
```diff
@@ -0,0 +1,35 @@
...
+            'span_name',
+            'span_kind',
+            'service_name',
+            'service_version',
+            'message',
+            'secure_session_id',
+            'span_id',
+            'trace_id',
+            'parent_span_id',
+            'duration'
```
SpennyNDaJets, line 28: Is this a good time time to sync these with the columns in https://github.com/highlight/highlight/blob/main/backend/clickhouse/traces.go#L49?

Think this includes `has_errors` and `environment`, but not sure if we want to include the others:
- would we need to include `uuid`, `project_id`, `timestamp`?
- `status_code` and `status_message`? Not sure we use these
- I don't believe `level` is a column we use in traces, but we are filtering out?

SpennyNDaJets (approved):
LGTM

backend/clickhouse/migrations/000084_rewrite_trace_attributes_mv.up.sql
mayberryzane: - I don't think we need to include uuid, project_id, timestamp since we don't need search suggestions for those
- not sure how `status_code` or `status_message` are being used rn - but we don't currently write those in a materialized view so it shouldn't be necessary to add them to the filter here
- agree that we don't need to filter out level, removing that

PR #7753, opened 2024-02-14
Improve display of search query errors

ccschmitz:
## Summary

Cleans up the display of errors in the UI.

* Instead of skipping whitespace in our grammar, we push it to a hidden channel allowing us to still get tokens for whitespace from the lexer.
* Updated the logic for creating the visual tags around filters. This no longer is directly linked to the expressions we get back from the parser. This allows us to handle the grouping better visually even though the parser can't break things into expressions for us.
* Moves the display of the errors above the search input. This required some changes to the `Tooltip` component allowing it to be wider and adjust the `z-index` to ensure it displays over the header links.
* Adds a red background to the token causing an error.
* Fixes a spacing issue we currently have in production when we have consecutive space (`" "`) characters.

Typed a number of queries in search on logs and traces. Ensured any examples I could think of looked and behaved as expected.

The only change that could impact the backend was the grammar change, but that should be a no-op still for all the backend logic since it only deals with parsing.

Will have @julian-highlight and @jay-khatri review the PR preview (see [Slack thread](https://highlightcorp.slack.com/archives/C0454PWLLLW/p1707945102098409)).


Vadman97 (approved):
wanted to call out some strange behavior i noticed with writing something more complex like `level=error OR (level=info AND service_name=private-graph)`
https://www.loom.com/share/8c4ffdfd01ca4fd5977c8525e7853216

PR #7766, opened 2024-02-15
report usage as part of phonehome metrics

Vadman97:
## Summary

Recent change to run metric monitor / alert monitor in self-hosted deploy (all workers mode)
would prevent refreshing materialized views from running, which would avoid reporting usage metrics.

CI build of docker image + VM hobby deploy.

new docker image release

no

### DIFF
--- a/backend/main.go
+++ b/backend/main.go
@@ -619,6 +619,9 @@ func main() {
 			go w.GetPublicWorker(kafkaqueue.TopicTypeBatched)(ctx)
 			go w.GetPublicWorker(kafkaqueue.TopicTypeDataSync)(ctx)
 			go w.GetPublicWorker(kafkaqueue.TopicTypeTraces)(ctx)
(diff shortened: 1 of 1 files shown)


backend/main.go
```diff
@@ -619,6 +619,9 @@ func main() {
 			go w.GetPublicWorker(kafkaqueue.TopicTypeBatched)(ctx)
 			go w.GetPublicWorker(kafkaqueue.TopicTypeDataSync)(ctx)
 			go w.GetPublicWorker(kafkaqueue.TopicTypeTraces)(ctx)
+			// for the 'All' worker, run alert / metric watchers
+			go w.StartLogAlertWatcher(ctx)
```
Vadman97, line 623: these calls block so we need to run them in a goroutine

SpennyNDaJets (approved):

PR #7773, opened 2024-02-16
Set up routing and flyout panel for related resources

ccschmitz:
## Summary

Lays the foundation for showing related resources in a flyout panel.

* Creates the resizable panel component for rendering related resources
* Creates a `useRelatedResource` hook for accessing/updating the related resource + panel information
* Updates the trace panel so that when it's resized the flame graph recalculates the layout

Note that this change does not have anything visible to the end user and the related resource pages will be built out in subsequent PRs.

https://www.loom.com/share/817905be29ed448b8a8c21ee6c4c86f2

I made some updates to the error and trace pages so that clicking to view the related resources would open them in the panel. I just had it open the existing page components in the panel, but I reverted those changes before putting this up for review.

N/A - all client-side changes and should be nothing visible to the end user in prod right now.

I did a design review with @julian-highlight on 2/26. Will make sure to have him review the new pages as they are implemented.


SpennyNDaJets (approved):
LGTM!

PR #7797, opened 2024-02-20
Improve logs related resource query

SpennyNDaJets:
## Summary
Update the `useGetLogsRelatedResourcesQuery` query to be more efficient
- switch from `traces_by_id` to `traces` table
- provide a date range to limit the number of partitions needed to be searched

1) Load the logs page
- [ ] the `useGetLogsRelatedResourcesQuery` returns with no errors
- [ ] Any logs with related correctly show the related trace button

Monitor performance in production

N/A

### DIFF
--- a/frontend/src/pages/LogsPage/useGetLogs.ts
+++ b/frontend/src/pages/LogsPage/useGetLogs.ts
@@ -119,6 +119,12 @@ export const useGetLogs = ({
 			project_id: project_id!,
 			log_cursors: data?.logs.edges.map((e) => e.cursor) || [],
 			trace_ids: logTraceIds,
+			date_range: {
+				start_date: moment(startDate)
+					.subtract(5, 'minutes')
+					.format(TIME_FORMAT),
+				end_date: moment(endDate).add(5, 'minutes').format(TIME_FORMAT),
+			},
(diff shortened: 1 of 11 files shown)


frontend/src/pages/LogsPage/useGetLogs.ts
```diff
@@ -106,19 +106,41 @@ export const useGetLogs = ({
...
+
+			if (!latestLogTime || latestLogTime < edge.node.timestamp) {
+				latestLogTime = edge.node.timestamp
 			}
-			return acc
-		}, [] as string[]) || []
+
+			if (!earliestLogTime || earliestLogTime > edge.node.timestamp) {
+				earliestLogTime = edge.node.timestamp
+			}
```
SpennyNDaJets, line 124: Maybe over engineered - chose to iterate through rather than using first and last timestamp incase we want to support sorting by other attributes in future

deltaepsilon: Consider using useMemo if this will be recalculated often.

deltaepsilon (approved):

mayberryzane (approved):
🆒

PR #7807, opened 2024-02-20
Improve request/response redactions

SpennyNDaJets:

### DIFF
--- a/sdk/client/src/listeners/network-listener/utils/utils.ts
+++ b/sdk/client/src/listeners/network-listener/utils/utils.ts
@@ -48,13 +48,33 @@ const sanitizeRequestResponsePair = (
 		requestResponseSanitizer,
 	}: SanitizeOptions,
 ): RequestResponsePair | null => {
+	// body keys are already be redacted at this point (see getBodyThatShouldBeRecorded)
 	let sanitizedPair: RequestResponsePair | null = pair
 
 	// step 1: pass through user defined sanitizer
 	if (requestResponseSanitizer) {
 		try {
+			try {
+				sanitizedPair.request.body = JSON.parse(
+					sanitizedPair.request.body,
+				)
+				sanitizedPair.response.body = JSON.parse(
+					sanitizedPair.response.body,
+				)
+			} catch (err) {}
+
 			sanitizedPair = requestResponseSanitizer(sanitizedPair)
-		} catch (err) {}
+		} catch (err) {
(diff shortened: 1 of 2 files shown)


sdk/client/src/listeners/network-listener/utils/utils.ts
```diff
@@ -48,13 +48,33 @@ const sanitizeRequestResponsePair = (
...
-		} catch (err) {}
+		} catch (err) {
+		} finally {
+			if (sanitizedPair) {
+				sanitizedPair.request.body = JSON.stringify(
+					sanitizedPair.request.body,
+				)
+				sanitizedPair.response.body = JSON.stringify(
+					sanitizedPair.response.body,
+				)
```
Vadman97, line 75: if `sanitizedPair.request.body` is a string and `requestResponseSanitizer` throws an error, wouldn't this double-stringify the string body? `sanitizedPair.request.body` can be a string on line 58, we then catch the json parse error, but then may have the `sanitizedPair.request.body` be a string (we shouldn't try stringify a string because that may quote things unexpectedly)

sdk/client/src/listeners/network-listener/utils/utils.ts
```diff
@@ -48,13 +48,33 @@ const sanitizeRequestResponsePair = (
 		requestResponseSanitizer,
 	}: SanitizeOptions,
 ): RequestResponsePair | null => {
+	// body keys are already be redacted at this point (see getBodyThatShouldBeRecorded)
 	let sanitizedPair: RequestResponsePair | null = pair
 
 	// step 1: pass through user defined sanitizer
 	if (requestResponseSanitizer) {
 		try {
+			try {
```
Vadman97, line 57: nested trys could be cleaned up into just trying the things that could actually error

Vadman97 (approved):
thanks for getting to this so quickly! 💪

sdk/client/src/listeners/network-listener/utils/utils.ts
SpennyNDaJets: Makes sense! I added a boolean to track if it was successfully converted to json to help with this scenario

sdk/client/src/listeners/network-listener/utils/utils.ts
SpennyNDaJets: Done!

PR #7808, opened 2024-02-20
React Server Components testbed + docs comments [HIG-4346]

deltaepsilon:
It turns out that RSCs have worked for a long while. They get filtered through `app/error.tsx` and caught by `appRouterSsrErrorHandler` from `@highlight-run/next/ssr`.

image


Vadman97 (approved):
cool! is this related to [server actions](https://nextjs.org/docs/app/building-your-application/data-fetching/server-actions-and-mutations#forms)? if it is, could we also add an example of a server action form with highlight error capture?

PR #7815, opened 2024-02-21
fix issues with mp metering

Vadman97:
## Summary

* For workspaces with an AWS MP subscription, report their usage rather than using stripe.
* Call the get entitlements api to check for entitlements
* Use the entitlements api to set workspace quotas per product

Local deploy, will be testing in production (only gated to AWS MP usage)

No

No


SpennyNDaJets (approved):
Code looks good to me and limited to AWS MP workspaces

PR #7825, opened 2024-02-22
add instructions for self-hosted MS Teams

mayberryzane:
## Summary
- adds section for self-hosted integrations and steps for configuring Teams

- testing in preview env

- no

- no


Vadman97 (approved):
nice! looks great; will share with the customer once this is live

PR #7833, opened 2024-02-23
propagate highlight trace id on logs

Vadman97:
## Summary

Logs do not propagate the trace id from spans, causing logs to be disconnected
from their parent span (as per the discord thread in linear).
Correctly propagates the trace id on logs based on the span context.

```bash
curl -d '{}' -H "x-highlight-request: abc123/def456" localhost:8000/not-found
```
image
image

New SDK patch version.

No


SpennyNDaJets (approved):
LGTM

PR #7835, opened 2024-02-23
Support matching regex in searches

SpennyNDaJets:
## Summary
Support matching regex in the logs and traces search, and add a shortcut operator.

https://www.loom.com/share/acd783b81a1d4df6825a8bb4f31f56e8?sid=75469963-6482-42bc-becd-710f19e87ea8

1) Navigate to the logs/traces page
2) Select a string key to search by
- [ ] "Matches" operator is displayed
- [ ] "Does not match" operator is displayed
4) Click on the "Contains" operator
- [ ] Correct operator is displayed `=//`
- [ ] Cursor is placed between `/`
5) Search for part of a string
- [ ] Correct results
6) Repeat for "does not match"
- [ ] "Does not contain" works as aspected

N/A

N/A

### DIFF
--- a/backend/parser/listener/listener.go
+++ b/backend/parser/listener/listener.go
@@ -21,6 +21,8 @@ var OperatorLike Operator = "LIKE"
 var OperatorNotLike Operator = "NOT LIKE"
 var OperatorILike Operator = "ILIKE"
 var OperatorNotILike Operator = "NOT ILIKE"
+var OperatorRegExp Operator = "REGEXP"
+var OperatorNotRegExp Operator = "NOT REGEXP"
 var OperatorContains Operator = "hasTokenCaseInsensitive"
 var OperatorGreaterThan Operator = ">"
 var OperatorGreaterThanOrEqualTo Operator = ">="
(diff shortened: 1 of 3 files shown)


backend/parser/listener/listener.go
```diff
@@ -314,7 +334,25 @@ func (s *searchListener[T]) appendRules(value string) {
...
+			if traceAttributeKey {
+				s.rules = append(s.rules, s.sb.Var(sqlbuilder.Buildf("NOT "+s.attributesColumn+"[%s] REGEXP %s", s.currentKey, value)))
+				s.ops = append(s.ops, &FilterOperation{
+					Key:      s.currentKey,
+					Column:   s.attributesColumn,
+					Operator: OperatorNotRegExp,
+					Values:   []string{value},
+				})
+			} else {
+				s.rules = append(s.rules, s.sb.Var(sqlbuilder.Buildf("NOT "+filterKey+" REGEXP %s", value)))
```
SpennyNDaJets, line 348: This differs from the other `NOT`'s because the NOT must come before the filter key. I'm not sure what the s.ops is used for below, but not sure if this edge case will break that. I did not notice any regressions when testing

mayberryzane: I wonder if we can simplify the code here by treating the `!=` operator as `NOT (<subquery>)` in all cases. I think the logic isn't 100% correct at the moment, for example `environment!=(prod OR "on-prem")` is treated as `environment!=prod OR environment!="on-prem"`.

Screen Shot 2024-02-26 at 9 37 31 AM

mayberryzane (approved):

mayberryzane:
also do we need to make changes to support this syntax in ingestion filters?
https://github.com/highlight/highlight/blob/main/backend/clickhouse/query.go#L366

backend/parser/listener/listener.go
SpennyNDaJets: This is a good callout! I've been looking at this change, and it feels a little risky and unrelated to these changes? I created [this card](https://linear.app/highlight/issue/HIG-4394/fix-nested-=-queries) to tackle this work separately

PR #7838, opened 2024-02-23
fix error grouping sessions / errors

mayberryzane:
## Summary
- fixes an issue `Missing columns: '_sample_factor'`
- don't need to sample in inner limit query - metric is only used for relative ordering, and multiplying all by the sample factor will not affect the order

- clicktested locally

- no

- no


Vadman97 (approved):
nice 🚢

PR #7854, opened 2024-02-27
Fix session alignment

SpennyNDaJets:
## Summary
The session secure id is not aligned with the other text in the trace and rows tables. They are also causing their rows to grow in height when they are not empty. Make some CSS tweaks.

## How did you test this change?
1) View the logs table
- [ ] Rows w/ and w/o sessions are the same height
- [ ] Session id is properly aligned in the cell
- [ ] Empty is properly aligned in the cell
2) View the traces table
- [ ] Rows w/ and w/o sessions are the same height
- [ ] Session id is properly aligned in the cell
- [ ] Empty is properly aligned in the cell

Screenshot 2024-02-27 at 12 02 40 PM
Screenshot 2024-02-27 at 12 02 19 PM

## Are there any deployment considerations?
N/A

## Does this work require review from our design team?
N/A

PR #7855, opened 2024-02-27
Display inline search result attributes

SpennyNDaJets:
## Summary
Show the matched attributes in a single line underneath each row to use the space more efficiently.

https://www.loom.com/share/922a8d28067f4f1ab774dd67db5000aa?sid=e8b38230-9420-4b5a-a211-816d3a866b79

1. View the logs table
2. Search by an attribute
- [ ] Attribute is displayed underneath each row
- [ ] No `;` is rendered
3. Add another search attribute
- [ ] Attribute is added underneath in the same row
- [ ] Separated by a `;`
4. Add another attribute
- [ ] Attribute is added underneath in the same row (breaks to new row if needed)
- [ ] Separated by a `;`

N/A

Yes - https://www.figma.com/file/rdBPNzn7Klk6RG1LHUCE5B/Screen-Designs?node-id=15015%3A163252&mode=dev

### DIFF
--- a/frontend/src/__generated/index.css
+++ b/frontend/src/__generated/index.css
@@ -6575,6 +6575,7 @@ body {
 }
 .g2tqod2 {
   border-bottom: 1px solid var(--_1pyqka9k) !important;
+  background-color: var(--_1pyqka92);
 }
 .szxd8q0 .szxd8q0 {
   padding-left: 22px;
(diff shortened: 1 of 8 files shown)


ccschmitz (approved):
Code + click test LGTM 😻 

One minor regression I noticed is that the `empty` text in cells without values isn't centered anymore. Might be something worth cleaning up quick before merging.

SpennyNDaJets:
> Code + click test LGTM 😻
> 
> One minor regression I noticed is that the `empty` text in cells without values isn't centered anymore. Might be something worth cleaning up quick before merging.

Not sure if this was pre-existing, but I have a PR to fix the session and align the session columns here which I will wait to merge - https://github.com/highlight/highlight/pull/7854

PR #7866, opened 2024-02-29
metrics page demo

mayberryzane:
## Summary
- adds new `metrics` page to demo creating a metric view
- metric view is pretty much complete for line graphs, have not worked on the other view types yet
- some more work to do for the editor panel
  - need to load keys by query instead of statically loading 10 keys
  - need to allow grouping by multiple keys
  - need to replace some of the dropdowns with the checkbox/switch component
- adds generic `metrics` and `keys` endpoints to query each of the different resource types without needing to use a separate Apollo hook

https://www.loom.com/share/54544cb21812477792f957967d801e1c

- clicktested locally

- no

- no


ccschmitz (approved):
Code + the click test you showed me during pairing look great 😻

PR #7867, opened 2024-02-29
check if teams id is nil

mayberryzane:
## Summary
- was causing a bug where a nil pointer would occur after creating a log alert

- clicktested alert creating

- no

- no


ccschmitz (approved):
Thanks for jumping on this one!

SpennyNDaJets (approved):

PR #7886, opened 2024-03-01
View related errors in flyout panel

ccschmitz:
## Summary

Builds out the related resource panel for errors and wires it up to display errors in the panel across the app. Here's where you can see errors in the panel:

* Errors in session dev tools
* ~~Error instances table~~ (this change was reverted - need to think about the UX more)
* Related errors in logs
* Related errors in traces

Also updates the related resource panel so you can pass pagination params to it to enable keyboard shortcuts + buttons to move through the items in a list.

Building the error panel required exporting some of the components from the errors page, so there's a bigger diff there from moving things around to make them exportable.

https://www.loom.com/share/47d07e872a3f4c328147fff0b5abff28

Click tested viewing errors in the following areas:

* Errors in session dev tools
* ~~Error instances table~~
* Related errors in logs
* Related errors in traces

Confirmed behavior of the panel worked as expected while viewing related resources. Could only click test locally since it is dependent on some backend changes.

Should be relatively safe to deploy from a technical perspective, but we'll want to keep tabs on any issues reported and ensure we aren't breaking anyone's workflows.

Will have @julian-highlight take a look and provide some feedback once it's out in production. He has been in the loop along the way, but possibly some refinements coming once we see this in the app.


SpennyNDaJets (approved):
Code and click test look good! Just looks like some trackingIds are missing in `pages/Player/MetadataBox/RelatedResourceButtons.tsx`

PR #7889, opened 2024-03-01
Spenny/sync key and values across searches

SpennyNDaJets:

### DIFF
--- a/frontend/src/components/Search/SearchForm/SearchForm.tsx
+++ b/frontend/src/components/Search/SearchForm/SearchForm.tsx
@@ -1,5 +1,5 @@
 import { useSavedSegments } from '@components/Search/useSavedSegments'
-import { GetLogsKeysQuery, GetTracesKeysQuery } from '@graph/operations'
+import { GetKeysQuery } from '@graph/operations'
 import {
 	Badge,
 	Box,
@@ -49,11 +49,10 @@ import {
 } from '@/components/Search/SearchForm/utils'
 import { parseSearch } from '@/components/Search/utils'
 import {
-	useGetLogsKeysLazyQuery,
-	useGetLogsKeyValuesLazyQuery,
-	useGetTracesKeysLazyQuery,
-	useGetTracesKeyValuesLazyQuery,
+	useGetKeysLazyQuery,
+	useGetKeyValuesLazyQuery,
 } from '@/graph/generated/hooks'
+import { ProductType } from '@/graph/generated/schemas'
 import { useDebounce } from '@/hooks/useDebounce'
 
 import * as styles from './SearchForm.css'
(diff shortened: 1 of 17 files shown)


backend/clickhouse/errors.go
```diff
@@ -552,47 +552,43 @@ func (client *Client) QueryErrorHistogram(ctx context.Context, projectId int, qu
 var ErrorObjectsTableConfig = model.TableConfig[modelInputs.ReservedErrorObjectKey]{
 	TableName: ErrorObjectsTable,
 	KeysToColumns: map[modelInputs.ReservedErrorObjectKey]string{
+		modelInputs.ReservedErrorObjectKeyBrowser:         "Browser",
+		modelInputs.ReservedErrorObjectKeyClientID:        "ClientID",
+		modelInputs.ReservedErrorObjectKeyEnvironment:     "Environment",
+		modelInputs.ReservedErrorObjectKeyHasSession:      "HasSession",
+		modelInputs.ReservedErrorObjectKeyOsName:          "OSName",
 		modelInputs.ReservedErrorObjectKeySecureSessionID: "SecureSessionID",
-		modelInputs.ReservedErrorObjectKeyRequestID:       "RequestID",
```
SpennyNDaJets, line 556: Synced with Clickhouse table:
- Added `Browser`, `ClientID`, `HasSession`, `OSName`
- Removed `RequestID`, `SpanId`, `LogCursor`, `Source`, `StackTrace`, `Payload`
- Moved `Event` and `Type` to Group reserved words
- Renamed `URL` -> `VisitedURL`, `Service.Name` -> `ServiceName`, `Service.Version -> ServiceVersion`

frontend/src/components/Search/SearchForm/SearchForm.tsx
```diff
@@ -308,7 +296,11 @@ export const Search: React.FC<{
 	if (showOperators) {
 		const operators =
 			keyMatch.type === 'Numeric'
-				? [...NUMERIC_OPERATORS, ...EXISTS_OPERATORS]
+				? [
+						...BOOLEAN_OPERATORS,
```
SpennyNDaJets, line 300: Not directly related, but added boolean operators since they should be used with numbers as well

SpennyNDaJets:
@mayberryzane wanted to make sure the changes to the error tables and keys would not have a regression on Grafana? Seems like these keys were not being used since there were no columns they would map to?

backend/clickhouse/errors.go
```diff
@@ -552,57 +552,66 @@ func (client *Client) QueryErrorHistogram(ctx context.Context, projectId int, qu
...
-var errorsSampleableTableConfig = sampleableTableConfig[modelInputs.ReservedErrorObjectKey]{
+var errorsSampleableTableConfig = sampleableTableConfig[modelInputs.ReservedErrorsJoinedKey]{
 	tableConfig: errorsJoinedTableConfig,
 	useSampling: func(time.Duration) bool {
 		return false
 	},
 }
 
 func ErrorMatchesQuery(errorObject *model2.BackendErrorObjectInput, filters listener.Filters) bool {
-	return matchesQuery(errorObject, ErrorObjectsTableConfig, filters)
```
SpennyNDaJets, line 603: I want to double check that this would not affect the behavior of filters? Seems like their were additional columns in this config that were removed, but tried to include the relevant ones in  `errorsJoinedTableConfig`. However, maybe we need to map these to the SQL table columns in a new list?

mayberryzane (approved):

Vadman97 (approved):

PR #7899, opened 2024-03-04
Upgrade Ariakit to v0.4.2

ccschmitz:
## Summary

Upgrades a number of things to improve our front end dev experience.

* Upgrades Araikit to v0.4.2.
* Updates our components to use [component providers](https://ariakit.org/guide/component-providers) rather than our own custom contexsts.
* Updates all components to use the `render` prop rather than `as`.
* Updates all usages in `frontend/` to use the new `useStore` and `useContext` methods.
* Upgrades Yarn.
* Updates our eslint config in the UI package to be consistent w/ `frontend` and runs autofix on all files.
  * Note that there is a fair amount of noise from this change because of all the icon files that were updated!

Click tested as many pages as I could find where we use forms, popovers, and menus, since we changed how we created/accessed their component stores.

All client-side changes, so should be fairly safe deploy as long as we're not seeing any issues with our UI components.

N/A - should not be any visual changes here.


frontend/src/components/Header/Header.tsx
```diff
@@ -733,7 +733,7 @@ export const Header: React.FC<Props> = ({ fullyIntegrated }) => {
 																		)
 																	},
 															  )}
-														<Divider className="mt-1 mb-0" />
+														<Divider className="mb-0 mt-1" />
```
ccschmitz, line 736: One of the eslint changes was to the tailwind class ordering. This was an autofix as a result of that.

frontend/vite.config.ts
```diff
@@ -55,7 +55,7 @@ export default defineConfig(({ mode }) => {
 			port,
 			https:
 				env.SSL === 'false'
-					? false
+					? undefined
```
ccschmitz, line 58: Using `false` started producing a type error and failed the build.

Vadman97 (approved):
gargantuan PR! clicktest looks great 🚢

PR #7912, opened 2024-03-05
Fix over fetching and wrong status on setup

ccschmitz:
## Summary

Fixes two issues:

1. The integration status in the top bar on the setup pages for sessions and errors was always saying "pending" and never said setup was complete.
2. We had an infinite loop fetching sessions and errors to figure out what to display in the integration bar.

This fixes the problem by ensuring the variables don't change for the session and error queries from one query to the next.

* Went to `/setup` locally and opened the network inspector. Note the continuous stream of requests hitting the sessions table in Clickhouse.
* Introduced my changes and reloaded to confirm there was only 1 request and that the status actually updates from pending to integrated.
* Same thing for errors.
* Ran the same checks on the Reflame preview.

N/A - this should reduce the number of queries we're seeing to sessions and errors from setup pages.

@Vadman97 this gave me a thought... do you think we should have any sort of alerting for spikes in requests like we probably saw from the setup page? If someone was sitting on the setup page I think we would have received hundreds of requests from them to the same endpoint.

N/A

### DIFF
--- a/frontend/src/pages/Setup/IntegrationBar.tsx
+++ b/frontend/src/pages/Setup/IntegrationBar.tsx
@@ -18,7 +18,7 @@ import {
 } from '@highlight-run/ui/components'
 import { useProjectId } from '@hooks/useProjectId'
 import moment from 'moment'
-import React, { useMemo } from 'react'
+import React, { useEffect, useMemo } from 'react'
 import { useLocation } from 'react-router-dom'
 
 import {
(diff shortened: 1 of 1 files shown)


frontend/src/pages/Setup/IntegrationBar.tsx
```diff
@@ -66,15 +66,18 @@ export const IntegrationBar: React.FC<Props> = ({ integrationData }) => {
 	const integrated = integrationData?.integrated
 	const ctaText = CTA_TITLE_MAP[area!]
 
+	const start = useMemo(() => moment().subtract(30, 'days').toISOString(), [])
+	const end = useMemo(() => moment().toISOString(), [])
```
SpennyNDaJets, line 70: Future improvement - So currently this is only calculated once? wondering if we should add some type of polling in case the user is sitting on the page waiting?

SpennyNDaJets (approved):
LGTM! No blockers but one future improvement

frontend/src/pages/Setup/IntegrationBar.tsx
ccschmitz: @SpennyNDaJets I ended up addressing this in [6e6e08d](https://github.com/highlight/highlight/pull/7912/commits/6e6e08d257e71c634a60a687c87b932f9a26bcec)!

PR #7928, opened 2024-03-06
default to inlining stylesheets

Vadman97:
## Summary

We hear many concerns about folks needing to inline stylesheets because of their website setup.
There isn't much of a downside to it (except for edge cases where stylesheets may change programmatically, in which case the setting can be disabled).

reflame

no

no

### DIFF
--- a/.changeset/breezy-poets-warn.md
+++ b/.changeset/breezy-poets-warn.md
@@ -1,0 +1,5 @@
+---
+'highlight.run': minor
+---
+
+inline stylesheets by default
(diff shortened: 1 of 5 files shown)


SpennyNDaJets (approved):
Code looks good!

> Are there any deployment considerations?
Is this a risky change? Are there any projects or alerts we should be keeping an eye on?

Vadman97:
> Code looks good!
> 
> > Are there any deployment considerations?
> 
> Is this a risky change? Are there any projects or alerts we should be keeping an eye on?

just potential CORS implications for some edge case setups, but would only affect new customers installing latest version of the sdk, so i don't think it's a huge concern

PR #7935, opened 2024-03-07
fix(sdk/rust): fix publish CI failing

mogery:
## Summary
This is a hotfix. Publishing CI failed because of unexpected behaviour of the used GitHub action (it iterates through every crate in the workspace instead of only publishing the ones we provide in the config). I've changed the Rust workspace to be compatible with this behaviour.

### DIFF
--- a/.github/workflows/rust-sdk.yml
+++ b/.github/workflows/rust-sdk.yml
@@ -37,19 +37,11 @@ jobs:
                   toolchain: stable
                   override: true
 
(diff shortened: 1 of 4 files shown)


mogery:
CI is exhibiting the correct behaviour now (only recognizing `highlightio` and `highlightio-actix` as publishable packages, excluding the examples).

Vadman97 (approved):

PR #7939, opened 2024-03-07
upgrade vite in all packages

mayberryzane:
## Summary
- upgrades vite version for frontend and all dependencies
- removes `terser` plugin as it was failing with an error in the generated code

- clicktested locally to make sure no general regressions

- no

- no


Vadman97 (approved):
nice! you could also add it to the top level package.json resolutions to ensure it is pinned everywhere

PR #7966, opened 2024-03-09
remove store from popover to fix hanging

mayberryzane:
## Summary
- error instance view hanging indefinitely whenever there's a sourcemap error
- not ideal to remove the store here but better than hanging, trying to see if there's a workaround

- clicktested locally

- no

- no


Vadman97 (approved):
thank you! makes sense to get a workaround out first

PR #7967, opened 2024-03-09
try gorm upgrades again

mayberryzane:
## Summary
- first go dependency upgrade https://github.com/highlight/highlight/pull/7937 was hanging during db migration because gorm was attempting to convert some integer columns to bigint. now that those are overridden by an integer struct tag, planning to merge in these dependency upgrades again

- tested locally by auto-migrating from previous version and in a fresh install to make sure both worked

- no

- no


SpennyNDaJets (approved):
LGTM!

> Are there any deployment considerations?

Do we need a rollout/rollback strategy incase anything goes wrong?

mayberryzane:
> LGTM!
> 
> > Are there any deployment considerations?
> 
> Do we need a rollout/rollback strategy incase anything goes wrong?

GORM automigrate won't do anything that would cause data loss, the biggest risk is an accidental DDL change that affects performance. Will be monitoring the rollout - if there are issues, we will revert ASAP to automigrate back to the original state. If there are issues after that, we have daily snapshots and can manually look at the delta to apply schema changes to the prod DB. Finally, if that causes issues, we can restore the entire DB from a snapshot and rewind the Kafka queues. This would be the worst case scenario as it would cause a few hour interruption where customers wouldn't see recent data, but would not cause data loss.

PR #7974, opened 2024-03-11
Animate related resource panel transitions

ccschmitz:
## Summary

Adds an animation to the related resource panel transitioning it in/out from the right side of the screen. There are also a couple bug fixes going out in this PR:

1. Prevents the text on related error buttons from spanning multiple lines in the errors list on a trace.
2. Fixes an issue where the hotkeys weren't behaving as expected when the trace dialog + the related resource dialog were being shown.
3. Updates the cursor on the reside handle to be `col-resize` instead of `ew-resize`.
4. Sets a max width you can drag the panel to.

**Before**
https://www.loom.com/share/2a01aa8400264fd9bf4884b051c65d5b

**After**
https://www.loom.com/share/d8951dff3e884546b468a3e492494ee0

Screen shared with Julian and click tested the until we were happy with the animation. Also click tested in PR preview to ensure it was looking good across all areas of the product and that keyboard shortcuts were behaving as expected.

Paired w/ Julian on these changes so he has seen them already.

### DIFF
--- a/frontend/src/__generated/index.css
+++ b/frontend/src/__generated/index.css
@@ -7423,13 +7423,26 @@ body {
   bottom: 0;
   z-index: 30000;
   position: fixed;
+  transition: transform 0.2s;
+  transform: translateX(100%);
+  will-change: transform;
+}
+._10o9hoe1[data-enter] {
+  transform: translateX(0);
 }
 ._10o9hoe2 {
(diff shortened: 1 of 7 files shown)


SpennyNDaJets (approved):
LGTM!

PR #7980, opened 2024-03-11
Fix error group queries by secure ID

ccschmitz:
## Summary

* Populates the `SecureID` field on the `error_groups` Clickhouse table.
* Fixes the query fetching errors associated with a network resource ("Errors" tab on the network resource panel.
* Clean up an unused argument in `getErrorQueryImpl` (`retentionDate`).
* Fixes a bug with the "Enter" key not working or selecting a command bar item (an issue @Vadman97 and I noticed while pairing).
* Updates the local storage key for the active right panel when viewing a session (Closes HIG-4407).

* Viewed an error
* View the related session
* Viewed the network resource associated with the error and confirmed the "Errors" tab was populated correctly
* Viewed the network panel and confirmed the `GetErrorGroupsClickhouse` query was errored on prod and working on this branch
* Opened the command bar, typed a query, and used the keyboard to navigate + press "Enter" to select an item
* Click tested to ensure the right panel on a session behaved as expected

Need to confirm the changes to `fieldMap` won't break anything. Question out to @mayberryzane on this.

N/A

### DIFF
--- a/backend/clickhouse/sessions.go
+++ b/backend/clickhouse/sessions.go
@@ -53,7 +53,7 @@ var fieldMap map[string]string = map[string]string{
 	"browser":           "Browser",
 	"visited_url":       "VisitedURL",
 	"timestamp":         "Timestamp",
-	"secure_id":         "ErrorGroupSecureID",
+	"secure_id":         "SecureID",
 	"service_name":      "ServiceName",
 	"service_version":   "ServiceVersion",
 	"Tag":               "ErrorTagTitle",
--- a/backend/clickhouse/errors.go
+++ b/backend/clickhouse/errors.go
@@ -22,6 +22,7 @@ type ClickhouseErrorGroup struct {
 	CreatedAt           int64
 	UpdatedAt           int64
 	ID                  int64
+	SecureID            string
 	Event               string
 	Status              string
 	Type                string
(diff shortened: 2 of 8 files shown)


Vadman97 (approved):

backend/clickhouse/sessions.go
```diff
@@ -53,7 +53,7 @@ var fieldMap map[string]string = map[string]string{
 	"browser":           "Browser",
 	"visited_url":       "VisitedURL",
 	"timestamp":         "Timestamp",
-	"secure_id":         "ErrorGroupSecureID",
+	"secure_id":         "SecureID",
```
ccschmitz, line 56: @mayberryzane do you see any issues with this change? Thinking this key is only being used for errors, not sessions, but wanted to double-check with you before changing. Also wondering if we should consider breaking up `fieldMap` to a `sessionFieldMap` and `errorFieldMap`.

ccschmitz: I dug around and saw the `ErrorGroupSecureID` was introduced in https://github.com/highlight/highlight/pull/6492/files#diff-3ed68c77331af5393b04d063f3a44152d8c2928410c0fb9627dd256adc4ae878R51. I don't believe this was ever working, so I don't see any harm in migrating this key to `SecureID`. I think I'll hold off on refactoring the field maps for now, but definitely something we could do in the future to eliminate some confusion here.

PR #7994, opened 2024-03-12
Zane/gorm 3

mayberryzane:
## Summary
- un-reverts the prior commit for go upgrades
- differences from last PR: https://github.com/highlight/highlight/commit/9ff6aa3898570cd5253e56cab7e1dd1bdd53f52f
  - fix issue with default constraints re-applied on every automigrate by removing them and replacing with `BeforeCreate`
  - remove `default:true` from `SystemConfiguration` PK as it triggered some `LastInsertId` bug in GORM
  - fix marshalling for `JSONB` -> `DiscordChannels` and similar (`value` is passed as a `string`)
  - add unit test to ensure all models can be created without error

- clicktested locally after automigrate to make sure app was generally working
- unit tested creation of all models

- same migration plan as in last commit - will monitor release, stop migration and revert if issues, manually patch DB if any performance regressions after revert

- no


backend/model/model.go
```diff
@@ -621,6 +619,23 @@ func (u *Workspace) BeforeCreate(tx *gorm.DB) (err error) {
...
+func (s *SystemConfiguration) BeforeCreate(tx *gorm.DB) (err error) {
+	if s.ErrorFilters == nil {
+		s.ErrorFilters = pq.StringArray{"ENOENT.*", "connect ECONNREFUSED.*"}
+	}
+	if s.IgnoredFiles == nil {
+		s.IgnoredFiles = pq.StringArray{".*/node_modules/.*", ".*/go/pkg/mod/.*", ".*/site-packages/.*"}
+	}
+	return
+}
+
```
Vadman97, line 638: is this only a problem for existing models, or should we use `BeforeCreate` for all future defaults?

backend/model/model.go
```diff
@@ -734,7 +749,7 @@ type Session struct {
 	// The version of Highlight's Firstload.
 	FirstloadVersion string `json:"firstload_version"`
 	// The client configuration that the end-user sets up. This is used for debugging purposes.
-	ClientConfig *string `json:"client_config" sql:"type:jsonb"`
+	ClientConfig *string `json:"client_config" sql:"type:jsonb" gorm:"type:jsonb"`
```
Vadman97, line 752: is the `sql` tag still necessary?

Vadman97 (approved):

backend/model/model.go
mayberryzane: The issue affected these defaults specifically, for example GORM was comparing `'1970-01-01'` against `1970-01-01` without quotes and thinking they were different defaults, but changing the default to remove quotes broke the default in Postgres. The other defaults seem to work ok so I figure it's something that will be fixed in GORM eventually. I think old versions of GORM didn't attempt to migrate the defaults.

backend/model/model.go
mayberryzane: probably not, looks like we only use the `sql:` tag for these `jsonb` fields so I'll try removing them all

PR #7998, opened 2024-03-12
Fix out of order session requests

SpennyNDaJets:
## Summary
Currently network resources and websocket events use relative times to the start of the performance api. This time is based on the the window's start time, which is reset when the page is navigated to or refreshed. This causes the events to be out of order when viewing a session.

Update the events to use an absolute time to be placed in the correct spot in the session.

https://www.loom.com/share/af1ebe59557648abaa28c38945abefd1?sid=a65948e2-938a-4f9c-8fb2-3c2c94011879

1) Start a session
2) Let the page idle, navigate to and away, refresh the page
3) Complete some network requests in the window
4) Close the tab to let the session process
5) View the processed session
6) Test viewing in absolute and relative times
- [ ] All network requests are in correct order
- [ ] Jumping to a network request jumps tot he right place
- [ ] All labels for session length and event timestamps are correct 

Let knows about how old versions of the SDK will be using `startTime` and `requestEnd`, but should no longer be used in code

N/A

### DIFF
--- a/frontend/src/pages/Player/RightPlayerPanel/components/NetworkResourcePanel/NetworkResourceLogs.tsx
+++ b/frontend/src/pages/Player/RightPlayerPanel/components/NetworkResourcePanel/NetworkResourceLogs.tsx
@@ -40,14 +40,20 @@ export const NetworkResourceLogs: React.FC<{
 	}>()
 	const requestId = resource.requestResponsePairs?.request?.id
 	const [query, setQuery] = useState('')
-	const startDate = useMemo(
-		() => new Date(sessionStartTime + resource.startTime - TIME_BUFFER),
-		[sessionStartTime, resource.startTime],
-	)
-	const endDate = useMemo(
-		() => new Date(sessionStartTime + resource.responseEnd + TIME_BUFFER),
-		[resource.responseEnd, sessionStartTime],
-	)
+	const startDate = useMemo(() => {
+		// startTime used in highlight.run <8.8.0 for websocket events and <7.5.4 for requests
+		const resourceStart =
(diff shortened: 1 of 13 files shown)


frontend/src/pages/Player/RightPlayerPanel/components/NetworkResourcePanel/NetworkResourceLogs.tsx
```diff
@@ -40,14 +40,20 @@ export const NetworkResourceLogs: React.FC<{
...
-		[sessionStartTime, resource.startTime],
-	)
-	const endDate = useMemo(
-		() => new Date(sessionStartTime + resource.responseEnd + TIME_BUFFER),
-		[resource.responseEnd, sessionStartTime],
-	)
+	const startDate = useMemo(() => {
+		// startTime used in highlight.run <8.8.0 for websocket events and <7.5.4 for requests
+		const resourceStart =
+			resource.startTimeAbs ?? sessionStartTime + resource.startTime
```
Vadman97, line 46: just checking the order of operations here
this is the same as `resource.startTimeAbs ?? (sessionStartTime + resource.startTime)` right?

Vadman97 (approved):
tested with new and old highlight.run client, on both relative and absolute times with and without page refreshes. works great in all cases! great work @SpennyNDaJets

frontend/src/pages/Player/RightPlayerPanel/components/NetworkResourcePanel/NetworkResourceLogs.tsx
SpennyNDaJets: Yep! Same think and the linter actually removes the `()` on save

PR #8002, opened 2024-03-12
more go dependency upgrades

mayberryzane:
## Summary
- `go get -u` + `go mod tidy` to upgrade minor version dependencies for backend, highlight-go, and e2e
- a few code changes as some APIs changed slightly

- clicktested locally to make sure everything was generally working

- can revert if issues

- no


Vadman97 (approved):
awesome, last time i tried this stuff broke because of the gorm / pg driver getting updated indirectly, so now that that is all fixed this seems safe to do

PR #8011, opened 2024-03-13
use backend.Dockerfile in prod deployment

mayberryzane:
## Summary
- `backend.Dockerfile` has some optimizations to make the image smaller by only including binaries
- use that Dockerfile and override the command in `private-graph-task.json` (we do this for all other task definitions)

- will monitor prod deployment

- GitHub action can fail, can revert if issues

- no


Vadman97 (approved):
sweet! if it's easier than reverting, you can push a temp commit to this branch that will run the aws deploy on pushes to this branch to trigger deploy while still in PR

PR #8013, opened 2024-03-13
use backend.Dockerfile

mayberryzane:
## Summary
- fixes a few issues with the original dockerifle update:
  - update task definitions to use the `/build/backend` path
  - root for Clickhouse migrations should be the `/build` folder now

- ran the Deploy to AWS step, confirmed services are all up

- can revert if issues

- no


Vadman97 (approved):
lgtm :ship:

PR #8028, opened 2024-03-14
redirect to workspace's default project on join

mayberryzane:
## Summary
- navigate to the default project for a workspace after accepting that workspace's invite link
- use reflame setting `"important": true` to fix tailwind not applying !important

- clicktested locally

- no

- no


SpennyNDaJets (approved):
Code looks good!

Click test notes:

> navigate to the default project for a workspace after accepting that workspace's invite link

I noticed when I invited myself, I think it flashes the project page, but then I get redirected to put more info in. Is this intended?

mayberryzane:
> I noticed when I invited myself, I think it flashes the project page, but then I get redirected to put more info in. Is this intended?

hmm I think that was probably the previous behavior too but probably not intended - this change affects users that already have a workspace to make sure they get redirected to their new one and not confused by being stuck in their existing one

PR #8060, opened 2024-03-19
Fix bugs applying filters from log/span attributes

ccschmitz:
## Summary

Fixes a few issues around applying attribute filters.

1. Applying attribute filters that contained `"` was broken because a) we didn't wrap the value in quotes properly, and b) our query language doesn't support escaping quotes.
2. Removes the `message` prefix when applying an attribute filter on logs since this would break the query.
3. Nothing would happen when trying to apply a `message` attribute filter when there was no query present.

1. Apply a `message` filter and confirm the `message=` prefix is not applied
1. Apply a `message` filter when the query input is empty
1. Apply a filter that contains quoted strings within it

Confirmed all broken behavior was working as expected locally, in prod, and on the customer's project who reported the error.

Should follow up with customer who reported the issue.

N/A


ccschmitz:
@SpennyNDaJets I noticed that this still doesn't fix the original issue reported in the ticket in the PR preview. I believe this is because they are trying to search on a custom attribute named `body`, which we end up intercepting and trying to filter against our `body` column attribute. Trying to decide whether to come up with a workaround for this or just let the customer know that `body` is a reserved attribute. Curious if you have any thoughts on this 🤔 

cc @Vadman97

ccschmitz:
@SpennyNDaJets @Vadman97 I was wrong about it conflicting with the reserved `body` attribute. The problem was with how we were escaping strings on the client and the server, which resulted in double quotes being escaped twice and the query not returning any results.

backend/parser/listener/listener.go
```diff
@@ -300,6 +300,10 @@ func (s *searchListener[T]) appendRules(value string) {
 		filterKey = fmt.Sprintf("toString(%s)", filterKey)
 	}
 
+	// Quotes are sometimes escaped on the client and need to be unescaped before
+	// being used in the query or they will be double escaped.
+	value = strings.ReplaceAll(value, `\"`, `"`)
```
ccschmitz, line 305: @mayberryzane curious if you can think of a better way to handle this, and quote escaping in general on queries 🤔

SpennyNDaJets (commented):
image

SpennyNDaJets (approved):

PR #8064, opened 2024-03-19
Use related resource panel to view associated traces

ccschmitz:
## Summary

Creates a new trace view for the related resource panel and uses it to view related traces in the following areas:

* Related trace on an error
* Related trace on logs

Note that we are not using the related resource panel on the traces page. We will [look at rolling that out in the future](https://www.figma.com/file/rdBPNzn7Klk6RG1LHUCE5B?type=design&node-id=14888-85455&mode=design#742454572). Tracking in [a Linear ticket](https://linear.app/highlight/issue/HIG-4447/update-trace-panel-on-traces-list).

This PR also has a slew of other updates:

* Uses the log viewer components to visualize trace attributes and moves the components to a new `JsonViewerV2` component.
* Fixes a bug where resizing wouldn't work when the mouse moved over the trace flame graph during resize. This was happening because the event handler was being intercepted and prevented from bubbling up.
* Moves analytics for related resources to a generic hook inside the related resource panel.
* Fixes a bug where drag to zoom wouldn't work on the flame graph if you release the mouse off the canvas for the flame graph.
* Removes the logic for showing the related error when viewing a related session from an error. We decided to remove this in our latest Waterloo session.
* Updates spacing on error panel to match what's in the mocks.
* Fixes matching attribute matching on custom attributes.

* View a related trace from an error
* View a related trace from a log
* View a trace on the traces page (ensure the old panel is used)
* Filters on custom attributes highlight correctly on logs

Will want to monitor any user frustration or issues with workflows breaking, but that will mostly be watching people use the related trace panel and keeping an eye out for any user reported issues.

Will need to discuss some of the aforementioned issues w/ Julian about rolling out on the trace list page.


frontend/src/components/JsonViewer/JsonViewerV2.tsx
```diff
@@ -0,0 +1,51 @@
...
+								queryKey={key}
+								queryParts={queryParts}
+							/>
+						)}
+					</Box>
+				)
+			})}
+		</Stack>
+	)
+}
```
ccschmitz, line 51: Should probably move supporting components here instead of keeping them in logs.

frontend/src/components/JsonViewer/utils.ts
```diff
@@ -0,0 +1,71 @@
...
+	AndOrExpression,
+	SearchExpression,
+} from '@/components/Search/Parser/listener'
+import { ReservedLogKey } from '@/graph/generated/schemas'
+
+const bodyKey = ReservedLogKey['Message']
+
+export const findMatchingAttributes = (
+	queryParts: Array<SearchExpression | AndOrExpression>,
+	logAttributes: object | string,
```
SpennyNDaJets, line 11: Nit: should we rename this to `attributes`?

SpennyNDaJets (approved):
Looks good to me!

I also viewed a trace on the session page. Just noticed it was tight regarding padding compared to the other trace views. Wanted to make sure it was intended
Screenshot 2024-03-21 at 2 40 23 PM

frontend/src/components/JsonViewer/utils.ts
ccschmitz: Good call - will update before merging!

ccschmitz:
> I also viewed a trace on the session page. Just noticed it was tight regarding padding compared to the other trace views. Wanted to make sure it was intended

@SpennyNDaJets thanks for catching that - will update before merging!

PR #8082, opened 2024-03-22
Support single quotes and backticks in queries

ccschmitz:
## Summary

Adds support for single quotes and backticks in quoted strings. This was not a requested feature, but something @mayberryzane and I decided to work on after https://github.com/highlight/highlight/pull/8060 which added support for double quotes inside a string.

This should be especially helpful for those copy/pasting values searching for logs.

Click tested a number of different queries using different types of quotes in search.

* `body="{\"key\":\"value\"}"`
* `body='{"key":"value"}'`
* ``body=`{"key":"value"}` ``

N/A - should only expand functionality of our search.

N/A


Vadman97 (approved):
nice!

PR #8088, opened 2024-03-23
Create issue modal Updates

### LINKED ISSUE #7970: Create issue modal Updates
New designs for the 'Link issue' functionality.  

Designs [here](https://www.figma.com/file/rdBPNzn7Klk6RG1LHUCE5B/Screen-Designs?type=design&node-id=15277-66270&mode=design&t=KP7f2ZFB710ZdYOW-11)

jemiluv8:
Add IssueModeSelector component
Use component in issue creation modals for issue mode selection

Fixes #7970
## Summary
Screenshot 2024-03-23 at 2 57 54 PM
Screenshot 2024-03-23 at 3 04 33 PM
Screenshot 2024-03-23 at 3 09 41 PM

https://github.com/highlight/highlight/assets/119384208/88e051af-7041-4ccd-8de2-c9ec065875b2

Click testing

No

Yes


frontend/src/components/IssueModeSelector/IssueModeSelector.tsx
```diff
@@ -0,0 +1,62 @@
...
+	onSelect: (p: T) => void
+	label: any
+	key: any
+	selected: boolean
+}) => {
+	const modeClass = selected ? styles.activeSwitch : styles.inactiveSwitch
+
+	return (
+		<div
+			className={clsx(styles.tagSwitch, modeClass)}
```
ccschmitz, line 20: Nit, but we try to use the object syntax when conditionally applying classes like this. A little more verbose, but maybe a little more readable.

```tsx
clsx(styles.tagSwitch, {
  [styles.activeSwitch]: selected,
  [styles.inactiveSwitch]: !selected,
})}
```

ccschmitz: Also, wondering if we could use `Box` here and apply many of these styles via props rather than using a separate stylesheet. Not a big deal, just a thought for future contributions 😄

frontend/src/components/IssueModeSelector/IssueModeSelector.module.css
```diff
@@ -0,0 +1,24 @@
...
+.activeSwitch {
+	background-color: var(--color-purple);
+	background: var(--color-purple);
+	color: var(--color-white);
+	box-shadow: 0px -1px 0px 0px rgba(0, 0, 0, 0.32) inset
+}
+
+.inactiveSwitch {
+	borderColor: var(--color-gray-300);
+	color: var(--_11tb6c5c);
```
ccschmitz, line 22: Where did this variable come from? I'm wondering if this is a hash produced during a build, which would be unreliable to reference long term. Is there another variable we could reference? Or else maybe it would be better to use the `color` prop on the `Text` component here.

```tsx
<Text color={selected ? 'secondaryContentText' : 'white'}>...</Text>
```

frontend/src/components/IssueModeSelector/IssueModeSelector.tsx
```diff
@@ -0,0 +1,62 @@
+import { Box, Text } from '@highlight-run/ui/components'
+import clsx from 'clsx'
+
+import styles from './IssueModeSelector.module.css'
```
ccschmitz, line 4: Nit, but we are using [vanilla-extract](https://vanilla-extract.style/) for managing styles going forward. This is fine for now, but for future reference it would be best to use a `.css.ts` file rather than a CSS module file.

ccschmitz (changes requested):
Code + click test look good! Thank you for the contributions!!

Most of my comments were not blockers, but the one about the CSS variable should probably be addressed before merging since I don't think we can rely on that `_11tb6c5c` variable existing.

frontend/src/components/IssueModeSelector/IssueModeSelector.module.css
jemiluv8: I thought it was a global css variable but the naming makes me believe as you do that it was generated by a build script whose name may change in the future - so cannot be relied upon.

Using the color prop on the Text component is the more “idiomatic” way. Will do that instead.

frontend/src/components/IssueModeSelector/IssueModeSelector.tsx
```diff
@@ -0,0 +1,79 @@
+import { Box, Text } from '@highlight-run/ui/components'
+import clsx from 'clsx'
+
+import * as styles from './IssueModeSelector.css'
+
+export const TagSwitch = <T extends string | number>({
```
jemiluv8, line 6: I'm interested in working on #8105. 
Do you have any suggestions regarding how
best to extract this component?

jemiluv8:
I have resolved all issues raised from the review.

PR #8091, opened 2024-03-25
Fix state reset on session transition

ccschmitz:
## Summary

In Waterloo last week we noticed that the view mode on sessions would get out of sync when transitioning sessions. e.g. you could be in comment mode, but when you click a session card in the feed you were transitioned to inspect mode and the session right panel was opened while the UI still showed you being in comment mode. We always want to show the comments panel when you are in comment mode, so this PR removes the state reset when clicking on a session card.

Note that [there is still logic](https://github.com/highlight/highlight/blob/db9259f8bb9877036735fe3c3cfef66af16ad066/frontend/src/pages/Player/RightPlayerPanel/RightPlayerPanel.tsx#L62-L68) to reset to the session view if you are in the event view when transitioning sessions.

* View a session
* Transition to comment mode
* View another session
* Ensure you are still in comment mode and that comment panel is open

N/A

N/A


frontend/src/components/FullCommentList/FullCommentList.tsx
```diff
@@ -24,41 +24,43 @@ const FullCommentList = ({
...
-			{!loading && comments.length === 0 ? (
-				<div className={styles.noCommentsContainer}>
-					<div className={styles.noCommentsTextContainer}>
-						<h2>There are no comments yet</h2>
-						<p>{noCommentsMessage}</p>
-					</div>
-					<ConnectHighlightWithSlackButton />
-				</div>
-			) : (
+			{!loading && (
```
ccschmitz, line 32: This change fixes the loading UI so it is centered vertically in the panel. We were showing the list component while in the loading state causing it to be pushed up from the center.

Vadman97 (approved):
reflame preview works great

PR #8096, opened 2024-03-25
add migration to save error object embeddings per group

mayberryzane:
## Summary
- adds `ErrorGroupEmbeddings` table to save average embedding per group
- adds script to migrate error group embeddings for project_id = 1
- script also creates an index and helper table so migration can run quicker in future, minimizing the amount of missed data when we cut over to use these group embeddings for all projects

- ran script multiple times locally, verified embeddings and counts are updated as more errors are added to the group

- will deploy a separate change to use these embeddings for project_id = 1 after migration

- no

### DIFF
--- a/backend/migrations/cmd/migrate-group-embeddings/main.go
+++ b/backend/migrations/cmd/migrate-group-embeddings/main.go
@@ -1,0 +1,100 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	log "github.com/sirupsen/logrus"
+	"gorm.io/gorm"
+
+	"github.com/highlight-run/highlight/backend/model"
+)
+
+func main() {
+	ctx := context.Background()
+
+	db, err := model.SetupDB(ctx, os.Getenv("PSQL_DB"))
+	if err != nil {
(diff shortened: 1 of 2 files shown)


backend/migrations/cmd/migrate-group-embeddings/main.go
```diff
@@ -0,0 +1,100 @@
...
+					on eoe.error_object_id = eo.id
+					where eoe.gte_large_embedding is not null
+					and eoe.id > ?
+					and eoe.id <= ?
+					group by eo.project_id, eo.error_group_id) a
+				on conflict (project_id, error_group_id)
+				do update set
+					gte_large_embedding = 
+						error_group_embeddings.gte_large_embedding * array_fill(error_group_embeddings.count::numeric / (error_group_embeddings.count + excluded.count), '{1024}')::vector
+						+ excluded.gte_large_embedding * array_fill(excluded.count::numeric / (error_group_embeddings.count + excluded.count), '{1024}')::vector,
```
Vadman97, line 74: TIL `excluded` is a keyword that references the row that failed to insert because of the conflict

Vadman97 (approved):

PR #8108, opened 2024-03-26
match embedding at group level for project 1

mayberryzane:
## Summary
- modify migration to run on batches of 100k embeddings (less memory use, can recover after crashing)
- for project_id = 1, use ErrorGroupingMethodGteLargeEmbeddingV3 as the matching method to match at the group level

- clicktested from buttons page locally, ensured new errors used the new matching method

- will monitor error grouping for project_id = 1 after deploying and migrate all projects' embeddings to the new table

- no

### DIFF
--- a/backend/migrations/cmd/migrate-group-embeddings/main.go
+++ b/backend/migrations/cmd/migrate-group-embeddings/main.go
@@ -11,6 +11,8 @@ import (
 	"github.com/highlight-run/highlight/backend/model"
 )
 
+var step = 100000
+
 func main() {
 	ctx := context.Background()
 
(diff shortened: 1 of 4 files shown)


backend/public-graph/graph/resolver.go
```diff
@@ -494,17 +494,34 @@ func (r *Resolver) GetTopErrorGroupMatchByEmbedding(ctx context.Context, project
...
+			map[string]interface{}{
+				"embedding": embedding,
+				"projectID": projectID,
+			}).Scan(&result).Error; err != nil {
+			return nil, e.Wrap(err, "error querying top error group match")
+		}
+	} else {
+		// an alternative query to consider: for M error objects of every error group,
+		// find the average score. then pick the error group
+		// with the lowest average score.
```
Vadman97, line 516: can remove?

backend/public-graph/graph/resolver.go
```diff
@@ -749,8 +787,12 @@ func (r *Resolver) HandleErrorAndGroup(ctx context.Context, errorObj *model.Erro
 			errorObj.ErrorGroupingMethod = model.ErrorGroupingMethodClassic
 		} else {
 			embedding = emb[0]
+			embeddingType := model.ErrorGroupingMethodGteLargeEmbeddingV2
+			if errorObj.ProjectID == 1 {
+				embeddingType = model.ErrorGroupingMethodGteLargeEmbeddingV3
+			}
```
Vadman97, line 793: clever using the method to transition to this by project

Vadman97 (approved):

backend/migrations/cmd/migrate-group-embeddings/main.go
```diff
@@ -11,6 +11,8 @@ import (
 	"github.com/highlight-run/highlight/backend/model"
 )
 
+var step = 100000
```
SpennyNDaJets, line 14: Nit: not sure if go has the same convention, but like keeping constants all capitalized to help with readability. If this is in fact a constant.

SpennyNDaJets (approved):
Nothing blocking but just a comment on some code styles

backend/migrations/cmd/migrate-group-embeddings/main.go
mayberryzane: I should change var -> const, but in go unexported constants are typically lowercase / camel case to avoid being exported

PR #8118, opened 2024-03-27
Related resource panel for sessions

ccschmitz:
## Summary

Sets up the related resource panel for sessions and rolls it out in the following areas:

* Errors
* Logs
* Traces

Here is a preview from my post in #demo last week:

https://www.loom.com/share/35b43e03d1d84bf0ad5146a5883bb2a5

Note that this version of the player has a few limitations.

1. You can't view the right panel
2. No commenting
3. Can't configure an alert (doesn't make sense outside search)

In order to get the player working outside the player page I abstracted a few elements of `PlayerPage` so I could use just what I needed. I also created a `useSessionParams` hook for accessing the URL params which checks the router props as well as the related resource query params. This ends up touching a lot of files because it updates everywhere we were using `useParams` previously in the player components.

I click tested all the player controls, dev tools, timeline interactions, resizing (viewport + panel), and anything else I could think of around functionality of the session player in the panel. Also tested in the following contexts.

**Errors**

* View an error
* Click the "Related session" button

**Logs**

* Visit logs search
* Search for logs where `session_secure_id EXISTS`
* Expand a log and click the button to view the related session

**Traces**

* Visit traces search
* Search for traces where `session_secure_id EXISTS`
* Open a trace and click the view session button

No backend considerations, but a change that could break some workflows for people, so will want to keep an eye on frustration in sessions and any reported issues in support.

Have shared the PR preview with Julian and he has signed off! ✅


packages/ui/src/components/ButtonIcon/styles.css.ts
```diff
@@ -16,6 +16,7 @@ export const variants = recipe({
 			cursor: 'pointer',
 			padding: 0,
 			outline: 'none',
+			flexShrink: 0,
```
ccschmitz, line 19: Not directly related to this change, but prevents the square buttons from being squished and changing shape.

frontend/src/pages/Player/utils/utils.test.ts
```diff
@@ -22,17 +22,6 @@ describe('sessionIsBackfilled', () => {
...
-		// the types say this will always be boolean, so... what is this test for? 🙃
-		session.identified = undefined as unknown as boolean
-		expect(sessionIsBackfilled(session)).toBe(false)
-	})
-
-	it('returns false if the "client_id" key is missing', () => {
-		session.client_id = ''
-		expect(sessionIsBackfilled(session)).toBe(false)
-	})
-
```
ccschmitz, line 35: Not related - just noticed these could be cleaned up at this point.

Vadman97 (approved):
looking great, all of the paths working well in the reflame preview. the session taking a while to load is still the case when you expand the related session panel, but that can be handled in a separate pr

frontend/src/pages/Player/PlayerPage.tsx
```diff
@@ -152,93 +146,18 @@ const PlayerPage = () => {
...
-			const widthScale = (targetWidth - style.PLAYER_PADDING_X) / width
-			const heightScale = (targetHeight - style.PLAYER_PADDING_Y) / height
-			let scale = Math.min(heightScale, widthScale)
-			// If calculated scale is close enough to 1, return to avoid
-			// infinite looping caused by small floating point math differences
-			if (scale >= 0.9999 && scale <= 1.0001) {
-				return false
-			}
+	const { resizeListener, centerColumnResizeListener, controllerWidth } =
+		useResizePlayer(replayer, playerWrapperRef, setScale)
```
ccschmitz, line 155: I believe something with this refactor caused an issue with the player canvas resizing. See this thread for more details: https://highlightcorp.slack.com/archives/C02N2AZS8BV/p1717007998091049

PR #8122, opened 2024-03-27
Fix antlr crashes

SpennyNDaJets:
## Summary
There is an edge case where the frontend crashes when typing in an `AND` after an open parenthesis. Fix the type error.

https://www.loom.com/share/474a6d9cb7bd4af2922cd64b347d02cd?sid=5bda8150-6ae1-4748-8dfc-322a70aaf8a9

1) Open the logs or traces search
2) Start searching for `source=(backend AND frontend)`
- [ ] Page does not crash
- [ ] No results should be shown

N/A

N/A


ccschmitz (approved):
Nice - thanks for fixing this! Just curious, do you understand what the issue was with using `and_op` vs `AND`? Not sure I understand why this works now 🤔

SpennyNDaJets:
> Nice - thanks for fixing this! Just curious, do you understand what the issue was with using `and_op` vs `AND`? Not sure I understand why this works now 🤔

I haven't confirmed, but my hunch is something about `and_op` causes a more lazily evaluated expression, resulting in this [search_expr](https://github.com/highlight/highlight/blob/main/antlr/SearchGrammar.g4#L27) being evaluated based on the order of operations. As a result, we call `exitKey_val_search_expr` on the "new default expression" (i.e. `message=`) which is throwing a type error.

I guess using the `AND` causes the evaluator to still think its a `top_col_expr` and runs the `exitKey_val_search_expr` on the entire value, (i.e. `source=(backend AND`)

PR #8131, opened 2024-03-28
Livestream banner

deltaepsilon:
## Summary
image


highlight.io/components/common/Navbar/Navbar.tsx
```diff
@@ -43,6 +43,36 @@ const LaunchWeekBanner = () => {
...
+				for&nbsp;
+				<a
+					target="_blank"
+					href="https://lu.ma/b0uz0fiz"
+					className={styles.launchWeekText}
+					style={{ textDecoration: 'underline' }}
+					rel="noreferrer"
+				>
+					Zoom
+				</a>
```
Vadman97, line 60: could you please use the `<Link>` tag with the same styling as the startup banner below?
see 
```
<Link
						href="/xxx"
						className="flex justify-center items-center w-full h-[40px] bg-color-primary-200 text-white hover:bg-opacity-90"
					>
```

Vadman97 (approved):

highlight.io/components/common/Navbar/Navbar.tsx
deltaepsilon: I think I got it!

PR #8133, opened 2024-03-28
Keep url in sync with session

SpennyNDaJets:
## Summary
The sessions url gets out of sync when jumping around. Make sure to set the url when the time is changed, and remove unnecessary handler that was reseting url to the url at time `0`.

Also:
- removes unused `CurrentUrlBar` component
- remove `console.log`

https://www.loom.com/share/5c085c2e2c5f4a68aacccb3141d32746?sid=4c52fbc2-e136-412a-b95f-a9a7993d1b29

1) Play a session
- [ ] Url stays in sync with session navigation events
2) Drag the time control to a new page
- [ ] Url updates to correct page url
3) Jump to an event 
- [ ] Url updates to correct page url

N/A

N/A


frontend/src/pages/Player/PlayerHook/PlayerState.ts
```diff
@@ -346,6 +346,10 @@ export const PlayerReducer = (
 			break
 		case PlayerActionType.setTime:
 			s.time = action.time
+			s.currentUrl = findLatestUrl(
+				getAllUrlEvents(events),
+				action.time + s.sessionMetadata.startTime,
+			)
```
Vadman97, line 352: a bit worried that doing this on each debounced rerender would be slow if there are a lot of events to search. doesn't seem from profiling like this is a bottleneck though, but something to keep in mind

Vadman97 (approved):
works much better with this preview! just a side note is that we should also update the client to emit a `navigate` event on session restart. looks like some sessions never have a url navigation event if they are starting a new session from a tab that was previously opened, like: https://app.highlight.io/1/sessions/VBSzxVPPAO8OMiQeBTFH4Vo1VTDk

frontend/src/pages/Player/PlayerHook/PlayerState.ts
SpennyNDaJets: For sure! I think this method (`findLatestUrl`) should be called a little more than the current state (called in `PlayerActionType.updateCurrentUrl`). It will still get called on start and pause, but this is just adding to when the user manually changes the time.

Vadman97: ah cool, thanks!

PR #8136, opened 2024-03-29
fix stacktrace processing for brotli compressed code

Vadman97:
## Summary

Our reflame frontend sourcemaps would not be processed because we
could not read brotli-encoded content.
* adds support for brotlied js/map files in the reader
* cleans up reflame query string trimming to make sure the qs parameter is kept

cc @lewisl9029 

unit test processing our frontend sourcemaps correctly.

no

no


backend/stacktraces/enhancer.go
```diff
@@ -87,6 +89,14 @@ func (n NetworkFetcher) fetchFile(ctx context.Context, href string) ([]byte, err
 		return nil, e.New("status code not OK")
 	}
 
+	if res.Header.Get("Content-Encoding") == "br" {
```
lewisl9029, line 92: Hmm I'll have to double check on my side as well, but I think this might actually be a bug in the Reflame CDN if we respond with a `br` encoded response when your request doesn't have a `Accept-Encoding` header with `br` in it.

Will add this to the backlog but I might not be able to get to it for a while since most browsers do support brotli so include it in accept-encoding. Appreciate y'all putting in this workaround in the meantime!

mayberryzane (approved):

PR #8137, opened 2024-03-29
Try worker support

lewisl9029:
## Summary

Not yet working due to lack of import map support in workers: https://github.com/WICG/import-maps/issues/2

Have an idea for a workaround using https://github.com/guybedford/es-module-shims but it's going to take a bit longer.


Vadman97:
thanks lewis! manually pulled this into #8058

PR #8155, opened 2024-04-01
truncate trace values > 65536

mayberryzane:
## Summary
- uses some logic for truncating large log values to apply to traces as well
- expecting this to solve our current issue with trace uploads where certain values are very large >10mb and failing to write to clickhouse

- tested the happy path locally to make sure traces are still ingested, will monitor the prod traces backlog after deployment

- no

- no

### DIFF
--- a/backend/clickhouse/trace_row.go
+++ b/backend/clickhouse/trace_row.go
@@ -7,6 +7,15 @@ import (
 	"github.com/google/uuid"
 )
 
+const TraceAttributeValueLengthLimit = 2 << 15
+
+func truncateValue(input string) string {
+	if len(input) > TraceAttributeValueLengthLimit {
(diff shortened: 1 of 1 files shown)


backend/clickhouse/trace_row.go
```diff
@@ -7,6 +7,15 @@ import (
 	"github.com/google/uuid"
 )
 
+const TraceAttributeValueLengthLimit = 2 << 15
+
+func truncateValue(input string) string {
+	if len(input) > TraceAttributeValueLengthLimit {
+		return input[:TraceAttributeValueLengthLimit] + "..."
```
Vadman97, line 14: should we log the project id here? i guess it will be obvious which trace attributes are truncated because of the `...` so perhaps it's not worth adding the large log volume

Vadman97 (approved):

PR #8176, opened 2024-04-02
Fix bug with mismatching left panel open state

ccschmitz:
## Summary

While investigating [a bug](https://linear.app/highlight/issue/HIG-4453/left-menu-not-always-being-shownhidden-correctly-on-player) where the state of `showLeftPanel` was providing different values in different components I realized this was [a known issue](https://github.com/streamich/react-use/issues/785) with the `useLocalStorage` hook from [react-use](https://github.com/streamich/react-use) we were relying on to keep this in sync across the app. I noticed we also have the `@rehooks/local-storage` package which provides a `useLocalStorage` hook that does not have this issue.

I added an eslint rule to prevent importing `useLocalStorage` from react-use and guide people toward importing it from `@rehooks/local-storage`, then I updated all the imports to import from the rehooks package. I also upgraded the `react-use` and `@rehooks/local-storage` packages.

One other change I made was to some logic for opening/closing panels depending on whether certain params were present. Some of it wasn't relevant anymore now that we don't show error details in the right panel and open the related resource panel for errors instead.

While investigating the issue with `useLocalStorage` from react-use I saw that [most of the active maintainers of react-use have stepped away from the project](https://github.com/streamich/react-use/issues/1974). We may want to consider migrating away from that in the future. Perhaps to something like [usehooks-ts](https://usehooks-ts.com/). I did explore this in this PR, but noticed some odd behavior and a slew of console errors when I transitioned to their hooks so I decided to not go down that path here.

* Click test to confirm the original bug was fixed
  * Go to an error w/ left panel open
  * Click related session
  * Confirm panel is closed and center content is not bumped out
* Click across the app to ensure nothing appears broken

N/A

N/A


frontend/src/components/QueryBuilder/QueryBuilder.tsx
```diff
@@ -1336,7 +1335,7 @@ function QueryBuilder(props: QueryBuilderProps) {
 		rules: serializedRules,
 	}: QueryBuilderState = JSON.parse(searchQuery)
 	const startingRules = deserializeRules(serializedRules)
-	const [isAnd, toggleIsAnd] = useToggle(serializedIsAnd)
+	const [isAnd, toggleIsAnd] = useState(serializedIsAnd)
```
ccschmitz, line 1338: Didn't see a real use case for `useToggle` since we were always passing in the value we want to assign.

frontend/src/components/QueryBuilder/QueryBuilder.tsx
```diff
@@ -1336,7 +1335,7 @@ function QueryBuilder(props: QueryBuilderProps) {
 		rules: serializedRules,
 	}: QueryBuilderState = JSON.parse(searchQuery)
 	const startingRules = deserializeRules(serializedRules)
-	const [isAnd, toggleIsAnd] = useToggle(serializedIsAnd)
+	const [isAnd, toggleIsAnd] = useState<boolean>(serializedIsAnd)
```
ccschmitz, line 1338: Didn't see a real use case for `useToggle` since we were always passing in the value we want to assign.

frontend/src/pages/Player/PlayerHook/PlayerHook.tsx
```diff
@@ -799,14 +796,6 @@ export const usePlayer = (): ReplayerContextInterface => {
 		state.sessionMetadata,
 	])
 
-	useEffect(() => {
-		const searchParamsObject = new URLSearchParams(location.search)
-		if (searchParamsObject.get(PlayerSearchParameters.errorId)) {
-			setShowLeftPanel(false)
-			setShowRightPanel(true)
-		}
-	}, [setShowLeftPanel, setShowRightPanel])
```
ccschmitz, line 808: Not showing the error in the right panel anymore because of related resource panel. Also, we're using the `PlayerSearchParameters.search` param to determine whether we should close the left panel now, so decided to clean up this logic entirely.

mayberryzane (approved):

PR #8177, opened 2024-04-02
Compare excluded env against error env

ccschmitz:
## Summary

Fixes a bug where we would compare the excluded environments on an error alert against the session associated with the error rather than the error's environment. This can be problematic when there isn't an associated session.

Click tested to confirm I was getting alerts and that they stopped once I added the excluded environments locally.

Will follow up with customers impacted by this bug to let them know it should be resolved, but no major considerations here.

N/A

### DIFF
--- a/backend/public-graph/graph/resolver.go
+++ b/backend/public-graph/graph/resolver.go
@@ -527,8 +527,8 @@ func (r *Resolver) GetTopErrorGroupMatchByEmbedding(ctx context.Context, project
 			if method == model.ErrorGroupingMethodGteLargeEmbeddingV3 {
 				if err := r.DB.WithContext(ctx).Exec(`
 					update error_group_embeddings
-					set gte_large_embedding = gte_large_embedding * array_fill(count::numeric / (count + 1), '{1024}')::vector 
-						+ @embedding * array_fill(1::numeric / (count + 1), '{1024}')::vector, 
+					set gte_large_embedding = gte_large_embedding * array_fill(count::numeric / (count + 1), '{1024}')::vector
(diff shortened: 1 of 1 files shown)


backend/public-graph/graph/resolver.go
```diff
@@ -1724,7 +1724,7 @@ func (r *Resolver) sendErrorAlert(ctx context.Context, projectID int, sessionObj
 			}
 			excluded := false
 			for _, env := range excludedEnvironments {
-				if env != nil && *env == sessionObj.Environment {
+				if env != nil && *env == errorObject.Environment {
```
ccschmitz, line 1727: @Vadman97 @mayberryzane does this change make sense to you? I think this resolves one of the issues discussed in [this Slack thread](https://highlightcorp.slack.com/archives/C062LLVR9LK/p1704963364911419), but wanted to confirm there wasn't a reason to be comparing against the session's environment rather than the error's. Assuming it's just because we didn't support environment on errors when the error alerts were implemented, but thought I'd double check. Thanks!

mayberryzane (approved):

backend/public-graph/graph/resolver.go
Vadman97: yup, that's exactly right. a frontend error's environment [will be set](https://github.com/highlight/highlight/blob/da48ded61786ebaccb54d74e59345a43beb1abcc/backend/public-graph/graph/resolver.go#L2744) to the session's environment anyways, and for backend errors using the backend sdk's reported environment makes much more sense

Vadman97 (approved):

PR #8178, opened 2024-04-02
Remove log alert excluded envs

SpennyNDaJets:
## Summary
Users should be using the query line to implement LogAlert environments - not the excluded environments.

1. Create a log alert
- [ ] Able to create log alert
2. Update a log alert
- [ ] Able to update log alert
3. Trigger a log alert
- [ ] Alert is sent to respective channel
4. Repeat for a non-log alert

All data has been migrated to use query instead of excluded environments
```
SELECT COUNT(*) FROM log_alerts WHERE excluded_environments != '' AND excluded_environments IS NOT NULL AND excluded_environments != '[]';
> 0
```

N/A


backend/private-graph/graph/schema.resolvers.go
```diff
@@ -9370,3 +9359,19 @@ type sessionAlertResolver struct{ *Resolver }
...
+//   - You have helper methods in this file. Move them out to keep these resolver files clean.
+func (r *logAlertResolver) ExcludedEnvironments(ctx context.Context, obj *model.LogAlert) ([]string, error) {
+	envs, err := obj.GetExcludedEnvironments()
+	if err != nil {
+		return nil, err
+	}
+	return lo.Map(envs, func(env *string, idx int) string {
+		return *env
+	}), nil
+}
```
ccschmitz, line 9377: Should be safe to delete this, right?

frontend/src/pages/Alerts/LogAlert/LogAlertPage.tsx
```diff
@@ -139,7 +135,7 @@ export const LogAlertPage = () => {
 				query: data?.log_alert.query,
 				name: data?.log_alert.Name,
 				belowThreshold: data?.log_alert.BelowThreshold,
-				excludedEnvironments: data?.log_alert.ExcludedEnvironments,
+				excludedEnvironments: [], // log alerts use query instead of excluded environments
```
ccschmitz, line 138: What's the reason for including this at all? Is it just because the `LogAlertForm` inherits from `AlertForm`? If yes, I think could we use something like `Omit<AlertForm, 'excludedEnvironments'>` to update the types and drop this.

ccschmitz (approved):
Looks great, but a couple quick questions on things we can potentially clean up still!

PR #8184, opened 2024-04-03
Fix alert styles

SpennyNDaJets:
## Summary
Long alert names cause weird flexbox issues

Screenshot 2024-04-03 at 11 23 54 AM

1) Create a new alert with a long name
2) View the alert on the alerts page
- [ ] UI is consistent across different alerts

N/A

N/A

### DIFF
--- a/frontend/src/pages/Alerts/Alerts.tsx
+++ b/frontend/src/pages/Alerts/Alerts.tsx
@@ -323,415 +323,15 @@ function AlertsPageLoaded({
 								{alertsAsTableRows.length > 0 ? (
 									<>
 										{alertsAsTableRows.map(
-											(record, idx) => {
-												return (
-													<Box
-														key={idx}
-														border="dividerWeak"
-														width="full"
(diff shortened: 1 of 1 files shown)


frontend/src/pages/Alerts/Alerts.tsx
```diff
@@ -752,6 +352,253 @@ function AlertsPageLoaded({
...
+									shape="basic"
+									emphasis="low"
+									iconRight={<IconSolidInformationCircle />}
+								></Tag>
+							}
+						>
+							{record.configuration.description}
+						</Tooltip>
+					</Box>
+					<Box display="flex" gap="8" flexShrink={0}>
```
SpennyNDaJets, line 439: Only change is here -> `flexShrink={0}`

SpennyNDaJets (commented):
Pulled out the row into its own component to help with readability, but only change is labeled down below

Vadman97 (approved):
:100:

PR #8192, opened 2024-04-04
Fix trace attributes not appearing

ccschmitz:
## Summary

Fixes an issue where trace attributes weren't being shown on a trace ([more context in Slack](https://highlightcorp.slack.com/archives/C02N2AZS8BV/p1712175890638669)). Also tweaks the attribute cleaning logic so we don't filter out values like `0` or `null` which could be useful to the end user.

Click tested [the problematic trace](https://app.highlight.io/29954/traces/a8a037c749e30845c9c4a8517d7255fa/d4de3069dd0405ca?query=service_name%21%3Dmagicsky-frontend%20process.executable.path%3D%2Fvar%2Flang%2Fbin%2Fnode) in the PR preview to confirm the changes. Also added some unit tests around the formatting/attribute cleaning logic.

N/A

N/A


Vadman97 (approved):
thanks for the quick fix!

PR #8193, opened 2024-04-04
Allow searching logs by `message` key

ccschmitz:
## Summary

You can't search logs using the `message` key on a filter. This is confusing because people see `message` in the log attributes and try to filter on it. This PR fixes that.

Performed a number of searches using the `message` key and confirmed the value was being mapped properly using a breakpoint in my IDE and checking the values.

Screenshot 2024-04-04 at 3 44 26 PM

N/A

N/A


SpennyNDaJets (approved):
LGTM

PR #8194, opened 2024-04-04
Aggressive caching for player

ccschmitz:
## Summary

Updates the caching strategy for requests that retrieve the data necessary to play a session. We went from the default caching or the `no-cache` policy to using the `cache-and-network` policy which will retrieve from the cache while fetching in the background and update the cache once data from the server is retrieved. This allows sessions to be immediately playable if they have already been loaded once and still fetches new data if it came in for a live session.

Also fixed a bug where we weren't destroying the replayer when transitioning away from the session panel to the player page. In order to do that we had to pass a ref around to access the player DOM rather than relying on querying the DOM for replayer elements and cleaning them up.

This PR also autoplays the session if it is opened in the related session panel and makes sure we are using the cached data to populate the session feed. Note that in many places we now need to check if `loading === true` and `data === undefined` to determine if we should show a loading state since `loading` could be `true` for the background request and we want to show the cached data on initial load.

* [ ] Click test going back and forth between sessions on session search. Ensure things load quickly with no new bugs.
* [ ] Click test going back from related resource session panel -> session player. Ensure things load quickly with no new bugs.
* [ ] No bugs transitioning back and forth to live sessions.
* [ ] Fetching event chunks that are already cached don't cause UI jank.
* [ ] Ensure live sessions still loading correctly

All client-side changes, but the handoff between session panel and player page should be much smoother. However, we should keep an eye out for edge cases and any bugs around the session replay.

N/A - should only be a performance improvement w/ no visible changes.


frontend/src/pages/Player/PlayerHook/PlayerState.ts
```diff
@@ -736,9 +748,6 @@ const processSessionMetadata = (
 		return s
 	}
 
-	if (s.replayerState < ReplayerState.Playing) {
-		s.replayerState = ReplayerState.Paused
-	}
```
ccschmitz, line 741: This didn't appear to be doing anything, and the `<` check may not have been accurate anymore since the `ReplayerState` enum was converted to using string values in https://github.com/highlight/highlight/pull/3308/files#diff-9e3458faf1153040ff59b48e4c06213925aaa30bd2920628945825651f76f2e9.

Vadman97 (commented):
https://www.loom.com/share/c3c22ea170a9465488c09af2bcb329ba
* sessions page stuck in loading state
* related session -> session page redownloads session-contents files

Vadman97 (approved):

PR #8197, opened 2024-04-05
Fix creating user on password login

SpennyNDaJets:
## Summary
When creating a new user on password auth, we are creating a new user without an email.

Solution:
- backend: do not allow a user to be created without a UID (token) being provided in the request
- frontend: set the token before trying to send any requests

For Password and Simple auth, sign up and sign in behave the same.
For Cloud app, can sign in with Google/Github from either page, but must be on correct page to sign up/in with a password.

https://www.loom.com/share/a936f08afdc34b9191bf428fa73115b5

1) In [this file](https://github.com/highlight-run/highlight/blob/33ad55d6cc9c74ba825bad3256e30cd0ed4e177a/frontend/src/util/onPrem/onPremUtils.ts), set `isOnPrem` to `true`
2) Use the `/docker/run-backend.sh` and `/docker/run-frontend.sh`script to use the `password` auth type
3) Create a new user on a fresh database
- [ ] Redirected to the "About you" page
- [ ] User created in DB with email
4) Fill out the "About you" information
- [ ] Redirected correctly to your workspace

Miscellaneous
- [ ] Can still sign in on Cloud app
- [ ] Can still sign up on Cloud app

Cloud app: N/A
Enterprise deployments: ensure this fixes issues with hobby app

N/A


frontend/src/pages/Auth/SignUp.tsx
```diff
@@ -77,24 +78,31 @@ export const SignUp: React.FC = () => {
...
+					auth.currentUser?.sendEmailVerification()
+				}
 
-			await createAdmin()
-			message.success('Account created succesfully!')
+				await createAdmin()
+				message.success('Account created succesfully!')
+			}
 
+			await fetchAdmin()
```
Vadman97, line 98: is this what ensures we block other requests until the token is set by the auth.tsx `/login` request?

Vadman97 (approved):
looks great! wanted to confirm that joining with the same email puts you back into the same workspace while signing in with a different user still creates a new workspace (that user 1 can't see).

frontend/src/pages/Auth/SignUp.tsx
Vadman97: ah awesome, thanks for getting to the bottom of this

PR #8202, opened 2024-04-08
metric graphs within a dashboard

mayberryzane:
## Summary
- adds the ability to add, edit, view graphs within a single dashboard
Screen Shot 2024-04-08 at 9 05 25 AM

- clicktested locally

- behind highlight admin flag
- dashboard overview page will be worked on next - for now a dashboard can be added directly in postgres

- @julian-highlight to review after this is released (includes backend changes)


frontend/src/pages/Graphing/ExpandedGraph.tsx
```diff
@@ -0,0 +1,126 @@
...
+							width="full"
+							height="full"
+							px="12"
+							py="16"
+						>
+							<Graph
+								title={g.title}
+								viewConfig={getViewConfig(
+									g.type,
+									g.display ?? undefined,
```
SpennyNDaJets, line 101: Nit: Is this necessary to do the `?? undefined`? Seems like we should be able to leave out

Possibly the `getViewConfig` types need to be updated to
```
export const getViewConfig = (
	viewType: string,
	display?: Maybe<string>,
	nullHandling?: Maybe<string>,
)
```

frontend/src/hooks/useDataTimeRange.ts
```diff
@@ -33,27 +34,32 @@ const setDataTimeRange = makeVar<DataTimeRange>({
 const useDataTimeRange = () => {
 	const timeRange = useReactiveVar(setDataTimeRange)
 
-	const setTimeRange = (start: string, end: string, absolute = false) => {
-		const startDate = moment(start).startOf('minute')
-		const endDate = moment(end).startOf('minute')
-		const lookback = moment.duration(endDate.diff(startDate)).asMinutes()
+	const setTimeRange = useCallback(
```
SpennyNDaJets, line 37: Thanks! Think this is solving some issues I was seeing 🙌

frontend/src/pages/Graphing/components/Graph.tsx
```diff
@@ -165,15 +210,121 @@ export const CustomXAxisTick = ({
...
+	limitMetric,
 	title,
-	loading,
 	viewConfig,
+	onShare,
+	onExpand,
+	onEdit,
 }: ChartProps<ViewConfig>) => {
+	const [graphHover, setGraphHover] = useState(false)
+	const queriedBucketCount = bucketByKey !== undefined ? bucketCount : 1
```
SpennyNDaJets, line 268: Not sure if this is relevant, but if `bucketByKey` is `null`, then `null !==  undefined` will return `true`

However, we can do `!=` to capture null and undefined values. This will still let `""` and `0` evaluate to true

SpennyNDaJets (approved):
Code looks good to me! Just a few questions but feel free to ignore

frontend/src/pages/Graphing/components/Graph.tsx
mayberryzane: I think in this case bucketByKey is `string | undefined` so the original check should work

PR #8207, opened 2024-04-08
Improve small span display on flame graph

ccschmitz:
## Summary

Tweaks the display of really small spans on the trace flame graph.

* Sets a minimum width for a span on the flame graph to make it easier to see and interact with spans that have a very short duration.
* Fixes the duration display for spans that report a `0` duration.

The previews below are looking at [this trace](https://app.highlight.io/1/traces/UGNvAHtaX3/ae30c0e32967e662) for reference.

**Before**

**After**

Click tested some traces in prod and contrasted them with the PR preview for the same trace (see before and after above).

N/A - client-side change only.

Will review the changes w/ Julian before merging.


SpennyNDaJets (approved):
I like it!

PR #8214, opened 2024-04-10
Fix search params being set in wrong place

ccschmitz:
## Summary

We have a bug where interacting with UI in the related resource panel will manipulate the main page's search query rather than the query used for the related resource in the panel. e.g. if you are on the traces page and click on "Apply as filter" on a log attribute from the logs' related resource panel it will apply the filter on the traces' query.

My workaround for this is to move some of the search state/logic into a `SearchContext` that can be added around search components, then use the context (via `useSearchContext`) to access a setter that will update the `query` just for that instance of the search.

Also makes a few small tweaks to fix some issues I noticed while click testing.

* Centers the reset ("X") button to the right of the search filters
* Sorts the root spans so we always show the longest one first if we have two that start around the same time (temporary fix improvement until we fix our timing issues)

* Go to the traces page
* Open a trace
* Click "View logs" on the trace panel
* Expand a log and click "Apply as filter" on one of the attributes
  * Make sure this doesn't update the traces' search query

Should only fix a client-side bug. No backend changes.

N/A


frontend/src/components/Search/SearchForm/SearchForm.tsx
```diff
@@ -627,7 +624,7 @@ export const Search: React.FC<{
 				/>
 
 				{isDirty && !disableSearch && (
-					<Box pt="6">
+					<Box style={{ paddingTop: 9 }}>
```
ccschmitz, line 627: Centers the reset button vertically, which appeared to be a few pixels off.

SpennyNDaJets (commented):
I noticed a lot of the log attributes were missing the filter by button. Not sure if this was expected, but only looks to be the case on the related-resource view

Screenshot 2024-04-16 at 4 36 13 PM

ccschmitz:
@SpennyNDaJets thanks for catching that! In that context I think we want to disable that feature entirely since we don't want people interacting with that search box. I just fixed that up and merged in your latest changes + handled conflicts. Should be all set!

SpennyNDaJets (approved):
Click test looks good to me!

PR #8220, opened 2024-04-10
report more client events

Vadman97:
## Summary

A customer suggested that our client sdk should report additional track events
depending on certain user actions. The change ensures that automatically reported attributes
such as `clickTextContent` become session attributes that can be visible in the replay events.

loom

changeset

no


SpennyNDaJets (commented):
Looks like some builds are failing on EventStream?

PR #8224, opened 2024-04-11
fix xhr monkeypatch breaking for URL objects

Vadman97:
## Summary

Our XHR monkeypatch would break with `URL` objects passed as a constructor, a valid argument,
because we would expect it to be a string where we can call `.toLowerCase()`
`@mux/mux-player` would use this type of constructor for the XHR calls under the hood, resulting
in an error internal to the library trying to fetch the video.

image
image

This PR fixes the monkeypatch to support the URL object.
Also improves the dev experience with highlight.run (bundled client) to support hot reloading.

See https://highlightcorp.slack.com/archives/C06NYBC45EJ/p1711674863357099

local testing with @mux/mux-player
image

changeset

no


SpennyNDaJets (approved):
LGTM!

PR #8249, opened 2024-04-15
Set width for flame graph

SpennyNDaJets:
## Summary
For network resources, it appears the width is not being updated when the ref component is mounted, resulting in missing flame graph data.

Video TBD

1) View [this trace](https://app.highlight.io/1/sessions/o1yhwFT9CRWjM7xYdvtI2Gt51guS?network-resource-id=264&page=1&query=and%7C%7Ccustom_processed%2Cis%2Ctrue&segment=%22undefined%22), and confirm the flame graph is correctly displayed
2) Confirm the trace can be viewed from the traces search page
3) Confirm the trace can be viewed from the logs search page

N/A

N/A

### DIFF
--- a/frontend/src/pages/Traces/TraceFlameGraph.tsx
+++ b/frontend/src/pages/Traces/TraceFlameGraph.tsx
@@ -20,7 +20,6 @@ import {
 } from 'react'
 
 import LoadingBox from '@/components/LoadingBox'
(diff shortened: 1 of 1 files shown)


SpennyNDaJets:
https://github.com/highlight/highlight/pull/8268

PR #8268, opened 2024-04-16
Fix flame graph not appearing on network resources

ccschmitz:
## Summary

Updates the behavior of tab content to not be mounted until the tab is shown. This is consistent with the behavior we had in our old `Tabs` component and fixes a bug where we don't calculate the `width` property on the trace flame graph when put behind a tab on the network resource panel.

In order to get this behavior working we also needed to upgrade Ariakit because there was a bug where the tabs would _never_ render when using the `unmountOnHide` prop.

View a fetch network resource in the session dev tools and ensure the trace flame graph is rendered correctly.

N/A

N/A


SpennyNDaJets (approved):
LGTM!

PR #8274, opened 2024-04-16
fall back to unversioned sourcemaps

Vadman97:
## Summary

Closes HIG-2599
Closes HIG-4506

new unit testing

testing the sourcemap generation

before:
Screenshot from 2024-04-16 14-23-00

after:
Screenshot from 2024-04-16 14-28-29
Screenshot from 2024-04-16 14-27-47

will be looking at % of errors enhanced with sourcemap before/after merging this change

no

### DIFF
--- a/backend/public-graph/graph/resolver.go
+++ b/backend/public-graph/graph/resolver.go
@@ -361,8 +361,8 @@ func (r *Resolver) GetErrorAppVersion(ctx context.Context, errorObj *model.Error
 	}
 
 	// guess version from error service
-	if errorObj.ServiceName != "" && errorObj.ServiceVersion != "" {
-		return pointy.String(fmt.Sprintf("%s-%s", errorObj.ServiceName, errorObj.ServiceVersion))
+	if errorObj.ServiceVersion != "" {
+		return pointy.String(errorObj.ServiceVersion)
 	}
 
 	return nil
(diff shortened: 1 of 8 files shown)


backend/public-graph/graph/resolver.go
```diff
@@ -356,17 +356,21 @@ func (r *Resolver) GetErrorAppVersion(ctx context.Context, errorObj *model.Error
...
-		if !e.Is(err, gorm.ErrRecordNotFound) {
-			return nil
-		}
+		Pluck("app_version", &session).Error; err == nil && session.AppVersion != nil {
+		return session.AppVersion
+	}
+
+	// guess version from error service
+	if errorObj.ServiceName != "" && errorObj.ServiceVersion != "" {
+		return pointy.String(fmt.Sprintf("%s-%s", errorObj.ServiceName, errorObj.ServiceVersion))
```
SpennyNDaJets, line 365: Curious why we need `ServiceName` here? Wouldn't `ServiceVersion` be the parallel to `session.AppVersion` (i.e. githash, although it can be any user inputted string)? 

Not blocking, assuming this may be hard to update for backwards compatibility, but was a little unsure how this was used.

SpennyNDaJets (approved):

backend/public-graph/graph/resolver.go
Vadman97: a very good point... i just added this recently and no one is using this versioning for backend errors, so i'll switch this to just `ServiceVersion`

PR #8281, opened 2024-04-17
Swap date picker on log alerts

SpennyNDaJets:
## Summary
The logs alert page was using the `PreviousDateRangePicker`, but needed to use the values from the query passed into it. Swap it out with the search `DateRangePicker` and use the `useSearchTime` hooks to allow updating the url.

https://www.loom.com/share/82611b06d77941cba8b46fa7cdba8424?sid=d008b64e-5711-44d5-bfe3-89cf482df085

1) Make a log query in which the time is updated
2) Create an alert from that query
- [ ] Params get moved over correctly
- [ ] Time is also ported over
- [ ] Reloading the page will reset any changes to log queries original params
- [ ] Time is held when reloading page

N/A

N/A

### DIFF
--- a/__generated/rr/rrweb/rr.js
+++ b/__generated/rr/rrweb/rr.js
@@ -1588,13 +1588,11 @@ function getTagName(n2) {
 function escapeRegExp(str) {
   return str.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
 }
+var MEDIA_SELECTOR = /(max|min)-device-(width|height)/;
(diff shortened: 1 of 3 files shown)


__generated/rr/rrweb/rr.js
Vadman97, line 1: i think this is reverting some changes? could you try sync submodules and rerun `yarn dev:frontend`?

Vadman97 (approved):

PR #8284, opened 2024-04-17
Fix all instances link

SpennyNDaJets:
## Summary
When clicking on the all instances link, it resets the search query for the overall search. Pass in the search to hold the current search values

1) Search for errors (not using the default search)
2) Click into and error and view all instances
- [ ] All instances page loaded
- [ ] Error search does not change

N/A

N/A

### DIFF
--- a/backend/scripts/migrate-error-segments/main.go
+++ b/backend/scripts/migrate-error-segments/main.go
@@ -257,12 +257,12 @@ func buildStringQuery(fieldName modelInputs.ReservedErrorsJoinedKey, op string, 
 		queryArray = append(queryArray, fmt.Sprintf("%s!=%s", fieldName, value))
 	case "matches":
 		for _, v := range val {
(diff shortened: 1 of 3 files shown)


backend/scripts/migrate-error-segments/main.go
```diff
@@ -257,12 +257,12 @@ func buildStringQuery(fieldName modelInputs.ReservedErrorsJoinedKey, op string,
 		queryArray = append(queryArray, fmt.Sprintf("%s!=%s", fieldName, value))
 	case "matches":
 		for _, v := range val {
-			value := fmt.Sprintf("\\%s\\", v)
+			value := fmt.Sprintf("/%s/", v)
```
SpennyNDaJets, line 260: Ran this in prod with this change, but wanted to officially change it since this will be the basis for the sessions migration

mayberryzane (approved):

PR #8298, opened 2024-04-18
Fix error group autoresolver data sync

SpennyNDaJets:
## Summary
The error group auto resolver passes in `nil` ids to the data sync queue. Pass in the param id, and refactor to never use the error group.

https://www.loom.com/share/83f8b158e4be4a6a93ef6cc2dccab35e?sid=43bad0d2-ba8d-4cc8-8d92-632dce2bda74

1) Search for `status=OPEN` errors
2) Update an error to be resolved
3) Refresh the search (may have a lag)
- [ ] Error no longer in search

N/A

N/A

### DIFF
--- a/backend/store/error_groups.go
+++ b/backend/store/error_groups.go
@@ -190,8 +190,25 @@ type UpdateErrorGroupParams struct {
 }
 
 func (store *Store) UpdateErrorGroupStateByAdmin(ctx context.Context,
-	admin model.Admin, params UpdateErrorGroupParams) error {
-	return store.updateErrorGroupState(ctx, &admin, params)
+	admin model.Admin, params UpdateErrorGroupParams) (*model.ErrorGroup, error) {
+	err := store.updateErrorGroupState(ctx, &admin, params)
+	if err != nil {
+		return nil, err
+	}
+
+	// For user-driven state updates, write the error group directly to Clickhouse.
+	// Write to the data sync queue as well to guarantee eventual consistency.
+	var errorGroup model.ErrorGroup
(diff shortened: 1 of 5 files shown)


backend/store/error_groups.go
```diff
@@ -229,28 +227,19 @@ func (store *Store) updateErrorGroupState(ctx context.Context,
...
 	})
 
 	if err != nil {
-		return errorGroup, err
-	}
-
-	// For user-driven state updates, write the error group directly to Clickhouse.
-	// Write to the data sync queue as well to guarantee eventual consistency.
-	if admin != nil {
-		err = store.clickhouseClient.WriteErrorGroups(ctx, []*model.ErrorGroup{&errorGroup})
```
SpennyNDaJets, line 243: Are we good to remove this? (c.c. @Vadman97 @mayberryzane) This creates a small lag in which if the user manually resolves an error, they may still see it in the search until the datasync queue catches up

Vadman97: @SpennyNDaJets do you know what kind of delay we see locally?

SpennyNDaJets: Its a little awkward, so thinking of adding this back. Since the page reloads, this error typically sits at the top with the "Resolved" tag.

Was also looking into sorting the errors by the last instance timestamp so this will also be an improvement, but part of a different PR

Vadman97: sounds good, i guess we should keep this then and just make sure we query the error group from postgres here?

SpennyNDaJets: Moved into the `UpdateErrorGroupStateByAdmin`

Vadman97 (approved):
nice!

PR #8303, opened 2024-04-19
Revert combobox changes

SpennyNDaJets:
## Summary
New flex value is causing combobox to collapse on words

Screenshot 2024-04-19 at 9 57 59 AM

Revert flex property

Need to add back solution for dashboards page

N/A


SpennyNDaJets:
c.c. @mayberryzane just reverting this change for now since dashboards is behind toggle, but happy to work on a longer term solution

mayberryzane (approved):

PR #8328, opened 2024-04-23
Swap out session query builder

SpennyNDaJets:
## Summary
Swap out the session query builder

Other notable changes/non-changes:
- Backend: swap out deleteSessions params for new search language
- removes the session search context wrapper
- moves the session saved searches to the new table
- removes command bar
- removes preloading sessions
- does not edit ingestion filters

https://www.loom.com/share/aee3267d8f4e45008367515d70cb4b43?sid=bc3c843d-0b21-4b5c-9dd5-3ef4f42d5ca5

Closes HIG-4464

TBD

Must migrate saved searches before deploying

Yes


ccschmitz (approved):
Code + click test look great! Nice work on this, @SpennyNDaJets!!

One thing I'm noticing in the filter suggestions are some keys that aren't valid (see screenshot below).

Screenshot 2024-04-30 at 2 26 06 PM

Maybe an issue with my local data, but wanted to mention it.

Wondering if we can/should prioritize cleaning up the dropdown list to make it more functional now that we're filtering sessions which could have a ton of filterable attributes. I also wonder if we should clean up things like the `loc_state` filter you mentioned in your screencast 🤔 - not something to address in this PR, but perhaps dig into with @julian-highlight as a future improvement.

Looking great overall!

PR #8345, opened 2024-04-25
Fix instance link

SpennyNDaJets:
## Summary
When clicking an error instance, the search gets cleared.

1) View the error search
2) Modify the search to not be the default search
3) Click into an error and view all instances
4) Click on an instance
- [ ] Instance loads
- [ ] Search does not change
- [ ] No new search requests made

N/A

N/A

### DIFF
--- a/__generated/rr/rrweb/rr.js
+++ b/__generated/rr/rrweb/rr.js
@@ -4648,7 +4648,7 @@ var CanvasManager = class {
         return hadLoadingTime;
       };
       const promises = [];
-      promises.push(...getCanvas(timestamp).filter(filterElementStartTime).map(this.snapshot));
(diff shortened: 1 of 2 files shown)


__generated/rr/rrweb/rr.js
```diff
@@ -4648,7 +4648,7 @@ var CanvasManager = class {
         return hadLoadingTime;
       };
       const promises = [];
-      promises.push(...getCanvas(timestamp).filter(filterElementStartTime).map((canvas) => this.snapshot(canvas)));
+      promises.push(...getCanvas(timestamp).filter(filterElementStartTime).map(this.snapshot));
```
Vadman97, line 4651: this is reverting the latest rrweb fix - i think you committed rrweb unintentionally before updating it in your branch

Vadman97 (approved):
preview works great

PR #8351, opened 2024-04-25
Prep backend for session search swap

SpennyNDaJets:
## Summary
In order to swap the session search with the new search UX, we need the following backend updates:
- Rebuild `sessions_joined_vw` to include latest session fields
- Add new boolean and creatable type for filters
- Add missing session keys (rename a few to map to field name)
- Add solution for custom filters (i.e. `sample` and `viewed_by_me`)

Tested in https://github.com/highlight/highlight/pull/8328

Most changes are not affect production. However, make sure the different search experiences are not affected (include graphana)

N/A

N/A


backend/clickhouse/migrations/000098_recreate_sessions_joined_vw.down.sql
```diff
@@ -0,0 +1,13 @@
...
+select ProjectID as ProjectId,
+    CreatedAt as Timestamp,
+    mapFromArrays(
+        arrayMap(x->splitByChar('_', x, 2) [2], FieldKeys),
+        arrayMap(
+            (k, kv)->substring(kv, length(k) + 2),
+            arrayZip(FieldKeys, FieldKeyValues)
+        )
+    ) as SessionAttributes,
+    *
```
Vadman97, line 12: isn't this the same as the schema in the `up` migration?

Vadman97 (approved):

backend/clickhouse/migrations/000098_recreate_sessions_joined_vw.down.sql
SpennyNDaJets: It is! Which is the same as `backend/clickhouse/migrations/000066_add_sessions_joined_vw.up.sql`. The issue is we added fields to sessions after we created the mv, so that mv doesn't contain those fields, but the logic is still sound. We did something similar to `errors_joined_vw`.

Vadman97: oh makes sense, thanks!

PR #8352, opened 2024-04-26
refactor worker session search

Vadman97:


backend/main.go
```diff
@@ -223,7 +223,7 @@ func main() {
 		samplingMap = map[trace.SpanKind]float64{
 			trace.SpanKindUnspecified: 1. / 1_000_000,
 			trace.SpanKindInternal:    1. / 1_000_000,
-			trace.SpanKindConsumer:    1. / 1_00,
+			trace.SpanKindConsumer:    util.ConsumerSpanSamplingRate(),
```
SpennyNDaJets, line 226: 🔥 I like the idea of making these env variables so we can switch them faster than needing a new PR/deploy

SpennyNDaJets (approved):
LGTM!

PR #8357, opened 2024-04-26
Migrate session segments

SpennyNDaJets:


Vadman97 (approved):
nice, looks reasonable

PR #8403, opened 2024-04-30
Fix performance issue with query editor on logs

ccschmitz:
## Summary

Editing a search query on the logs page is sluggish and causes some jank in the UI. This is because we try to update the text highlight for query matches on every log row with every keystroke. This PR works around this issue by passing the needed query data down via props to the logs table and only passing the top-level `query` variable that is updated when the form is actually submitted.

* Open the logs page in production and enter some query data. Note the sluggishness when typing in the input.
* Open the PR preview for this branch. Things should be much more responsive now!

N/A - all client-side changes.

N/A - no visual change from this.


Vadman97 (approved):
way better performance! noticed a few stutters still but they are quite rare now, the main perf issue is definitely fixed :tada: 

Screenshot from 2024-05-01 08-56-38

PR #8409, opened 2024-05-01
Error alert search queries

SpennyNDaJets:
## Summary
Allow users to specify queries on the error alerts
- Removes (and migrates) the excluded environments
- Removes regex expressions for new alerts and alerts with no regex expressions

https://www.loom.com/share/244777ebf646411bbca64be9f3ea3091?sid=0ba759d4-0d15-4ee8-840c-a540842b4510

1) Create a new error alert with a query
  a) Add a threshold of 1
  b) Add a slack/discord channel or email
2) Kick an error off that matches the query
- [ ] Alert notifies correct channel
3) Kick an error off that does NOT match the query
- [ ] Alert does not notify the channel

Run migration first

N/A


backend/private-graph/graph/schema.resolvers.go
```diff
@@ -3494,28 +3484,21 @@ func (r *mutationResolver) UpdateErrorAlert(ctx context.Context, projectID int,
...
-		projectAlert.ExcludedEnvironments = envString
+		projectAlert.ChannelsToNotify = channelsString
 	}
 
-	if slackChannels != nil {
-		channelsString, err := r.MarshalSlackChannelsToSanitizedSlackChannels(slackChannels)
-		if err != nil {
-			return nil, err
-		}
+	if regexGroups != nil {
```
SpennyNDaJets, line 3496: Moved this out from the `if slackChannels != nil` check 🤔

backend/private-graph/graph/schema.resolvers.go
```diff
@@ -3557,7 +3542,7 @@ func (r *mutationResolver) UpdateErrorAlert(ctx context.Context, projectID int,
 		Model: model.Model{
 			ID: errorAlertID,
 		},
-	}).Where("project_id = ?", projectID).Updates(projectAlert).Error; err != nil {
+	}).Select("query").Where("project_id = ?", projectID).Updates(projectAlert).Error; err != nil {
```
SpennyNDaJets, line 3545: Allow empty queries to update appropriately

frontend/src/pages/Alerts/ErrorAlert/ErrorAlertPage.tsx
```diff
@@ -402,9 +392,19 @@ export const ErrorAlertPage = () => {
...
 									<AlertTitleField />
-									<ErrorAlertForm />
+									<SearchContext
+										initialQuery={initialQuery}
+										onSubmit={handleUpdateInputQuery}
+									>
+										<ErrorAlertForm
+											hideRegexExpression={
+												isCreate ||
+												!alert?.RegexGroups?.length
```
SpennyNDaJets, line 404: Only show regex expression for older alerts that have non empty regex expressions

mayberryzane (approved):
thanks for video and explanation on not migrating the regex field, makes sense to me! on that same topic - is it possible to filter on the stack trace right now? if not, it could be something useful to have, or in the future add some fake key that we can use to filter on both event and stack trace at the same time to replace this

SpennyNDaJets:
> thanks for video and explanation on not migrating the regex field, makes sense to me! on that same topic - is it possible to filter on the stack trace right now? if not, it could be something useful to have, or in the future add some fake key that we can use to filter on both event and stack trace at the same time to replace this

At the moment, it is not possible - we are storing the stack trace in clickhouse, so would have to add there or come up with something more creative. I created a card here: https://linear.app/highlight/issue/HIG-4609/allow-filtering-by-stack-trace-matching

PR #8410, opened 2024-05-01
tooltips for graphing editor

mayberryzane:
## Summary
- adds tooltips with each linking to the graphing fields docs page
- adds hovercard support without adding hovercards yet
Screen Shot 2024-05-01 at 8 50 17 AM

- clicktested locally

- no

- @julian-highlight


frontend/src/pages/Graphing/GraphingEditor.tsx
```diff
@@ -105,22 +108,48 @@ const SidebarSection = (props: PropsWithChildren) => {
...
 	name: string
 	enabled?: boolean
 	setEnabled?: (value: boolean) => void
+	tooltip?: string
 }>) => {
 	return (
 		<Box width="full" display="flex" flexDirection="column" gap="4">
 			<Box display="flex" flexDirection="row" gap="6">
 				<Label label={label} name={name} />
+				{tooltip !== undefined && <InfoTooltip text={tooltip} />}
```
SpennyNDaJets, line 152: Wonder if we can just do `!tooltip` here? May just help with future with bugs where tooltip is `null` or `""` and we don't want to display?

mayberryzane: I don't think it's a huge change either way but I prefer the `!== undefined` because
1. it can't be null here and if its type updates to include null this line would throw a type error
2. to me, the expected behavior when setting `tooltip=""` would be to show a blank tooltip rather than no tooltip
3. I've run into short circuiting bugs with `&&` which could happen if tooltip's type ever changes to include `0` or `false`

PR #8413, opened 2024-05-01
fix count distinct for metrics api

Vadman97:
## Summary

Fixes `CountDistinct` operator not working for non-numeric values.

https://www.loom.com/share/8a00fd6e20d84a20a853796ea87fc625

no

no

### DIFF
--- a/backend/clickhouse/query.go
+++ b/backend/clickhouse/query.go
@@ -596,8 +596,13 @@ func readWorkspaceMetrics[T ~string](ctx context.Context, client *Client, sample
 	} else {
 		config = sampleableConfig.tableConfig
 	}
+	var isCountDistinct bool
 	for _, metricType := range metricTypes {
+		if metricType == modelInputs.MetricAggregatorCountDistinct {
+			isCountDistinct = true
+		}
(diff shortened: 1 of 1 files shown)


backend/clickhouse/query.go
```diff
@@ -611,11 +616,13 @@ func readWorkspaceMetrics[T ~string](ctx context.Context, client *Client, sample
 		OrderForwardNatural,
 	)
 
-	var metricExpr string
-	if col, found := keysToColumns[T(strings.ToLower(column))]; found {
-		metricExpr = fmt.Sprintf("toFloat64(%s)", col)
-	} else {
-		metricExpr = fmt.Sprintf("toFloat64OrNull(%s[%s])", attributesColumn, fromSb.Var(column))
```
Vadman97, line 618: i consolidated both to `toFloat64OrNull` - this means now that if a `keysToColumns` value is not numeric we would not error but drop those points from the aggregation, which seems safer to me

mayberryzane (approved):
looks good to me, thanks for taking a look 🙏

PR #8414, opened 2024-05-01
Don't allow inviting existing admins or invitees

ccschmitz:
## Summary

We currently allow people to send invites to existing admins and people who have already been invited to a workspace. This PR adds some validations and error handling to ensure this doesn't happen. The checks for existing emails are case-insensitive.

Screenshot 2024-05-01 at 2 31 06 PM

Screenshot 2024-05-01 at 2 31 24 PM

* Try inviting an existing admin and ensure you get an error.
* Try inviting someone who has already been invited and ensure you get an error.
* Ensure emails that don't match admins or existing invitees are created correctly.

Adds some DB load when creating a workspace invite, but only a few reads.

N/A

### DIFF
--- a/backend/private-graph/graph/schema.resolvers.go
+++ b/backend/private-graph/graph/schema.resolvers.go
@@ -882,6 +882,30 @@ func (r *mutationResolver) SendAdminWorkspaceInvite(ctx context.Context, workspa
 		}
 	}
 
+	// Check if an invite to the email address already exists
+	var existingInvite model.WorkspaceInviteLink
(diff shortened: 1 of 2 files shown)


SpennyNDaJets (approved):
Looks good! I wonder if we can also prevent null emails from being created? It doesn't send an invite, but not sure how this can be taken advantage of
Screenshot 2024-05-02 at 10 43 15 AM

ccschmitz:
@SpennyNDaJets I added `required` to the input to prevent submitting empty values!

PR #8415, opened 2024-05-01
Add search query to error instance search

SpennyNDaJets:
## Summary
Allow users to search error instances

https://www.loom.com/share/7aacda9ccd244bb783705ccaec09e730?sid=dc41f87e-5824-4754-bc7e-95c48bc3a863

1) Load the errors page
2) Search for errors using non default query
3) Click on an error and view its instances
- [ ] Initial query is the error group query
- [ ] Only instances matching query are returned
4) Edit the instance query
- [ ] Only instance query is affected
- [ ] Only instances matching query are returned
5) Edit the group query
- [ ] Only the groups are affected
6) Change out of the instance view and back in
- [ ] Initial query is the error group query
- [ ] Only instances matching query are returned

N/A

Yes


backend/clickhouse/errors.go
```diff
@@ -698,6 +698,43 @@ func (client *Client) QueryErrorObjectsHistogram(ctx context.Context, projectId
...
+	offset := (pageInt - 1) * count
+
+	sb, err := readErrorsObjects(params, projectId)
+	if err != nil {
+		return nil, 0, err
+	}
+	sb.Where(sb.Equal("ErrorGroupID", errorGroupId))
+
+	sb.Select("ID, count() OVER() AS total")
+	sb.OrderBy("UpdatedAt DESC, ID DESC")
```
mayberryzane, line 715: should these be ordered by Timestamp? There could be a big delay between when an error instance happens and when it's finally saved in our db

backend/store/error_groups.go
```diff
@@ -3,92 +3,39 @@ package store
...
-	} else if params.Before != nil {
-		query = query.Order("error_objects.id ASC").Where("error_objects.id > ?", *params.Before)
-	} else {
-		query = query.Order("error_objects.id DESC")
-	}
+	query := store.db.WithContext(context.TODO()).
+		Where(&model.ErrorObject{ErrorGroupID: errorGroup.ID}).
+		Where("id IN (?)", ids).
+		Limit(LIMIT + 1).
+		Order("error_objects.id DESC")
```
mayberryzane, line 26: I think the above ordering comment applies here too

mayberryzane (approved):
video looks good! some small comments about ordering but nothing blocking

backend/clickhouse/errors.go
SpennyNDaJets: Good call! Updated here and the frontend to use the timestamp value as well

PR #8416, opened 2024-05-01
use optimistic cache updates in metrics ui

mayberryzane:
## Summary
- fixes a bug where a new graph may not be shown until refreshing the page
- uses apollo's `cache.modify` when creating/editing/deleting graphs and reordering or renaming a dashboard
  - this optimistically modifies the cache with the updated state, while making another query to `GetVisualization` in case of an issue (similar to apollo's 'cache-and-network' mode)
https://www.loom.com/share/e8db544807ed430681450afef840887f
- svg mask seems to be a source of perf issues - replace with clipPath

- clicktested locally

- no

- no

### DIFF
--- a/frontend/src/pages/Graphing/components/BarChart.tsx
+++ b/frontend/src/pages/Graphing/components/BarChart.tsx
@@ -42,10 +42,10 @@ const RoundedBar = (id: string, isLast: boolean) => (props: BarProps) => {
 				height={Math.max((height ?? 0) - 1.5, 0)}
 				stroke="none"
 				fill={fill}
-				mask={`url(#barmask-${id}-${x})`}
+				clipPath={`url(#barmask-${id}-${x})`}
 			/>
 			{isLast && (
-				<mask id={`barmask-${id}-${x}`}>
+				<clipPath id={`barmask-${id}-${x}`}>
 					<rect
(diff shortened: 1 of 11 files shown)


frontend/src/pages/Graphing/components/BarChart.tsx
```diff
@@ -42,10 +42,10 @@ const RoundedBar = (id: string, isLast: boolean) => (props: BarProps) => {
 				height={Math.max((height ?? 0) - 1.5, 0)}
 				stroke="none"
 				fill={fill}
-				mask={`url(#barmask-${id}-${x})`}
+				clipPath={`url(#barmask-${id}-${x})`}
```
Vadman97, line 45: neat

Vadman97 (approved):
https://www.loom.com/share/d04573bd6a814d85b9bbb64665308960

branch looks great locally. i do notice this tiny flicker after saving, seems the graph component is rerendering? not a blocker tho

mayberryzane:
> branch looks great locally. i do notice this tiny flicker after saving, seems the graph component is rerendering? not a blocker tho

didn't see flickering locally but it could be the id of the graph changing (from some temp id to the real one) since I use id as the react key. taking a look

PR #8457, opened 2024-05-03
improvement to errors_joined_vw perf

mayberryzane:
## Summary
- small change to `errors_joined_vw` replaces the line
`SELECT eg.ProjectID as ProjectId,` with `SELECT ProjectID as ProjectId,`
- this filters using the error_objects' primary key, and since it is the larger table in the join this makes a big perf difference

- applied this in prod after testing

- no

- no


mayberryzane:
> don't quite understand why we had those duplicate error groups with the same id... but the view change looks great

@Vadman97 I think there was a change at some point to send higher granularity timestamps w/ microsecond instead of second precision, and because `CreatedAt` is part of the order by key, for some error groups we ended up w/ two rows

PR #8462, opened 2024-05-03
add range select, make time selector replace history

mayberryzane:
## Summary
- add the concept of a reference area to the common `Graph` file - when the x-axis is a timestamp, allow selecting a range in the graph to adjust the dates
- the date pickers were adding to the history whenever the date was changed, but pressing back would not revert to the previous date - for now, I'm making it replace the history
https://www.loom.com/share/59d9bb3c0af243f18bc161dd9f5b65c0

- clicktested locally

- no

- @julian-highlight to review range selecting functionality - I just used a recharts example as inspiration but let me know if I should adjust the colors or anything

### DIFF
--- a/frontend/src/__generated/index.css
+++ b/frontend/src/__generated/index.css
@@ -7477,6 +7477,9 @@ body {
 .dg2vg6c {
   width: min-content;
 }
+.dg2vg6d {
+  user-select: none;
+}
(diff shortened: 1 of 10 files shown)


Vadman97 (approved):
lgtm!

PR #8467, opened 2024-05-06
Remove error segment code

SpennyNDaJets:
## Summary
Now that error search and segments have been migrated and working for the past month, we can clean up the legacy error segment code.

- [ ] Error search still works
- [ ] Saved error segments still work

Harder to revert error experience

N/A


mayberryzane (approved):
🙌

PR #8477, opened 2024-05-06
send billing emails to role = admin only

mayberryzane:
## Summary
- `AdminEmailAddresses` used in billing should return only users with `wa.role = 'ADMIN'`

- unit tested to ensure only admins in the input workspace are returned
- ran query against Highlight workspace to ensure non-admins were filtered out

- no

- no

### DIFF
--- a/backend/model/model.go
+++ b/backend/model/model.go
@@ -328,6 +328,7 @@ func (w *Workspace) AdminEmailAddresses(db *gorm.DB) ([]struct {
 			INNER JOIN admins a
 			ON wa.admin_id = a.id
 			WHERE wa.workspace_id = ?
+			AND wa.role = 'ADMIN'
 			AND NOT EXISTS (
 				SELECT *
 				FROM email_opt_outs eoo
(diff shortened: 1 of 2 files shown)


backend/model/model.go
```diff
@@ -328,6 +328,7 @@ func (w *Workspace) AdminEmailAddresses(db *gorm.DB) ([]struct {
 			INNER JOIN admins a
 			ON wa.admin_id = a.id
 			WHERE wa.workspace_id = ?
+			AND wa.role = 'ADMIN'
```
Vadman97, line 331: could we also audit that all sensitive emails we send use this `AdminEmailAddresses` list?

Vadman97 (approved):
sweet, and thanks for adding a test!

backend/model/model.go
mayberryzane: @Vadman97 I checked other references to `workspace_admins` - I think this is the only sensitive email, the other ones are the digest and session insights emails, which should be fine to send out to everyone in the project, right?

PR #8486, opened 2024-05-06
[HIG-4610] use dataMin, dataMax for x-axis, dot fixes

mayberryzane:
## Summary
- change line and bar chart components to use `dataMin, dataMax` as the x-axis domain instead of `auto, auto`
- fix dot logic which had bugs when multiple series were displayed (needed to check both undefined and null)
- fix categorical line charts not showing data

- clicktested locally

- no

- no


SpennyNDaJets (approved):
Click test looks good to me!

PR #8494, opened 2024-05-07
Metric polling

SpennyNDaJets:
## Summary
Add polling to the metrics page to refresh data every minute

https://www.loom.com/share/d520058cc11d4437929ab0acdf8c779c

1. Load the metrics page with a preset time (i.e. "Last 15 minutes")
- [ ] Metrics are refreshed after ~1 minute (starts after successfully receiving metrics)
- [ ] Metrics are refreshed after another ~1 minute (starts after successfully receiving metrics)
2. Change the time to a fixed time range
- [ ] Metrics are not refreshed after a minute
3. Change time back to a preset time 
- [ ] Metrics are refreshed after ~1 minute (starts after successfully receiving metrics)
4. Expanded a graph
- [ ] Metrics are refreshed after ~1 minute (starts after successfully receiving metrics)
5. Edit a graph
- [ ] Metrics are refreshed after ~1 minute (starts after successfully receiving metrics)

N/A

N/A

### DIFF
--- a/frontend/src/pages/Graphing/components/Graph.tsx
+++ b/frontend/src/pages/Graphing/components/Graph.tsx
@@ -19,12 +19,12 @@ import { vars } from '@highlight-run/ui/vars'
 import clsx from 'clsx'
 import _ from 'lodash'
 import moment from 'moment'
-import { useEffect, useMemo, useState } from 'react'
+import { useCallback, useEffect, useMemo, useRef, useState } from 'react'
 import { ReferenceArea } from 'recharts'
 import { CategoricalChartFunc } from 'recharts/types/chart/generateCategoricalChart'
 
 import { TIME_FORMAT } from '@/components/Search/SearchForm/constants'
-import { useGetMetricsQuery } from '@/graph/generated/hooks'
+import { useGetMetricsLazyQuery } from '@/graph/generated/hooks'
 import { Maybe, MetricAggregator, ProductType } from '@/graph/generated/schemas'
 import {
 	BarChart,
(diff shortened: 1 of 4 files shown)


frontend/src/pages/Graphing/components/Graph.tsx
```diff
@@ -338,6 +341,8 @@ export const getViewConfig = (
 	return viewConfig
 }
 
+const POLL_INTERVAL_VALUE = 60000
```
SpennyNDaJets, line 344: Currently set this to 1 minute - I don't think it makes sense to go less than this since we are using the start the minutes and end of the minute to fetch the metric data, resulting in the cache being hit for fetches within the minute.

We could look to increase this for longer time frames, but think it is still useful to see up-to-date data

frontend/src/pages/Graphing/components/Graph.tsx
```diff
@@ -359,37 +364,102 @@ const Graph = ({
...
-			params: {
-				date_range: {
-					start_date: moment(startDate)
-						.startOf('minute')
-						.format(TIME_FORMAT),
-					end_date: moment(endDate)
-						.startOf('minute')
-						.format(TIME_FORMAT),
+	const pollTimeout = useRef<number>()
+	const [pollInterval, setPollInterval] = useState<number>(0)
```
mayberryzane, line 375: could you do this with the apollo pollInterval setting or did you need more flexibility? https://www.apollographql.com/docs/react/data/queries/#polling

frontend/src/pages/Graphing/components/Graph.tsx
mayberryzane: Not a huge deal, but I think a longer poll interval would be nice for long time periods (6+ hours) because the graph won't change much (each bucket size is likely 5+ minutes) and it will add extra load on clickhouse. Separately, is there anything easy we could do to avoid refreshing while a user is interacting with the graph?

mayberryzane (approved):

frontend/src/pages/Graphing/components/Graph.tsx
SpennyNDaJets: I think more flexibility was needed - the issue I was running into was the request would keep requesting the same start and end times.

So the polling function is actually just "rebasing" the start and end times, which then triggers a new request. Happy to look back into this though if we feel it would be better

frontend/src/pages/Graphing/components/Graph.tsx
SpennyNDaJets: Increased this interval to 5 minutes for data showing 6+ hours

SpennyNDaJets: > Separately, is there anything easy we could do to avoid refreshing while a user is interacting with the graph?

I'd have to look more into this. I don't have a great method off the top of my head, but we could disable the polling on the edit page?

PR #8496, opened 2024-05-07
set minDate based on retention, add new preset

mayberryzane:
## Summary
- add another preset for the earliest retention date, set minDate based on the same
- leave `custom` mode when date picker menu closes

- clicktested locally with different retention settings

- no

- no


frontend/src/components/Search/SearchForm/utils.ts
```diff
@@ -167,3 +178,79 @@ export const buildTokenGroups = (tokens: SearchToken[]) => {
 
 	return tokenGroups
 }
+
+export const useRetentionPresets = (productType: ProductType) => {
```
Vadman97, line 182: looks like the `DateTimePicker` component still has a fixed time range for the UI - could we fix that too? 

Screenshot from 2024-05-07 12-44-26

Screenshot from 2024-05-07 12-45-01

Vadman97 (approved):
thanks for the quick turnaround :tada:

frontend/src/components/Search/SearchForm/utils.ts
mayberryzane: was inconsistent - sometimes had the min date updated and sometimes not, seems to be fixed with a dependency upgrade

PR #8497, opened 2024-05-07
Allow users to sort traces

ccschmitz:
## Summary

Allows users to sort traces by clicking the column headers of the traces table.

https://www.loom.com/share/cd09c436b33548d0bdf6add86439a98e

* Click table headers to trigger sorting
  * Ensure results are sorted correctly
  * Ensure that scrolling loads the correct results
* Ensure scrolling backward on logs loads the correct results
* Ensure all results load as expected for logs and traces

We will want to monitor performance when using sorting. My testing was in development without a lot of data, but it could have a more drastic negative impact on performance in production.

Have been discussing the UX of table headers w/ @julian-highlight in https://linear.app/highlight/issue/HIG-4525/improvements-to-table-headers which will include some additional changes to table headers.


backend/clickhouse/query.go
```diff
@@ -39,13 +39,30 @@ func readObjects[TObj interface{}, TReservedKey ~string](ctx context.Context, cl
...
+	sortDirection := modelInputs.SortDirectionDesc
+	if params.Sort != nil {
+		sortColumn = params.Sort.Column
+		sortDirection = params.Sort.Direction
 	}
 
+	if col, found := config.KeysToColumns[TReservedKey(sortColumn)]; found {
+		sortColumn = col
+	} else {
+		sortColumn = fmt.Sprintf("%s['%s']", config.AttributesColumn, sortColumn)
```
mayberryzane, line 52: I think we need to make sure sortColumn is used as a parameter instead of interpolated as a string since it's user input e.g. `fmt.Sprintf("%s['%s']", config.AttributesColumn, sb.Var(sortColumn))`

backend/clickhouse/query.go
```diff
@@ -39,13 +39,30 @@ func readObjects[TObj interface{}, TReservedKey ~string](ctx context.Context, cl
...
+
+	forwardDirection := "DESC"
+	backwardDirection := "ASC"
+	if pagination.Direction == modelInputs.SortDirectionAsc || sortDirection == modelInputs.SortDirectionAsc {
+		forwardDirection = "ASC"
+	} else {
+		backwardDirection = "DESC"
+	}
+
+	orderForward := fmt.Sprintf("%s %s, UUID %s", sortColumn, forwardDirection, forwardDirection)
```
mayberryzane, line 63: same here and on the line below

mayberryzane (changes requested):

backend/clickhouse/query.go
ccschmitz: Right - makes sense!

cc @SpennyNDaJets since we were going back and forth on this in pairing.

backend/clickhouse/query.go
```diff
@@ -961,3 +944,36 @@ func repr(val reflect.Value) string {
...
+	sortDirection := modelInputs.SortDirectionDesc
+	if params.Sort != nil {
+		sortColumn = params.Sort.Column
+		sortDirection = params.Sort.Direction
+	}
+
+	if col, found := config.KeysToColumns[TReservedKey(sortColumn)]; found {
+		sortColumn = col
+	} else {
+		sortColumn = fmt.Sprintf("%s[%s]", config.AttributesColumn, sb.Var(sortColumn))
```
ccschmitz, line 964: @mayberryzane I think this is the only place we ended up needed to parameterize since we would know it's a mapped column if we hit the other side of this condition. Trying to wrap the string in `Var` again also seemed to break SQL building.

backend/clickhouse/query.go
ccschmitz: We ended up not needing to use `Var` here since we can be confident the original value passed in was already escaped or existed in our key to column mapping.

ccschmitz:
@mayberryzane ready for another look whenever you have time!

mayberryzane (approved):

PR #8500, opened 2024-05-08
support visualizing trace metrics

Vadman97:
## Summary

Shows trace metric event keys as graphable keys in the metrics UI.
This will allow creating a web vitals dashboard.

Screenshot from 2024-05-09 18-28-08

Screenshot from 2024-05-09 18-28-34

when we extract metrics into their own product in the future, we should 
refactor to remove some of this logic.

no


Vadman97:
@mayberryzane updated per your feedback!

frontend/src/pages/Graphing/components/Graph.tsx
```diff
@@ -166,6 +166,16 @@ const durationUnitMap: [number, string][] = [
 	[24, 'd'],
 ]
 
+const timeMetrics = {
```
mayberryzane, line 169: :+1: nice, was going to do something similar for session length / active length

mayberryzane (approved):

frontend/src/pages/Graphing/components/Graph.tsx
Vadman97: @mayberryzane any extra keys i should add here in this pr?

PR #8503, opened 2024-05-08
resolve stripe rate limiting

Vadman97:
## Summary

* upgrades stripe API to latest version, supporting cross-api-version webhooks to ensure we do not miss events that recently failed that are retrying on the old api version
* adds caching to stripe API call used by the frontend `Header` component that is causing us to be rate limited
* fixes private graph span sampling to make sure we don't record all private graph traces

manual deploy to prod
Screenshot 2024-05-08 at 18 48 29

no

no


SpennyNDaJets (approved):
Code makes sense to me

PR #8515, opened 2024-05-09
Fix expand link in related resource panel

ccschmitz:
## Summary

Fixes a bug where clicking the expand button in the related resource panel on a trace from the traces page wouldn't work. Also updates to use a `Link` instead of of `navigate` and an onClick handler which allows you to command + click or right click to open in a new tab.

* Open a trace from the traces page
* Click the expand button - nothing should happen :)
* View other related resources across the app and click the expand button to ensure they work as expected

N/A - client-side change only.

N/A - no change in UI.


frontend/src/components/RelatedResources/TracePanel.tsx
```diff
@@ -13,9 +16,10 @@ export const TracePanel: React.FC<{ resource: RelatedTrace }> = ({
 	resource,
 }) => {
 	const { projectId } = useNumericProjectId()
-	const path = `/${projectId}/traces/${resource.id}`
 	const { highlightedSpan, loading, selectedSpan, traces } = useTrace()
 	const span = selectedSpan || highlightedSpan
+	const relatedResourceParam = encodeURIComponent(JSON.stringify(resource))
+	const path = `/${projectId}/traces?${RELATED_RESOURCE_PARAM}=${relatedResourceParam}`
```
Vadman97, line 22: looks like something is up with the navigation

https://www.loom.com/share/8787ba9f66104de69f898496da96d495

Vadman97 (approved):
looks good pending the other product -> related trace issue

frontend/src/components/RelatedResources/TracePanel.tsx
ccschmitz: Thanks for catching that, @Vadman97! I fixed that issue but am discussing the UX here with @julian-highlight. We are thinking we may hide the expand button entirely on the related trace panel since 1) we don't have the ability to "view in context" like we do with logs, and 2) the UI doesn't really change when expanding a trace because it just links to the traces page with the same panel open.

ccschmitz: @Vadman97 we ended up removing the expand button on related traces. Keeping the updates to use a `Link` instead of `onClick` + `navigate`, though.

PR #8516, opened 2024-05-09
Remove duplicate sessions in sessions_joined_vw

SpennyNDaJets:
## Summary
The `sessions_joined_vw` has duplicates because it is not using the `FINAL` keyword when creating the view. Recreate the view with the `FINAL` keyword 

1) Load the sessions search
- [ ] Search works with filters
- [ ] 10 sessions per page when appropriate
- [ ] Results match filters

N/A

N/A

### DIFF
--- a/backend/clickhouse/migrations/000100_recreate_sessions_joined_vw.up.sql
+++ b/backend/clickhouse/migrations/000100_recreate_sessions_joined_vw.up.sql
@@ -1,0 +1,13 @@
+DROP VIEW IF EXISTS sessions_joined_vw;
+CREATE VIEW IF NOT EXISTS sessions_joined_vw AS
+select ProjectID as ProjectId,
+    CreatedAt as Timestamp,
+    mapFromArrays(
+        arrayMap(x->splitByChar('_', x, 2) [2], FieldKeys),
+        arrayMap(
+            (k, kv)->substring(kv, length(k) + 2),
+            arrayZip(FieldKeys, FieldKeyValues)
+        )
+    ) as SessionAttributes,
+    *
+from sessions FINAL SETTINGS splitby_max_substrings_includes_remaining_string = 1;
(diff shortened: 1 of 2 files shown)


backend/clickhouse/migrations/000100_recreate_sessions_joined_vw.up.sql
```diff
@@ -0,0 +1,13 @@
...
+    CreatedAt as Timestamp,
+    mapFromArrays(
+        arrayMap(x->splitByChar('_', x, 2) [2], FieldKeys),
+        arrayMap(
+            (k, kv)->substring(kv, length(k) + 2),
+            arrayZip(FieldKeys, FieldKeyValues)
+        )
+    ) as SessionAttributes,
+    *
+from sessions FINAL SETTINGS splitby_max_substrings_includes_remaining_string = 1;
```
Vadman97, line 13: curious what the performance implications of this `final` are vs. using a group by on the session id and picking the latest by `updatedat`

Vadman97 (approved):

backend/clickhouse/migrations/000100_recreate_sessions_joined_vw.up.sql
SpennyNDaJets: I may need @mayberryzane's expertise here - I do not the final is what we do for errors_joined_vw, so would want to keep this consitent

https://github.com/highlight/highlight/blob/main/backend/clickhouse/migrations/000099_errors_joined_vw_perf.up.sql#L5

SpennyNDaJets: Synced with Zane - we believe the performance will be similar since it will be doing something similar under the hood. We do think the FINAL implementation may be better since this view also powers the grafana and metrics queries

PR #8518, opened 2024-05-09
UX tweaks for query builder

ccschmitz:
## Summary

* Only shows errors after submitting a query.
* Allow empty spaces at the beginning of the query string. This required using a monospace font in order to get the leading space to expand the correct width. Discussed tradeoffs with @julian-highlight and decided to move forward with monospace font for the input + the typeahead results (screenshot below).
* Trigger submit when a user selects an `EXISTS` or `NOT EXISTS`.

Screenshot 2024-05-16 at 9 31 15 AM

* Enter a query that has an error and ensure it is not highlighted until you blur the input.
  * Ensure the state is reset if you fix the error and introduce a new one.
* Enter empty space at the beginning of the query string and ensure formatting looks correct.
* Select an `EXISTS` or `NOT EXISTS` value and ensure the form submits and behaves as expected.
* Enter a number of different queries and ensure the search form behaves as expected.

N/A - all client-side chahnges.

Yes, will need to review the font update w/ @julian-highlight. Might be something we don't want to do and just tolerate in order to use the font we want.


ccschmitz:
I talked through the font change on the query builder with @julian-highlight. He is going to look the preview over and update the mocks and massage things a bit after he gets a chance to play with the monospace font.

mayberryzane (approved):

PR #8527, opened 2024-05-10
Wrap operators in quotes

SpennyNDaJets:
## Summary
- Wrap the contains and match operators in `"` to support the usage of reserved chars within it.
- Align the error/X with the bottom row

https://www.loom.com/share/d0d0b4d4acb74a139a91bfcd4cf91fb6?sid=441c1a9d-74f0-4b0a-b5a4-0de1f27988f4

1) Load the sessions page
2) Use the contains operator
- [ ] The * are wrapped in "
3) Select a value that breaks to a new line
- [ ] X aligned on bottom row
- [ ] Any error aligned on bottom row

N/A

N/A

### DIFF
--- a/frontend/src/__generated/index.css
+++ b/frontend/src/__generated/index.css
@@ -5933,25 +5933,25 @@ body {
 ._10xh0c2c {
   background-color: #ffffff;
   border-radius: 999px;
+  bottom: 8px;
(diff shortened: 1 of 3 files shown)


Vadman97 (approved):
preview looks great

PR #8549, opened 2024-05-13
Allow users to sort traces (take 2)

ccschmitz:
## Summary

This PR reverts the revert of #8497 and introduces some fixes + improvements.

* We started using the `HighlightType` column rather than `TraceAttributes['highlight.type']` for default trace filters, which should provide a performance improvement.
* We only hit the sampling table for the inner query when selecting traces.
* Adds the sort param to the queries fetching additional results.

* Load the traces page and ensure results + stats in the top look correct
  * Inspect SQL statements to ensure the `traces` table is being hit
* Click on a column header to sort results
  * Ensure results are sorted correctly
  * Inspect SQL and ensure the `traces_sampling` table is hit for the inner query and `traces` is hit for the outer query
* Ensure all other querying behavior works as expected, including logs

Will monitor as this goes out again to ensure it doesn't cause any more issues.

Will have @julian-highlight click test in prod. We may do some follow up on the UX to trigger the dropdown menus on table headers.

### DIFF
--- a/backend/clickhouse/traces.go
+++ b/backend/clickhouse/traces.go
@@ -46,6 +46,7 @@ var traceKeysToColumns = map[modelInputs.ReservedTraceKey]string{
 	modelInputs.ReservedTraceKeyMetricValue:     "MetricValue",
 	modelInputs.ReservedTraceKeyEnvironment:     "Environment",
 	modelInputs.ReservedTraceKeyHasErrors:       "HasErrors",
+	modelInputs.ReservedTraceKeyTimestamp:       "Timestamp",
 }
 
 var traceColumns = []string{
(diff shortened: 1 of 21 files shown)


backend/clickhouse/traces.go
```diff
@@ -46,6 +46,8 @@ var traceKeysToColumns = map[modelInputs.ReservedTraceKey]string{
 	modelInputs.ReservedTraceKeyMetricValue:     "MetricValue",
 	modelInputs.ReservedTraceKeyEnvironment:     "Environment",
 	modelInputs.ReservedTraceKeyHasErrors:       "HasErrors",
+	modelInputs.ReservedTraceKeyTimestamp:       "Timestamp",
+	modelInputs.ReservedTraceKeyHighlightType:   "HighlightType",
```
Vadman97, line 50: so `highlight_type` will now be a searchable attribute, cool!

Vadman97 (approved):

SpennyNDaJets (commented):
Maybe not necessary now, but noticed a couple things:
- when I try to visit a link already sorted, it seems the sort gets removed (e.g. https://localhost:3000/1/traces?sort_column=duration&sort_direction=DESC)
- do we want to remove polling when the sort column is not timestamp? Maybe a little awkward, and not sure if the results will change on reload due to using the sampling table?

ccschmitz:
@SpennyNDaJets thanks for the click test!

> when I try to visit a link already sorted, it seems the sort gets removed (e.g. https://localhost:3000/1/traces?sort_column=duration&sort_direction=DESC)

Will look into this one!

> do we want to remove polling when the sort column is not timestamp? Maybe a little awkward, and not sure if the results will change on reload due to using the sampling table?

Yeah, I wasn't totally sure about that one... Curious what @Vadman97 thinks 🤔 - might leave it in for now, but something we should probably discuss.

PR #8562, opened 2024-05-14
Fix unquoting queries with non escape characters

SpennyNDaJets:
## Summary
The `strconv.Unquote` is intended for use in cases where you are processing go source code, so there are some rough edges when trying to unquote a string passed in from the frontend. Mainly, it cannot parse `\d` or \w` and other regex expressions since they are not valid go escape keys.

Update the method to trim the quotes and replace any internal quotes with the non escaped version - like we previously did for single quotes `'`

https://www.loom.com/share/e30f51d823d541e0a78d194139b95a85?sid=c5f92a46-e7df-4709-a30a-12b648bd0649

1) Load one of the search experiences
2) Search with the `matches` operator
3) Add a search expression such as a number (`/n`) or word (`/w`) to the search within the quotes `"`
- [ ] Same results as without quotes

N/A

N/A

### DIFF
--- a/backend/parser/listener/listener_test.go
+++ b/backend/parser/listener/listener_test.go
@@ -1,0 +1,37 @@
+package listener
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+var testCases = []struct {
+	input          string
+	expectedOutput string
+}{
+	{
+		"test",
+		"test",
+	},
+	{
+		`"test"`,
+		"test",
+	},
+	{
+		`"\w\w"`,
+		"\\w\\w",
+	},
+	{
+		`'\w\w'`,
+		"\\w\\w",
+	},
+}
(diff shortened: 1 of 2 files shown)


ccschmitz (approved):
Nice! Thanks for cleaning this up!

backend/parser/listener/listener_test.go
```diff
@@ -0,0 +1,37 @@
...
+	},
+}
+
+func TestUnquote(t *testing.T) {
+	for _, tc := range testCases {
+		// Call the function with the test case
+		output := Unquote(tc.input)
+		assert.Equal(t, tc.expectedOutput, output)
+	}
+}
```
Vadman97, line 37: would be a cool candidate for a fuzz test

Vadman97 (approved):

PR #8569, opened 2024-05-14
enterprise feature: project-level membership

mayberryzane:
## Summary
- adds an enterprise ability to set project-level access for users with member role
- changes invites to default to member (onboarding, invite through members page, workspace secret link)
- renames some of the resolver methods to be more clear about what kind of permissions are being checked
https://www.loom.com/share/8b4de7ae89b44e13ad0c88055f2bec13

- clicktested locally
- unit test for various access level resolver methods and `Projects` resolver

- no

- @julian-highlight


ccschmitz (approved):
Code + click test are looking good!

One thing I noticed is that non-admins still have access to some of the admin UI even though the page doesn't really work for them. For example, if I'm a project-level member, I can still access `/w/${workspaceId}` even though I see an error.

Screenshot 2024-05-21 at 2 37 16 PM

Think it would be worthwhile to only define those routes for admins inside `AppRouter`?

I also think we might be linking to some admin-only pages for members too. e.g. `LogsOverageCard`, the session overage CTA, the [default settings page redirect](https://github.com/highlight/highlight/blob/12530cf2087cba77eb876ce2414cc4c3247c7b65/frontend/src/pages/SettingsRouter/SettingsRouter.tsx#L167-L169), and possibly a few more (I tried to audit the places I saw `/w/${` in the app.

PR #8570, opened 2024-05-14
[HIG-4638] revive loading overlay, show old data while loading

mayberryzane:
## Summary
- brings back the loading overlay, styles it similarly to the "No data found" badge
- continues to show old data while new data is loading

- clicktested locally

- no

- @julian-highlight

### DIFF
--- a/frontend/src/pages/Graphing/components/Graph.tsx
+++ b/frontend/src/pages/Graphing/components/Graph.tsx
@@ -8,6 +8,7 @@ import {
 	IconSolidChartSquareLine,
 	IconSolidDocumentReport,
 	IconSolidDotsHorizontal,
+	IconSolidLoading,
 	IconSolidPencil,
 	IconSolidTable,
 	IconSolidTrash,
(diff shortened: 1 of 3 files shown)


frontend/src/pages/Graphing/components/Graph.tsx
```diff
@@ -838,7 +833,31 @@ const Graph = ({
...
+								label="Loading"
+								iconStart={
+									<IconSolidLoading
+										className={loadingIcon}
+										color={vars.theme.static.content.weak}
+									/>
+								}
+							/>
+						</Stack>
+					)}
```
Vadman97, line 860: noticed that this `loading` state is an overlay on top of the chart even if the chart is loaded

should we do something like blur the chart if it already has data?
Screenshot from 2024-05-14 16-31-56

Vadman97 (approved):

frontend/src/pages/Graphing/components/Graph.tsx
mayberryzane: yeah I'm not 100% sure - I tried w/ blur and other background effects but I thought they were kind of distracting. cc @julian-highlight for input here

julian-highlight: We could just reduce the opacity of everything apart from the title down to 10%. This way we're sure nothing is blocking sight

image

PR #8596, opened 2024-05-16
Update CTA for traces empty state

ccschmitz:
## Summary

Updates the empty state CTA for traces so they are bit more specific and helpful. Here is what we currently show in every instance.

Screenshot 2024-05-16 at 4 02 57 PM

This PR adds a new CTA if we know they have integrated. The content varies depending on whether they have a query string entered.

Screenshot 2024-05-16 at 1 01 16 PM

Screenshot 2024-05-16 at 1 01 31 PM

This PR also makes some updates to the `Callout` component so you don't have to pass so much boilerplate to use a custom icon. It also leverages the component for the empty states instead of creating our own.

* Go to the traces page and enter a random query that you know won't produce any results
* Ensure the CTA is correct
* Remove the query and select a very specific time range that you know won't produce any results
* Ensure the CTA is correct
* Test on a project that has not integrated tracing
* Ensure the CTA is correct
* Click around the app to other places you know there `Callout` is used and make sure they look alright

N/A - click-side only change.

N/A


SpennyNDaJets (approved):

PR #8627, opened 2024-05-20
Fix unit length search

SpennyNDaJets:
## Summary
Some time related keys have different units. Fix the search to convert to the correct unit:
- Duration -> ns
- Length -> ms
- ActiveLength -> ms
- Default -> ns

Open question: what is a session's duration and what is its units?

https://www.loom.com/share/f66d69ae0df544f5bb70dcf1a0b8c30a

1) Search for sessions by length with units
- [ ] Correct sessions returned
2) Search for sessions by active_length with units
- [ ] Correct sessions returned
3) Search for traces by duration with units
- [ ] Correct traces returned

N/A

N/A


ccschmitz (approved):
Code + click test LGTM!

PR #8628, opened 2024-05-20
stop polling for more resources after 15 mins

mayberryzane:
## Summary
- stop polling after 15 mins
- set a `pollingExpired` variable, show "load more results" after this time has passed

- clicktested locally w/ shorter polling expiry time

- no

- no

### DIFF
--- a/frontend/src/pages/LogsPage/LogsTable/LogsTable.tsx
+++ b/frontend/src/pages/LogsPage/LogsTable/LogsTable.tsx
@@ -336,7 +336,7 @@ const LogsTableInner = ({
 						<Box width="full">
 							<AdditionalFeedResults
 								maxResults={MAX_LOGS}
-								more={moreLogs}
+								more={moreLogs ?? 0}
 								type="logs"
 								onClick={() => {
 									clearMoreLogs()
--- a/frontend/src/pages/LogsPage/LogsTable/LogsTable.tsx
+++ b/frontend/src/pages/LogsPage/LogsTable/LogsTable.tsx
@@ -128,6 +128,7 @@ type LogsTableInnerProps = {
 	handleAdditionalLogsDateChange?: () => void
 	selectedColumns?: LogCustomColumn[]
 	setSelectedColumns?: (columns: LogCustomColumn[]) => void
+	pollingExpired?: boolean
 }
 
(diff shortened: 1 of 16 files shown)


Vadman97 (approved):

frontend/src/pages/LogsPage/LogsTable/LogsTable.tsx
```diff
@@ -148,7 +150,9 @@ const LogsTableInner = ({
 }: LogsTableInnerProps) => {
 	const bodyRef = useRef<HTMLDivElement>(null)
 	const enableFetchMoreLogs =
-		!!moreLogs && !!clearMoreLogs && !!handleAdditionalLogsDateChange
+		(!!moreLogs || pollingExpired) &&
```
SpennyNDaJets, line 153: Just to make sure I'm understanding this, we will show the button to refresh if polling has expired?

SpennyNDaJets: See we are handing this case in `frontend/src/components/FeedResults/FeedResults.tsx`

frontend/src/util/search.ts
```diff
@@ -1,11 +1,24 @@
...
 // usePollQuery polls a lazy query function and returns a 'more' count
-export function usePollQuery<T, U>({
+export function usePollQuery<
+	T,
+	U extends
+		| {
+				params: QueryInput
+		  }
+		| {
+				query: ClickhouseQuery
```
SpennyNDaJets, line 19: This shouldn't be needed with the new searches (i.e. only `QueryInput` is used), but if its giving a type error maybe we should make a card to clean it up?

SpennyNDaJets (approved):

frontend/src/util/search.ts
mayberryzane: ahh this is old code before your new changes were merged in, let me see if I can change this

frontend/src/pages/LogsPage/LogsTable/LogsTable.tsx
mayberryzane: yes, and instead of showing "50+ more results" or nothing in case no more had been loaded, it will show "load more results"

PR #8633, opened 2024-05-21
[SUP-27] cors config updates

mayberryzane:
## Summary
- Highlight doesn't use credentials for public graph requests, but we should not allow credentials in case this changes in future
- move cors configs within the public and private routers only - I don't think it's necessary to use cors on other routes (`/health`, `/otel`) and we should probably have separate public/private configs for hobby deploys too rather than using a single one for the `All` runtime - @Vadman97 can you confirm this last bullet point?

- clicktested locally to confirm functionality was working properly, tested cross-origin requests

- no

- no

### DIFF
--- a/backend/main.go
+++ b/backend/main.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"flag"
 	"fmt"
-	"github.com/highlight-run/highlight/backend/assets"
 	"html/template"
 	"io"
 	"math/rand"
@@ -14,6 +13,8 @@ import (
 	"strings"
 	"sync"
 	"time"
+
+	"github.com/highlight-run/highlight/backend/assets"
 
 	"github.com/aws/aws-sdk-go-v2/config"
 	"github.com/aws/aws-sdk-go-v2/service/marketplacemetering"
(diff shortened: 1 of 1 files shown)


ccschmitz (approved):
I know you had an open question out for @Vadman97, but this looks good to me!

backend/main.go
```diff
@@ -183,20 +184,29 @@ func enhancedHealthCheck(ctx context.Context, db *gorm.DB, rClient *redis.Client
...
-		// From the highlight frontend, only the url is whitelisted.
-		isRenderPreviewEnv := strings.HasPrefix(origin, "https://frontend-pr-") && strings.HasSuffix(origin, ".onrender.com")
-		// Is this an AWS Amplify environment?
-		isAWSEnv := strings.HasPrefix(origin, "https://pr-") && strings.HasSuffix(origin, ".d25bj3loqvp3nx.amplifyapp.com")
-		isReflamePreview := origin == "https://preview.highlight.io"
-
-		if origin == frontendURL || origin == "https://app.highlight.run" || origin == "https://app.highlight.io" || origin == landingStagingURL || isRenderPreviewEnv || isAWSEnv || isReflamePreview {
-			return true
-		}
-	} else if runtimeParsed == util.PublicGraph || runtimeParsed == util.All {
```
Vadman97, line 197: > we should probably have separate public/private configs for hobby deploys too rather than using a single one for the All runtime?

yeah, we can detect the hobby deploy via the `util.IsInDocker()` check, split out the configs and then use the `REACT_APP_PRIVATE_GRAPH_URI` and `REACT_APP_PUBLIC_GRAPH_URI` values for the respective origin checks. with the normal hobby deploy tho, the private and public graphs run together, but assuming the runtime flag is used we should use the corresponding origin

Vadman97 (approved):

PR #8640, opened 2024-05-21
Require re-authentication after removing 2FA

ccschmitz:
## Summary

As a security measure, we should be requiring users to sign in again after removing 2FA from their account.

* Enroll in 2FA
* Remove 2FA
* Ensure you are signed out and that the notification message makes sense

N/A - minor client-side change.

N/A - text-only change in the UI.


mayberryzane:
@ccschmitz for this I think we need users to enter their MFA code in order to remove MFA

PR #8643, opened 2024-05-21
Replace antd toasts

SpennyNDaJets:
## Summary
Swap out the antd toast element with a Highlight toast element

https://www.loom.com/share/1fc13dede6ff45909ec06ac4c2c46be2?sid=47010324-a73e-4222-a612-1cadaa3f936c

1) View the alerts page
2) Toggle an alert on and off
- [ ] Toast is displayed appropriately
3) View a session on the sessions page
4) Try to download a session
- [ ] First alert displayed appropriately
- [ ] Second alert displayed appropriately (if not enterprise)

N/A

@julian-highlight


experiments/chrome-extension/src/App.tsx
```diff
@@ -4,8 +4,9 @@ import 'antd/dist/antd.css' // or 'antd/dist/antd.less'
 import { FaCopy } from 'react-icons/fa'
 
 import { useChromeStorageLocal } from 'use-chrome-storage'
-import { InputNumber, Checkbox, message } from 'antd'
+import { InputNumber, Checkbox } from 'antd'
 import { useClipboard } from 'use-clipboard-copy'
+import { toast } from '@components/Toaster'
 
 const Core = () => {
```
SpennyNDaJets, line 11: Not sure what this is, but looks to be broken anyways?

frontend/src/pages/Buttons/Buttons.tsx
```diff
@@ -261,7 +261,7 @@ export const Buttons = () => {
 					<button
 						className={commonStyles.submitButton}
 						onClick={() => {
-							toast('Event has been created')
+							toast.success('Event has been created')
```
SpennyNDaJets, line 264: Can we swap this over? @Vadman97

frontend/src/components/Toaster/index.tsx
```diff
@@ -0,0 +1,166 @@
...
+			<Stack direction="row" justifyContent="space-between" width="full">
+				<Stack direction="column" gap="8">
+					<Box display="flex" alignItems="flex-start" pt="4">
+						<Text weight="bold">{toast.message}</Text>
+					</Box>
+
+					{!!toast.content && (
+						<Box display="flex">{toast.content}</Box>
+					)}
+				</Stack>
```
ccschmitz, line 153: Quick thought on the spacing in here... Wonder if we want to wrap `toast.content` in `Text` as well so we can control the spacing more precisely. Also wondering if the button will be too close to the text if there is no `content` right now.

Finally, in a text-only toast that is multiline it feels like the spacing on the bottom is off since it's less that the top.

image

Looking at the mocks, it appears we have the same padding on the top and the bottom.

Screenshot 2024-05-22 at 2 55 07 PM

Should that `pt="4"` actually be `py="4"`?

ccschmitz (approved):
Some minor comments on styling of the toast, but looks great overall! Love how many `antd` imports are removed here 🎉

frontend/src/components/Toaster/index.tsx
SpennyNDaJets: Also, the button in this case is added through the `content`. Not sure if that changes anything

SpennyNDaJets: Screenshot 2024-05-22 at 5 06 19 PM

PR #8645, opened 2024-05-21
fix dashboard crash for tooltip for invalid point

mayberryzane:
## Summary
- Previously, tooltips returned null if there was no data at a point. This was changed to fix a flickering issue, but certain inputs started crashing. Don't call getTickFormatter for these points as the `value` to format may be undefined.
- also fix `isEmpty` check so loading state doesn't have grey lines displayed

- clicktested locally with crash scenario

- no

- no


SpennyNDaJets (approved):

Vadman97:
thanks @mayberryzane !

PR #8648, opened 2024-05-21
rewrite log_attributes mvs to write directly to log_key_values table

mayberryzane:
## Summary
- saw `log_attributes` merges accounting for ~10% of total merge time, we can write directly to `log_key_values` instead like we've done for `trace_key_values` previously

- ran migrations locally, validated that new entries were added to `log_key_values`

- no, will follow up by deleting `log_attributes` to reclaim some storage

- no


ccschmitz (approved):

Vadman97:
@mayberryzane is this still relevant?

PR #8661, opened 2024-05-23
avoid rendering css animation when loading page is hidden

Vadman97:
## Summary

As reported by a customer over discord,
Loading page would use `motion` to update CSS of the elements even when the page is not rendered.

reflame

### DIFF
--- a/frontend/src/components/Loading/Loading.tsx
+++ b/frontend/src/components/Loading/Loading.tsx
@@ -56,14 +56,17 @@ export const LoadingPage = React.memo<{ show?: boolean; className?: string }>(
 	({ show, className }) => {
 		const { loadingState } = useAppLoadingContext()
 		const speedFactor = 0.1
+		const shouldShow =
+			show ||
+			[
+				AppLoadingState.LOADING,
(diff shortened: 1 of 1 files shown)


frontend/src/components/Loading/Loading.tsx
```diff
@@ -56,14 +56,17 @@ export const LoadingPage = React.memo<{ show?: boolean; className?: string }>(
...
+		if (!shouldShow) return null
 
 		return (
 			<AnimatePresence>
-				{(show ||
-					[
-						AppLoadingState.LOADING,
-						AppLoadingState.EXTENDED_LOADING,
-					].includes(loadingState)) && (
+				{shouldShow ? (
```
SpennyNDaJets, line 69: Is this needed if we are returning early above?

SpennyNDaJets (approved):
Not sure I know how to repo, but just the one comment

frontend/src/components/Loading/Loading.tsx
Vadman97: thanks, good catch!

PR #8666, opened 2024-05-24
[HIG-4672] update issue linking modals w/ new styles

mayberryzane:
## Summary
- use new components in issue linking modals
- set type="button" for buttons - by default, these act as type="submit"
- reuses my `OptionDropdown` component from the graphing editor - should replace all of these with the new select component when that's ready
- closes HIG-4672

- clicktested locally

- no

- @julian-highlight


packages/ui/src/components/Button/Button.tsx
```diff
@@ -56,6 +56,7 @@ export const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
 					cssClass,
 				)}
 				ref={ref}
+				type="button"
```
ccschmitz, line 59: Could we make this accept a prop value for `type` and fallback to `"button"` as the default? In some cases we may want to use `type="submit"`.

ccschmitz (approved):
Looks great, but one quick thing we may want to update on the main `Button` component before merging.

packages/ui/src/components/Button/Button.tsx
mayberryzane: I think `buttonProps` includes a `type` prop which should be overriding this already

PR #8670, opened 2024-05-24
Custom controls for number input

ccschmitz:
## Summary

Adds support for numbers on the `Form.Input` component w/ some custom UI for stepping the number up/down.

Screenshot 2024-05-29 at 7 32 50 AM

Updates the places we were importing from `antd` and removes the old `InputNumber` component we were using as well (see screenshot above from rage clicks form). Anywhere we were already using `Form.Input[type="number"]` (e.g. metrics) is updated to the new styles as well.

Screenshot 2024-05-29 at 7 33 26 AM

* Added a story in our UI package
* Updated the project settings page to use the number input and confirmed behavior was working as expected in my local click tests
  * Also confirmed the change looked good and wasn't a step back in terms of UI polish

Will have @julian-highlight click test the Reflame preview + provide feedback before merging.


Vadman97 (approved):
look great! tested on the alerts page where we use the input

PR #8671, opened 2024-05-24
New Select component in UI package

ccschmitz:
## Summary

Creates a new `Select` component in the UI package and sets it up in a few places in the app. There is more work to do to swap out other instances of antd selects which will be handled in future PRs.

Check out the Storybook stories in Reflame or locally and play around with it to see how it works. I think the stories show off most of the features, but here's a list of what's currently supported as well:

* Basic single selects
* Multi select
* Filterable selects
* Controlled selects (value managed externally)
* "tags" display mode
* Custom value rendering (used under the hood for tags)
* A `loading` prop if data is being fetched for options when initially rendered
* A `clearable` prop for selects that you can set to an empty value

I wrote a number of Storybook stories that I clicked tested thoroughly. I also used the new in our frontend a few places to ensure it worked as expected.


packages/ui/src/components/Button/Button.stories.tsx
```diff
@@ -54,13 +42,7 @@ export const ButtonVariants = () => {
...
 								key={`b-${idx}-${jdx}-d`}
 								size={size[size.length - 1]}
-								iconLeft={
-									jdx % emphasis.length !== 0 ? (
-										<IconSolidCheveronDown />
-									) : (
-										<></>
-									)
-								}
+								iconLeft={<IconSolidCheveronDown />}
```
ccschmitz, line 45: Fixed these stories which weren't rendering icons correctly

packages/ui/package.json
```diff
@@ -102,15 +103,15 @@
...
+		"eslint-plugin-storybook": "^0.8.0",
 		"eslint-plugin-tailwindcss": "^3.6.1",
 		"eslint-plugin-unused-imports": "^1.1.2",
 		"fs-extra": "^11.1.0",
 		"globby": "^13.1.3",
 		"npm-run-all": "^4.1.5",
 		"prop-types": "^15.8.1",
 		"react-dom": "^18.2.0",
-		"storybook": "^8.0.4",
+		"storybook": "^8.2.6",
```
ccschmitz, line 114: Some of the Storybook config changes came from running the auto upgrade task from the Storybook CLI.

frontend/src/pages/Alerts/LogAlert/LogAlertPage.tsx
SpennyNDaJets, line 1: Very nitty: but when you change the Alert frequency, the width of the select changes here

frontend/src/pages/WorkspaceTeam/components/AutoJoinForm.tsx
SpennyNDaJets, line 1: Seeing an error that may be related when trying to access:
- https://preview.highlight.io/w/1/settings
- https://preview.highlight.io/w/1/team

```
TypeError: Cannot read properties of undefined (reading 'length')
```

SpennyNDaJets (changes requested):
Looking great! Just left a couple comments from click test:

1. On `ErrorAlertPage`, not able to select an option
Screenshot 2024-08-07 at 4 10 37 PM

2. Also saw it was used on the SessionAlertForm, but wondering if `configuration.canControlThreshold ` should just be cleaned up?

frontend/src/pages/WorkspaceTeam/components/AutoJoinForm.tsx
ccschmitz: Thanks for catching this! Just fixed that up. It was an issue with auto join form.

frontend/src/pages/Alerts/LogAlert/LogAlertPage.tsx
ccschmitz: Fixed by setting `span="6"` on each wrapping `Column` - good catch!

ccschmitz:
> ... wondering if `configuration.canControlThreshold ` should just be cleaned up?

Maybe so, but I might hold off on that in this PR. Will defer to you on that one with the work you're already doing on alerts 😸 

I think everything else is cleaned up, so this is ready for another look! Thanks for the review!

packages/ui/src/components/Form/styles.css.ts
```diff
@@ -83,6 +83,7 @@ export const inputVariants = recipe({
 		},
 	},
 	defaultVariants: {
+		size: 'small',
```
ccschmitz, line 86: Noticed the height of form inputs was slightly off from the mocks. All form inputs were supposed to be 28px tall, which this enforces.

SpennyNDaJets (approved):
Click test and code look good to me!

ccschmitz:
Thanks for the review on this, @SpennyNDaJets! Apologies that this got so large...

PR #8679, opened 2024-05-28
Fix noise in Ruby SDK

ccschmitz:
## Summary

Users reported seeing the following error from our Ruby SDK:

```
OpenTelemetry error: invalid event attribute value type Hash for key 'log.message' on span 'highlight-ctx'
```

I was able to reproduce this error by passing a hash for `message` when recording a log. I added a test case that would fail because of this error being called and then updated our logic to call `to_s` on the `message` so we can be sure it's always a string.

Added some logging in our e2e blog example app + some tests to ensure things are working as expected.

Also, the first step I took in fixing this was [adding a failing test](https://github.com/highlight/highlight/actions/runs/9275741519/job/25521158933) which I confirmed my later changes got to pass.

Bumped the SDK version and will let users know to upgrade who have reported being impacted by this.

N/A


sdk/highlight-ruby/highlight/lib/highlight.rb
```diff
@@ -86,13 +87,14 @@ def record_log(session_id, request_id, level, message, attrs = {})
...
           LOG_SEVERITY_ATTRIBUTE => H.log_level_string(level),
-          LOG_MESSAGE_ATTRIBUTE => message,
+          LOG_MESSAGE_ATTRIBUTE => message.to_s,
           CODE_FILEPATH => caller_info[0],
           CODE_LINENO => caller_info[1],
           CODE_FUNCTION => function
         }.merge(attrs))
       end
     end
+    # rubocop:enable Metrics/AbcSize
```
ccschmitz, line 97: Unrelated, but I was getting [a CI failure](https://github.com/highlight/highlight/actions/runs/9275823170/job/25521424360) from Rubocop violations here that I don't think were introduced by my changes.

Vadman97 (approved):
awesome, thanks chris!

PR #8696, opened 2024-05-30
Console logs table

SpennyNDaJets:
## Summary
Convert the console logs table to use the same styles as the logs table.

Screenshot 2024-06-03 at 12 45 30 PM

## How did you test this change?
1) View a session and open the console logs in the dev tools
2) Play a session
- [ ] Table tracks the log
- [ ] Can jump to a log
- [ ] Can filter the logs appropriately

https://www.loom.com/share/85e57e9e293740ed8295dc990a18dc9f?sid=c2387319-0302-45cd-9e4d-42397a7dbfe2

## Are there any deployment considerations?
N/A

## Does this work require review from our design team?
@julian-highlight

PR #8697, opened 2024-05-30
Console logs table backend

SpennyNDaJets:
## Summary
Update the backend to fetch have all the log columns available for graph - this is needed to replace the console log table with the logs table in https://github.com/highlight/highlight/pull/8696

Want to get this in separately to test frontend performance 

1. Load a session
- [ ] Able to fetch console logs appropriately in the dev tools

N/A

N/A

### DIFF
--- a/backend/clickhouse/logs.go
+++ b/backend/clickhouse/logs.go
@@ -193,11 +193,9 @@ func (client *Client) ReadLogs(ctx context.Context, projectID int, params modelI
 
 // This is a lighter weight version of the previous function for loading the minimal about of data for a session
 func (client *Client) ReadSessionLogs(ctx context.Context, projectID int, params modelInputs.QueryInput) ([]*modelInputs.LogEdge, error) {
-	selectCols := []string{"Timestamp", "UUID", "SeverityText", "Body"}
-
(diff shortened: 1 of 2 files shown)


backend/clickhouse/logs.go
```diff
@@ -223,10 +221,19 @@ func (client *Client) ReadSessionLogs(ctx context.Context, projectID int, params
...
 		var result struct {
-			Timestamp    time.Time
-			UUID         string
-			SeverityText string
-			Body         string
+			Timestamp       time.Time
+			UUID            string
+			SeverityText    string
+			Body            string
+			LogAttributes   map[string]string
```
Vadman97, line 228: i would expect fetching `LogAttributes` rather than just `Body` / a few of the main columns to be quite a bit slower, so i'm curious how this changes things

Vadman97 (approved):

backend/clickhouse/logs.go
SpennyNDaJets: Do we want the user to expand to see all the attributes? If not I think these changes can be ignored (will sync with @julian-highlight)

Screenshot 2024-05-31 at 11 42 35 AM

PR #8698, opened 2024-05-30
support a custom callback endpoint for firebase

Vadman97:
## Summary

adds a new `/auth/action` endpoint to the frontend which performs firebase auth based on their callbacks.
this will allow us to update our firebase config to use app.highlight.io with our styling for auth requests
such as resetting password, recovering email, and verifying email, rather than using the firebase hosted app.

Closes SUP-29

manual local testing with firebase token from prod
https://www.loom.com/share/739d8020915c45b3a9d90a7aa5e0c26f

update https://console.firebase.google.com/u/1/project/highlight-f5c5b/authentication/emails?pli=1 to app.highlight.io once released

no

### DIFF
--- a/.reflame.config.jsonc
+++ b/.reflame.config.jsonc
@@ -396,6 +396,7 @@
     },
     "firebase": {
       "entryPoints": [
+        "/auth",
         "/compat/auth",
         "/compat/app"
       ],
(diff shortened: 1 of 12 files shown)


SpennyNDaJets (approved):
Code and test video look good to me!

PR #8700, opened 2024-05-30
feat: added option to disable frontend analytics

### LINKED ISSUE #8680: self hosted deploy sends analytics data to hubspot / rudderstack

ezhil56x:
## Summary

Closes #8680
/claim #8680


Vadman97 (changes requested):
rather than changing every place that `analytics.track` is called, can we update the `analytics.ts` implementation for `track` and the other methods to do nothing if analytics are disabled?

ezhil56x:
@Vadman97 Updated my PR

frontend/src/constants.ts
```diff
@@ -7,3 +7,4 @@ export const FRONTEND_URI =
 	import.meta.env.REACT_APP_FRONTEND_URI ||
 	window.location.protocol + '//' + window.location.host
 export const AUTH_MODE = import.meta.env.REACT_APP_AUTH_MODE
+export const DISABLE_ANALYTICS = import.meta.env.REACT_APP_DISABLE_ANALYTICS
```
Vadman97, line 10: thanks for the changes, now - you'll need to make sure this env var is passed thru from the hobby deployment. could you update the `docker/compose.hobby.yml` and `docker/.env` / `docker/env.sh` accordingly?

Vadman97 (changes requested):

ezhil56x:
@Vadman97 Updated my PR

ezhil56x:
@Vadman97 Anymore changes to be done?

docker/.env
```diff
@@ -22,6 +22,7 @@ PSQL_DOCKER_HOST=postgres
 PSQL_PASSWORD=
 PSQL_PORT=5432
 PSQL_USER=postgres
+REACT_APP_DISABLE_ANALYTICS=true
```
Vadman97, line 25: we should default this to `false`
```suggestion
REACT_APP_DISABLE_ANALYTICS=false
```

docker/env.sh
```diff
@@ -33,6 +33,7 @@ else
 fi
 export BUILD_ARGS="--build-arg GOARCH=${GOARCH}
 --build-arg REACT_APP_AUTH_MODE=${REACT_APP_AUTH_MODE}
+--build-arg REACT_APP_DISABLE_ANALYTICS=${REACT_APP_DISABLE_ANALYTICS}
```
Vadman97, line 36: since this is a frontend var that would need to work in the built frontend, we'll also need to update `docker/frontend-entrypoint.py` to update the value.

Vadman97 (commented):
looks like just one more change necessary. could you also please verify the changes locally and record a loom.com demo testing the new environment variable?

docker/env.sh
ezhil56x: In `frontend-entrypoint.py`, we are replacing the values. For URLs that may be good as they are hard coded. I don't think we need to add `REACT_APP_DISABLE_ANALYTICS` to `frontend-entrypoint.py`

ezhil56x:
@Vadman97 I think it works as expected now.

ezhil56x:
@Vadman97 Anymore changes to be done?

Vadman97:
apologies for the delay @ezhil56x 
i have another PR in flight that will help simplify the places where the build args have to be updated
https://github.com/highlight/highlight/pull/8715

will merge this shortly after that

ezhil56x:
@Vadman97 Resolved conflicts

Vadman97 (approved):
looks good, thanks!

ezhil56x:
@Vadman97 Can I add the env variables to turbo.json. I think that might fix the build failure

Vadman97:
@ezhil56x yes pls, could you give that a try?

ezhil56x:
@Vadman97 Can you approve the ci? I couldn't check the build failure of past runs

PR #8706, opened 2024-05-30
Set up OTEL browser auto instrumentation

ccschmitz:


frontend/src/pages/Traces/TraceProvider.tsx
```diff
@@ -132,6 +132,19 @@ export const TraceProvider: React.FC<React.PropsWithChildren<Props>> = ({
...
+			const selectedSpan = data?.trace?.trace.find(
+				(span) => span.spanID === spanId,
+			)
+
+			if (selectedSpan) {
+				setSelectedSpan(selectedSpan as FlameGraphSpan)
+			}
+		}
+		// eslint-disable-next-line react-hooks/exhaustive-deps
+	}, [spanId])
```
ccschmitz, line 146: Unrelated, but noticed that the selected span wasn't updating when clicking different spans on the traces page.

deploy/otel-collector.yaml
```diff
@@ -20,9 +20,21 @@ receivers:
                 include_metadata: true
                 cors:
                     allowed_origins:
-                        - '*'
+                        # Can't use '*' with credentials because it fails CORS
+                        # validation on the preflight request.
```
ccschmitz, line 24: @Vadman97 maybe not relevant here, but thought it made sense to keep the configs identical.

sdk/client/src/otel/index.ts
```diff
@@ -0,0 +1,319 @@
...
+	}
+
+	const recordedBody = getBodyThatShouldBeRecorded(
+		parsedBody,
+		networkRecordingOptions?.networkBodyKeysToRedact,
+		networkRecordingOptions?.bodyKeysToRecord,
+		headers as Headers,
+	)
+
+	// TODO: We don't seem to be able to access this sometimes. Figure out why.
```
ccschmitz, line 305: Need to look into this, but figured I'd handle in a follow up PR. It might require creating our own instrumentation and waiting until the requests are actually finished to submit spans so that we can get the full response data.

sdk/highlight-go/otel.go
```diff
@@ -167,6 +168,12 @@ func StartOTLP() (*OTLP, error) {
 		return nil, err
 	}
 
+	propagator := propagation.NewCompositeTextMapPropagator(
+		propagation.TraceContext{},
+		propagation.Baggage{},
+	)
+	otel.SetTextMapPropagator(propagator)
```
ccschmitz, line 175: This was required for span propagation to work from the client to the server.

deploy/otel-collector.yaml
```diff
@@ -20,9 +20,21 @@
...
+                        - 'https://*'
+                        - 'http://*'
```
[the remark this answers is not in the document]
ccschmitz: @Vadman97 think we could restrict to only `https`?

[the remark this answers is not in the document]
Vadman97: don't think it's a concern for the cors config

sdk/highlight-go/otel.go
Vadman97: could we check that our other SDKs configure this too? for customers on node.js backend, ruby, etc

e2e/dotnet/HighlightConfig.cs
Vadman97, line 1: does this need to set up a propagator like you did in highlight.go?

sdk/client/src/otel/index.ts
```diff
@@ -0,0 +1,319 @@
...
+let provider: WebTracerProvider
+
+export const initializeOtel = (config: OtelConfig) => {
+	if (provider !== undefined) {
+		console.warn('OTEL already initialized. Skipping...')
+		return
+	}
+
+	// TODO: Is there a better way to determine if we're in dev mode? Seems
+	// import.meta.env.DEV is set to dev when building with dev:frontend.
```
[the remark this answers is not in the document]
Vadman97: do we need this check? why don't we just keep the fallback always as `https://otel.highlight.io:4318`, requiring something to be provided via `config.endpoint` otherwise?

sdk/client/src/otel/index.ts
```diff
@@ -0,0 +1,319 @@
...
+			],
+		}),
+	})
+
+	// Export spans to console for debugging
+	if (isDev) {
+		provider.addSpanProcessor(
+			new SimpleSpanProcessor(new ConsoleSpanExporter()),
+		)
+	}
```
Vadman97, line 88: maybe rather than using `isDev`, we could have a `config.debug` flag, like we do in the main client sdk?

sdk/client/src/otel/index.ts
```diff
@@ -0,0 +1,319 @@
...
+		compression: 'gzip' as any,
+	})
+
+	const spanProcessor = new CustomSpanProcessor(exporter, {
+		tracingOrigins: config.tracingOrigins,
+	})
+	provider.addSpanProcessor(spanProcessor)
+
+	// Can customize by passing a config object as 2nd param.
+	provider.addSpanProcessor(new BatchSpanProcessor(exporter))
```
Vadman97, line 104: any tweaks needed to the config, like we do on the backend sdks to batch larger?

sdk/highlight-go/highlight.go
```diff
@@ -247,12 +249,15 @@ func InterceptRequest(r *http.Request) context.Context {
...
 	highlightReqDetails := r.Header.Get("X-Highlight-Request")
+
 	ids := strings.Split(highlightReqDetails, "/")
-	if len(ids) < 2 {
-		return ctx
+	if len(ids) >= 2 {
+		ctx = context.WithValue(ctx, ContextKeys.SessionSecureID, ids[0])
+		ctx = context.WithValue(ctx, ContextKeys.RequestID, ids[1])
+	} else {
+		ctx = otel.GetTextMapPropagator().Extract(ctx, propagation.HeaderCarrier(r.Header))
```
Vadman97, line 258: does this need to happen in any other backend sdks?

sdk/firstload/src/index.tsx
```diff
@@ -112,6 +112,26 @@ const H: HighlightPublicInterface = {
 			}
 			init_called = true
 
+			if (options?.enableOtelTracing) {
+				// Use dynamic import to avoid bundling the opentelemetry dependencies
+				import('@highlight-run/client/src/otel/index').then(
```
Vadman97, line 117: this is importing the src file, does it get bundled? should we add the otel bundle to the client vite bundling? otherwise, we can probably avoid building client altogether by updating line 137 below to use the src file as well

sdk/client/src/otel/index.ts
Vadman97: interesting... could our fetch instrumentation ever collide with otel's?

Vadman97 (approved):
looks great generally, awesome work. lmk if you want to walk thru any of my comments over a call / our pairing

sdk/highlight-go/otel.go
ccschmitz: Yeah, we should set this up so that we are sending/receiving the necessary data for propagation in all our SDKs. I can file a ticket to follow up for each of them.

e2e/dotnet/HighlightConfig.cs
ccschmitz: I don't believe so. The only reason we needed to set up the propagator in the Go SDK right was so that we could use the propagation data that we configure in the client SDK, which should only be sending to the Go backend.

sdk/highlight-go/highlight.go
ccschmitz: Don't believe so. Same reasoning as https://github.com/highlight/highlight/pull/8706#discussion_r1646675080, but let me know if you think I'm off there.

sdk/firstload/src/index.tsx
ccschmitz: Hmm... Not sure I know what you mean here. Maybe something we can follow up on in pairing tomorrow?

sdk/client/src/otel/index.ts
ccschmitz: I was going to stick with the defaults for now, but wanted to make it known we could customize things like batch size if we decide we need to. That sound alright to you?

sdk/firstload/package.json
```diff
@@ -70,7 +70,10 @@
 	},
 	"size-limit": [
 		{
-			"path": "dist/**.js",
+			"path": [
+				"dist/*.js",
+				"!dist/*.umd.js"
+			],
```
ccschmitz, line 76: @Vadman97 it looked like the UMD bundle was including all necessary code to run Highlight and doesn't benefit from dynamic imports. I didn't see a way to trim it down any further, so decided to exclude it from this check. Do you think that is reasonable?

e2e/dotnet/HighlightConfig.cs
Vadman97: right, but what if a customer uses the client SDK on the frontend to talk to their .NET backend, should they propagate the trace context to the .NET trace?

sdk/client/src/otel/index.ts
Vadman97: yup!

sdk/highlight-go/highlight.go
Vadman97: i think this does affect other customers tho since they may want to propagate the session ID to the backend trace, so their backend trace is associated to the session. maybe we can test that with an e2e app (like a customer react + python opentelemetry native no highlight)

sdk/firstload/src/index.tsx
Vadman97: i think im realizing that since firstload is built by vite, we can reference the `src` file both here and line 137, rather than having to `vite build` the client and then use the bundled file here. https://www.loom.com/share/15ce6ce813e24f848839c069bebe5234

sdk/firstload/package.json
ccschmitz: Sounds good! FWIW, I don't think the main bundle size changed much. You can see the new chunk that will be conditionally loaded, but the other files are around the same size.

**Before**
Screenshot 2024-06-25 at 5 13 03 PM

**After**
Screenshot 2024-06-25 at 5 14 23 PM

Vadman97: neat! yeah post gzip not a huge network difference

sdk/firstload/src/index.tsx
ccschmitz: Oh, that's a great point! We should definitely look at simplifying this!

ccschmitz: https://linear.app/highlight/issue/HIG-4770/eliminate-dependency-on-highlight-runclient

sdk/highlight-go/otel.go
ccschmitz: https://linear.app/highlight/issue/HIG-4771/configure-otel-propagation-in-all-sdks

e2e/dotnet/HighlightConfig.cs
ccschmitz: Followed up on this w/ @Vadman97 on a call. We're going to update each SDK to propagate trace context in https://linear.app/highlight/issue/HIG-4771/configure-otel-propagation-in-all-sdks.

PR #8709, opened 2024-05-31
Introduce support for handling asgi servers for django integration.

### LINKED ISSUE #8684: Django errors not processed by highlight
**Describe the bug**
I noticed that some errors are not showing up in highlight and found [this line](https://github.com/highlight/highlight/blob/d630e9132c3df7e292baa08132bbf379dafd47e2/sdk/highlight-py/highlight_io/integrations/django.py#L46), which seems to limit error handling to wsgi workers. We recently switched to asgi, so I think this is the cause of the problem.

**To Reproduce**

1. Set up a django project and run it with asgi server
2. Make a request that raises an exception
3. Go to highlight and no error there

**Expected behavior**
Error being handled any way

**Screenshots**

**Additional context**

jemiluv8:
## Summary
Modify django integration to support asgi servers.

The existing django integration is a thin wrapper around the django WSGIHandler that is used to record
traces and logs from requests being processed by the framework. Django also supports asgi and a similar 
"instrumentation" is required to support handling asgi servers (and thus handlers).

/claim #8684
fixes #8684 

- [ ] Modify django integration file with relevant code to enable it work with ASGI handlers.
- [ ] Setup e2e example using relavant project id
- [ ] Run e2e django example using uvicorn ```poetry run python -m gunicorn highlight_django.wsgi:application ```
- [ ] Visit the django application url at localhost:8000/example
- [ ] Reload until i get an error 
- [ ] Check highlight UI to verify that relevant errors, logs and traces are created

No

No

### DIFF
--- a/sdk/highlight-py/highlight_io/integrations/django.py
+++ b/sdk/highlight-py/highlight_io/integrations/django.py
@@ -8,6 +8,7 @@ try:
     from django.conf import settings as django_settings
     from django.core import signals
     from django.core.handlers.wsgi import WSGIHandler
+    from django.core.handlers.asgi import ASGIHandler
 
     try:
         from django.urls import resolve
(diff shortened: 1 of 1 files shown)


jemiluv8:
@Vadman97, I feel like I might be missing something from this issue.
I haven't been able to successfully replicate the "error not showing up on highlight"
from the related issue. I thought my poetry setup/cache may have been the issue so
I moved the e2e django test application outside the workspace and used the
current/published version of the highlight-io package(0.8.2). It didn't make a difference.
It worked as expected (I saw errors, logs, traces on the highlight web UI).

When I use gunicorn(wsgi) and uvicorn(asgi) to run the django application
I still see the errors from UI. Attached are some relevant screenshots.

**The Integration Update**
Wrapping the ASGI handler to support asgi django applications made sense.
I'm wondering if you've been able to replicate the issue "locally".

## Uvicorn logs
Screenshot 2024-06-03 at 12 55 52 PM

## Current setup outside monorepo
Screenshot 2024-06-03 at 1 10 55 PM

PR #8713, opened 2024-06-04
move ADMIN_PASSWORD auth handler inside route

mayberryzane:
## Summary
- CORS was breaking for `ADMIN_PASSWORD` auth because the handler was defined outside the `r.Route` function that applied the CORS config

- tested locally w/ `ADMIN_PASSWORD` auth enabled, validated problem before, fix after. will confirm cypress test passes

- no

- no


Vadman97 (approved):
should any other routes move inside to get the cors config? eg. the msft teams bot endpoint

mayberryzane:
> should any other routes move inside to get the cors config? eg. the msft teams bot endpoint

I looked at those but I don't think it's necessary as those are all webhooks and shouldn't need to be accessed from browsers

PR #8716, opened 2024-06-04
Grafana query variables

SpennyNDaJets:
## Summary
Allow users to query the backend for attributes that can be used as variables.

https://www.loom.com/share/c060b69eee0e487cb5cc26f6c0bf013a?sid=a9dcc254-2c93-4f0e-bb77-f0df3a5a12f7

1. Click to add a variable
2. Select the query type 
3. Enter a table
4. Enter a Key
- [ ] Preview of options displayed at bottom
5) Apply the variable
6) Edit a graph
7) Add the variable to the query
- [ ] Able to query with variable value

N/A

N/A

### DIFF
--- a/sdk/highlightinc-highlight-datasource/pkg/plugin/datasource.go
+++ b/sdk/highlightinc-highlight-datasource/pkg/plugin/datasource.go
@@ -94,6 +94,8 @@ type QueryKey struct {
 	Type string
 }
 
+type QueryKeyValue string
+
 type KeyType string
 
 const (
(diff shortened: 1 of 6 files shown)


sdk/highlightinc-highlight-datasource/pkg/plugin/datasource.go
```diff
@@ -196,6 +223,66 @@ func (d *Datasource) CallResource(ctx context.Context, req *backend.CallResource
 		if err != nil {
 			return err
 		}
+
+	case "traces-key-values":
+		var q struct {
+			TracesKeyValues []QueryKeyValue `graphql:"traces_key_values(project_id: $project_id, date_range: $date_range, key_name: $query)"`
```
mayberryzane, line 229: I think it would be nice to use the polymorphic `key_values` resolver to avoid most of the code in each of these switch cases. The above code could also be refactored to use the `keys` resolver. I think both of those polymorphic resolvers were added after this original code

sdk/highlightinc-highlight-datasource/src/datasource.ts
```diff
@@ -78,4 +78,15 @@ export class DataSource extends DataSourceWithBackend<HighlightQuery, HighlightD
       limitColumn: columnOptions[0].value,
     };
   }
+
+  async metricFindQuery(query: HighlightVariableQuery, options?: any) {
+    if (!query.table || !query.key) {
+      return [];
+    }
+
+    console.log('metricFindQuery', query);
```
mayberryzane, line 87: console.log

sdk/highlightinc-highlight-datasource/src/components/VariableQueryEditor.tsx
```diff
@@ -0,0 +1,56 @@
...
+  };
+
+  const handleKeyBlur = () => {
+    onChange({ ...query, key: keyText });
+  };
+
+  return (
+    <>
+      <InlineFieldRow>
+        <InlineField label="Table" labelWidth={10}>
```
mayberryzane, line 33: Nit but might be nice to call this "resource" to match its name in the query editor

mayberryzane (approved):

PR #8725, opened 2024-06-05
more dashboard polish items

mayberryzane:
## Summary
- [HIG-4715] makes inputs span full width
- [HIG-4655] fix word wrapping in tooltips
- [HIG-4671] add cloned chart directly after the chart it clones
- [HIG-4712] remove tooltip max height

- clicktested locally

- no

- @julian-highlight


packages/ui/src/components/Tooltip/Tooltip.tsx
```diff
@@ -76,6 +76,7 @@ export const TooltipContent: React.FC<
 			p="4"
 			borderRadius="6"
 			shadow="small"
+			wordBreak="break-word"
```
Vadman97, line 79: padding here feels too little, is this per the designs? cc @julian-highlight 

Screenshot from 2024-06-05 17-20-49

Vadman97 (approved):

packages/ui/src/components/Tooltip/Tooltip.tsx
mayberryzane: yes I think I need a few more pixels gap between rows, will fix

PR #8738, opened 2024-06-06
[HIG-4636] use new graph component in logs / traces histogram

mayberryzane:
## Summary
- changes the histograms on the logs page, traces page, and log alerts page to use the new charts

- clicktested locally

- no

- @julian-highlight for review in the preview env - one question regarding loading states - do we want to have a similar loading state like we used to with grey fake bars / lines?


frontend/src/pages/LogsPage/LogsPage.tsx
```diff
@@ -166,8 +167,9 @@ const LogsPageInner = ({ timeMode, logCursor, presetDefault }: Props) => {
 
 	const { projectId } = useNumericProjectId()
 	const { data: histogramData, loading: histogramLoading } =
-		useGetLogsHistogramQuery({
+		useGetMetricsQuery({
```
Vadman97, line 170: the logs histogram looks really spread out on my machine. maybe we could make each bar fill more space? or we could keep it more consistent with the traces page by adding a second graph on the right (maybe showing number of logs by service name?) so that each one has less horizontal space 

Screenshot from 2024-06-07 08-35-33

Vadman97 (approved):
log alerts and traces page look great - awesome to have the time on the x axis now

frontend/src/pages/LogsPage/LogsPage.tsx
mayberryzane: right now, the bars have a max width (same w/ the metrics graphs) and above that the gap between bars gets larger. Agree that in this case with a small height and large width it looks funny. Another option is to query for more buckets? Will bring this up in demos

mayberryzane: updated to 90, looks better imo
Screen Shot 2024-06-07 at 9 25 17 AM

PR #8739, opened 2024-06-07
[HIG-4723] return all projects in workspace for super users

mayberryzane:
## Summary
- fix issue where the project dropdown for other workspaces was empty for super users
- attempted fix for test flakiness https://github.com/highlight/highlight/actions/runs/9407237601/job/25912686105?pr=8722#step:5:836 
  - only the tests involving the demo project were failing - there's another test which can set a different project id so I think if that test runs first it was causing this test to fail. explicitly set `DEMO_PROJECT_ID = 0` to avoid that

- clicktested locally

- no

- no


Vadman97 (approved):
:bow:

PR #8746, opened 2024-06-09
update sendResponse snippet in cloudflare sdk documentation

cmgriffing:
## Summary

There seemed to be 2 issues with the snippet for `sendResponse` on the Cloudflare SDK documentation.

1. `return` when creating the response would prevent any following code from running
2. The function signature for `sendResponse` only takes in one argument, the response object. Not all 4 arguments being passed in the snippet.

From index.d.ts:
```ts
interface HighlightInterface {
    init: (request: Request, env: HighlightEnv, ctx: ExecutionContext, service?: string) => WorkersSDK;
    isInitialized: () => boolean;
    consumeError: (error: Error) => void;
    sendResponse: (response: Response) => void;
    setAttributes: (attributes: ResourceAttributes) => void;
    recordMetric: (metric: Metric) => void;
}
```

This is a documentation change, so no testing was done. Please verify as you can.

No migration or deployment concerns.

I do not think so.

### DIFF
--- a/docs-content/sdk/cloudflare.md
+++ b/docs-content/sdk/cloudflare.md
@@ -19,7 +19,7 @@ with no impact on performance..
 
 <section className="section">
   <div className="left">
-    <h3>H.init</h3> 
+    <h3>H.init</h3>
     <p>H.init() configures the highlight SDK and records console log methods. The session is inferred based on the incoming network request headers.</p>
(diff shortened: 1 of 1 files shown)


Vadman97 (approved):
thanks @cmgriffing ! looks like these docs were out of date and referencing an old API we had for `sendResponse`

PR #8749, opened 2024-06-10
[SUP-52] ensure unauthed users don't get a NPE, fix join workflow

mayberryzane:
## Summary
- return a default workspace id when the user is unauthenticated or not in the project
- use `errorPolicy: 'all'` so `joinable_workspaces` is available even though other resolvers (`project` and `workspace`) throw an authentication error

- clicktested unauthenticated and auto-join use cases locally

- no

- no


Vadman97 (approved):
TIL about errorPolicy

PR #8754, opened 2024-06-12
Add search docs to AI query Suggestion Prompts

jay-khatri:
## Summary
Adds search docs to AI query Suggestion Prompts. I'm using the `embed` directive to read files in at build time so we don't have to `os.readfile` for each call to the endpoint.

Also, this updates all examples to the prompt to use `=` instead of `:` in queries.


backend/prompts/prompts.go
```diff
@@ -0,0 +1,20 @@
+package prompts
+
+import (
+	_ "embed"
+)
+
+//go:embed search_cleaned.md
+var SearchSyntaxDocs string
```
Vadman97, line 8: TIL!

Vadman97 (approved):

PR #8768, opened 2024-06-14
Improve logs table performance

SpennyNDaJets:
## Summary
Improve performance of logs table by:
- conditionally matchedAttributes attributes
- limit rerenders by reducing changes in `columnData`'s dependency array

1) View the logs table
2) Open the dev tools and start recording the performance
3) Scroll down a few pages
4) Open a logs row
5) Scroll down a few pages

N/A

N/A


Vadman97 (approved):
noticeably better perf! scrolling still has a bit of lag (eg. this page with a few of the rows expanded, slight lag as you scroll https://app.highlight.io/1/logs?query=trace_id%3D68c8a1a2d5a7b002004f89bca613e284), but way better than before.

PR #8769, opened 2024-06-14
[SUP-48] support 7 day retention for sessions / errors

mayberryzane:
## Summary
- add `SevenDays` retention option
- change the Stripe logic to not require Stripe prices if an override is set. Not creating Stripe prices for 7 day retention as this is a custom agreement.
- load `retention_period` / `errors_retention_period` from the workspace in the app context - show loading screen until this is loaded
- filter presets to remove any that are larger than the retention period. For 7 day retention, presets default to 7 days.
- apollo caching improvements to reduce network calls and get rid of white screen transitions (became a problem after requiring workspace details to be loaded)
  - remove `workspace_for_project` and instead use a nested `project.workspace` 
  - create a `Project` fragment to standardize the project fields that are loaded so new fields don't have to be retrieved. This doesn't affect performance since all of these fields are already returned from the DB

- clicktested locally

- do not have to make any changes in Stripe
- will run `update workspaces set retention_period = 'SevenDays' where id = x` after this is deployed
- have already added S3 lifecycle rules for the `SevenDays` option


Vadman97 (approved):
makes sense to me!

PR #8774, opened 2024-06-17
[HIG-4714] longer rounding for longer time intervals for better caching

mayberryzane:
## Summary
- round to nearest 5 minutes when time range >= 4 hours to limit refetching when time range is likely a fraction of a bucket and improve the experience when navigating between pages as data doesn't have to be reloaded 

- testing in preview

- no

- no

### DIFF
--- a/frontend/src/pages/Graphing/components/Graph.tsx
+++ b/frontend/src/pages/Graphing/components/Graph.tsx
@@ -497,7 +497,7 @@ const Graph = ({
 
 		const newStartFetch = presetStartDate(selectedPreset)
 		const newPollInterval =
-			moment().diff(newStartFetch, 'hours') > 5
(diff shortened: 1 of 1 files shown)


frontend/src/pages/Graphing/components/Graph.tsx
```diff
@@ -497,7 +497,7 @@ const Graph = ({
 
 		const newStartFetch = presetStartDate(selectedPreset)
 		const newPollInterval =
-			moment().diff(newStartFetch, 'hours') > 5
+			moment().diff(newStartFetch, 'hours') >= 4
```
Vadman97, line 500: i dont think this will help with navigating between `/logs` and `/traces` since the time range is always reset to last 15 minutes when you click the respective product link from the header. does this only affect the background polling?

Vadman97 (approved):

PR #8776, opened 2024-06-17
fix public graph deployment issue

mayberryzane:
## Summary
- public graph deployment error since Friday - from the logs it appears to be caused by a nil map reference here:
Screen Shot 2024-06-17 at 11 15 16 AM

https://github.com/highlight/highlight/commit/072137b69b7dfee81aca288cdaed5c4d621f8aac#diff-1e34df331d996d37ce29f5f628641b29174beec6e2e2a4a3b84bf21fa39598d2R53
- I think this code is supposed to use `msg` instead of `message`

- ensured logs / traces were still working locally

- no, will monitor public graph deploy

- no

### DIFF
--- a/sdk/highlight-go/log/console_messages_parser.go
+++ b/sdk/highlight-go/log/console_messages_parser.go
@@ -2,9 +2,10 @@ package hlog
 
 import (
 	"encoding/json"
+	"strconv"
+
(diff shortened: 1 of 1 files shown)


Vadman97 (approved):
thanks for catching this

PR #8792, opened 2024-06-18
Bug fix - Reduce log searches

SpennyNDaJets:
## Summary
Using the `query` and `queryParts` from the `SearchContext` is not the same as using the `useQueryParam` as the query changes with each keystroke, resulting in lots of searches.

Revert this change back

1) View the logs
2) Start typing in the search box
- [ ] No search on key type
- [ ] Search on blur
- [ ] Search on enter

N/A

N/A


Vadman97 (approved):
log page perf still looks good!

PR #8813, opened 2024-06-19
[SUP-49] add daily job to delete sessions according to a SESSION_RETENTION_DAYS env var

mayberryzane:
## Summary
- in the `All` runtime, start a job which runs on start up and every 24 hours that runs session deletion logic to remove any session data from all projects older than `SESSION_RETENTION_DAYS`

- tested locally with different `SESSION_RETENTION_DAYS`, confirmed the appropriate sessions were deleted from clickhouse, postgres, and the local filesystem

- no

- no


backend/worker/worker.go
```diff
@@ -1109,6 +1110,24 @@ func (w *Worker) StartLogAlertWatcher(ctx context.Context) {
...
+	}
+	sessionRetentionDays, err := strconv.Atoi(os.Getenv("SESSION_RETENTION_DAYS"))
+	if err != nil {
+		log.WithContext(ctx).Error("Error parsing SESSION_RETENTION_DAYS, skipping SessionDeleteJob")
+		return
+	}
+	if sessionRetentionDays <= 0 {
+		log.WithContext(ctx).Error("sessionRetentionDays <= 0, skipping SessionDeleteJob")
+		return
+	}
```
ccschmitz, line 1127: Do we have any sort of alerting for job run failures? Just wondering if these errors could get lost in the noise since they run kind of infrequently.

ccschmitz (approved):
Code looks good, but I did not click test - let me know if you'd like me to!

Was also wondering, what happens to things associated with these sessions after the records are removed? e.g. errors, sessions, or traces that have the `session_secure_id` stored on them.

backend/lambda-functions/deleteSessions/handlers/handlers.go
```diff
@@ -234,3 +197,63 @@ func (h *handlers) SendEmail(ctx context.Context, event utils.QuerySessionsInput
...
+		return
+	}
+
+	now := time.Now()
+	endDate := now.AddDate(0, 0, -1*retentionDays)
+	startDate := endDate.AddDate(-10, 0, 0)
+
+	for _, id := range projectIds {
+		projectId := id
+		go h.DeleteSessions(ctx, projectId, startDate, endDate, "viewed_by_anyone=false")
```
Vadman97, line 257: perhaps a comment here or a constant to make it clear why we're adding this `viewed_by_anyone=false` filter?

backend/private-graph/graph/schema.resolvers.go
```diff
@@ -4327,14 +4328,19 @@ func (r *mutationResolver) DeleteSessions(ctx context.Context, projectID int, pa
...
-		Email:        email,
-		FirstName:    firstName,
-		Params:       params,
-		SessionCount: sessionCount,
-		DryRun:       util.IsDevOrTestEnv(),
-	})
+	if util.IsInDocker() {
+		deleteHandlers := delete_handlers.InitHandlers(r.DB, r.ClickhouseClient, nil, r.StorageClient)
+		deleteHandlers.DeleteSessions(ctx, projectID, params.DateRange.StartDate, params.DateRange.EndDate, params.Query)
+	} else {
```
Vadman97, line 4334: would be nice to create an e2e test case for this, validating the session no longer shows up once deleted. relatively big effort to add a functional test like that so perhaps we just ticket that for the future?

backend/storage/storage.go
```diff
@@ -473,6 +475,52 @@ func (f *FilesystemClient) SetupHTTPSListener(r chi.Router) {
 	})
 }
 
+func (f *FilesystemClient) DeleteSessionData(ctx context.Context, projectId int, sessionId int) error {
+	if f.fsRoot == "" {
+		return errors.New("f.fsRoot cannot be empty")
```
Vadman97, line 480: good idea lol so we don't `rm /`

Vadman97 (approved):

backend/private-graph/graph/schema.resolvers.go
mayberryzane: :+1: ticketed

backend/worker/worker.go
mayberryzane: agree it's probably worth some better visibility for this and other jobs (e.g. alert evaluation) - ticketing to look into this as part of the alerts work

PR #8820, opened 2024-06-20
Show that data is sampled when sorting traces

ccschmitz:
## Summary

Add some UI to indicate that data from traces is sampled when sorting by the non-default column.

Screenshot 2024-06-20 at 1 16 23 PM

Also adds the UI package back to the Turbo build for `dev:frontend`.

https://www.loom.com/share/5b2802748fe34c5595d4d15c3ea08ec1

N/A

Need to review copy w/ @jay-khatri and @julian-highlight. See [Slack thread](https://highlightcorp.slack.com/archives/C0454PWLLLW/p1718907479112799) on this.


package.json
```diff
@@ -39,7 +39,7 @@
 		"dev": "run-p --print-label --race 'dev:*'",
 		"dev:backend": "cd backend && make start",
 		"dev:components": "yarn turbo run dev --filter @highlight-run/component-preview...",
-		"dev:frontend": "doppler run -- yarn turbo run dev --concurrency 32 --filter frontend... --filter=!rrweb... --filter=!@highlight-run/ui...",
+		"dev:frontend": "doppler run -- yarn turbo run dev --concurrency 32 --filter frontend... --filter=!rrweb...",
```
ccschmitz, line 42: Kind of unrelated, but brought this back since I realized it was disabled a while back because the Storybook build was broken, but now that's fixed, so I didn't see any harm in building it as part of `dev:frontend`. It's come up as a pain point a few times too in retros 😄

frontend/src/pages/Traces/TracesPage.tsx
```diff
@@ -271,23 +275,47 @@ export const TracesPage: React.FC = () => {
...
 							>
 								{metricsLoading ? (
-									<HistogramLoading />
+									<Box py="4">
+										<HistogramLoading />
+									</Box>
 								) : (
 									<>
-										<Text size="xSmall" color="weak">
+										<Tag
```
Vadman97, line 289: a bit weird that the card has a hover state that makes it look like a button when it's not clickable

Screenshot from 2024-06-21 11-16-24
image

Vadman97 (approved):
pulled locally to click test  - works great!

frontend/src/pages/Traces/TracesPage.tsx
ccschmitz: Yeah, I agree. @julian-highlight I think this is a problem with using a `Tag` for something that is purely visual. Perhaps a `Badge` would make more sense here? I will swap out for a badge, but I think used tag initially because it's what I saw in the mocks.

PR #8824, opened 2024-06-21
Add err handling to handle irrelevant queries with Ai search

jay-khatri:
## Summary
- adds an interface for the openai `chatcompletionresponse` and stubs it for `resolver_test`
- adds a prompt entry that handles 'irrelevant queries'
- removes the skipped test in `resolver_test` now that we're using the stub
- update some placeholders


backend/main.go
```diff
@@ -385,6 +386,7 @@ func main() {
 		StepFunctions:          sfnClient,
 		OAuthServer:            oauthSrv,
 		IntegrationsClient:     integrationsClient,
+		OpenAiInterface:        &openai_interface.OpenAiImpl{},
```
Vadman97, line 389: nit: OpenAIClient might be better naming

backend/private-graph/graph/resolver_test.go
```diff
@@ -376,6 +379,14 @@ func TestResolver_GetAIQuerySuggestion(t *testing.T) {
 			productType: modelInputs.ProductTypeLogs,
 			query:       "all logs with level error, between 4pm yesterday and just now.",
 		},
+		"irrelevant query": {
+			productType: modelInputs.ProductTypeLogs,
+			query:       openai_interface.IrrelevantQuery,
```
Vadman97, line 382: i'd move the `IrrelevantQuery` variable to the test, since it's only relevant to the test

backend/openai_interface/openai_interface.go
```diff
@@ -0,0 +1,67 @@
...
+
+func (o *OpenAiImpl) CreateChatCompletion(client *openai.Client, request openai.ChatCompletionRequest) (openai.ChatCompletionResponse, error) {
+	if client == nil {
+		return openai.ChatCompletionResponse{}, errors.New("openai client is nil")
+	}
+	return client.CreateChatCompletion(context.Background(), request)
+}
+
+type OpenAiTestImpl struct {
+}
```
Vadman97, line 32: can this type live in the test file, since it's only relevant to the test?

backend/openai_interface/openai_interface.go
```diff
@@ -0,0 +1,67 @@
...
+
+type OpenAiImpl struct {
+}
+
+func (o *OpenAiImpl) CreateChatCompletion(client *openai.Client, request openai.ChatCompletionRequest) (openai.ChatCompletionResponse, error) {
+	if client == nil {
+		return openai.ChatCompletionResponse{}, errors.New("openai client is nil")
+	}
+	return client.CreateChatCompletion(context.Background(), request)
```
Vadman97, line 28: pass the context thru from the caller, so the openAI inference can be traced back to the calling graphql operation

Vadman97 (approved):
10/10 tests

backend/private-graph/graph/schema.resolvers.go
```diff
@@ -8783,10 +8784,13 @@ And specifically, for the %s product, you can refer to the following documentati
...
 		},
+		{
+			request:  openai_interface.IrrelevantQuery,
+			response: `{"query":"","date_range":{"start_date":"","end_date":""}}`,
+		},
 	}
 
-	resp, err := client.CreateChatCompletion(
-		context.Background(),
+	resp, err := r.OpenAiInterface.CreateChatCompletion(
```
SpennyNDaJets, line 8793: Seeing a `"openai client is nil"` error from hitting this resolver, not sure if it may be a local issue

SpennyNDaJets (commented):
Looking good! Just want to make sure that you were able to get a successful response in the comment below

I added some fixes to the error state on the frontend here: https://github.com/highlight/highlight/pull/8828
Happy to merge those in as part of this PR or separately

PR #8831, opened 2024-06-21
Fix events autoscrolling jumping around

SpennyNDaJets:
## Summary
The events on a session are not scrolling correctly.

Update the algorithm to use a time based search, similar to logs, errors and network resources. Add a throttle for each of these resources to limit the browser computations.

https://www.loom.com/share/4b88dd5d7e2a44238a6ba27b9688561e?sid=86b4665b-84e7-4274-89a4-bb7333f82000

1. View a session
2. View the events
- [ ] Events are being scrolled through accurately
- [ ] Logs are being scrolled through accurately
- [ ] No concerns when running dev tools performance anaylizer
 
N/A

N/A

### DIFF
--- a/frontend/src/pages/Player/PlayerHook/PlayerHook.tsx
+++ b/frontend/src/pages/Player/PlayerHook/PlayerHook.tsx
@@ -9,7 +9,6 @@ import {
 	useMarkSessionAsViewedMutation,
 } from '@graph/hooks'
 import { GetSessionQuery } from '@graph/operations'
-import { usefulEvent } from '@pages/Player/components/EventStreamV2/utils'
 import {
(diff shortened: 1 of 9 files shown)


SpennyNDaJets:
Tagging @Vadman97 incase there are any concerns with the frontend performance - I think it feels the same

Vadman97 (approved):
thanks spenny, i think this is definitely an improvement from the current state (the scrolling is smoother and much more responsive even when the player is updating quickly. i do think we have a regression in the player recently that causes a lot of repaints (see below, with each repaint taking too long), but that's currently the behavior in `main`

Screenshot from 2024-06-24 09-12-22
Screenshot from 2024-06-24 09-12-29

PR #8832, opened 2024-06-21
Additional logging and tracing for batch kafka workers

SpennyNDaJets:
## Summary
Having some trouble figuring out where messages may or may not have been drop, so add logging to help ensure everything is moving smoothly. Also change the kafka batch working resource to be the span name to help with optimized search.

1. Ensure logs and traces can be seen locally

Screenshot 2024-06-24 at 11 39 04 AM

Come back and clean up logs after debugging if appropriate

N/A

### DIFF
--- a/backend/worker/kafka_worker.go
+++ b/backend/worker/kafka_worker.go
@@ -142,10 +142,7 @@ func (k *KafkaBatchWorker) log(ctx context.Context, fields log.Fields, msg ...in
 }
 
 func (k *KafkaBatchWorker) flush(ctx context.Context) error {
-	k.log(ctx, nil,
-		log.Fields{"message_length": len(k.messages)},
-		"KafkaBatchWorker flushing messages",
-	)
+	k.log(ctx, log.Fields{"message_length": len(k.messages)}, "KafkaBatchWorker flushing messages")
 
 	s, _ := util.StartSpanFromContext(ctx, fmt.Sprintf("worker.kafka.%s.flush", k.Name))
 	s.SetAttribute("BatchSize", len(k.messages))
--- a/backend/worker/kafka_worker.go
+++ b/backend/worker/kafka_worker.go
@@ -127,6 +127,12 @@ type KafkaWorker struct {
 }
 
 func (k *KafkaBatchWorker) flush(ctx context.Context) error {
+	log.WithContext(ctx).WithFields(
+		log.Fields{
+			"worker_name":    k.Name,
+			"message_length": len(k.messages)},
+	).Info("KafkaBatchWorker flushing messages")
+
 	s, _ := util.StartSpanFromContext(ctx, fmt.Sprintf("worker.kafka.%s.flush", k.Name))
 	s.SetAttribute("BatchSize", len(k.messages))
 	defer s.Finish()
@@ -190,6 +196,17 @@ func (k *KafkaBatchWorker) flush(ctx context.Context) error {
 			log.WithContext(ctx).Errorf("unknown message type received by batch worker %+v", lastMsg.GetType())
 		}
 	}
+	log.WithContext(ctx).WithFields(
+		log.Fields{
+			"worker_name":       k.Name,
+			"session_ids":       syncSessionIds,
+			"error_group_ids":   syncErrorGroupIds,
+			"error_object_ids":  syncErrorObjectIds,
+			"log_rows_length":   len(logRows),
+			"trace_rows_length": len(traceRows),
+		},
+	).Info("KafkaBatchWorker organized messages")
+
 	k.messages = []kafka_queue.RetryableMessage{}
 
 	readSpan.SetAttribute("MaxIngestDelay", time.Since(oldestMsg).Seconds())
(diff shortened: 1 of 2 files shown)


backend/worker/kafka_worker.go
```diff
@@ -127,7 +127,13 @@ type KafkaWorker struct {
 }
 
 func (k *KafkaBatchWorker) flush(ctx context.Context) error {
-	s, _ := util.StartSpanFromContext(ctx, util.KafkaBatchWorkerOp, util.ResourceName(fmt.Sprintf("worker.kafka.%s.flush", k.Name)))
+	log.WithContext(ctx).WithFields(
+		log.Fields{
+			"worker_name":    k.Name,
+			"message_length": len(k.messages)},
+	).Info("KafkaBatchWorker flushing messages")
```
Vadman97, line 134: how often does this happen? might want to sample (ie. how we do it in the `log` method)

backend/worker/kafka_worker.go
```diff
@@ -190,12 +196,23 @@ func (k *KafkaBatchWorker) flush(ctx context.Context) error {
...
+	log.WithContext(ctx).WithFields(
+		log.Fields{
+			"worker_name":       k.Name,
+			"session_ids":       syncSessionIds,
+			"error_group_ids":   syncErrorGroupIds,
+			"error_object_ids":  syncErrorObjectIds,
+			"log_rows_length":   len(logRows),
+			"trace_rows_length": len(traceRows),
+		},
+	).Info("KafkaBatchWorker organized messages")
```
Vadman97, line 208: sample this too

backend/worker/kafka_worker.go
```diff
@@ -490,7 +521,15 @@ func (k *KafkaBatchWorker) flushDataSync(ctx context.Context, sessionIds []int,
...
 
-		chSpan, _ := util.StartSpanFromContext(ctx, util.KafkaBatchWorkerOp, util.ResourceName(fmt.Sprintf("worker.kafka.%s.flush.clickhouse", k.Name)))
+		chSpan, _ := util.StartSpanFromContext(ctx, fmt.Sprintf("worker.kafka.%s.flush.clickhouse", k.Name))
+
+		log.WithContext(ctx).WithFields(
+			log.Fields{
+				"worker_name":     k.Name,
+				"sessions_length": len(allSessionObjs),
+			},
+		).Info("KafkaBatchWorker writing sessions")
```
Vadman97, line 531: probably all logs can be sampled. based on partition id for instance

Vadman97 (approved):
:tada:

backend/worker/kafka_worker.go
```diff
@@ -126,8 +126,25 @@ type KafkaWorker struct {
 	WorkerThread int
 }
 
+func (k *KafkaBatchWorker) log(ctx context.Context, fields log.Fields, msg ...interface{}) {
+	if k.lastPartitionId != nil {
+		return
+	}
+
+	partitionId := *k.lastPartitionId
+	if partitionId == 408 || partitionId == 58 || partitionId == 289 {
```
SpennyNDaJets, line 135: I added these to match `KafkaWorker.log`, but not sure if these should be different or more?

backend/worker/kafka_worker.go
```diff
@@ -619,6 +673,7 @@ type KafkaBatchWorker struct {
 	Name                string
 	TracingDisabled     bool
 
-	lastFlush time.Time
-	messages  []kafkaqueue.RetryableMessage
+	lastFlush       time.Time
+	messages        []kafkaqueue.RetryableMessage
+	lastPartitionId *int
```
SpennyNDaJets, line 678: Added `lastPartitionId` to the `KafkaBatchWorker` to use as a filtering key

backend/worker/kafka_worker.go
Vadman97: batched topics only have 100 partitions, so we should do something based on that
```suggestion
	if partitionId % 25 == 0 {
```

backend/worker/kafka_worker.go
```diff
@@ -126,8 +126,25 @@ type KafkaWorker struct {
 	WorkerThread int
 }
 
+func (k *KafkaBatchWorker) log(ctx context.Context, fields log.Fields, msg ...interface{}) {
+	if k.lastPartitionId != nil {
+		return
+	}
```
Vadman97, line 132: i think this is flipped? you want to return if the lastPartitionId is nil, to avoid dereferencing it
```suggestion
	if k.lastPartitionId == nil {
		return
	}
```

backend/worker/kafka_worker.go
Vadman97: since the flush will has messages across partitions, maybe it makes sense to just flush randomly but sampled? ie. you can have a `flushCounter int64` here and then the `log()` call will log every `N=100?` times?

Vadman97 (approved):

backend/worker/kafka_worker.go
SpennyNDaJets: Thats an interesting idea! Correct me if I'm wrong but would that result in random logs being logged - I think one of the benefits of using the `lastPartitionId` is we will get all the logs for a single flush and can see where it may be dropping, vs random logs across multiple flushes

PR #8851, opened 2024-06-25
Write to clickhouse from autoresolver

SpennyNDaJets:
## Summary
Since we are writing to clickhouse using their `async_insert` we do not need to queue up on kafka and can write directly from the autoresolver.

1) Resolve an error manually
- [ ] Clickhouse has up to date error

Keep an eye on clickhouse writes 

N/A

### DIFF
--- a/backend/worker/kafka_worker.go
+++ b/backend/worker/kafka_worker.go
@@ -640,11 +640,13 @@ func (k *KafkaBatchWorker) ProcessMessages(ctx context.Context) {
 			defer receiveCancel()
 			task := k.KafkaQueue.Receive(receiveCtx)
 			s1.Finish()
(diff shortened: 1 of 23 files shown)


backend/worker/kafka_worker.go
```diff
@@ -646,8 +646,6 @@ func (k *KafkaBatchWorker) ProcessMessages(ctx context.Context) {
 
 			k.lastPartitionId = &task.GetKafkaMessage().Partition
 
-			k.log(ctx, log.Fields{"last_flush": k.lastFlush, "message_length": len(k.messages)}, "KafkaBatchWorker received messages")
```
SpennyNDaJets, line 649: Tangent change, but remove this noisy log

mayberryzane (approved):

PR #8866, opened 2024-06-27
Enable caching for highlight.io#build

ccschmitz:
## Summary

When running the app via `yarn dev:frontend` we rebuild the `highlight.io` package every time because we have `cache: false` set for the `dev` task. However, the contents of this package don't change often and it's not even a critical part of the main app. This PR sets `cache: true` for the `highlight.io#build` step which allows the cache to be hit and save a lot of time starting up the app locally.

Screenshot 2024-06-27 at 10 10 34 AM

Ran a `dev:frontend` build with and without the setting. Confirmed that with caching enabled the cache was leveraged locally and the time to start up the app was much faster.

No, I don't believe so. I think the cache is not leveraged in CI, so this only impacts local builds.

N/A

### DIFF
--- a/turbo.json
+++ b/turbo.json
@@ -23,7 +23,8 @@
 		"highlight.io#build": {
 			"dependsOn": ["^build"],
 			"outputs": [".next/**"],
-			"env": ["GRAPHCMS_TOKEN"]
+			"env": ["GRAPHCMS_TOKEN"],
+			"cache": true
 		},
 		"highlight.io#lint": {
 			"dependsOn": ["^build"]


ccschmitz:
@Vadman97 curious what you think about this. Do you know if we need to add the [`--force` flag](https://turbo.build/repo/docs/reference/run#--force) to avoid using the cache in prod builds (or should we use the cache in prod?)?

turbo.json
```diff
@@ -23,7 +23,8 @@
 		"highlight.io#build": {
 			"dependsOn": ["^build"],
 			"outputs": [".next/**"],
-			"env": ["GRAPHCMS_TOKEN"]
+			"env": ["GRAPHCMS_TOKEN"],
+			"cache": true
```
Vadman97, line 27: confused why this helps since the `cache: false` is only for `dev`. i would expect the `frontend:dev` to not be cached, but depend on `highligt.io#build` which should already be cached? `highlight.io#dev` still isn't cached, but that shouldn't matter?

Vadman97 (approved):
thanks chris! another thing to think about is how we can remove the package.json dependency of `frontend` on `highlight.io`

turbo.json
ccschmitz: I think it helps because `dev:frontend` runs `turbo run dev`, so it uses the config under `dev` in `turbo.json`. That's my understanding at least, but I could definitely be misunderstanding 😅

Vadman97:
i think it's fine to use cache in prod, but we might not always leverage it because prod uses different env values for some of the keys, resulting in a different cache key (prod retries would be cached tho

PR #8871, opened 2024-06-27
fix datasync queue

Vadman97:
## Summary

Testing a hypothesis that autoresolver error group updates may be
missing from clickhouse because we were writing to the datasync queue
async and not waiting for a batched write flush. This is particularly noticeable
with the autoresolver where the job would run one-off and not
wait for the datasync queue write to complete.

deploying to prod

no

no


SpennyNDaJets (approved):
Looks good to me!

PR #8875, opened 2024-06-28
Fix date range picker

SpennyNDaJets:
## Summary
Make the following updates to the date picker
- No overflow on focus state
- Allow inputs without the year
- Always show time
- Veritcally align boxes
- Fix error state for end date

Also swaps out final instance of `TimeRangePicker` and cleans up respective code 

https://www.loom.com/share/96f62d308ba8488983932c09da7bb7bb?sid=8465255b-d54e-491e-af77-3aa69f02e37b

1. View the logs/traces page
2. Select a custom time
- [ ] Can select custom date
- [ ] Can type in date
- [ ] Incorrect dates and times show error
3. View an error and click into the metrics tab
- [ ] Able to select date

N/A

N/A


packages/ui/src/components/DatePicker/DateInput.tsx
Vadman97, line 1: two issues i still notice 
* using full month names like `july` or `june` does not work, requiring `jul` and `jun` respectively
* hitting `enter` on the date/time inputs doesn't update the selected range, requiring unfocusing

Vadman97 (approved):
much better across the products

PR #8882, opened 2024-06-30
Harold UI improvements

SpennyNDaJets:
## Summary
Make the following UI updates to Harold:
- [x] Clicking 'escape' once i get into the query text UI should take me out of it (essentially click 'cancel' for me)
- [x] Write the error message to the user (so they know if there prompt is wrong or if its an error with highlight)
- [x] We shouldn't cache the query text if you accept or edit a query.
- [x] Clicking 'generate query…' should force my cursor into the input box
- [x] Make this outline a solid purple (same color as the icon)

1) Navigate to the logs page
2) When typing in a generated query ensure each of the following check points above is satisfied

N/A

N/A

### DIFF
--- a/frontend/src/__generated/index.css
+++ b/frontend/src/__generated/index.css
@@ -6318,8 +6318,7 @@ body {
 }
 ._1k082jh0:after {
   content: "";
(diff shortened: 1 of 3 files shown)


jay-khatri (approved):
did a clicktest on the relfame preview, looks great.

PR #8889, opened 2024-07-01
[SUP-63] fix logging from background tasks

### LINKED ISSUE #8827: Logging doesn't work inside FastAPI background tasks
**Describe the bug**
Logs inside background tasks of fastapi are not getting captured.

**To Reproduce**
```
import logging

import highlight_io
from fastapi import BackgroundTasks, FastAPI, Request
from highlight_io.integrations.fastapi import FastAPIMiddleware

H = highlight_io.H(
    "<>",
    instrument_logging=True,
    service_name="my-app",
    service_version="git-sha",
    environment="production",
)

app = FastAPI()
app.add_middleware(FastAPIMiddleware)


def test_fn():
    logging.error("Logging inside background task")


@app.get("/")
async def root(request: Request, background_tasks: BackgroundTasks):
    logging.info("Before adding background task")
    background_tasks.add_task(test_fn)
    logging.info("After adding background task")
```

**Expected behavior**
Expected to see "Logging inside background task" but I don't see that. I only see the logs in the root function

**Screenshots**

**Additional context**
I suspect it's due to the custom middleware. If I remove the middleware line `app.add_middleware(FastAPIMiddleware)`, I can see the logs.

However, I can't remove it because I want to capture any unforseen exceptions etc to be visible in `Errors` section, and this happens only when I add the middleware instrumentation

mayberryzane:
## Summary
- we fail to send logs to Highlight in FastAPI background tasks with our `FastAPIMiddleware` because the current span has ended (the request returned before the background task started)
- one workaround implemented here is to create a new span in the `LogRecordFactory` if the current span has stopped recording
- it also works to remove the `is_recording` check in the log hook - unsure if this causes issues elsewhere

closes #8827

- used e2e test to validate background task's logs were sent

- no

- no

### DIFF
--- a/sdk/highlight-py/highlight_io/sdk.py
+++ b/sdk/highlight-py/highlight_io/sdk.py
@@ -480,7 +480,7 @@ class H(object):
 
         def factory(*args, **kwargs) -> LogRecord:
             span = otel_trace.get_current_span()
-            if span != INVALID_SPAN:
+            if span != INVALID_SPAN and span.is_recording():
                 manager = contextlib.nullcontext(enter_result=span)
             else:
                 manager = self.trace("highlight.log")
--- a/e2e/python/highlight_fastapi/main_2.py
+++ b/e2e/python/highlight_fastapi/main_2.py
@@ -2,14 +2,16 @@ import asyncio
 import sys
 import highlight_io
 import uvicorn
-from fastapi import FastAPI, Request
+from fastapi import FastAPI, Request, BackgroundTasks
 from highlight_io.integrations.fastapi import FastAPIMiddleware
+import logging
+
 
 # `instrument_logging=True` sets up logging instrumentation.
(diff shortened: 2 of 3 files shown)


Vadman97 (approved):

sdk/highlight-py/highlight_io/sdk.py
```diff
@@ -480,7 +480,7 @@ def _instrument_logging(self):
 
         def factory(*args, **kwargs) -> LogRecord:
             span = otel_trace.get_current_span()
-            if span != INVALID_SPAN:
+            if span != INVALID_SPAN and span.is_recording():
```
Vadman97, line 483: could you bump `pyproject.toml` patch version to release a new SDK version?

sgondala:
Hi, the latest version of highlight doesn't install. 

Can you pin the right dependencies?

https://github.com/highlight/highlight/issues/8937

PR #8898, opened 2024-07-02
Fix duplicate errors reported on traces in Ruby SDK

ccschmitz:
## Summary

Cleans up the explicit reporting of errors during traces in our Ruby SDK because we already have [logic for creating errors/logs on traces that have error events](https://github.com/highlight/highlight/blob/b0ed16c1ef093e58712d608cb647cebf102cc6b1/backend/otel/otel.go#L223-L265).

* Spun up the e2e Rails app blog example app and was easily able to reproduce the duplicate errors when hitting the `/articles` path.
* Removed the code calling `record_expection` and reloaded then confirmed I still saw the errors coming in, but wasn't seeing duplicates anymore.

Need to [notify any customers impacted by this bug](https://highlightcorp.slack.com/archives/C0721TM6H6C/p1719952551690369?thread_ts=1718233023.899189&cid=C0721TM6H6C).

N/A


sdk/highlight-ruby/highlight/lib/highlight.rb
```diff
@@ -57,9 +57,6 @@ def trace(session_id, request_id, attrs = {})
         HIGHLIGHT_TRACE_ATTRIBUTE => request_id
       }.merge(attrs).compact) do |_span|
         yield
-      rescue StandardError => e
-        record_exception(e)
```
Vadman97, line 61: ah, so the otel instrumentation already automatically reports the exception event when one is raised, so we don't need to explicitly create another span

Vadman97 (approved):
awesome, thanks for digging into this!

sdk/highlight-ruby/highlight/lib/highlight.rb
ccschmitz: Yeah, exactly!

PR #8907, opened 2024-07-03
avoid processing multiple ruby span events

Vadman97:
## Summary

Ruby SDK produces multiple span events resulting in duplicate exceptions being recorded.

Prod deploy monitoring customer project.

no, will debug why this is an issue only with ruby sdk

no


ccschmitz (approved):
Makes sense as a workaround for now! Thanks for digging into this!

PR #8938, opened 2024-07-10
Fix re-searching sessions

SpennyNDaJets:
## Summary
Session search researches after 10 seconds when the endDate is no longer near now - use the selectedPreset to determine time to add to the endDate.

1. View the sessions page
- [ ] No re-searching after ~10 seconds
- [ ] Caching improvements are still intact 

N/A

N/A


ccschmitz (approved):
Thanks for fixing this!

PR #8944, opened 2024-07-10
[SUP-72] fix env replacements

mayberryzane:
## Summary
- `frontend-entrypoint` should replace env variables that were set during the frontend build - this wasn't happening because the released docker builds have production values instead of dev ones
- the `AUTH_MODE` checks were being inlined by rollup? - fix that by using a switch statement because rollup doesn't seem to touch that

- ran `doppler -c prod_aws run -- yarn build` and `python frontend-entrypoint.py` locally to validate that a `constants` file was created with all 4 variables and that all 4 were replaced
- will test this PR's docker image once it's built 

- will publish a new docker image after deploying

- no

### DIFF
--- a/docker/frontend-entrypoint.py
+++ b/docker/frontend-entrypoint.py
@@ -23,7 +23,7 @@ def main():
             data = re.sub('https://pri\.highlight\.io', private, data)
         if public:
             data = re.sub('http://localhost:8082/public', public, data)
-            data = re.sub('https://pub\.highlight\.run', private, data)
+            data = re.sub('https://pub\.highlight\.run', public, data)
         if frontend:
             data = re.sub('http://localhost:3000', frontend, data)
             data = re.sub('https://app\.highlight\.io', frontend, data)
--- a/docker/frontend-entrypoint.py
+++ b/docker/frontend-entrypoint.py
@@ -12,7 +12,7 @@ def main():
     frontend = os.environ.get('REACT_APP_FRONTEND_URI')
     auth = os.environ.get('REACT_APP_AUTH_MODE')
     use_ssl = os.environ.get('SSL') != 'false'
-    print("replacing", {"private": private, "public": public, "frontend": frontend})
+    print("replacing", {"private": private, "public": public, "frontend": frontend, "auth": auth})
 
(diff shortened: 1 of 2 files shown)


docker/frontend-entrypoint.py
```diff
@@ -12,18 +12,21 @@ def main():
...
     with open(CONSTANTS_FILE, 'r') as f:
         data = f.read()
         if auth:
             data = re.sub('firebase', auth, data)
         if private:
             data = re.sub('http://localhost:8082/private', private, data)
+            data = re.sub('https://pri\.highlight\.io', private, data)
         if public:
             data = re.sub('http://localhost:8082/public', public, data)
+            data = re.sub('https://pub\.highlight\.run', public, data)
```
ccschmitz, line 26: Should this be `.run` or `.io`? If `.run` is right, just curious why we are using that instead of `.io`?

ccschmitz (approved):
Code here looks good to me, but not sure I understand the issue with the switch vs conditional 🤔 - mind sharing a little more info on that?

Also, do you think I need to do a similar replacement for `OTLP_ENDPOINT` [in this PR](https://github.com/highlight/highlight/pull/8706/files#diff-960508675c48f0ea375f7edf25c366b38b88b9810d17c6ecaac6d338d9bbd13aR10-R11)?

docker/frontend-entrypoint.py
mayberryzane: the value is `highlight.run` in our production docker image - we can probably update that at some point

mayberryzane:
> Code here looks good to me, but not sure I understand the issue with the switch vs conditional 🤔 - mind sharing a little more info on that?
> 
> Also, do you think I need to do a similar replacement for `OTLP_ENDPOINT` [in this PR](https://github.com/highlight/highlight/pull/8706/files#diff-960508675c48f0ea375f7edf25c366b38b88b9810d17c6ecaac6d338d9bbd13aR10-R11)?

@ccschmitz the switch vs conditional thing is a hack - when the conditional was used, since the env var was constant, the other branches were removed in the generated code

you probably need to replace `OTLP_ENDPOINT` too to allow users to overwrite this value

PR #8951, opened 2024-07-11
[SUP-73] faster to call slack api separately for each channel type

mayberryzane:
## Summary
- due to how Slack applies the limit 1000 param before filtering, for slack workspaces with many channels, it was taking 60+ iterations and many were returning <10 results
- somehow, calling the api separately for each channel type fixes this

- tested locally against large slack workspace

- no, will re-validate after deploying

- no

### DIFF
--- a/backend/private-graph/graph/resolver.go
+++ b/backend/private-graph/graph/resolver.go
@@ -3477,31 +3477,37 @@ func (r *Resolver) GetSlackChannelsFromSlack(ctx context.Context, workspaceId in
 		slackClient := slack.New(*workspace.SlackAccessToken)
 		existingChannels, _ := workspace.IntegratedSlackChannels()
 
-		getConversationsParam := slack.GetConversationsParameters{
-			ExcludeArchived: true,
-			Limit:           1000,
-			// public_channel is for public channels in the Slack workspace
-			// private is for private channels in the Slack workspace that the Bot is included in
-			// mpim is for multi-person conversations in the Slack workspace that the Bot is included in
-			Types: []string{"public_channel", "private_channel", "mpim"},
-		}
 		allSlackChannelsFromAPI := []slack.Channel{}
 
-		// Slack paginates the channels/people listing.
-		for {
-			channels, cursor, err := slackClient.GetConversations(&getConversationsParam)
-			getConversationsParam.Cursor = cursor
-			if err != nil {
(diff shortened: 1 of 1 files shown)


backend/private-graph/graph/resolver.go
```diff
@@ -3477,31 +3477,37 @@ func (r *Resolver) GetSlackChannelsFromSlack(ctx context.Context, workspaceId in
...
+				}
 
-			if getConversationsParam.Cursor == "" {
-				break
+				allSlackChannelsFromAPI = append(allSlackChannelsFromAPI, channels...)
+
+				if getConversationsParam.Cursor == "" {
+					break
+				}
+				// delay the next slack call to avoid getting rate limited
```
SpennyNDaJets, line 3508: Curious how this may affect the rate limit? Will we be making 3x calls?

backend/private-graph/graph/resolver.go
```diff
@@ -3477,31 +3477,37 @@ func (r *Resolver) GetSlackChannelsFromSlack(ctx context.Context, workspaceId in
...
+				Types:           []string{channelType},
 			}
 
-			allSlackChannelsFromAPI = append(allSlackChannelsFromAPI, channels...)
+			// Slack paginates the channels/people listing.
+			for {
+				channels, cursor, err := slackClient.GetConversations(&getConversationsParam)
+				getConversationsParam.Cursor = cursor
+				if err != nil {
+					return nil, e.Wrap(err, "error getting Slack channels from Slack.")
```
SpennyNDaJets, line 3500: Just want to check if we should still return early here vs log and continue incase only one channelType fails?

SpennyNDaJets (approved):
Looks good! Just a couple questions that wanted to make sure were raised

backend/private-graph/graph/resolver.go
mayberryzane: for small workspaces, yes, but for large ones we'll actually end up making fewer calls as there will be fewer pages to iterate over. Since these large workspaces are probably more impactful for our rate limit, I think this is better

backend/private-graph/graph/resolver.go
mayberryzane: I would prefer to return early in case it's an issue with the slack api - that way we'll just fall back to whatever was already saved on the workspace instead of overwriting it with potentially fewer slack channels

PR #8957, opened 2024-07-12
fix cross-origin iframe parent recording start

mayberryzane:
## Summary
- looks like some earlier refactoring broke iframe recording - the child iframe would wait indefinitely for a message from the parent as `this._recordStop` was always true

- set up cross-origin iframe recording locally, validated issue before / fixed after
https://www.loom.com/share/142e5b5fd764453fb6691b86315b4895

- will release new client version w/ changeset

- no

### DIFF
--- a/.changeset/small-peas-knock.md
+++ b/.changeset/small-peas-knock.md
@@ -1,5 +1,0 @@
----
-'highlight.run': patch
----
-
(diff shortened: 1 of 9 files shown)


Vadman97 (approved):
add a changeset?

PR #8959, opened 2024-07-12
Client OTEL updates

ccschmitz:
## Summary

Adds a couple methods to our JavaScript SDK for capturing spans.

* `startSpan` - creats a span that automatically ends when the callback function returns.
* `startManualSpan` - creates a span that needs to be ended manually.

I also swapped out some of our timing instrumentation for play/pause/seek operations on the player. These were submitting metrics via `analytics.track`, but now they are creating spans.

This PR also contains a few other updates:

* Bumps the version of the `highlight.run` SDK (I forgot to do this in https://github.com/highlight/highlight/pull/8706). 
* Adds string replacement logic to `docker/frontend-entrypoint.py` for the OTEL endpoint.
* Started using the `StackContextManager` which should do a better job of connecting spans created in async operations.
* Pulls out the timing information from span events created by `DocumentLoadInstrumentation` and assigns them as span attributes, which makes them usable in metrics (resolves [HIG-4789](https://linear.app/highlight/issue/HIG-4789/ensure-timing-events-are-captured-correctly)).
* Prevents trace propagation headers from being applied to requests if the request is not supposed to be recorded (resolves [HIG-4798](https://linear.app/highlight/issue/HIG-4798/remove-trace-propagation-headers-on-spans-that-shouldnt-be-tracked)).

* Load the app locally, which should trigger the `adminFetch` span.
* Play a session and click around to trigger play, pause, and seek oprations.
* Go to `/traces` and confirm the traces are appearing as expected.

Note that I used different methods of capturing spans to confirm the different forms behave as expected:

* Calling `startSpan`
* Calling `startManualSpan` and calling `span.end()` inside the callback
* Calling `startManualSpan` and returning the `span` from the callback, assigning it to a variable, and calling `span.end()` outside the callback

Also click tested to verify that the following span types had the timing attributes assigned correctly:

* `documentLoad`
* `documentFetch`
* `resourceFetch`

Finally, I confirmed the `traceparent` and `tracestate` headers were being removed from requests headers for requests that aren't recorded. e.g. all the `*dataplane*` URLs for Rudderstack requests.

Will verify these things work as expected in production.

N/A


docker/frontend-entrypoint.py
```diff
@@ -11,6 +11,7 @@ def main():
     public = os.environ.get('REACT_APP_PUBLIC_GRAPH_URI')
     frontend = os.environ.get('REACT_APP_FRONTEND_URI')
     auth = os.environ.get('REACT_APP_AUTH_MODE')
+    otel = os.environ.get('REACT_APP_OTLP_ENDPOINT')
```
mayberryzane, line 14: you may want to check the generated `constants.js` file to make sure this isn't being elided

SpennyNDaJets (approved):
Code looks good, but was not able to complete code test 🤔
- ensured `REACT_APP_OTLP_ENDPOINT` was set
- ran `yarn build:sdk`

mayberryzane (approved):
LGTM!

docker/frontend-entrypoint.py
ccschmitz: Confirmed it's in there! Here's what I see in a local build:

```ts
var _ = { BASE_URL: "/", MODE: "production", DEV: !1, PROD: !0, SSR: !1 };
const o = _.REACT_APP_PRIVATE_GRAPH_URI ?? window.location.origin + "/private",
  t = _.REACT_APP_PUBLIC_GRAPH_URI || "https://pub.highlight.io",
  P =
    _.REACT_APP_FRONTEND_URI ||
    window.location.protocol + "//" + window.location.host,
  R = _.REACT_APP_AUTH_MODE,
  A = _.REACT_APP_OTLP_ENDPOINT || "https://otel.highlight.io:4318";
export { R as A, P as F, A as O, o as P, t as a };
```

PR #8991, opened 2024-07-22
[HIG-4596] related resources from metric dashboards

mayberryzane:
## Summary
- some small backend changes for error objects:
  - allow error objects to be queried either by error object id or project id
  - return errorGroupSecureId
- frontend implementation of related resources
- for now, clicking on an individual line or bar will open the right panel with the correct resource
- logs uses the existing logs list view
- traces / sessions / errors introduces a new `RelatedResourceList` - borrows heavily from the existing `TracesTable` but allows for different columns to be used depending on the resource

- clicktested locally

- releasing behind admin flag
- following this up with HIG-4816

- @julian-highlight


SpennyNDaJets (approved):
LGTM!

mayberryzane:
monorepo step failing - continuing because there are no new npm versions

PR #8998, opened 2024-07-23
add other common emails to denylist

mayberryzane:
## Summary
- restrict other common email domains in auto-join logic

- this is just a config change

- no

- no

### DIFF
--- a/backend/private-graph/graph/resolver.go
+++ b/backend/private-graph/graph/resolver.go
@@ -223,7 +223,27 @@ func (r *Resolver) getCustomVerifiedAdminEmailDomain(admin *model.Admin) (string
 	}
 
 	// this is just the top 10 email domains as of June 6, 2016, and protonmail.com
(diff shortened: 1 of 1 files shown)


backend/private-graph/graph/resolver.go
```diff
@@ -223,7 +223,27 @@ func (r *Resolver) getCustomVerifiedAdminEmailDomain(admin *model.Admin) (string
...
+		"wanadoo.fr":     true,
+		"orange.fr":      true,
+		"qq.com":         true,
+		"icloud.com":     true,
+		"live.com":       true,
+		"me.com":         true,
+		"proton.me":      true,
+		"duck.com":       true,
+		"mail.ru":        true,
+		"163.com":        true,
```
adelowo, line 245: probably time to move this to a constant/file definition of it's own?

SpennyNDaJets (approved):

PR #9010, opened 2024-07-24
Improvements to Ruby SDK and Rails demo app

ccschmitz:
## Summary

Creates a new Rails demo app with a page that has buttons to trigger the creation of traces, errors, and logs. Also makes some updates to the Ruby SDK:

* Does not error if methods are called and SDK is not initialized (helpful to code that is run in isolation during tests)
* Adds a `start_span` method that allows you to create a new span with just a name and attributes (the old `trace` method required a session and request ID to be passed, which didn't make sense outside of the controller action)
* Fixes a bug with the logger so its API is compatible with the silent logger (this was causing an issue in the old demo app loading assets)
* Adds some convenience methods to improve the ergonomics of using the SDK, allowing you to call methods directly on the `Highlight` module (e.g. call `Highlight.start_trace` rather than `Highlight::H.instance.start_trace`)
* Enables all available auto instrumentations.

Closes SUP-68

testing using local ruby SDK and local firstload SDK
https://www.loom.com/share/572698c9f81f49579ab3171fe35003d5

image

Screenshot 2024-08-13 at 14 25 43

new ruby sdk version

no


ccschmitz:
@Vadman97 I just cleaned up a few things in this PR in preparation for a handoff to you. I also tossed in 6bc543330e04dd0a44da854b607bc85e7f7eea88 which includes a `<meta>` tag attempting to handoff the server trace to the client. Hadn't successfully connected any client and server traces that way yet, but I think that's how it's supposed to work 😸

ccschmitz:
@Vadman97 was also wondering if you think we should move the Rails app from the SDK directory to the root-level `e2e` directory?

sdk/client/src/otel/index.ts
```diff
@@ -285,7 +285,7 @@ const getSpanName = (
 	let parsedBody
 	const urlObject = new URL(url)
 	const pathname = urlObject.pathname
-	let spanName = `${method} - ${pathname}`
+	let spanName = `${method.toUpperCase()} - ${pathname}`
```
Vadman97, line 288: @ccschmitz this change is unrelated, right?

ccschmitz: Oh yeah, sorry... Can back this change out probably

docker/collector.yml
Vadman97, line 1: @ccschmitz i'm noticing a difference between this file and `deploy/otel-collector.yaml`, our prod version. notably in `allowed_headers` - should i update the `deploy` one?

image

sdk/client/src/otel/index.ts
Vadman97: all good, can also have this here but i'll add a changelog to release the client

docker/collector.yml
ccschmitz: Ah, yes, I think we'll need it in the deploy one as well. Good catch!

.github/workflows/ruby.yml
```diff
@@ -14,6 +14,9 @@ jobs:
     build:
         name: Ruby SDK
         runs-on: ubuntu-latest
+        strategy:
+            matrix:
+                ruby: ['3.0', '3.3']
```
ccschmitz, line 19: Nice moves 🕺

e2e/ruby/rails/demo/Gemfile
```diff
@@ -68,7 +68,6 @@ group :test do
   # Use system testing [https://guides.rubyonrails.org/testing.html#system-testing]
   gem "capybara"
   gem "selenium-webdriver"
-  gem "webdrivers"
 end
 
-gem "highlight_io", path: "../../../highlight"
+gem "highlight_io", path: "../../../../sdk/highlight-ruby/highlight"
```
ccschmitz, line 73: Maybe we should not change this path?

sdk/client/src/utils/storage.ts
```diff
@@ -34,15 +61,21 @@ export const setStorageMode = (m: Mode) => {
...
+export const setCookieWriteEnabled = (enabled: boolean) => {
+	cookieWriteEnabled = enabled
+}
+
 export const getItem = (key: string) => {
 	return getPersistentStorage().getItem(key)
 }
 
 export const setItem = (key: string, value: string) => {
+	cookieStorage.setItem(key, value)
```
ccschmitz, line 73: Sorry if I'm missing something here, but why write everything to cookies too?

e2e/ruby/rails/demo/Gemfile
Vadman97: this was from moving the e2e ruby app to the `e2e` directory; the sdk didn't move. does that make sense?

ccschmitz: Oh, I see - sorry!

sdk/client/src/utils/storage.ts
ccschmitz: Ah, right - makes sense!

Vadman97 (approved):
💪

PR #9022, opened 2024-07-25
Use local DB to query geolocation data from IP

ccschmitz:
## Summary

Swaps out our geolocation lookup service for a local geolocation DB that we can query to get geo data for an IP address. The data comes from [MaxMind's GeoLite 2 Free Geolocation Data](https://dev.maxmind.com/geoip/geolite2-free-geolocation-data).

I attempted to click test by creating a new session, but when making a request from `localhost` it uses a loopback IP (`::1`) which couldn't be used for querying the IP address DB. Then I hard-coded a static IP and it geocoded my session correctly.

Will need to double-check results of new sessions and ensure data is loaded as expected.

N/A


mayberryzane:
could we also add a migration file to backfill missing geolocations for last week?

ccschmitz:
@mayberryzane good call - tracking that in https://linear.app/highlight/issue/HIG-4834/migration-to-backfill-geolocation-on-sessions!

PR #9028, opened 2024-07-26
replace metric montior w/ metric alert

mayberryzane:
## Summary
- piggyback on existing metric-monitor service, but use the `StartMetricAlertWatcher` entrypoint instead

- have tested the 'StartMetricAlertWatcher' entrypoint locally, will monitor service after prod deployment

- no

- no


SpennyNDaJets (approved):

mayberryzane:
ignoring monorepo check because no js changes

PR #9041, opened 2024-07-30
fix highlight.run getSessionUrl only returning session ID 

Vadman97:
## Summary

* Fixes `getSessionURL` to return the full session url
* Updates `getSessionDetails` to return the URL with timestamp without blocking for the client.

new cypress test additions

changeset

no


ccschmitz (approved):
Nice! Looks good to me, but did not click test. Let me know if you want me to pull down and click test, but looks like the new test covers the change pretty well!

PR #9080, opened 2024-08-01
Fix error finding geolocation database

ccschmitz:
## Summary

When we build our backend for production the file with the geolocation database wasn't included or available alongside the app. This PR bundles the geolocation database in the app. It also fills in the `State` attribute of locations correctly.

I hard-coded my IP address and ran the code then kicked off a new session. It correctly geocoded my location.

Screenshot 2024-08-01 at 3 01 06 PM

I also checked the before and after build size of the app. It does bloat the build by 50mb, but don't know of a reason that would be a problem.

```sh
# before
-rwxr-xr-x@  1 chris  staff  105632242 Aug  1 14:27 backend

# after
-rwxr-xr-x@  1 chris  staff  157645506 Aug  1 14:58 backend
```

Need to confirm the larger build isn't a problem and then verify geolocation data is gathered correctly in production. I'll spot check when this goes out.

N/A


backend/public-graph/graph/resolver.go
```diff
@@ -963,6 +964,7 @@ func GetLocationFromIP(ctx context.Context, ip string) (location *Location, err
 
 	location = &Location{
 		City:      record.City.Names["en"],
+		State:     record.Subdivisions[0].Names["en"],
```
mayberryzane, line 967: is `Subdivisions` guaranteed to be non-empty? should we have a check to avoid a panic here?

mayberryzane (approved):

backend/public-graph/graph/resolver.go
ccschmitz: Yeah, sorry, we actually have handling for it below, I just missed it while stepping through code on a breakpoint.

https://github.com/highlight/highlight/pull/9080/files#diff-37a818bb25fd3cee29d865b3306a81501a1371a5ba2e6f1927464f21a56b20a1L972-L974

PR #9087, opened 2024-08-02
fix otel webjs tracer creation when otel code has not loaded

Vadman97:
## Summary

Ensures that calling `H.startSpan()` before the otel library has loaded resolves 
and calls the `fn` but without a span, preventing a delay and crash of the library.

render preview of our frontend

changeset

no

### DIFF
--- a/.changeset/calm-years-repeat.md
+++ b/.changeset/calm-years-repeat.md
@@ -1,0 +1,5 @@
+---
+'highlight.run': patch
+---
+
+fix otel webjs startSpan crashing when otel code has not loaded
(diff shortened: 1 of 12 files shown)


ccschmitz (approved):

.changeset/calm-years-repeat.md
```diff
@@ -0,0 +1,5 @@
+---
+'highlight.run': patch
+---
+
+fix otel webjs startSpan crashing when otel code has not loaded
```
ccschmitz, line 5: @Vadman97 do we need to bump the version number for this? Also, do we need to add an entry in the changelog?

Vadman97: yep, that's what the `yarn changeset version` command will do with this file. normally i release the SDK in a separate PR, but you're right that I should do that here so we can get the SDK change out asap.

PR #9094, opened 2024-08-02
[HIG-4823] fix exists queries for sessions

mayberryzane:
## Summary
- for exists / not exists queries when the attributes column is a list, check if the key exists in the array without comparing the value

- clicktested locally with grouped results

- no

- no

### DIFF
--- a/backend/parser/listener/listener.go
+++ b/backend/parser/listener/listener.go
@@ -79,7 +79,12 @@ func (s *SearchListener) getAttributeFilterExpr(op Operator, value any) sqlbuild
 		postfix = ")"
 	}
 	if s.attributesList {
-		return sqlbuilder.Buildf(fmt.Sprintf("notEmpty(arrayFilter((k, v) -> k = %%s AND %sv%s %s %%s, %s))", prefix, postfix, op, s.attributesColumn), s.currentKey, value)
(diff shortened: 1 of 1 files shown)


backend/parser/listener/listener.go
```diff
@@ -79,7 +79,12 @@ func (s *SearchListener) getAttributeFilterExpr(op Operator, value any) sqlbuild
 		postfix = ")"
 	}
 	if s.attributesList {
-		return sqlbuilder.Buildf(fmt.Sprintf("notEmpty(arrayFilter((k, v) -> k = %%s AND %sv%s %s %%s, %s))", prefix, postfix, op, s.attributesColumn), s.currentKey, value)
+		// For NOT EXISTS queries, return true if there is no matching key in the array.
+		if value == "" {
+			return sqlbuilder.Buildf(fmt.Sprintf("empty(arrayFilter((k, v) -> k = %%s, %s))", s.attributesColumn), s.currentKey)
```
Vadman97, line 84: isn't the value `""` for both `exists` and `not exists` queries?

Vadman97 (approved):

backend/parser/listener/listener.go
mayberryzane: yes, was trying to explain why it's returning `empty` instead of `not empty` here cause it was a little confusing

PR #9118, opened 2024-08-07
Add migration for backfilling sessions with missing geolocation data

ccschmitz:
## Summary

Adds a migration for backfilling sessions that were missing geolocation data. After #9080 new sessions should be getting location data assigned correctly, but still all sessions created between when the old service went down sometime on 7/19/24 and when that PR was merged are missing location data.

All sessions fail to geocode properly in local dev now, so I set the IP address of some sessions manually then ran the migration and confirmed it set the location correctly.

This PR won't actually run the migration. We will need to kick that off manually. I have this on my list of things to go over in my next pairing w/ @mayberryzane.

We should notify any customers who had questions about this while geolocation data wasn't available.

N/A


ccschmitz:
@Vadman97 @mayberryzane I was working on this migration and realizing no sessions were being updated for me. I tracked it down to the fact we aren't actually storing IP addresses on most workspaces because of the workspace setting this is behind: https://github.com/highlight/highlight/pull/6343/files#diff-37a818bb25fd3cee29d865b3306a81501a1371a5ba2e6f1927464f21a56b20a1R1159-R1161

I don't think this will help most accounts since the default value is `false` for `StoreIP`, but can't hurt to get this out and run it.

As a follow up, do you think we should change the default or add a setting in the UI for managing whether IPs are stored?

Vadman97 (approved):

mayberryzane (approved):
looks good! two things that might help - you can filter by workspaces where StoreIP = true if that's less common to speed up the query, you also probably want to submit the sessions to the datasync queue to make sure their geo data is updated in clickhouse for anyone trying to search

see: https://github.com/highlight/highlight/blob/main/backend/migrations/cmd/backfill-errors-clickhouse-v2/main.go

backend/migrations/cmd/backfill-geolocation-on-sessions/main.go
```diff
@@ -0,0 +1,82 @@
...
+				"latitude":  location.Latitude,
+				"longitude": location.Longitude,
+			}).Error; err != nil {
+				log.WithContext(ctx).Errorf("Error updating session_id:%d : %v\n", session.ID, err)
+				continue
+			}
+
+			if err := kafkaDataSyncProducer.Submit(ctx, strconv.Itoa(session.ID), &kafka_queue.Message{Type: kafka_queue.SessionDataSync, SessionDataSync: &kafka_queue.SessionDataSyncArgs{SessionID: session.ID}}); err != nil {
+				log.WithContext(ctx).Errorf("Error syncing session to Clickhouse - session_id:%d : %v\n", session.ID, err)
+			}
```
ccschmitz, line 67: @mayberryzane does this look right to you for submitting the sessions to the datasync queue? I wasn't seeing this working locally, but wasn't sure if the datasync queue maybe doesn't work the same in dev?

backend/migrations/cmd/backfill-geolocation-on-sessions/main.go
```diff
@@ -0,0 +1,82 @@
...
+		}
+
+		return nil
+	}
+
+	if err := db.Debug().
+		Joins("JOIN projects AS p ON p.id = sessions.project_id").
+		Joins("JOIN all_workspace_settings AS aw ON aw.workspace_id = p.workspace_id").
+		Where("aw.store_ip = ? AND sessions.country = ? AND sessions.state = ? AND sessions.created_at >= ? AND sessions.IP != ?", true, "", "", "2024-07-19 00:00:00", "").
```
ccschmitz, line 76: @mayberryzane this query look better to you?

ccschmitz:
@mayberryzane would appreciate your eyes on this one more time and if you have any thoughts on the datasync queue logic. Thanks!

PR #9124, opened 2024-08-09
stop recording data when the highlightrun sdk uploads cannot keep up

Vadman97:
## Summary

stop recording data when the highlight.run sdk uploads cannot keep up with `PushPayload`
this typically happens due to canvas / video element recording not being able to
upload quickly enough given upload BW restrictions. stopping recording will
prevent an OOM that would otherwise happen due to garbage collection not being
able to release the canvas / video element bitmaps.

requested by a customer that is using the video element serialization feature

https://www.loom.com/share/1b49bc8236d34c58b2ba00390237b74d

minor version

no

### DIFF
--- a/.changeset/sixty-pans-tan.md
+++ b/.changeset/sixty-pans-tan.md
@@ -1,0 +1,5 @@
+---
+'highlight.run': minor
+---
+
+stop recording if pushpayload cannot keep up with uploading data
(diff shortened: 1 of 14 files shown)


ccschmitz (approved):
Nice! Did not click test, but the demo + code look great!

PR #9133, opened 2024-08-12
add constant interval bucketing, table sorting

mayberryzane:
## Summary
- adds the ability to bucket by a constant interval
- adds a click-to-sort on table columns

- clicktested locally

- no

- @julian-highlight


backend/model/model.go
```diff
@@ -1413,6 +1413,7 @@ type Graph struct {
 	GroupByKey        *string
 	BucketByKey       *string
 	BucketCount       *int
+	BucketInterval    *int
```
Vadman97, line 1416: how does this get used?

Vadman97 (approved):

backend/model/model.go
mayberryzane: wasn't being used in original PR but fixed now - this is just for saving / loading the graph's bucket interval setting

PR #9142, opened 2024-08-14
Update sign up options

SpennyNDaJets:
## Summary
On sign up, the selects are not working -> update to the new select method

Screenshot 2024-08-14 at 12 29 40 PM

1. Create a new account
- [ ] Able to select role
- [ ] Able to select team size
- [ ] Able to select where you heard about us

N/A

N/A


ccschmitz (approved):
Looks good - thank you for fixing this!

I think there is still a regression around the form validation (the selects don't seem to be required anymore), which I can look into further!

PR #9143, opened 2024-08-14
Support user search in clickhouse

SpennyNDaJets:


backend/clickhouse/migrations/000111_recreate_sessions_joined_vw.up.sql
SpennyNDaJets, line 1: Fixes issue with Fingerprint
Screenshot 2024-08-14 at 3 33 38 PM

PR #9158, opened 2024-08-15
switch to traces_sampling_new table

mayberryzane:
## Summary
- from perf testing, new table performs comparably to both old use cases
- new table requires querying by `Timestamp` and `farmHash64(TraceId)` to be performant
- for now, timestamp filter is just +/- 1 hour of the current span - can make this logic smarter in the future if necessary

- clicktested locally for traces metrics and single trace view

- will monitor perf in prod after deploying - can revert if issues

- no


SpennyNDaJets (approved):
LGTM

PR #9171, opened 2024-08-16
Fix session alert graph

SpennyNDaJets:
## Summary
Update the session alert to display a relevant bar graph instead of the confusing line graph

Other changes:
- Add information callout for session and error alerts
- remove function for sessions and error alerts
- Swap source and graph type on metric
- Fix bug where useEffect was reset locked options (switch between session -> error)

https://www.loom.com/share/38d33d69c01e4a989aa0c70485ab0a19?sid=5eca44e4-ce41-4300-8bb4-50e596c73edd

1. Make a session alert
- [ ] Displays bar graph
2. Switch between sources
- [ ] Default filters are reverted

N/A

N/A


frontend/src/pages/Alerts/AlertGraph/index.tsx
```diff
@@ -0,0 +1,150 @@
...
+	endDate,
+	selectedPreset,
+	updateSearchTime,
+}) => {
+	const { projectId } = useProjectId()
+	const sessionsProduct = productType === ProductType.Sessions
+
+	const viewConfig = sessionsProduct
+		? getViewConfig('Bar chart', 'Stacked', 'Zero')
+		: getViewConfig('Line chart', 'Line', 'Zero')
```
Vadman97, line 77: if a custom has >1000 sessions per time interval of the session alert, could there be any issues with rendering that many individual bars / with the `group by LIMIT` on the backend? just wanted to check that case will work / that we shouldn't render 1 bar for all the number of sessions

Vadman97 (approved):

PR #9198, opened 2024-08-22
Skip large s3 file downloads

SpennyNDaJets:
## Summary
A session was causing unresponsive pages due to a 50MB file download. Skip large file downloads that are larger than 32MB. This was due to a large network resource file, as session recording are chunked.

N/A - local uses file system client

1. Test on broken sessiomn
2. Ensure other parts are not broken
3. Create card to look into breaking down network resources

N/A

### DIFF
--- a/backend/storage/storage.go
+++ b/backend/storage/storage.go
@@ -54,7 +54,7 @@ var (
 const (
 	MIME_TYPE_JSON          = "application/json"
 	CONTENT_ENCODING_BROTLI = "br"
-	MAX_DOWNLOAD_SIZE       = 32 * 1024 * 1024 // 32GB
+	MAX_DOWNLOAD_SIZE       = 32 * 1024 * 1024 // 32MB
 )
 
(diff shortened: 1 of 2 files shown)


backend/storage/storage.go
```diff
@@ -54,6 +54,7 @@ var (
 const (
 	MIME_TYPE_JSON          = "application/json"
 	CONTENT_ENCODING_BROTLI = "br"
+	MAX_DOWNLOAD_SIZE       = 10 * 1024 * 1024 // 10MB
```
Vadman97, line 57: makes sense to do this for live sessions, though this might be a bit of a low limit. wonder if we should do something in the UI to inform the user of why the live mode session might be not showing the network?

Vadman97 (approved):
description says gb :) should be mb

SpennyNDaJets:
Closing in favor of https://github.com/highlight/highlight/pull/9209

PR #9209, opened 2024-08-23
Limit size of fetched resources

SpennyNDaJets:

### DIFF
--- a/backend/private-graph/graph/schema.resolvers.go
+++ b/backend/private-graph/graph/schema.resolvers.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/DmitriyVTitov/size"
 	"github.com/PaesslerAG/jsonpath"
 	mpeTypes "github.com/aws/aws-sdk-go-v2/service/marketplaceentitlementservice/types"
 	"github.com/aws/aws-sdk-go-v2/service/marketplacemetering"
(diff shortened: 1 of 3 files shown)


backend/private-graph/graph/schema.resolvers.go
```diff
@@ -5815,6 +5816,17 @@ func (r *queryResolver) Resources(ctx context.Context, sessionSecureID string) (
...
 
+	resourceSize := size.Of(resources)
+	log.WithContext(ctx).WithFields(
+		log.Fields{
+			"size":            resourceSize,
+			"sessionSecureID": sessionSecureID,
+		}).Info("[Resources] Fetched resources size")
+
+	if resourceSize > MaxDownloadSize {
+		return nil, fmt.Errorf("resource size (%v) exceeds max download size", resourceSize)
```
Vadman97, line 5827: could we check that the frontend shows the existing error popup? 
https://github.com/highlight/highlight/blob/7c3431b68472e118217cd49b5215e2a039b5eaed/frontend/src/pages/Player/Toolbar/DevToolsWindowV2/NetworkPage/NetworkPage.tsx#L501-L521

Vadman97 (approved):

backend/private-graph/graph/schema.resolvers.go
SpennyNDaJets: Updated

Screenshot 2024-08-27 at 10 25 07 AM

PR #9212, opened 2024-08-24
Add function to filter quickStartContent and use in getStaticProps

ysadamt:
## Summary
Filter out unnecessary Quick Start documentation content with regex, reducing ISR payloads (~50kb to ~15kb)

Fixes GRO-118

|Before|After|
|--|--|
|image|image|

Click through each quick start doc page

No

No

### DIFF
--- a/highlight.io/pages/docs/[[...doc]].tsx
+++ b/highlight.io/pages/docs/[[...doc]].tsx
@@ -491,7 +491,7 @@ export const getStaticProps: GetStaticProps<DocData> = async (context) => {
 						mdxOptions: {
 							remarkPlugins: [remarkGfm],
 						},
-					})
+				  })
 				: null,
 			markdownTextOG: newContent,
 			slug: currentDoc.simple_path,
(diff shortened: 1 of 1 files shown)


highlight.io/pages/docs/[[...doc]].tsx
```diff
@@ -326,6 +326,34 @@ interface TocEntry {
...
+		const filteredQuickStartContent = quickStartContentMatches.reduceRight(
+			(obj, key) => ({ [key]: obj }),
+			getFilteredValue(quickStartContent, quickStartContentMatches),
+		)
+
+		return filteredQuickStartContent
+	} else {
+		return quickStartContent
+	}
+}
```
Comment (Vadman97) [inline highlight.io/pages/docs/[[...doc]].tsx:355]: looks like this works, nice! couldn't you also `JSON.parse(newContent)` and just build the object based on the keys that are required? that might be a bit more robust and simpler

Vadman97 (approved):

highlight.io/pages/docs/[[...doc]].tsx
Comment (ysadamt) [reply highlight.io/pages/docs/[[...doc]].tsx:355]: since `newContent` is in the format of 
```
<QuickStart content={quickStartContent["client"]["js"]["react"]}/>
```
I believe `JSON.parse` in this context would give an error?

Comment (Vadman97) [reply highlight.io/pages/docs/[[...doc]].tsx:355]: ohh my bad, you're using the regex to capture `quickStartContent["client"]["js"]["react"]` and then using the `getFilteredValue` and reduction to get the filtered object. makes sense!

PR #9213, opened 2024-08-26
session queries - join only relevant fields at query time

mayberryzane:
## Summary
- only load data for fields involved in the query to reduce the amount of data
- uses same format as existing `SessionAttributePairs` column - array of tuple(Key, Value) so filters don't change

- tested performance by querying against prod - pretty large speedup especially for short time range queries
- clicktested sessions metrics and sessions list locally

- no, can revert if performance or correctness issues

- no


backend/clickhouse/query.go
```diff
@@ -771,11 +771,20 @@ func (client *Client) ReadMetrics(ctx context.Context, input ReadMetricsInput) (
 		return nil, err
 	}
 
+	// Base fields for sessions queries, needed if body filter is used - should clean this up, maybe make these top-level?
+	attributeFields := []string{"email", "device_id"}
```
SpennyNDaJets, line 775: I noticed if I do a query `email EXISTS` or `email=spencer@highlight.io` that it returns no sessions. Not sure if that is intended?

SpennyNDaJets (commented):
Just one possible regression below, but otherwise looks good

backend/clickhouse/query.go
mayberryzane: should be fixed now! was an issue with the sessions list, tested with email and other fields

backend/clickhouse/errors.go
```diff
@@ -164,74 +163,6 @@ func (client *Client) WriteErrorObjects(ctx context.Context, objects []*model.Er
 	return nil
 }
 
-func getErrorQueryImplDeprecated(tableName string, selectColumns string, query modelInputs.ClickhouseQuery, projectId int, groupBy *string, orderBy *string, limit *int, offset *int) (string, []interface{}, error) {
```
SpennyNDaJets, line 167: Just check if these are still needed for external APIs (e.g. Grafana)?

SpennyNDaJets (approved):
Code and click test LGTMPS

backend/clickhouse/errors.go
mayberryzane: yeah I'm not 100% sure - there were some issues with the tests though so I'm just separating the deprecated code removal into its own PR to avoid delaying things

PR #9224, opened 2024-08-28
Create event search tables in clickhouse

SpennyNDaJets:
## Summary
Create the tables to start collect session events and its respective keys and values.
- `session_events` table to insert aside session `fields`
- `session_events_vw` to combine `session_events` with relevant `sessions` data
- `event_keys` and `event_key_values` to collect recommendations for search
- `materialized views to start inserting into `event_keys` and `event_key_values`

Build passes

N/A

N/A


backend/clickhouse/migrations/000111_create_session_events_table.up.sql
```diff
@@ -0,0 +1,15 @@
...
+    ProjectID Int32,
+    SessionID Int64,
+    SessionCreatedAt DateTime64(6),
+    Timestamp DateTime64(6),
+    Event String,
+    Attributes Map(LowCardinality(String), String)    
+) ENGINE = MergeTree
+ORDER BY (
+    ProjectID,
+    SessionID,
```
mayberryzane, line 12: I think there are 2 order by clauses that would work best here, either:
1. `ORDER BY (ProjectID, Timestamp)`
2. `ORDER BY (ProjectID, SessionCreatedAt, SessionID)`

The first would be the fastest way to filter on the relevant `session_events`, since we only care about events that happened within a time range regardless of the session's start date. Because the order by doesn't match the `session` table's, it requires all sessions between (start - 4 hours, end) to be loaded into memory before joining, which could be expensive for >30 day queries on the few projects with huge amounts of sessions.
The second will be less performant for small time ranges because it will always require reading (start - 4 hours, end) session events and then filtering on timestamp. But Clickhouse does not need to load anything into memory as the rows in both tables are already sorted: https://clickhouse.com/blog/clickhouse-fully-supports-joins-full-sort-partial-merge-part3#full-sorting-merge-join

backend/clickhouse/migrations/000118_write_event_identifier.up.sql
```diff
@@ -0,0 +1,18 @@
...
+    `Day` DateTime,
+    `Value` String,
+    `Count` UInt64
+) AS
+SELECT ProjectID as ProjectId,
+    'identifier' AS Key,
+    toStartOfDay(CreatedAt) AS Day,
+    Identifier AS Value,
+    count() AS Count
+FROM sessions
```
mayberryzane, line 13: for any top-level session properties, could we just look up the values from the `fields` table instead? these MVs are doing some duplicate work writing these to both locations

mayberryzane:
changes look good! some comments on the schema - lmk if you would like to discuss anything over a call!

backend/clickhouse/migrations/000111_create_session_events_table.up.sql
SpennyNDaJets: I think the second option makes sense since I assume most users would like to see this over days rather than hours

mayberryzane (approved):

PR #9225, opened 2024-08-28
Support event search metrics and alerts

SpennyNDaJets:
## Summary
Support using the event search as a metric and alerting product.

https://www.loom.com/share/e956d626fe9e4375b88c3f80ac5178aa?sid=e5c758b4-79ed-4ee8-bc26-7b38fd7255bd

1. View the metrics product
- [ ] Events is listed as a source
2. Click on events
- [ ] Able to filter by recommended keys and values
- [ ] Able to create an events graph
3. View alerts
- [ ] Events is listed as a source
4. Click on events
- [ ] Able to filter by recommended keys and values
- [ ] Able to create a events alert

Only rollout to Highlight project to start, but start collecting events for other projects

N/A


backend/public-graph/graph/resolver.go
```diff
@@ -2290,6 +2327,28 @@ func (r *Resolver) AddTrackProperties(ctx context.Context, sessionID int, events
...
+			}
+
+			sessionEvents = append(sessionEvents,
+				&clickhouse.SessionEventRow{
+					Event:      eventName,
+					Timestamp:  event.Timestamp,
+					Attributes: attributes,
+				},
+			)
+
```
Vadman97, line 2351: essentially all track events will be recorded, but what about click events? would they have an event name? i'm cool with starting with just the track events, but perhaps we could report other event types as well in the future

backend/public-graph/graph/resolver.go
```diff
@@ -2290,6 +2327,28 @@ func (r *Resolver) AddTrackProperties(ctx context.Context, sessionID int, events
 				return e.New("error deserializing track event properties")
 			}
 
+			attributes := make(map[string]string)
+			for k, v := range propertiesObject {
+				attributes[k] = fmt.Sprintf("%v", v)
+			}
+
+			// make event name the event key and delete from map
+			eventName := attributes["event"]
```
Vadman97, line 2336: what is the fallback name?

Vadman97 (approved):

backend/public-graph/graph/resolver.go
SpennyNDaJets: I'll create a card to tackle the click events - may be a subset of all events that we want to include?

backend/public-graph/graph/resolver.go
SpennyNDaJets: Good call! I'll update to have a default, maybe "unknown_event", but wondering if we should log these cases since its not an ideal case

PR #9227, opened 2024-08-28
improve session events filtering logic

Vadman97:
## Summary

* Fixes filtering logic for session events that would incorrectly include events when a search term was set
* Shows comments in the session events feed supporting filtering
* Uses AND logic for multiple search terms in the events feed
* Fixes rudderanalytics attributes to avoid sending `id` term which was throwing a warning
* Improves frontend `identify` call in our app to report more attributes

https://www.loom.com/share/38ee44b7ded344f690b883ec24ca3ca2

Screenshot from 2024-08-28 12-27-45

no

no

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -73,7 +73,7 @@
 		"react-virtuoso": "^2.13.1",
 		"recharts": "^2.12.1",
 		"rrweb": "workspace:*",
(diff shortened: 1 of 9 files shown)


SpennyNDaJets (approved):
Code and click test LGTM!

PR #9229, opened 2024-08-28
Speed up yarn install on Vercel

ccschmitz:
## Summary

Only install the dependencies needed for a production build of highlight.io when installing NPM deps in Vercel.


ccschmitz:
Don't think this will actually work...

PR #9240, opened 2024-09-01
Fix livestream banner typo

ysadamt:
## Summary

- Change month from "Septemper" to "September"
- Fixes GRO-129

Visual check:
| Before | After |
|--|--|
|image|image|

No

No


ccschmitz (changes requested):
@ysadamt thank you for the fix! I think we just need to back out the changes to the `opentelemetry-sdk-workers` and `rrweb` packages and then we can merge!

Vadman97 (approved):

Vadman97:
thanks for the fix @ysadamt !

PR #9262, opened 2024-09-05
fix related trace query for session network requests

Vadman97:
## Summary

Correctly query time range for session related traces.

before
Screenshot 2024-09-05 at 15 20 11

after
image

no

no


ccschmitz (approved):
Thank you for fixing! 🙏

PR #9267, opened 2024-09-06
correct inconsistent billing overage charges

Vadman97:
## Summary

* Skip invalid billing queries from the frontend.
* Ensure workspace details are returned for billing page logic to correctly show billing interval.

image

no

no


backend/private-graph/graph/middleware.go
```diff
@@ -29,7 +29,7 @@ var (
 	workspaceTokenHandler APITokenHandler
 )
 
-var HighlightAdminEmailDomains = []string{"@highlight.run", "@highlight.io"}
+var HighlightAdminEmailDomains = []string{"@highlight.io"}
```
mayberryzane, line 32: do we have to get rid of `highlight.run` here? I've always been using that account

mayberryzane (approved):

backend/private-graph/graph/middleware.go
Vadman97: my bad, will pull that out into a separate PR (motivation is that if we ever stop paying for our highlight.run DNS, this could be problematic)

PR #9269, opened 2024-09-06
correctly update rewrites for /highlight-events proxy when existing rewrites are defined

Vadman97:
## Summary

When next.js rewrites are defined, update existing ones correctly
rather than replacing when the definition is a `Rewrite[]`.

local debugger

before
Screenshot 2024-09-06 at 18 56 50

after
Screenshot 2024-09-06 at 18 55 36

changeset

no


ccschmitz (approved):
Nice!

PR #9276, opened 2024-09-09
Add Google analytics conversion events

ccschmitz:
## Summary

Adds GA tracking for some key events.

Ran through some of the conversion events and checked that they appeared in `window.dataLayer` correctly + showed up in the event stream for tag manager on the preview deploy.

Screenshot 2024-09-10 at 8 43 32 AM

N/A

N/A

### DIFF
--- a/frontend/src/pages/Auth/SignUp.tsx
+++ b/frontend/src/pages/Auth/SignUp.tsx
@@ -87,6 +87,11 @@ export const SignUp: React.FC = () => {
 					provider: additionalUserInfo.providerId,
 				})
 
+				analytics.trackGaEvent('sign_up', {
+					email: user.email!,
+					provider: additionalUserInfo.providerId,
+				})
+
 				if (!user?.emailVerified) {
 					auth.currentUser?.sendEmailVerification()
 				}
--- a/frontend/src/util/analytics.ts
+++ b/frontend/src/util/analytics.ts
@@ -78,6 +78,7 @@ const identify = (email: string, traits?: rudderanalytics.apiObject) => {
 		},
 	])
 	hsq.push(['trackPageView'])
+
 	// `id` is a reserved keyword in rudderstack and it's recommended to use a
 	// static property for the user ID rather than something that could change
 	// over time, like an email address.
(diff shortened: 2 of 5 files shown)


frontend/src/util/analytics.ts
```diff
@@ -93,11 +94,21 @@ const page = (name: string, properties?: rudderanalytics.apiObject) => {
...
 }
 
+const trackGaEvent = (event: string, properties?: Record<string, any>) => {
+	if (window.dataLayer) {
+		window.dataLayer.push({
+			event,
+			...properties,
+		})
+	}
+}
```
Vadman97, line 104: should we just add this to the `track` function and have all track events reported as GA events? or is that too noisy

frontend/src/pages/Auth/SignUp.tsx
```diff
@@ -87,6 +87,11 @@ export const SignUp: React.FC = () => {
 					provider: additionalUserInfo.providerId,
 				})
 
+				analytics.trackGaEvent('sign_up', {
+					email: user.email!,
+					provider: additionalUserInfo.providerId,
```
Vadman97, line 92: any other user attributes we couldsend in this obj? maybe their name? workspace id if it's available at this moment?

frontend/src/util/analytics.ts
Vadman97, line 1: last thing we want to also `track` is the demo page email submissions

Vadman97 (approved):

frontend/src/pages/Auth/SignUp.tsx
ccschmitz: I don't think we have any of that yet because all they've submitted is their email and password 🤔

frontend/src/util/analytics.ts
ccschmitz: Yeah, I wasn't sure about that. I remember being much more strict about sending new events into GA vs product analytics tools, so I'm inclined not to do that yet, but maybe something we should ask GR0 about. I added it to a doc I'll reference for our chats with them here: https://docs.google.com/document/d/1F2-kB-HScsfgEVe9Ou-PlEc_szGcdReRocOKecHWlGU/edit?usp=sharing

PR #9286, opened 2024-09-10
FIX: Deprecated Button.tsx and renamed index.tsx to Button.tsx #5753

### LINKED ISSUE #5753: Deprecate frontend/src/components/Button/Button/Button.tsx
We have `frontend/src/components/Button/index.tsx` which is the defacto way of using a button in the product (it wraps our ui library button and requires a tracking id).

However, we also have `frontend/src/components/Button/Button/Button.tsx` (imported as `@components/Button/Button/Button`) which uses ant design. This component is the first thing that shows up in vscode's cmd+p finder for "button" (the defacto one is near impossible to find given it's named `index.tsx`).

Requirements:

* Rename `frontend/src/components/Button/index.tsx` to `frontend/src/components/Button/Button.tsx`
* Remove code `frontend/src/components/Button/Button/Button.tsx` and update all references to the defacto button

poswalsameer:
## Summary

This PR fixes the issue of extra Button.tsx as mentioned in issue #5753. Removed the code file of Button.tsx and renamed index.tsx to Button.tsx, also making sure that the import statements are updated with the new Button.tsx component.

Fixes #5753


SpennyNDaJets (changes requested):
Looking good! I think there are still some references to the deleted `Button/Button/Button.tsx`. Think they are caught with the "Missing import" from reflame. Seems like the goal is to move these over to use the `Button/Button.tsx` component?

https://github.com/highlight/highlight/pull/9286/files#file-frontend-src-components-errorstate-errorstate-tsx-L12`

poswalsameer:
> Looking good! I think there are still some references to the deleted `Button/Button/Button.tsx`. Think they are caught with the "Missing import" from reflame. Seems like the goal is to move these over to use the `Button/Button.tsx` component?
> 
> [https://github.com/highlight/highlight/pull/9286/files#file-frontend-src-components-errorstate-errorstate-tsx-L12`](https://github.com/highlight/highlight/pull/9286/files#file-frontend-src-components-errorstate-errorstate-tsx-L12%60)

Thanks for replying. I will surely look into the codebase and will see if there's still something left out. Will try to do it asap.

PR #9290, opened 2024-09-10
Swap console logs table

SpennyNDaJets:
## Summary
Swap out the console logs table with the same table used for logs

https://www.loom.com/share/70f8babb2519414b96aee6c7e6d9e196

1. View a sessions logs
- [ ] Able to change search of logs
- [ ] Able to load more logs (if applicable)
- [ ] Able to expand logs
- [ ] Able to fetch in log viewer
- [ ] Able to autoscroll

Keep an eye on responsiveness

N/A


frontend/src/pages/Player/Toolbar/DevToolsWindowV2/ConsolePage/ConsolePage.tsx
```diff
@@ -1,96 +1,92 @@
...
+				const { scrollHeight, scrollTop, clientHeight } =
+					containerRefElement
 
-		if (filter !== '') {
-			return data.sessionLogs.filter((logEdge) => {
-				if (!logEdge.node.message) {
-					return false
+				if (scrollHeight - scrollTop - clientHeight < 100) {
+					fetchMoreForward()
 				}
```
Vadman97, line 63: makes sense, and with the logs results being driven by the query, searching for something that matches results towards the end of the logs would be good since the search would be server side. something to implement for the network request trace search with the new table

Vadman97 (approved):
looks good!

frontend/src/pages/Player/Toolbar/DevToolsWindowV2/ConsolePage/ConsolePage.tsx
```diff
@@ -115,139 +111,21 @@ export const ConsolePage = ({
...
-							}}
-						/>
-					)}
-				/>
-			) : (
-				<Box p="8" height="full">
-					<EmptyDevToolsCallout kind={Tab.Console} filter={filter} />
-				</Box>
-			)}
+			<ConsoleTable
```
Vadman97, line 116: just did a click test and this does seem quite laggy when resizing, but i'm also concerned about the performance when playing with autoscrolling. will do some more performance profiling rn - let's wait to merge this

Vadman97 (changes requested):

frontend/src/pages/Player/Toolbar/DevToolsWindowV2/ConsolePage/ConsolePage.tsx
Vadman97: Screenshot from 2024-09-24 10-37-03

looks like just playing a session, even with the dev tools minimized and auto scroll turned off, we aren't able to keep up with the time state updates (which happen every 100ms). we'll need to make sure the new logs page can render faster

Vadman97: Screenshot from 2024-09-24 10-53-59
looks like the console page rendering is roughly 30% of all player rendering time

Vadman97: ok... just tested on main and it seems like we have the same not great performance, so i think it's safe to say that the issue isn't as a result of this change
Screenshot from 2024-09-24 10-59-59

Vadman97 (approved):

PR #9299, opened 2024-09-12
Fix log page ordering

SpennyNDaJets:
## Summary
There is a race condition that causing use to double fetch logs on the infinite scroll. Fix this by adding a debounce and cleaning up some of the outdated code

https://www.loom.com/share/be2c80f36b6140deac8e384651168989

1. Log the logs page
2. Scroll to the bottom of the table
- [ ] Only one request sent
- [ ] No duplicate logs in table
4. Repeat for traces
- [ ] Only one request sent
- [ ] No duplicate traces in table

N/A

N/A

### DIFF
--- a/frontend/src/pages/LogsPage/useGetLogs.ts
+++ b/frontend/src/pages/LogsPage/useGetLogs.ts
@@ -161,38 +161,43 @@ export const useGetLogs = ({
 
 	// eslint-disable-next-line react-hooks/exhaustive-deps
 	const fetchMoreForward = useCallback(
-		debounce(async () => {
-			const { hasNextPage, endCursor } = data?.logs.pageInfo || {}
-			if (!hasNextPage || loadingAfter || !endCursor) {
-				return
-			}
+		debounce(
+			async () => {
+				const { hasNextPage, endCursor } = data?.logs.pageInfo || {}
+				if (!hasNextPage || loadingAfter || !endCursor) {
+					return
+				}
 
-			setLoadingAfter(true)
+				setLoadingAfter(true)
 
-			await fetchMore({
-				variables: { after: endCursor },
-				updateQuery: (prevResult, { fetchMoreResult }) => {
-					return {
-						logs: {
-							...prevResult.logs,
-							edges: [
-								...prevResult.logs.edges,
-								...fetchMoreResult.logs.edges,
-							],
-							pageInfo: {
-								...prevResult.logs.pageInfo,
-								hasNextPage:
-									fetchMoreResult.logs.pageInfo.hasNextPage,
-								endCursor:
-									fetchMoreResult.logs.pageInfo.endCursor,
+				await fetchMore({
+					variables: { after: endCursor },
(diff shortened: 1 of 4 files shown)


frontend/src/pages/LogsPage/useGetLogs.ts
```diff
@@ -36,41 +30,35 @@ export const useGetLogs = ({
...
-	useEffect(() => {
-		setWindowInfo(initialWindowInfo)
-	}, [query, startDate, endDate])
 
 	const { data, loading, error, refetch, fetchMore } = useGetLogsQuery({
 		variables: {
 			project_id: project_id!,
 			at: logCursor,
-			direction: Types.SortDirection.Desc,
+			direction: sortDirection,
```
SpennyNDaJets, line 51: This is tangental but needed to swap over the console log table which is ordered by timestamp ASC

frontend/src/pages/Traces/TracesPage.tsx
```diff
@@ -108,7 +108,7 @@ export const TracesPage: React.FC = () => {
 		startDate: searchTimeContext.startDate,
 		endDate: searchTimeContext.endDate,
 		skipPolling,
-		sortColumn,
+		sortColumn: sortColumn || undefined,
```
ccschmitz, line 111: Nit, but isn't this equivalent to `sortColumn,`?

frontend/src/pages/Traces/useGetTraces.ts
```diff
@@ -153,74 +133,42 @@ export const useGetTraces = ({
...
-	const fetchMoreBackward = useCallback(async () => {
-		if (!windowInfo.hasPreviousPage || loadingBefore) {
-			return
-		}
-
-		const firstItem = data && data.traces.edges[0]
-		const firstCursor = firstItem?.cursor
+	// eslint-disable-next-line react-hooks/exhaustive-deps
+	const fetchMoreForward = useCallback(
+		debounce(async () => {
```
ccschmitz, line 138: Noticed a delay here after hitting the bottom of the page, I think because we wait for the 500ms before executing the function. What do you think about setting `leading: true` in the options so it is invoked immediately when hit the first time?

https://lodash.com/docs/4.17.15#debounce

ccschmitz (approved):
One question about a config option for `debounce`, but code + click test looks good otherwise!

frontend/src/pages/Traces/TracesPage.tsx
SpennyNDaJets: Technically it can be `null` so this was just to be type happy, but open to other suggestions

frontend/src/pages/Traces/useGetTraces.ts
SpennyNDaJets: Good catch! Lets do that

SpennyNDaJets: Just for history, had to break out the request into its own callback to avoid the function being recreated when the state changes, resulting in a race condition and duplicate request sent

PR #9300, opened 2024-09-12
Don't record body of OTeL network requests

ccschmitz:
## Summary

Updates the logic for filtering out requests to be ignored so it does not record requests to the OTeL collector. This will keep OTeL requests out of the Network panel, similar to what we do for requests to `pub.highlight.io`.

Local preview click test to ensure that OTeL endpoint requests are:

* Not showing up in the network panel by default
* Showing up in the network panel when the OTeL value is in `tracingOrigins`

Should notify any users who were requesting this modification that it went live.

N/A


sdk/client/src/listeners/first-load-listeners.tsx
```diff
@@ -155,10 +155,15 @@ export class FirstLoadListeners {
 		sThis: FirstLoadListeners,
 		options: HighlightClassOptions,
 	): void {
-		sThis._backendUrl =
+		const _backendUrl =
 			options?.backendUrl ||
 			import.meta.env.REACT_APP_PUBLIC_GRAPH_URI ||
 			'https://pub.highlight.run'
+		const otlpEndpoint =
+			options.otlpEndpoint ?? 'https://otel.highlight.io:4318'
```
ccschmitz, line 163: @Vadman97 is this correct to include the port on this URL, or can/should we leave it off?

Vadman97: hmm probably better to leave it off, since https://otel.highlight.io (port 443) is also valid

Vadman97 (approved):
Screenshot from 2024-09-26 22-02-20

tested locally - private graph recorded, but not otel / public graph

sdk/client/src/listeners/first-load-listeners.tsx
```diff
@@ -155,10 +155,14 @@ export class FirstLoadListeners {
 		sThis: FirstLoadListeners,
 		options: HighlightClassOptions,
 	): void {
-		sThis._backendUrl =
+		const _backendUrl =
 			options?.backendUrl ||
 			import.meta.env.REACT_APP_PUBLIC_GRAPH_URI ||
 			'https://pub.highlight.run'
+		const otlpEndpoint = options.otlpEndpoint ?? 'https://otel.highlight.io'
```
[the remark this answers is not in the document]
ccschmitz: Yeah, good point - thanks!

PR #9302, opened 2024-09-12
Increase session keys ttl

SpennyNDaJets:
## Summary
Customers with more than 30 day session retention will not see any key recommendations when looking at dates older than 30 days. Update the `session_keys` ttl to be 3 years. Backfill and swap to be completed tomorrow in https://github.com/highlight/highlight/pull/9314

Note: future improvement is to dynamically set ttl by project.

1. Run migrations
- [ ] Tables created
- [ ] Data backfilled

Merge in day before https://github.com/highlight/highlight/pull/9314

n/A


mayberryzane (approved):
LGTM!

PR #9303, opened 2024-09-12
Only alert OPEN errors

SpennyNDaJets:
## Summary
Error alerts send notifications for errors that are ignored. Add a default filter to only alert OPEN errors.

Note: Resolved errors should be reopened in a different process.

1. Create an error alert
2. Cause an error
- [ ] Notification received
3. Ignore the error
4. Cause the error again
- [ ] No notification received

N/A

N/A

### DIFF
--- a/backend/jobs/metric-alerts/metric-alerts.go
+++ b/backend/jobs/metric-alerts/metric-alerts.go
@@ -28,7 +28,7 @@ const maxWorkers = 40
 const alertEvalFreq = time.Minute
 
 var defaultAlertFilters = map[modelInputs.ProductType]string{
-	modelInputs.ProductTypeErrors: "status=OPEN",
+	modelInputs.ProductTypeErrors: "status=OPEN ",
 }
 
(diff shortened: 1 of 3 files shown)


backend/jobs/metric-alerts/metric-alerts.go
```diff
@@ -27,6 +27,10 @@ import (
 const maxWorkers = 40
 const alertEvalFreq = time.Minute
 
+var defaultAlertFilters = map[modelInputs.ProductType]string{
+	modelInputs.ProductTypeErrors: "status=OPEN ",
```
Vadman97, line 31: wonder what the expectation is with folks using alerting, perhaps something we can add to the UI to make it clear that error alerts by default only trigger for open errors?

Vadman97 (approved):

backend/jobs/metric-alerts/metric-alerts.go
SpennyNDaJets: I can add that to the info - I think this does keep it in line with the old alerts: https://github.com/highlight/highlight/blob/main/backend/public-graph/graph/resolver.go#L1798

PR #9304, opened 2024-09-13
Number input behavior tweaks

ccschmitz:
## Summary

Tweaks the behavior of the number input. Some changes were at the component level and others in the logic where the component was being used.

* Changes the `type` of the input control buttons for incremeneting/decrementing the number so that they don't try to submit the form.
* Adds functionality to increment/decrement by 10x when the `shift` key is held while clicking the buttons.
* Fixes the color of the button icons and makes them match the mocks + the icons used in select dropdowns.
* Updates logic in the graph editor to not convert to an integer before setting the value that controls the input. This allows the user to clear the input entirely rather than it always showing a `0` for empty values.

Click tested locally and in a PR preview on the graph editor.

N/A

Will have @julian-highlight review, but the original changes were per his request.


SpennyNDaJets (approved):

PR #9314, opened 2024-09-16
Swap session keys table

SpennyNDaJets:
## Summary
Swap over to the new sessions keys table

N/A

✅  Run migration manually in clickhouse before merge

N/A


backend/clickhouse/migrations/000121_swap_session_keys_table.up.sql
```diff
@@ -0,0 +1,22 @@
...
+GROUP BY ProjectId,
+    Key,
+    Day,
+    Type;
+
+DROP VIEW session_keys_mv;
+DROP TABLE session_keys;
+
+RENAME TABLE session_keys_new_mv TO session_keys_mv;
+RENAME TABLE session_keys_new TO session_keys;
```
SpennyNDaJets, line 22: Looks like this results in the following error:
```
Target table 'test.session_keys_new' of view 'test.session_keys_mv' doesn't exists. To ignore this view use setting ignore_materialized_views_with_dropped_target_table: While executing WaitForAsyncInsert
```
Should we avoid renming and update the references to use `session_keys_new`?

mayberryzane: I don't know if you should do this as a migration file because running this will probably take a long time because of the insert. Would probably be better to run directly in prod. You can probably set the `ignore_materialized_views_with_dropped_target_table` setting and it should be fine. Although you might have to recreate it anyway if the table name is changing (since the materialized view is referencing the "_new" table)

mayberryzane (approved):
file looks good to me w/ small fixes to the `rename`. think this script should be ran directly against prod and not in the migrations though

PR #9318, opened 2024-09-16
chunk network resources

Vadman97:
## Summary

Updates the session dev tools `network` page to load data from `Traces`.
Updates the Resource context to use traces to load session network data across the product.
Doesn't yet switch to the new traces table to render network resources, but should make it easy to do so in the future.

* tested with otel frontend traces
* tested with frontend network requests recorded as traces

https://www.loom.com/share/3a252f8576e84405aa5ac6ded29868dd

monitoring in prod, will follow up to clean up network request files

no

### DIFF
--- a/__generated/rr/rrweb/rr.js
+++ b/__generated/rr/rrweb/rr.js
@@ -1,17 +1,11 @@
 // ../rrweb/packages/rrweb/dist/rrweb.js
 var __defProp = Object.defineProperty;
 var __defNormalProp = (obj, key, value) => key in obj ? __defProp(obj, key, { enumerable: true, configurable: true, writable: true, value }) : obj[key] = value;
-var __publicField = (obj, key, value) => {
-  __defNormalProp(obj, typeof key !== "symbol" ? key + "" : key, value);
-  return value;
-};
+var __publicField = (obj, key, value) => __defNormalProp(obj, typeof key !== "symbol" ? key + "" : key, value);
 var _a;
 var __defProp$1 = Object.defineProperty;
 var __defNormalProp$1 = (obj, key, value) => key in obj ? __defProp$1(obj, key, { enumerable: true, configurable: true, writable: true, value }) : obj[key] = value;
-var __publicField$1 = (obj, key, value) => {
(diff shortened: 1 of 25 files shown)


__generated/rr/rrweb/rr.js
SpennyNDaJets, line 1: Are these changes anticipated?

SpennyNDaJets (approved):
Click test looks good to me!

SpennyNDaJets (changes requested):
One thing I noticed is that we are still only returning 50 traces? Looks like the pagination is based on `1_000` but the cursor is not respecting that limit

frontend/src/pages/Player/ResourcesContext/ResourcesContext.tsx
```diff
@@ -98,112 +166,48 @@ const buildResources = (resources: NetworkResource[]) => {
...
+		query: `secure_session_id=${session?.secure_id} AND highlight.type=http.request`,
+		projectId,
+		startDate: moment(session?.created_at).toDate(),
+		endDate: moment(session?.created_at).add(4, 'hours').toDate(),
+		traceCursor: undefined,
+		sortDirection: SortDirection.Asc,
+		sortColumn: 'timestamp',
+		skip: !session?.secure_id,
+		skipPolling: true,
+		limit: 1_000,
```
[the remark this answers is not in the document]
Vadman97: yup meant to say 1000

__generated/rr/rrweb/rr.js
Vadman97: unrelated, reverting

Vadman97:
> One thing I noticed is that we are still only returning 50 traces? Looks like the pagination is based on `1_000` but the cursor is not respecting that limit -> think we just need to pass it into `getConnection`

thanks for the callout @SpennyNDaJets and for the help with the fix. just updated this!

Vadman97:
new testing https://www.loom.com/share/8016f403dd6a4ac08a1a61886c8e187d

SpennyNDaJets (approved):
LGTM

PR #9335, opened 2024-09-17
Add docs on pointing OTeL browser instrumentation to Highlight

ccschmitz:
## Summary

* Adds a page explaining how to point the OTeL browser instrumentation directly to Highlight's OTeL collector.
* Updates the client API docs to cover `startSpan` and `startManualSpan`.
* Updates the logic for finding the related session on the trace panel so that it works on `documentLoad` spans which don't have the session ID on the root level span (this is random - just noticed while going over a YouTube video w/ @jay-khatri).

Click tested the docs to verify the new page looks good and ran a quick test in an E2E app to point traces to Highlight to ensure instructions were correct.

N/A

N/A

### DIFF
--- a/docs-content/getting-started/7_native-opentelemetry/5_browser.md
+++ b/docs-content/getting-started/7_native-opentelemetry/5_browser.md
@@ -14,8 +14,8 @@ If you want to send OpenTelemetry data to Highlight from [OTeL's Browser instrum
 
 In order to associate spans with your project and sessions in Highlight, you'll need to add a few attributes:
 
-* `highlight.project_id` - Allows us to associate traces with your project.
-* `highlight.session_id` - Allows us to associate traces with a specific session in your project.
+* `highlight.project_id` - Allows us to associate traces with your project. This can be accessed from [your Highlight account](https://app.highlight.io/setup).
+* `highlight.session_id` - Allows us to associate traces with a specific session in your project. This can be accessed as the `sessionSecureID` attribute of the object returned from [H.getSessionData()](https://highlight.io/docs/sdk/client#HgetSessionDetails).
 * `service.name` (optional) - Specify the service name you want to assign on your traces. This makes it easier to filter to all traces from this instrumentation.
 
 You can set these attributes when creating your trace provider:
@@ -23,7 +23,6 @@ You can set these attributes when creating your trace provider:
 ```ts
 const provider = new WebTracerProvider({
   resource: new Resource({
-    // TODO: Explain how to get the project and session IDs.
     'highlight.project_id': 'your-project-id',
     'highlight.session_id': 'your-session-id',
     'service.name': 'your-service-name', // optional
--- a/docs-content/getting-started/7_native-opentelemetry/5_browser.md
+++ b/docs-content/getting-started/7_native-opentelemetry/5_browser.md
@@ -1,0 +1,46 @@
+---
+title: Browser Instrumentation
+slug: browser-otel-instrumentation
+heading: OpenTelemetry Browser Instrumentation
+metaTitle: Send OpenTelemetry Browser Instrumentation Data to Highlight
+createdAt: 2024-09-17T00:00:00.000Z
+updatedAt: 2024-09-17T00:00:00.000Z
+quickstart: true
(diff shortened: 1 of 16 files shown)


docs-content/getting-started/7_native-opentelemetry/5_browser.md
```diff
@@ -0,0 +1,47 @@
...
+* `service.name` (optional) - Specify the service name you want to assign on your traces. This makes it easier to filter to all traces from this instrumentation.
+
+You can set these attributes when creating your trace provider:
+
+```ts
+const provider = new WebTracerProvider({
+  resource: new Resource({
+    // TODO: Explain how to get the project and session IDs.
+    'highlight.project_id': 'your-project-id',
+    'highlight.session_id': 'your-session-id',
```
ccschmitz, line 28: @Vadman97 have any recommendations on how to tell people to access this? I didn't see anything that allowed you to access just the session ID, only the session URL.

ccschmitz: Ended up adding `sessionSecureID` to the object returned from `H.getSessionDetails()`. @Vadman97 let me know if you see any issues with that!

Vadman97: yeah that makes sense to me - the `H.init()` call also returns it but probably a simpler API to use the `getSessionDetails()` call

docs-content/sdk/client.md
```diff
@@ -245,10 +245,91 @@ slug: client
   </div>
 </section>
 
+<section className="section">
+  <div className="left">
+    <h3>H.startSpan</h3>
```
Vadman97, line 250: one thing i noticed with the `startSpan` and `startManualSpan` API is that it doesn't allow returning the value of the callback in some cases. should we make it clear that returning the return value of the `cb` is / is not an option?

Vadman97 (approved):
🚢

docs-content/sdk/client.md
ccschmitz: Good callout, @Vadman97! I ended up fixing that in https://github.com/highlight/highlight/pull/9335/commits/984435d30f5529a92c9dc2c694a86d59986c9ce0. Mind giving those changes a quick look? Will discard your previous review - thanks for looking this over!

sdk/firstload/src/index.tsx
```diff
@@ -380,25 +381,25 @@ const H: HighlightPublicInterface = {
 	): any => {
 		const tracer = typeof getTracer === 'function' ? getTracer() : undefined
 		if (!tracer) {
+			console.log('::: no tracer')
```
Vadman97, line 384: remove

sdk/firstload/src/index.tsx
```diff
@@ -380,25 +381,25 @@ const H: HighlightPublicInterface = {
...
-		const wrapCallback = async (
-			span: Span,
-			callback: (span: Span) => any,
-		) => {
+		const wrapCallback = (span: Span, callback: (span: Span) => any) => {
+			let result: any
 			try {
-				const result = await callback(span)
-				return result
+				result = callback(span)
```
Vadman97, line 399: if the callback is a promise, does that mean line 401 (`span.end()`) will happen before the promise is finished? seems like a regression in expected behavior

Vadman97 (approved):

docs-content/sdk/client.md
Vadman97: reviewed!

sdk/firstload/src/index.tsx
ccschmitz: Ah, you're right! Thanks for catching that. I think there was also a bug with our types for this function which should have said it is returning a promise.

ccschmitz: Just addressed this and added a test for async callbacks in https://github.com/highlight/highlight/pull/9335/commits/3099b9fe042aa312e84dbf41b3e6c3ecfdf8e46f

Vadman97 (approved):
makes sense to me

PR #9338, opened 2024-09-17
New default alerts

SpennyNDaJets:
## Summary
In the setup page, we create old default alerts for the new user if they request. Update these to be the new alerts.

Three alerts:
- New user sessions
- Error alerts
- Error logs

Additional notes
- Also removes being able to create legacy alerts.
- Removes routes from router

1. Create a new user/workspace
2. Sign up for the new alerts via the alert menu option
- [ ] Alerts created successfully in db (note that settings in UI may not be visible -> fix in https://github.com/highlight/highlight/pull/9336)
- [ ] Alert is denoted with `default: true`
- [ ] Correct destinations created
3. Create a manual alert
- [ ] Alert successfully created
- [ ] Alert is denoted with `default: false`


Vadman97 (approved):
tested locally, worked great 💪

PR #9345, opened 2024-09-18
Disable function column for count

SpennyNDaJets:
## Summary
Display the function column to help users know that there are other functions they can use

Screenshot 2024-09-19 at 11 41 19 AM

1. Create a graph
- [ ] Function column is disabled when count is selected
- [ ] Function column is enabled when other funcction selected
2. Create a log/trace alert
- [ ] Function column is disabled when count is selected
- [ ] Function column is enabled when other funcction selected

N/A

N/A


frontend/src/pages/Alerts/AlertForm/index.tsx
```diff
@@ -488,23 +488,22 @@ export const AlertForm: React.FC = () => {
...
-												/>
-											)}
+											<Combobox
+												selection={functionColumn}
+												setSelection={setFunctionColumn}
+												label="metric"
+												searchConfig={
+													searchOptionsConfig
+												}
+												disabled={
```
Vadman97, line 498: in addition to making the combox disabled, should it have some placeholder value like `Rows` to make it more obvious that we're referring to the count of results?

Vadman97 (approved):

PR #9368, opened 2024-09-24
[Ruby SDK] Improve docs and handling of environment variable

ccschmitz:
## Summary

Had a report of the Ruby SDK not emitting logs. Believe it's because the initialization code was passing `Rails.env` as the `environment` to `Highlight.init`. This does not return a string and was causing errors in the OTeL initialization preventing logs/traces/errors from being captured. This PR updates our E2E app to use `Rails.env` and calls `to_s` on that arg to ensure it's a string.

Also updates the logs to show using the newer `broadcast_to` option for sending logs to an additional logger besides Rails's.

Click tested the E2E app and confirmed OTeL data was flowing after this change.

N/A

N/A


SpennyNDaJets (approved):

e2e/ruby/rails/demo/Gemfile
```diff
@@ -6,7 +6,7 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 ruby '~> 3.3.4'
 
 # Bundle edge Rails instead: gem "rails", github: "rails/rails", branch: "main"
-gem 'rails', '~> 7.0.8', '>= 7.0.8.1'
+gem 'rails', '~> 7.1.0'
```
ccschmitz, line 9: Had to upgrade to use `broadcast_to`, which I thought might be part of the issue I was troubleshooting.

PR #9369, opened 2024-09-24
Upgrade go to v1.23.1

ccschmitz:
## Summary

Was investigating [an error I was seeing locally](https://highlightcorp.slack.com/archives/C02CJANPHQS/p1727195920289109) running the backend and saw somewhere upgrading Go might resolve the issue. It does seem to have fixed my local build, and didn't see any harm in upgrading so thought I'd propose the change.

Also updates our Go E2E app to use the same version of Go as everywhere else.

Local backend build using `make start` from `./backend` and then click tested locally. Also ran the go E2E app to ensure that worked too.

N/A

N/A

### DIFF
--- a/backend/go.mod
+++ b/backend/go.mod
@@ -1,6 +1,6 @@
 module github.com/highlight-run/highlight/backend
 
-go 1.22.2
+go 1.23.1
 
 replace golang.org/x/crypto => golang.org/x/crypto v0.17.0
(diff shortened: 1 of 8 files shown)


Vadman97 (approved):
may also need to run `go work sync` and then again run `go mod tidy` in the respective go packages

PR #9371, opened 2024-09-24
fix errors sort order

mayberryzane:
## Summary
- after clickhouse query, there was also an order by for the sql query
- remove that and replace with in-memory sort according to `LastOccurrence`

- clicktested locally

- no

- no


ccschmitz (approved):
Looks good!

PR #9374, opened 2024-09-25
Update metric event UI

SpennyNDaJets:
## Summary
Update the event UI to allow users to more directly select event

1. Create a new graph
2. Select the event source
3. Add filters
- [ ] Correctly displays graph
4. Save graph
- [ ] Correctly reloads graph

N/A

N/A

### DIFF
--- a/frontend/src/pages/Graphing/EventSelection/index.tsx
+++ b/frontend/src/pages/Graphing/EventSelection/index.tsx
@@ -1,0 +1,180 @@
+import _ from 'lodash'
+import moment from 'moment'
+import React, { useEffect, useMemo, useState } from 'react'
+import { useDebounce } from 'react-use'
+
+import { TIME_FORMAT } from '@/components/Search/SearchForm/constants'
+import { LabeledRow } from '@/pages/Graphing/LabeledRow'
+import { Box, Card, ComboboxSelect, Text } from '@highlight-run/ui/components'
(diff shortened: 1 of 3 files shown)


frontend/src/pages/Graphing/EventSelection/index.tsx
```diff
@@ -0,0 +1,180 @@
...
+					flexDirection="column"
+					gap="12"
+				>
+					<LabeledRow label="Event type" name="eventType">
+						<OptionDropdown<EventType>
+							options={EVENT_TYPES}
+							selection={eventType}
+							setSelection={setEventType}
+						/>
+					</LabeledRow>
```
Vadman97, line 146: just wanted to check here but i think the group by for events isn't working correctly. guessing unrelated to these UI changes though

https://www.loom.com/share/e1fdbf9514de4f31b00a0339fb5b4530

Vadman97 (approved):

frontend/src/pages/Graphing/EventSelection/index.tsx
SpennyNDaJets: Think we synced here - `email` is tied to the session and not the event, so its only finding events in which `email` is present in the attributes passed to the track event. We have a card here to add in the session attributes: https://linear.app/highlight/issue/HIG-4937/allow-searching-session-attributes-on-events

Vadman97 (approved):

PR #9377, opened 2024-09-26
cleanup raw events as part of session deletion job

mayberryzane:
## Summary
- enforces raw event cleanup by removing any folders in the `raw-events` directory > 1 day old
- applies only when the `SESSION_RETENTION_DAYS` env var is set

- ran locally + debugged

- no

- no

### DIFF
--- a/backend/lambda-functions/deleteSessions/handlers/handlers.go
+++ b/backend/lambda-functions/deleteSessions/handlers/handlers.go
@@ -219,7 +219,6 @@ func (h *handlers) DeleteSessions(ctx context.Context, projectId int, startDate 
 
 	if err := h.storageClient.CleanupRawEvents(ctx, projectId); err != nil {
 		log.WithContext(ctx).Error(err)
-		return
 	}
 
 	if len(batches) == 0 {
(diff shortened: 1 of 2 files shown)


backend/lambda-functions/deleteSessions/handlers/handlers.go
```diff
@@ -217,6 +217,11 @@ func (h *handlers) DeleteSessions(ctx context.Context, projectId int, startDate
 		return
 	}
 
+	if err := h.storageClient.CleanupRawEvents(ctx, projectId); err != nil {
+		log.WithContext(ctx).Error(err)
```
Vadman97, line 221: should we have some more resilient error handling here so that we still delete sessions if this fails?
Screenshot from 2024-09-26 09-44-02

Vadman97 (approved):

backend/lambda-functions/deleteSessions/handlers/handlers.go
mayberryzane: 👍 updated to catch not exists errors, removed the `return` and am just logging errors if something happens while cleaning up `raw-events`

PR #9378, opened 2024-09-26
Hide mouse cursor until replayer is ready

ccschmitz:
## Summary

Fixes an issue where the mouse would be shown initially very large and in the center of the replayer, then reposition to the top left corner once the replayer's size was adjusted. This PR hides the cursor until we get a mouse event that which contains positioning data for the mouse to display it in the correct spot on the replayer.

* On production, view a session and click around playing different sessions. Note the flicker on the replayer where you see the cursor incorrectly positioned briefly.
* Perform the same test on the PR preview and ensure the behavior looks better.

N/A

N/A

### DIFF
--- a/frontend/src/__generated/index.css
+++ b/frontend/src/__generated/index.css
@@ -6059,6 +6059,9 @@ body {
   position: relative;
   width: 100%;
 }
+.hide-mouse-cursor .replayer-mouse {
+  display: none;
+}
 ._1gpgayi5 iframe {
   border-radius: 4px;
   border: var(--_1pyqka9k) solid 1px;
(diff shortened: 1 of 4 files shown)


ccschmitz:
@jay-khatri @julian-highlight pinging in case you want to click test and ensure this is the right UX.

mayberryzane (approved):

ccschmitz:
Was chatting w/ @Vadman97 and he mentioned https://github.com/highlight/highlight/pull/8118/files#r1622775292 - need to investigate this.

ccschmitz:
@Vadman97 check out the logic update in https://github.com/highlight/highlight/pull/9378/commits/64fa2b382c0ac7ffd74254cd752621e3f9178bac. I didn't revert to the previous behavior, but I think this is much cleaner than what we were doing trying to multiply values to get to the desired scale.

https://www.loom.com/share/bcdce0e152a34c02a060cf5efdf21db9

PR #9400, opened 2024-09-30
Use new event key recommendations

SpennyNDaJets:
## Summary
Pass in the event name to be able to get event specific key and value recommendations.

https://www.loom.com/share/17eb2f77a4ed4ab4aa24a9b090735b80?sid=66db46e6-550e-41d8-bb1a-5a777e95e370

1. Create a new event graph
2. Select an event
3. Add filters
- [ ] Event specific keys are provided
- [ ] Generic keys are provided
- [ ] Event specific values are provided
- [ ] Generic values are provided
4. Make a graph with logs
- [ ] Fetch keys correctly
- [ ] Fetch values correctly
5. Search traces
- [ ] Fetch keys correctly
- [ ] Fetch values correctly

N/A

N/A

### DIFF
--- a/backend/clickhouse/events.go
+++ b/backend/clickhouse/events.go
@@ -17,8 +17,8 @@ import (
 
 const SessionEventsTable = "session_events"
 const SessionEventsView = "session_events_vw"
-const EventKeysTable = "event_keys"
-const EventKeyValuesTable = "event_key_values"
+const EventKeysTable = "event_keys_new"
+const EventKeyValuesTable = "event_key_values_new"
 
 var eventKeysToColumns = map[string]string{
 	string(modelInputs.ReservedEventKeyBrowserName):         "BrowserName",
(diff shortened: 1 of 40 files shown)


SpennyNDaJets:
Note: elected to switch over to new tables rather than trying to swap and recreate materialized views, but happy to discuss

backend/clickhouse/query.go
```diff
@@ -338,7 +344,7 @@ func KeysAggregated(ctx context.Context, client *Client, tableName string, proje
 	return keys, rows.Err()
 }
 
-func KeyValuesAggregated(ctx context.Context, client *Client, tableName string, projectID int, keyName string, startDate time.Time, endDate time.Time, query *string, limit *int) ([]string, error) {
+func KeyValuesAggregated(ctx context.Context, client *Client, tableName string, projectID int, keyName string, startDate time.Time, endDate time.Time, query *string, limit *int, additionalWhereClauses *[]string) ([]string, error) {
```
SpennyNDaJets, line 347: Opted to pass in the where clauses as a string since `event` is specific to one search. However, may be better to pass in another way?

backend/clickhouse/events.go
```diff
@@ -161,12 +166,17 @@ func (client *Client) EventsKeys(ctx context.Context, projectID int, startDate t
...
-func (client *Client) EventsKeyValues(ctx context.Context, projectID int, keyName string, startDate time.Time, endDate time.Time, query *string, limit *int) ([]string, error) {
+func (client *Client) EventsKeyValues(ctx context.Context, projectID int, keyName string, startDate time.Time, endDate time.Time, query *string, limit *int, event *string) ([]string, error) {
 	if eventBooleanKeys[keyName] {
 		return []string{"true", "false"}, nil
 	}
 
-	return KeyValuesAggregated(ctx, client, EventKeyValuesTable, projectID, keyName, startDate, endDate, query, limit)
+	whereClauses := []string{}
+	if event != nil && *event != "" {
+		whereClauses = append(whereClauses, fmt.Sprintf("Event = %s OR Event = ''", *event))
```
Vadman97, line 176: just checking that this is not vulnerable to sql injection since the `event` is unchecked user input, but i think the `whereClauses` in `KeyValuesAggregated` is checked by `sb.Var`

Vadman97 (approved):

backend/clickhouse/events.go
SpennyNDaJets: Good catch! Updated to pass in the event into `KeysAggregated` and handle it in there with `sb.Var`

backend/clickhouse/query.go
SpennyNDaJets: Updated to pass in the `event` into `KeysAggregated` and handle it in there with `sb.Var` to avoid sql injection converns

PR #9405, opened 2024-10-01
WordPress plugin, E2E app, and docs

ccschmitz:
## Summary

Creates a WordPress plugin for configuring Highlight session recording.

image

Created an E2E example app and click tested the plugin there. Ensured recordings were sent correctly to my local instance of Highlight and that configuration for the client recordings could be controlled properly from the plugin options page.

We will need to submit the plugin to [the WordPress plugin directory](https://wordpress.org/plugins/).

N/A


sdk/highlight-wordpress/highlight-io/readme.txt
Vadman97, line 1: could we add a note in this file on how someone would zip and upload the package when we need to update it?

docs-content/general/7_integrations/wordpress-integration.md
```diff
@@ -0,0 +1,68 @@
...
+- **Easy Setup**: Install and configure the plugin in just a few clicks.
+- **Session Recording**: Capture user interactions on your WordPress site.
+- **Error Monitoring**: Automatically detect and log errors with all the information you need to troubleshoot.
+- **Network Recording**: Record network requests for deeper insights.
+- **Console Log Recording**: Capture and review console logs to help identify and fix issues quickly.
+
+## Installation
+
+1. Log in to your WordPress admin panel.
+2. Go to "Plugins" > "Add New".
```
Vadman97, line 23: maybe here we could link to the wordpress plugin list?

Vadman97 (approved):
looks great! what dictates the backlink we get from the wordpress domain? is there some doc we control or just their plugin registry makes a link back to us?

sdk/highlight-wordpress/highlight-io/readme.txt
ccschmitz: Sure thing!

ccschmitz:
> what dictates the backlink we get from the wordpress domain

I believe the plugin directory uses the links we give it here on the plugin details page. e.g. the link back to Highlight would be next to the avatar, under the plugin title on this page: https://wordpress.org/plugins/easy-pdf-restaurant-menu-upload/

docs-content/general/7_integrations/wordpress-integration.md
ccschmitz: I think clicking "Add New" will give them some UI to explore in the directory.

PR #9409, opened 2024-10-02
Fix trace redirection loop

SpennyNDaJets:
## Summary
The traces page on demo is looping in redirects causing lots of network requests to be sent and cancelled. Fix the use effect to no longer loop.

1. Visit the demo traces page (`/demo/traces`)
- [ ] No redirection loops

N/A

N/A


ccschmitz (approved):
Thank you!!

PR #9412, opened 2024-10-03
sample sessions by user when possible

Vadman97:
## Summary

Update session sampling logic to use the user identifier as the key when possible to be able to capture all sessions of a given user

unit test

no, single customer using session sampling

no


SpennyNDaJets (approved):
LGTM!

Related question: are there any concerns with sampling of sessions affecting downstream metrics (i.e. events)?

Vadman97:
@SpennyNDaJets 
> Related question: are there any concerns with sampling of sessions affecting downstream metrics (i.e. events)?

shouldn't be, because user events are still ingested for an excluded session. the only sessions that are dropped client-side are outside-billing-quota sessions

PR #9414, opened 2024-10-03
fix search suggestions when no defaultValueOptions provided

mayberryzane:
## Summary
- when `defaultValueOptions` was undefined, was skipping values entirely

- clicktested locally

- no

- no


Vadman97 (approved):
preview looks great!

PR #9418, opened 2024-10-03
Un-override header hierarchy on documentation pages

ysadamt:
## Summary

- Use normal header hierarchy instead of overwritten headings using custom classes
- Instead of overwriting `globals.scss` styles which breaks some of the stylings on the landing
- Fixes GRO-122

- Check heading sizes on localhost compared to hosted version

No

No


highlight.io/pages/docs/[[...doc]].tsx
```diff
@@ -1033,7 +1033,10 @@ export default function DocPage({
 												EnterpriseSelfHostCalendlyComponent,
 												DocsCardGroup,
 												h1: (props) => (
-													<h4 {...props} />
+													<h1
+														{...props}
+														className="docH1"
+													/>
 												),
```
Comment (Vadman97) [inline highlight.io/pages/docs/[[...doc]].tsx:1040]: looks like this might be working, but the main titles on pages are still h3 tags whereas we want them to be h1s

Screenshot from 2024-10-04 17-52-53

Vadman97 (changes requested):

Vadman97 (approved):

PR #9420, opened 2024-10-04
use new select component, enable multiselect / multi-group

mayberryzane:
## Summary
- switch the graphing editor / variable comboboxes over to use the new select component
- fix some styling with the select component, add a loading spinner while fetching results
- change graph `groupByKeys` and variable `defaultValues` to support multiple values

- clicktested locally

- defaultValues change is backwards compatible with existing format
- groupByKeys change needs a migration - will run this directly after automigrate:
```
update graphs
set group_by_keys = ARRAY[group_by_key]
where group_by_key is not null
```
(until this migration runs, existing graphs using group by will lose their grouping)

- @julian-highlight fyi


Vadman97 (approved):
makes sense to me

Issue #9042, opened 2024-07-30
Ticket enhancement: Automatically set up `<HighlightInit backendUrl=...>` using the proxy when one is set up via the `withHighlightConfig` wrapper


Vadman97:
the highlight next SDK can set up a backend proxy endpoint with the `withHighlightConfig` wrapper, but the `<HighlightInit>` frontend component does not use the configured proxy

maybe you already were looking into this with the docs improvements, but i'll ticket an enhancement from [https://discord.com/channels/1026884757667188757/1265564202269937664](https://discord.com/channels/1026884757667188757/1265564202269937664) to automatically set up the `<HighlightInit backendUrl=...>` to use the proxy when one is set up via the `withHighlightConfig` wrapper?

> [Discord - Group Chat That’s All Fun & Games](https://discord.com/channels/1026884757667188757/1265564202269937664)

Vadman97:
came up again
https://discord.com/channels/1026884757667188757/1078929798828666981/1290972453367189579

Abiji-2020:
> came up again https://discord.com/channels/1026884757667188757/1078929798828666981/1290972453367189579

I cannot open the channel

Rutik7066:
@Vadman97  I am having trouble setting up the development environment and need guidance on testing changes for this issue. Your help would be greatly appreciated.

Rutik7066:
@Vadman97  I successfully set up a local dev environment now  I am trying to do e2e test for nextjs  but I am getting error for typegen 
![Screenshot 2024-11-02 151003](https://github.com/user-attachments/assets/14d0e5f4-79df-4b6f-970d-a188771850c1)

Rutik7066:
@Vadman97 @jay-khatri  PR is ready to review.

Rutik7066:
@Vadman97 Please check the PR. The backend URL is now `/highlight-events`. It took me some time to understand why we need to set it 😂.

PR #9425, opened 2024-10-07
fix: auto backend url setup

rajeshj11:
## Summary

/claim #9042

### DIFF
--- a/sdk/highlight-next/src/util/with-highlight-config.ts
+++ b/sdk/highlight-next/src/util/with-highlight-config.ts
@@ -137,10 +137,10 @@ const getHighlightConfig = async (
 				source: '/:path*.map',
 				destination: '/404',
 			}
-
+			const backendUrl = defaultOpts.sourceMapsBackendUrl
 			const highlightRewrite = {
 				source: '/highlight-events',
-				destination: 'https://pub.highlight.io',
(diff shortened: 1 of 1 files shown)


sdk/highlight-next/src/util/with-highlight-config.ts
```diff
@@ -137,10 +137,10 @@ const getHighlightConfig = async (
 				source: '/:path*.map',
 				destination: '/404',
 			}
-
+			const backendUrl = defaultOpts.sourceMapsBackendUrl
 			const highlightRewrite = {
 				source: '/highlight-events',
-				destination: 'https://pub.highlight.io',
+				destination: backendUrl || 'https://pub.highlight.io',
```
Vadman97, line 143: The change required in the ticket is a bit different. Essentially, as part of the `withHighlightConfig` rewrite, we want to make it so the `<HighlightInit />` component automatically updates the client-side `H.init` [call](https://github.com/highlight/highlight/blob/e6a5bcea42b986a4da2fecd5234e18bf0758a39a/sdk/highlight-next/src/next-client.tsx#L26-L29) with the `backendUrl`

[Here's another full example app](https://github.com/highlight/highlight/tree/main/e2e/nextjs) if that helps.

Vadman97 (changes requested):

PR #9426, opened 2024-10-07
Fix session related resource player error

SpennyNDaJets:
## Summary
The SessionPanel is crashing - `useResizePlayer` calls `useReplayerContext` , so it must be called within the `ReplayerContextProvider`. Move the call into a Base component of the SessionPanel.

1. View traces
2. Use the search `secure_session_id EXISTS`
3. Click on the session id to play
- [x] Able to play session in panel
- [x] No crash

N/A

N/A


ccschmitz (approved):
Thank you for fixing 🙏

Vadman97 (approved):
works great in preview

PR #9433, opened 2024-10-08
Flatten objects to pass in as attributes

SpennyNDaJets:
## Summary
Python SDK can't handle logging multiple level `Dicts`. Flatten the `Dict` with a `.` separator to pass into the OpenTelemetry API.

Screenshot 2024-10-08 at 2 20 14 PM

1. Add a log to a Python E2E app that includes a multi level `Dict`
2. Start the Python E2E app
3. Hit the endpoint
- [ ] Log is processed
- [ ] Log displays `Dict` correctly

N/A

N/A


sdk/highlight-py/highlight_io/sdk.py
```diff
@@ -436,7 +437,8 @@ def log_hook(self, span: Span, record: logging.LogRecord, formatted: str = ""):
             attributes["highlight.trace_id"] = request_id
             attributes["highlight.session_id"] = session_id
             if isinstance(record.args, dict):
-                attributes.update(record.args)
+                addedAttributes = flatten_dict(record.args, sep=".")
+                attributes.update(addedAttributes)
```
Vadman97, line 441: would it be easy to update the [sdk unit test](https://github.com/highlight/highlight/blob/946928c433150c0272f8d3695b7874475f0d2204/sdk/highlight-py/tests/test_sdk.py#L77-L85) as well to flex this logic?

Vadman97 (approved):

sdk/highlight-py/highlight_io/sdk.py
Vadman97: that would also ideally improve the coverage of the new `sdk/highlight-py/highlight_io/utils/dict.py` file

PR #9442, opened 2024-10-10
Enable event search in beta

SpennyNDaJets:
## Summary
Enable the event search in beta
- remove noisy log
- limit the frontend group by and bucket by to 100 (i.e. backend restrictions)

Screenshot 2024-10-10 at 11 34 31 AM

1. Create a graph/alert for events
- [ ] Beta tag is present
- [ ] Not able to group by more than 100

N/A

N/A


frontend/src/pages/Graphing/GraphingEditor.tsx
```diff
@@ -79,6 +79,8 @@ import { useRetentionPresets } from '@/components/Search/SearchForm/hooks'
 type BucketBy = 'None' | 'Interval' | 'Count'
 const BUCKET_BY_OPTIONS: BucketBy[] = ['None', 'Interval', 'Count']
 
+const MAX_BUCKET_SIZE = 100
```
SpennyNDaJets, line 82: Do we want to up this? For users grouping by unique user or session, they would be limited to 100 values on the graph - maybe this is okay and we should write exceptions for users and session groups?

mayberryzane: IMO I feel like trying to graph so many buckets is a code smell since it creates charts with tons of small data points, and users should instead be pointed toward using metric drilldown instead, e.g. if they need to see each user contributing to the user count. Maybe there's stuff we could do to improve or customize the drilldowns.

mayberryzane (approved):

PR #9446, opened 2024-10-10
Fix click event processing

SpennyNDaJets:
## Summary
When a click event fails to be processed, it cancels the writing of all the other events in the batch to fields and session_events. Fix the click processing to correctly unmarshal the properties, but continue in processing events if an error is returned.

1. Complete a session
2. Create an event graph
- [ ] Able to query custom events accurately
- [ ] Able to query clicks accurately
- [ ] Able to query navigation events accurately

N/A

N/A


backend/public-graph/graph/resolver.go
```diff
@@ -2362,27 +2363,35 @@ func (r *Resolver) AddSessionEvents(ctx context.Context, sessionID int, events *
...
+					continue
 				}
 			}
 
 			if clickEvent {
-				propertiesObject := make(map[string]string)
+				propertiesObject := make(map[string]interface{})
 				if err := json.Unmarshal([]byte(payloadStr), &propertiesObject); err != nil {
-					return e.New("error deserializing track event properties")
+					log.WithContext(ctx).Error("error deserializing click event properties")
```
Vadman97, line 2381: could we log the payloadStr to get to the bottom of the errors?

Vadman97 (approved):

PR #9449, opened 2024-10-11
Fix python attributes dict warning

SpennyNDaJets:
## Summary
We are still passing in dictionaries as attributes to Otel, which is not allowed. Flatten the dict in other places with a `.`, which will help Otel rebuild the object.

Screenshot 2024-10-11 at 12 25 04 PM

N/A

N/A

N/A


.vscode/settings.json
SpennyNDaJets, line 1: Didn't mean to update this in https://github.com/highlight/highlight/pull/9433

Vadman97 (approved):

PR #9450, opened 2024-10-11
support showing active hours in session feed ui

Vadman97:
## Summary

Add a session feed configuration option that allows rendering the session total time rather than count.
Automatically enabled by default for certain workspaces.

reflame
Screenshot from 2024-10-11 15-02-46

Screenshot from 2024-10-11 15-52-46

no

no

### DIFF
--- a/frontend/src/components/Search/SearchForm/SearchForm.tsx
+++ b/frontend/src/components/Search/SearchForm/SearchForm.tsx
@@ -56,7 +56,6 @@ import {
 import { ProductType, SavedSegmentEntityType } from '@/graph/generated/schemas'
 import { useDebounce } from '@/hooks/useDebounce'
 import { useApplicationContext } from '@/routers/AppRouter/context/ApplicationContext'
-import { formatNumber } from '@/util/numbers'
 
 import { AiSearch } from './AiSearch'
 import * as styles from './SearchForm.css'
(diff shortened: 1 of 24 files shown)


frontend/src/components/Search/SearchForm/SearchForm.tsx
```diff
@@ -263,14 +262,11 @@ const SearchForm: React.FC<SearchFormProps> = ({
...
 											<LoadingBox />
-										) : (
-											resultCount != null && (
-												<Text color="weak">
-													{formatNumber(resultCount)}{' '}
-													results
-												</Text>
-											)
-										)}
+										) : resultFormatted ? (
```
SpennyNDaJets, line 265: Not sure this is the place for the comment, but noticed this text when 0 results - don't think it needs to block though

Screenshot 2024-10-15 at 10 23 28 AM

frontend/src/pages/Sessions/SessionsFeedV3/hooks/useSessionFeedConfiguration.ts
```diff
@@ -23,6 +27,17 @@ export const useSessionFeedConfiguration = () => {
...
+	const [resultFormat, setResultFormat] =
+		useLocalStorage<SESSION_FEED_RESULT_FORMAT>(
+			`${LOCAL_STORAGE_KEY_PREFIX}ResultFormat`,
+			'Count',
+		)
+
+	useEffect(() => {
+		if (sessionResultsVerbose) {
+			setResultFormat('Count/Length/ActiveLength')
+		}
```
SpennyNDaJets, line 39: Not sure how big a deal this is unless we need to roll back, but turning off the toggle still uses the updated result format for any users that have it saved in local storage since we don't reset it

SpennyNDaJets (approved):
Looks good to me! Just a couple callouts

PR #9452, opened 2024-10-11
data export for metric charts and tables

Vadman97:
## Summary

Supports exporting dashboard charts via a new `Download CSV` button.

https://www.loom.com/share/0e6a4af7a784484087b7ba6713012efb

no

@julian-highlight if you want to check the download button


ccschmitz (approved):
Code + click look great! My only feedback is that the format of the timestamp is a little hard to work with in a spreadsheet. I tried importing into Mac's Sheets and Google Sheets, and in both cases there wasn't an easy way to convert the number to a timestamp. Wondering if there's anything we can/should do to make that easier?

Vadman97:
good idea @ccschmitz , let me try formatting in a way that excel would be able to parse easily

Vadman97:
@ccschmitz got timestamps parsing now in excel

Screenshot 2024-10-14 at 19 11 20

PR #9454, opened 2024-10-14
support larger groups for funnels and tables

Vadman97:
## Summary

* Allow larger group limit for certain customers who want to get a larger set of results
* For funnels, use the max group size to allow funnel steps to filter on one another

local deploy

no

no

### DIFF
--- a/frontend/src/pages/Alerts/AlertConfigurationCard/AlertConfigurationConstants.ts
+++ b/frontend/src/pages/Alerts/AlertConfigurationCard/AlertConfigurationConstants.ts
@@ -4,8 +4,8 @@ const HOURS_PER_DAY = 24
 const DAYS_PER_WEEK = 7
 
 const SECONDS_PER_HOUR = SECONDS_PER_MINUTE * MINUTES_PER_HOUR
-const SECONDS_PER_DAY = SECONDS_PER_HOUR * HOURS_PER_DAY
-const SECONDS_PER_WEEK = SECONDS_PER_DAY * DAYS_PER_WEEK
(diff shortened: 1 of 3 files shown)


frontend/src/pages/Alerts/AlertConfigurationCard/AlertConfigurationConstants.ts
```diff
@@ -76,11 +76,31 @@ export const FREQUENCIES = [
 		value: `${SECONDS_PER_DAY * 3}`,
 		id: '3d',
 	},
+	{
+		name: '5 days',
```
SpennyNDaJets, line 80: Tagging @mayberryzane - are the alerts on the backend only looking back at a max of 1 hour currently? Think its okay for the graphing interval, but wondering if we should have a copy or manipulate the list for alerts?

SpennyNDaJets (approved):
Looks good! Just the one question

frontend/src/pages/Alerts/AlertConfigurationCard/AlertConfigurationConstants.ts
mayberryzane: during alert evaluation we aggregate data +/- 2 hours from the current date, then use the alert's threshold window to filter. I think there could be some issues for longer time ranges after a change is made to the alerts setting until there's `thresholdWindow` amount of data saved

Vadman97: :+1: will pull this out into a new dashboard frequencies array, tho it sounds like we may already have issues with this with the larger ranges?

PR #9456, opened 2024-10-15
Fix alert page crash

SpennyNDaJets:
## Summary
Alert page crashing due to trying to call useContext from outside wrapper -> wrap the form in the GraphContextProvider

- [ ] Able to create a new alert
- [ ] Able to view/edit a current alert

N/A

N/A


Vadman97 (approved):
my bad, thanks for catching this bug

PR #9459, opened 2024-10-15
fix live mode

Vadman97:


Vadman97:
superceded by https://github.com/highlight/highlight/pull/9610

PR #9468, opened 2024-10-16
fix writing session click events as fields

Vadman97:
## Summary

* Ensure that click events are written as session fields

session showing up with `clickTextContent` filter
Screenshot from 2024-10-16 15-22-31

no

no

### DIFF
--- a/backend/public-graph/graph/resolver.go
+++ b/backend/public-graph/graph/resolver.go
@@ -2384,7 +2384,10 @@ func (r *Resolver) AddSessionEvents(ctx context.Context, sessionID int, events *
 
 				attributes := make(map[string]string)
 				for k, v := range propertiesObject {
-					attributes[k] = fmt.Sprintf("%v", v)
+					attributes[k] = fmt.Sprintf("%.*v", SESSION_FIELD_MAX_LENGTH, v)
+					if len(attributes[k]) > 0 {
+						fields[k] = attributes[k]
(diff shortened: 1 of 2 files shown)


Vadman97:
@SpennyNDaJets friendly bump :)

backend/public-graph/graph/resolver.go
```diff
@@ -2384,7 +2384,10 @@ func (r *Resolver) AddSessionEvents(ctx context.Context, sessionID int, events *
 
 				attributes := make(map[string]string)
 				for k, v := range propertiesObject {
-					attributes[k] = fmt.Sprintf("%v", v)
+					attributes[k] = fmt.Sprintf("%.*v", SESSION_FIELD_MAX_LENGTH, v)
+					if len(attributes[k]) > 0 {
+						fields[k] = attributes[k]
```
SpennyNDaJets, line 2389: Are we concerned with collisions in the fields object? What if multiple click events write to fields with a `clickSelector`? I believe we will only get the last one

SpennyNDaJets (approved):
Just the one question, but otherwise GTG

backend/public-graph/graph/resolver.go
Vadman97: good point, guess that's a general problem with this function? fields should be an array rather than a map since we allow multiple values for the same field

PR #9474, opened 2024-10-17
anomaly alerts

mayberryzane:
## Summary
- create predictions service as a Python Flask app running on port 5001, add it to docker compose
- add anomaly as a threshold type option - when selected in the UI, it will show upper / lower confidence intervals with a grey band and highlight anomalous datapoints in red (grouped data shows anomalous datapoints but omits the confidence intervals)
- alert evaluation loads data for the preceding 100 intervals to make a prediction for the current interval, then compares the results for the last interval against the prediction

- tested running new service w/ `start_infra.sh`
- clicktested anomaly settings UI
- testing anomaly evaluation w/ debugger

- isn't necessary to migrate threshold type and condition on existing alerts (added some backfilling logic), but will do so anyway
```
update alerts
set threshold_type = 'Constant', threshold_condition = case when below_threshold then 'Below' else 'Above' end
```
- need to deploy predictions service to ECS

- @julian-highlight fyi


frontend/src/pages/Graphing/components/LineChart.tsx
```diff
@@ -40,6 +42,32 @@ export type LineChartConfig = {
...
+	const { payload } = props
+
+	if (!payload || !payload[key]) {
+		return false
+	}
+
+	if (
+		payload[YHAT_LOWER_KEY] &&
+		payload[YHAT_LOWER_KEY][key] &&
+		payload[key] < payload[YHAT_LOWER_KEY][key]
```
SpennyNDaJets, line 55: Nit: I think we can simplify this to something like:
```
payload[key] < payload[YHAT_LOWER_KEY]?.[key]
```

frontend/src/pages/Graphing/components/LineChart.tsx
```diff
@@ -40,6 +42,32 @@ export type LineChartConfig = {
...
+	) {
+		return true
+	}
+
+	if (
+		payload[YHAT_UPPER_KEY] &&
+		payload[YHAT_UPPER_KEY][key] &&
+		payload[key] > payload[YHAT_UPPER_KEY][key]
```
SpennyNDaJets, line 63: Same as above

backend/jobs/metric-alerts/metric-alerts.go
```diff
@@ -188,45 +195,78 @@ func processMetricAlert(ctx context.Context, DB *gorm.DB, MailClient *sendgrid.C
 		groupByKey = groupBy[0]
 	}
 
+	bucketsInner := buckets.Buckets
```
SpennyNDaJets, line 198: Not sure if this is a local issue, but getting a null pointer reference trying to start the backend here
```
github.com/highlight-run/highlight/backend/jobs/metric-alerts.processMetricAlert({0x1060466d0, 0x10a817160}, 0x140030923f0, 0x1400700e050, 0x14005f4db00, 0x14006dbaa20, 0x0)
	/Users/spenceramarantides/repos/highlight/backend/jobs/metric-alerts/metric-alerts.go:198 +0x14cc
github.com/highlight-run/highlight/backend/jobs/metric-alerts.WatchMetricAlerts.func1.1()
	/Users/spenceramarantides/repos/highlight/backend/jobs/metric-alerts/metric-alerts.go:50 +0x68
github.com/highlight-run/workerpool.(*WorkerPool).SubmitRecover.func1()
```

SpennyNDaJets (commented):
Overall code is looking good! Just the one `AlertGraph` to update and want to double check the null pointer is just my local issue

backend/jobs/metric-alerts/metric-alerts.go
mayberryzane: I might just have an env variable set differently for myself, but I think this is coming from the `lambdaClient` usage (which is probably nil locally). will remove the lambda client here anyway since I refactored that when I switched over to a flask app

frontend/src/pages/Graphing/components/Graph.tsx
```diff
@@ -475,6 +485,8 @@ const getCustomTooltip =
 			label = frozenTooltip.activeLabel
 		}
 
+		console.log('payload', payload)
```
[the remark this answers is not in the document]
mayberryzane: thanks for catching this!

frontend/src/pages/Alerts/AlertGraph/index.tsx
```diff
@@ -113,37 +122,52 @@ export const AlertGraph: React.FC<Props> = ({
...
+								? undefined
+								: [groupByKey]
 						}
 						setTimeRange={updateSearchTime}
 						bucketByKey="Timestamp"
 						bucketCount={sessionsProduct ? 50 : undefined}
 						bucketByWindow={
 							sessionsProduct ? undefined : thresholdWindow
 						}
+						predictionSettings={
```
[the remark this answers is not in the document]
mayberryzane: 👍 confirmed anomaly graphs are shown on the alert detail page after merging

backend/model/model.go
```diff
@@ -1961,10 +1961,12 @@ type Alert struct {
...
-	BelowThreshold    *bool
-	ThresholdValue    *float64
-	ThresholdWindow   *int
-	ThresholdCooldown *int
+	BelowThreshold     *bool
+	ThresholdValue     *float64
+	ThresholdWindow    *int
+	ThresholdCooldown  *int
+	ThresholdType      modelInputs.ThresholdType
+	ThresholdCondition modelInputs.ThresholdCondition
```
SpennyNDaJets, line 1969: Not sure whats going on, but was seeing these errors when loadin/saving an alert. The code looks good to me, and confirmed in postgres that these columns exists although they were null. Do we need a default here or to migrate data?

```
Cannot query field "threshold_type" on type "Alert". Did you mean "threshold_value" or "threshold_window"?
Cannot query field "threshold_condition" on type "Alert". Did you mean "threshold_window" or "threshold_cooldown"?
```

mayberryzane: those look like graphql backend errors and I see those fields in the `generated.go` file so I'm not sure what's going on, I was able to load/save alerts locally. are you getting any errors running the backend locally?

mayberryzane: was originally planning on requiring a migration (see sql script in PR description), but just added some code to handle this as well for on-prem / hobby customers

SpennyNDaJets (approved):
Great! Click test and code look good to me!

PR #9486, opened 2024-10-22
Event drilldown support

SpennyNDaJets:
## Summary
Support drilling down to sessions when a user clicks on an event data point. Show all the sessions that the events occur in.

Note: There are one session to many events, so the number of sessions returned from the result may not match the number of events.

https://github.com/user-attachments/assets/09f3f32b-4314-4aa2-a962-aec336b24c25

1. Create a graph using events as the source.
2. Click into a data point to open the side panel
- [ ] Sessions populate side panel
- [ ] Able to scroll through sessions
- [ ] Able to play a session

N/A

N/A


ccschmitz (approved):
Code + click test LGTM!

PR #9487, opened 2024-10-22
Update 3_app-router.md

vanisov:
This is such a stupid change and its super small, NOT priority but my OCD just doesn't let me NOT make this change xD

## Summary

It's a very small detail, but because i've been reviewing the docs (trying to integrate it myself), I noticed that it was missing the ending single-quote.
Screenshot 2024-10-21 at 5 59 34 PM
Not s big fix, not too important, but can save someone a headache if they are copy and pasting the code directly and can't figure out what's erroring.
No tests needed

I don't think so? It depends how you handle your docs.

Not really


Vadman97 (approved):
thank you, much appreciated!

PR #9490, opened 2024-10-22
Fix error calling H.getSession* after using forceNew to start a new session

ccschmitz:
## Summary

We were storing the `sessionSecureID` at the top level of the firstload package as well as on the instance of the `Highlight` class we returned from calling `initalize` (stored on `highlight_obj`). We were referencing the `sessionSecureID` from firstload in `H.getSessionURL` and `H.getSessionDetails`, but the data we were storing and trying to retrieve from localStorage was coming from `highlight_obj`. Also, we weren't updating the `sessionSecureID` in firstload when calling `H.start`, so this wasn't finding the latest session data if that was called.

This PR updates the `H.getSession*` methods so they always pull values from `highlight_obj` ensuring that we don't have stale values anywhere.

I added a `H.start({ forceNew: true })` call immediately after our `H.init` call on the client. This helped me repro the error reported by the customer, then I ensured this change fixed that issue.

https://www.loom.com/share/c78a6815fa374991a3e8cf4ad02699bd

https://www.loom.com/share/0c4d0174122d4c93aa140097ad5fefeb

Will let customers impacted by this know that the fix has gone out, but should be a noop for most users.

N/A


.changeset/cyan-fireants-matter.md
```diff
@@ -0,0 +1,5 @@
+---
+'highlight.run': patch
+---
+
+fix `H.getSession*` methods using stale session ID
```
ccschmitz, line 5: @Vadman97 should I do a release as part of this PR since we want the change out immediately?

Vadman97: sure, especially since we don't have any other changesets in `.changesets` currently, so the release would only have this pr. you can run `yarn changeset version` on this branch (also `yarn reflame-build`) and push that - once merged to main that should release the npm package

ccschmitz: Done!

sdk/firstload/src/index.tsx
Vadman97, line 1: should we also move the `sessionSecureID` to inside the `init` function since we should not be using it anywhere else in this file?

Vadman97 (approved):
thanks for figuring this out, chris!

sdk/firstload/src/index.tsx
ccschmitz: Yeah, good call. Just updated that!

ccschmitz:
@Vadman97 I made some changes to clean up the `onHighlightReady` implementation and added some specs for it. Would appreciate if you could give that another look since it's a critical code path.

The reason I refactored it is because I noticed an issue with the previous logic where it would get into an infinite loop when calling `getSessionDetails`. The issue was something to do with how it called `onHighlightReady` and `getSessionURL` which also called `onHighlightReady` in its implementation.

sdk/firstload/src/__tests__/setup.ts
```diff
@@ -1,4 +1,5 @@
 import { vi } from 'vitest'
+import 'vitest-canvas-mock'
```
ccschmitz, line 2: Unrelated, but fixes some errors we were seeing in the console when running tests.

Vadman97 (approved):
makes a lot of sense! the `onHighlightReady` refactor definitely makes the logic easier to read, though i can't tell how it's different in terms of the effect compared to before (other than the new immediate callback execution at the beginning, which still shouldn't be any different compared to before?). :ship:

PR #9505, opened 2024-10-25
Move session event processing to kafka queue

SpennyNDaJets:
## Summary
Move the session event processing to the DataSyncQueue so it does not block push payload

1. Run the app and click around
- [ ] Confirm events are still being processed and inserted into ClickHouse

1. How will this additional load add to the DataSync Queue?
2. How will it scale with Production data?

N/A

### DIFF
--- a/backend/public-graph/graph/resolver.go
+++ b/backend/public-graph/graph/resolver.go
@@ -327,7 +327,7 @@ func (r *Resolver) SubmitSessionEvents(ctx context.Context, sessionID int, event
 
 	err := r.DataSyncQueue.Submit(ctx, "", messages...)
 	if err != nil {
-		return e.Wrap(err, "failed to submit otel project traces to public worker queue")
+		return e.Wrap(err, "failed to submit session events to public worker queue")
 	}
 
 	return nil
--- a/backend/public-graph/graph/resolver.go
+++ b/backend/public-graph/graph/resolver.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/highlight-run/highlight/backend/env"
 	"github.com/highlight-run/highlight/backend/geolocation"
+	kafkaqueue "github.com/highlight-run/highlight/backend/kafka-queue"
 	"github.com/oschwald/geoip2-golang"
 
 	"go.opentelemetry.io/otel/codes"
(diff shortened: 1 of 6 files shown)


backend/public-graph/graph/resolver.go
```diff
@@ -295,37 +296,38 @@ func (r *Resolver) AppendProperties(ctx context.Context, sessionID int, properti
...
+		}
+
+		messages = append(messages, &kafkaqueue.SessionEventRowMessage{
+			Type:            kafkaqueue.PushSessionEvents,
+			SessionEventRow: sessionEvent,
 		})
 	}
 
-	err := r.Clickhouse.WriteSessionEventRows(ctxW, sessionEvents)
+	err := r.DataSyncQueue.Submit(ctx, "", messages...)
```
Vadman97, line 328: just want to think through this, but having a "" key means that messages for the same session may be inserted in order. shouldn't be a problem though, since the events are ordered by timestamp

backend/worker/kafka_worker.go
```diff
@@ -453,6 +470,21 @@ func (k *KafkaBatchWorker) flushTraces(ctx context.Context, traceRows []*clickho
 	return nil
 }
 
+func (k *KafkaBatchWorker) flushSessionEvents(ctx context.Context, sessionEventRows []*clickhouse.SessionEventRow) error {
+	span, ctxT := util.StartSpanFromContext(ctx, fmt.Sprintf("worker.kafka.%s.flush.clickhouse", k.Name), util.WithHighlightTracingDisabled(true))
+	span.SetAttribute("NumTraceRows", len(sessionEventRows))
+	span.SetAttribute("PayloadSizeBytes", binary.Size(sessionEventRows))
+	err := k.Worker.PublicResolver.Clickhouse.WriteSessionEventRows(ctxT, sessionEventRows)
```
Vadman97, line 477: we should change `WriteSessionEventRows` to not use async inserts since this batch will always be large now. that will help reduce load on clickhouse since our worker is ensuring the events are batched. this is what we do for logs / traces which are batched by the kafka worker: https://github.com/highlight/highlight/blob/00589a79d338415b469c533a44f8c4f5589f81d3/backend/clickhouse/traces.go#L186-L210

Vadman97 (approved):
when this is ready to deploy, we may need to update the datasync worker batch settings in case we need to flush the events in larger batches

backend/public-graph/graph/resolver.go
SpennyNDaJets: Sounds good! I originally had the secure session id here, but removed it because I thought it wouldn't need to be partitioned in any specific order?

backend/worker/kafka_worker.go
SpennyNDaJets: Make sense! I was curious if there was a difference in why errors and sessions should be written asynchronously? Because I believe they are on the same DataSyncQueue.

Vadman97: i think we just never switched those over after introducing the datasync queue :thinking: 
@mayberryzane do you remember if there was more context?

PR #9507, opened 2024-10-25
fix @highlight-run/next sdk breaking HMR with turbopack next.js 14

Vadman97:
## Summary

Per https://github.com/vercel/next.js/issues/70424, Next.js 14 HMR can be broken by
`@opentelemetry/instrumentation` because of the `require-in-the-middle` import that would be bundled in.
Keep the dependency external and have Next.js keep it external to avoid breaking HMR.

updating highlight.io to test locally

no

no


mayberryzane (approved):

juandl:
New version breaks nestjs too

PR #9511, opened 2024-10-28
Fix helper_method error when using Ruby SDK in Rails API-only mode

ccschmitz:
## Summary

Fixes an error we were seeing when initializing the Ruby SDK in a Rails app that is set up in [API-only mode](https://guides.rubyonrails.org/api_app.html). Also adds an E2E example app and some tests that I used to verify the changes were working as expected.

New E2E app with tests to verify the changes.

Need to notify users who were impacted by these errors to upgrade to the latest version of the SDK.

N/A


e2e/ruby/rails/api-only/Gemfile
Vadman97, line 1: would it be easy to add this e2e dir to the e2e test, so that we know in CI that this app runs? we would just need to add a new [Dockerfile](https://github.com/highlight/highlight/blob/148fa7803779745d7ae048ca13750a787649a8f3/e2e/ruby/Dockerfile#L22-L21) and [update the runner script](https://github.com/highlight/highlight/blob/0c478bf06283cd9f67faa2b18011dd3b38291194/e2e/tests/src/app_runner.py#L175-L185)

ccschmitz: Yeah, happy to look into this!

ccschmitz: Actually, I think this would be better to run the integration tests for the E2E app in CI rather than spin up the entire app with the app runner. I just made that change in 054866250c865d735130154b5276e90d2c528c64 - let me know if you think that is enough for now!

Vadman97: ah, makes sense to me!

PR #9513, opened 2024-10-28
persist table sort order by id

mayberryzane:
## Summary
- use local storage variables to store the sort column and asc/desc by visualization id

- clicktested locally

- no

- no

### DIFF
--- a/frontend/src/pages/Graphing/components/Table.tsx
+++ b/frontend/src/pages/Graphing/components/Table.tsx
@@ -8,16 +8,17 @@ import {
 } from '@highlight-run/ui/components'
 import clsx from 'clsx'
 import _ from 'lodash'
-import { useState } from 'react'
 
 import {
 	getTickFormatter,
 	GROUP_KEY,
 	InnerChartProps,
 	SeriesInfo,
+	VizId,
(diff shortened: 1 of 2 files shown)


frontend/src/pages/Graphing/components/Table.tsx
```diff
@@ -56,8 +58,14 @@ export const MetricTable = ({
 
 	const showMetricFn = series.join('') === ''
 
-	const [sortColumn, setSortColumn] = useState(-1)
-	const [sortAsc, setSortAsc] = useState(true)
+	const [sortAsc, setSortAsc] = useLocalStorage<boolean>(
+		`sort-asc-${visualizationId}`,
```
SpennyNDaJets, line 62: If this is "undefined", will it be set for all not saved visualizations?

mayberryzane: yep, ideally it would persist after saving, but this is a bit complicated because we're saving the cached graph with a temp id and then the id is replaced after apollo refetches the visualization, but there's not an easy way to link the two in order to replace the local storage key. I figured this approach was ok for now

SpennyNDaJets (approved):

PR #9526, opened 2024-10-28
Respect `tracingOrigins` network instrumentation context propagation

ccschmitz:
## Summary

Ensures that the `traceparent` header is only assigned on requests to URLs that match the `tracingOrigins` setting on the client config. Previously, we were adding the `traceparent` header to all requests.

Added some unit tests for the logic behind the new config setting and tested locally by inspecting network traffic with different settings - see results below.

**`['pri.highlight.io', 'localhost:8082/private']` (default - header included)**
Screenshot 2024-10-28 at 4 11 41 PM

**`['pri.highlight.io']` (header not included)**
Screenshot 2024-10-28 at 4 11 15 PM

**`true` (header included)**
Screenshot 2024-10-28 at 4 12 02 PM

Need to notify some customers about this change since they have been waiting on the fix.

N/A

### DIFF
--- a/.changeset/four-lizards-mix.md
+++ b/.changeset/four-lizards-mix.md
@@ -1,0 +1,5 @@
+---
+'highlight.run': patch
+---
+
+respect traceOrigins setting for context propagation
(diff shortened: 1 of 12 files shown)


Vadman97 (approved):
awesome, local clicktest looks good!

Screenshot from 2024-10-28 18-00-00

PR #9527, opened 2024-10-28
Fix span sorting in trace waterfall view

ccschmitz:
## Summary

Fixes an issue where spans were being rendered in the wrong order on the waterfall view of the flame graph. This change ensures that spans are always rendered in order of when they started, which makes more sense in this context than the other sorting method we use, which optimizes for placement on a flame graph.

Click tested by viewing a trace in the waterfall UI before and after making this change.

**Before**
Screenshot 2024-10-28 at 4 28 07 PM

**After**
Screenshot 2024-10-28 at 4 27 46 PM

I will notify customers who reported this issue in Discord that it is fixed.

N/A


Vadman97 (approved):
:muscle:

PR #9558, opened 2024-10-31
remove mouseMoveState for perf

mayberryzane:
## Summary
- this state was being tracked to determine whether or not to show a pointer for drilldown - this will cause a pointer to always be shown but will speed up perf for complex graphs significantly due to not having to rerender on mouse move

- clicktested locally w/ complex group by

- no

- no


SpennyNDaJets (approved):
Click test looks good to me!

PR #9560, opened 2024-10-31
fix object.hasown reference for node 14

Vadman97:
## Summary

* node 14 does not support `Object.hasOwn` - use `Object.prototype.hasOwnProperty` instead
* next 14 requires externalization of `require-in-the-middle` to support HMR, but this would break the node sdk in node 14

local node 14 deploy

before
Screenshot from 2024-10-31 13-30-46

after
Screenshot from 2024-11-01 15-48-31

no

no

### DIFF
--- a/.github/workflows/turbo.yml
+++ b/.github/workflows/turbo.yml
@@ -102,7 +102,7 @@ jobs:
                   NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
 
             - name: Publish npm packages
-              # if: github.ref == 'refs/heads/main'
+              if: github.ref == 'refs/heads/main'
               id: changesets-publish
               uses: changesets/action@v1
               with:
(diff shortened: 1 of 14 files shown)


.github/workflows/turbo.yml
```diff
@@ -102,7 +102,7 @@ jobs:
                   NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
 
             - name: Publish npm packages
-              if: github.ref == 'refs/heads/main'
+              # if: github.ref == 'refs/heads/main'
```
mayberryzane, line 105: reminder to uncomment?

mayberryzane (approved):

.github/workflows/turbo.yml
Vadman97: ty :pray:

PR #9561, opened 2024-10-31
support winston json formatted log messages

Vadman97:
## Summary

Supports ingesting structured json messages that have a message that is json.
This now unwraps the inner json to be top level attributes, in addition to serializing the top json payload as the message.

Screenshot from 2024-10-31 14-00-00

no

no


ccschmitz (approved):
🆒

PR #9578, opened 2024-11-01
Switch events to async writes

SpennyNDaJets:
## Summary
Revert back to async write while investigating why batch writes are causing no events to populate

Able to process events and create event graphs/alerts

N/A

N/A


Vadman97 (approved):
since this will be called from the datasync queue with a larger batch size, this may be good enough since it won't increase async_insert load since the batch already will be large

PR #9593, opened 2024-11-05
Clean up error fields

SpennyNDaJets:
## Summary
Error Fields are no longer being used, but we are still creating them - clean this up

1. Create an error on the frontend
- [ ] Error created succesfully
2. Create an error on the backend
- [ ] Error created succesfully

N/A

N/A


Vadman97 (approved):
nice! we should drop the old error fields table(s) in production after we make sure everything looks good

PR #9594, opened 2024-11-05
create and exchange new logs table, save timestamp with full precision

mayberryzane:
## Summary
- creates new logs table with DateTime(9) precision timestamps
- exchanges old and new logs tables
- write new log rows without truncating timestamps

- verified new logs are written and returned by the `GetLogs` endpoint with full precision

- new logs table has already been created in prod and backfilled with data. this will exchange it with the existing logs table.
- will follow up by deleting the existing logs table once everything looks good

- no


ccschmitz (approved):
👀👍🥈🫵

Vadman97 (approved):

PR #9598, opened 2024-11-06
Enable browser OTeL by default

ccschmitz:
## Summary

Removes the `enableOtelTracing` option for the `H.init` in the Highlight client SDK and enables it by default. Also avoids adding the fetch and XHR instrumentations when network recording is not enabled.

Click tested locally by testing that existing functionality still worked as expected for local sessions:

* Traces could be viewed for user interactions, document load, and fetch/xhr requests
* Network resource details could be viewed from session dev tools

Also ensured the network fetch and XHR instrumentations were not added when `networkRecording` was not enabled.

Maybe want to reach out to existing customers or announce the the update somehow. Will discuss with @Vadman97.

N/A


package.json
Vadman97, line 1: unrelated but fine to include here

sdk/client/src/listeners/network-listener/utils/utils.ts
```diff
@@ -321,17 +321,13 @@ function makeId(length: number) {
...
-	if (useOtelTraceId) {
-		const context = getActiveSpan()
-		const traceId = context?.spanContext().traceId
-		return [getNetworkSessionSecureID(), traceId ?? requestId]
-	}
-
-	return [getNetworkSessionSecureID(), requestId]
+	const context = getActiveSpan()
+	const traceId = context?.spanContext().traceId
+	return [getNetworkSessionSecureID(), traceId ?? requestId]
```
Vadman97, line 330: should there always be an active span here? if not, should we start a span to avoid creating our own `id` format here? then we can remove the `requestId = makeId...` line too

sdk/highlight-wordpress/highlight-io/highlight.js
Vadman97, line 1: isn't this just the firstload index.umd.cjs file? should we `.gitignore` it and remove it?

Vadman97 (commented):
clicktested the network request recording on/off - looks great :pray:

sdk/highlight-wordpress/highlight-io/highlight.js
ccschmitz: Yeah, but you can't load any external scripts in WP plugins, apparently. Not sure why. Maybe a CSP thing?

ccschmitz: Sorry, you're right - I'll remove from the repo and ignore it!

sdk/client/src/listeners/network-listener/utils/utils.ts
ccschmitz: We discussed this on a call and decided to keep it around for now.

Vadman97 (approved):

PR #9600, opened 2024-11-06
Add Prisma auto instrumentation to Node SDK

ccschmitz:


mayberryzane (approved):

.github/workflows/e2e.yml
```diff
@@ -188,6 +191,9 @@ jobs:
                   python-version: '3.10'
                   cache: 'poetry'
 
+            - name: Install Doppler CLI
+              uses: dopplerhq/cli-action@v3
+
```
ccschmitz, line 196: Unrelated, but I believe this is why we're sometimes seeing the error about doppler not being installed. I think it's ultimately called via `env.sh`, and I'm not sure how it would be installed without this step.

PR #9605, opened 2024-11-06
Add timestamps to fields

SpennyNDaJets:
## Summary
Add timestamp to fields so they can be ordered within a session - to be used when querying events by session fields.

1. Click around in a session
2. Open up the local clickhouse browser
- [ ] Fields have a timestamp
- [ ] Processing Click fields with different timestamp than session start
- [ ] Processing Navigate fields with different timestamp than session start
- [ ] Processing Custom event fields with different timestamp than session start

Migrate data to set all previous fields to have timestamp equal to session created at

N/A


.changeset/selfish-games-tell.md
```diff
@@ -0,0 +1,5 @@
+---
+'highlight.run': minor
```
Vadman97, line 2:
```suggestion
'highlight.run': patch
```
since this is not affecting the end user behavior (since we are dropping the `visited-url` events on our backend but still recording new fields from the navigate events)

Vadman97 (approved):
tested locally to make sure `visited-url` still populated after the new client; looks good!

PR #9613, opened 2024-11-07
save alert and graph settings in url

mayberryzane:
## Summary
- stores the current settings for alerts and metrics as a base64 encoded `settings` url param
- if a user navigates or closes / reopens window, they will be shown the same settings
- if a user exits the settings page via the `cancel` or `save` buttons, the settings param will be cleared and the user will be shown the default settings next time

- clicktested locally

- no

- no

### DIFF
--- a/frontend/src/pages/Alerts/AlertForm/index.tsx
+++ b/frontend/src/pages/Alerts/AlertForm/index.tsx
@@ -16,7 +16,7 @@ import {
 } from '@highlight-run/ui/components'
 import { useParams } from '@util/react-router/useParams'
 import { Divider } from 'antd'
-import React, { PropsWithChildren, useMemo, useState } from 'react'
+import React, { PropsWithChildren, useEffect, useMemo, useState } from 'react'
 import { Helmet } from 'react-helmet'
 import { useNavigate, useSearchParams } from 'react-router-dom'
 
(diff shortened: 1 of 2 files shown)


frontend/src/pages/Alerts/AlertForm/index.tsx
```diff
@@ -256,6 +298,31 @@ export const AlertForm: React.FC = () => {
...
+		groupByKey,
+		thresholdValue,
+		thresholdCondition,
+		thresholdType,
+		thresholdWindow,
+		thresholdCooldown,
+		destinations,
+	}
+
+	const settingsEncoded = btoa(JSON.stringify(settings))
```
ccschmitz, line 317: Nit, but would it be more performant to only create the settings + do the encoding inside the `useEffect`? Or do we need this as a dependency to trigger the logic?

frontend/src/pages/Graphing/GraphingEditor.tsx
```diff
@@ -564,6 +631,15 @@ export const GraphingEditor: React.FC = () => {
 		fetchedLimitMetric,
 	}
 
+	const settingsEncoded = btoa(JSON.stringify(settings))
```
ccschmitz, line 634: Same question here as above.

ccschmitz (approved):
Code + click test look great! One quick question to see if we can avoid encoding the string on every render, but very minor and maybe necessary to have this work the way we want.

PR #9618, opened 2024-11-08
gate teams and jira as enterprise features

mayberryzane:
## Summary
- on the integrations page, put the Jira and Teams toggles behind the enterprise feature gate

- clicktested locally with and without settings enabled

- after deploying, will run this script to enable the enterprise settings for anyone already using jira / teams
```
update all_workspace_settings 
set enable_teams_integration = true 
where workspace_id in (
	select id from workspaces where microsoft_teams_tenant_id is not null
)

update all_workspace_settings 
set enable_jira_integration = true 
where workspace_id in (
	select workspace_id from integration_workspace_mappings where integration_type = 'Jira'
)
```

- no


SpennyNDaJets (approved):
LGTM

PR #9630, opened 2024-11-14
Add URL attributes to spans

ccschmitz:
## Summary

Adds `url.*` properties to network request spans from the XHR and fetch OTeL instrumentations. Follows the [URL semantic conventions](https://opentelemetry.io/docs/specs/semconv/attributes-registry/url/) for attribute naming, adding the following properties:

* `url.original`
* `url.domain`
* `url.path`
* `url.query_params`

Built the latest version of `firstload` and opened the app locally. Verified the new attributes were showing when viewing traces as well (see screenshot).

Screenshot 2024-11-15 at 9 59 07 AM

N/A

N/A


sdk/firstload/package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "highlight.run",
-	"version": "9.6.0",
+	"version": "9.6.1",
```
Vadman97, line 3: don't think we need to bump this manually since you're adding a `.changeset`

Vadman97 (approved):

PR #9632, opened 2024-11-15
Allow user to disable highlight log traces

SpennyNDaJets:
## Summary
For every log in the Python SDK, we emit a `highlight.log` trace. This can become noisy, so allow users to disable/enable this feature in the SDK.

### New SDK option
- enable_log_traces: `False` by default

1. Start the Django app
2. Send a request to the Django app
- [ ] A `highlight.log` trace is not recorded for each log event 
3. Add `enable_log_traces=True` to the Highlight configuration
- [ ] A `highlight.log` trace is recorded for each log event 
-
Minor package bump for change in behavior

N/A

### DIFF
--- a/docs-content/sdk/python.md
+++ b/docs-content/sdk/python.md
@@ -54,6 +54,10 @@ slug: python
       <p>If enabled, Highlight will record log output from the logging module.</p>
     </aside>
     <aside className="parameter">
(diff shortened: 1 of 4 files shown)


mayberryzane (approved):

SpennyNDaJets:
Closing for otel.go solution

PR #9639, opened 2024-11-19
Feature flag highlight.log traces

SpennyNDaJets:
## Summary
The `highlight.log` traces in the tracing project can be noisy and not very helpful for many users, but some users request them. Feature flag the ingestion of these traces to allow users to hide/show the `highlight.log` traces.

1. Start the Django app
2. Send a curl request to the app
- [ ] No log traces recorded
2. Update the `AllWorkspaceSetting.EnableLogTraceIngestion` to `true` in Postgres
3. Restart the app and backend
4.  Send a curl request to the app
- [ ] Log traces recorded

Screenshot 2024-11-19 at 10 26 28 AM

Fix boolean for appropriate organizations

N/A

### DIFF
--- a/backend/otel/otel.go
+++ b/backend/otel/otel.go
@@ -308,8 +308,11 @@ func (o *Handler) HandleTrace(w http.ResponseWriter, r *http.Request) {
 						log.WithContext(ctx).WithError(err).Error("failed to get workspace settings")
 						continue
 					}
-
-					if settings == nil || !settings.EnableLogTraceIngestion {
+					if settings == nil {
+						log.WithContext(ctx).Error("no workspace settings found")
+						continue
+					}
+					if !settings.EnableLogTraceIngestion {
 						continue
 					}
 				}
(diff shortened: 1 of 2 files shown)


backend/otel/otel.go
```diff
@@ -300,9 +300,18 @@ func (o *Handler) HandleTrace(w http.ResponseWriter, r *http.Request) {
...
+				// skip log spans
+				if span.Name() == highlight.LogrusSpanName || span.Name() == highlight.LogSpanName {
+					// skip unless workspace setting is on
+					settings, err := o.resolver.Store.GetAllWorkspaceSettingsByProject(ctx, fields.projectIDInt)
+					if err != nil {
+						log.WithContext(ctx).WithError(err).Error("failed to get workspace settings")
+						continue
+					}
+
+					if settings == nil || !settings.EnableLogTraceIngestion {
```
mayberryzane, line 312: nit: I would log the `settings == nil` case as an error since I don't think it's expected?

mayberryzane (approved):

PR #9641, opened 2024-11-19
dynamic sampling

mayberryzane:
## Summary
- `SAMPLE N` queries are not working in current version
- also hard to predict if sampling is needed without checking source tables
- make `EXPLAIN ESTIMATE` query to any relevant tables to get a row estimate, then calculate the sampling ratio manually if the estimate is greater than the sampling threshold

- tested metrics locally w/ all product types

- no

- no


SpennyNDaJets (approved):
Local click test and code look good to me!

PR #9642, opened 2024-11-19
Write fields to clickhouse separately than postgres

SpennyNDaJets:


SpennyNDaJets (commented):
Note: Was going to hold off on merging until back in the office on 12/1 to monitor

Duplicate session properties are written in `IndexSessionClickhouse` when a user refreshes the page - can these values change? Maybe we can not appendProperties on existing sessions?

backend/kafka-queue/types.go
```diff
@@ -37,6 +37,7 @@ const (
 	PushLogsFlattened                      PayloadType = iota
 	PushTracesFlattened                    PayloadType = iota
 	PushSessionEvents                      PayloadType = iota
+	PushSessionFields                      PayloadType = iota
```
Vadman97, line 40: is there a difference between a `SessionEvent` and a `SessionField`? seems to me like the 2 can be consolidated

backend/public-graph/graph/resolver.go
```diff
@@ -326,19 +335,35 @@ func (r *Resolver) SubmitSessionEvents(ctx context.Context, sessionID int, event
 		})
 	}
 
-	err := r.DataSyncQueue.Submit(ctx, "", messages...)
+	err := r.DataSyncQueue.Submit(ctx, strconv.Itoa(session.ID), messages...)
```
Vadman97, line 338: good idea to order the events for a given session, but technically mightt not matter since the inserted events use the timestamp of the event.

backend/public-graph/graph/resolver.go
```diff
@@ -326,19 +335,35 @@ func (r *Resolver) SubmitSessionEvents(ctx context.Context, sessionID int, event
...
 	if err != nil {
 		return e.Wrap(err, "failed to submit session events to public worker queue")
 	}
 
 	return nil
 }
 
-func (r *Resolver) AppendFields(ctx context.Context, fields []*model.Field, session *model.Session) error {
+func (r *Resolver) AppendFields(ctx context.Context, chfields []*clickhouse.ClickhouseField, session *model.Session) error {
```
Vadman97, line 346: we recently introduced `SubmitSessionEvents` as a way to write some fields as special events to clickhouse, right? seems that this `AppendFields` and `SubmitSessionEvents` can be consolidated since what they are doing is very similar (taking fields and writing them to clickhouse, either to `fields` table or to the `events` table). 

maybe we don't need 2 schemas in clickhouse for fields vs. events either?

Vadman97 (commented):
i think we can rethink the session fields vs. session events storage in clickhouse to consolidate, since the 2 are effectively the same thing. then we can streamline the 2 write paths

backend/public-graph/graph/resolver.go
mayberryzane: this was just a small thing I noted while pair programming w/ Spenny - I think it's better to partition by session id in case of an ingestion delay so the session either has all events or no events and isn't missing random ones in between

PR #9649, opened 2024-11-21
Invite user UI refresh + new Modal component

ccschmitz:
## Summary

Adds a new `Modal` component to the UI package and uses it in a new user invite modal on the "Members" workspace settings page.

Also changes the toast component to render toasts as dialogs so they appear on top of other items rendered in portals. This was causing issues because you couldn't interact with the toasts when triggered from a modal because their stacking context was below the modal.

https://www.loom.com/share/949482f003144a8c80bc6478ad1607b1

Click tested locally, and should be able to click test in the PR preview as well since these are all client-side changes.

N/A

Will get @julian-highlight's thoughts on the disabled UI + migrating from alerts to toasts.


mayberryzane (approved):
looks great!

PR #9651, opened 2024-11-21
unauthenticate user with invalid token

Vadman97:
## Summary

* When the OIDC token expires, brings the user back to the sign_in page.
* Allow validating user domain.
* Fix invite user to workspace flow on self-hosted deploys.

https://www.loom.com/share/118aec461c664b0fae31470d4627ce6f

new docker image release

no


ccschmitz (approved):
Nice - looks great! ✨ 

Another thought, maybe for a future PR, is that we could add an error link to the Apollo client that redirects users to the sign in page when we encounter an expired token.

PR #9657, opened 2024-11-25
add special case for project id where clause

mayberryzane:
## Summary
- Vadim discovered this query was faster using `ProjectId=1` instead:
```
select distinct Timestamp, UUID
from logs
where ProjectId in (1)
and Timestamp < now()
and Timestamp > now() - interval '30 days'
order by Timestamp DESC, UUID DESC
limit 51
```

- clicktested locally, ensured query was generated with equals

- no, can revert if issues

- no

### DIFF
--- a/backend/clickhouse/query.go
+++ b/backend/clickhouse/query.go
@@ -190,7 +190,12 @@ func makeSelectBuilder(
 
 	sb.Select(selectCols...)
 	sb.From(config.TableName)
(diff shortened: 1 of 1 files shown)


Vadman97 (approved):

Vadman97:
sick, exactly what i was playing with last week https://github.com/highlight/highlight/pull/9654

PR #9661, opened 2024-11-25
remove noisy otel logs

Vadman97:
## Summary

* suppress noisy log from otel ingest caused by cases where exception is not expected

Deploying from PR

no

no


ccschmitz (approved):
🤫

PR #9669, opened 2024-11-27
Add keyboard shortcut to select filter + add new

ccschmitz:
## Summary

Adds a keyboard shortcut to select a result and inject a space to immediately start entering a new filter.

https://www.loom.com/share/6e5936ee483e4e27aafd8b56413578ea

Also fixes a couple of bugs:

* Suggested keys wouldn't be shown when the entire query is deleted
* Keys were being queried with the wrong term and shown briefly when starting to enter the value on a filter

Start typing a query, and when you select a result from the suggestions, press `cmd/ctrl+enter` to see the new behavior. Contrast that with selecting a result without cmd/ctrl.

Also, play around with deleting a query to ensure it returns you to the state where the results are shown and unfiltered.

N/A

N/A


SpennyNDaJets (approved):
image

PR #9670, opened 2024-11-27
support multiple variables for graphs

mayberryzane:
## Summary
- instead of `column` and `metric_types`, backend supports a list of `expressions` (`aggregator`, `column` pairs)
- refactor how metadata is stored within graphing data to support multiple series for the same group key
- multiple variables are supported in line charts, bar charts, and tables
  - small caveat for line charts and bar charts: axis formatting uses the formatting and scale of the first series. can extend in the future to have two y-axes (left and right)

- clicktested locally with all graph types, bucketing / grouping permutations, drilldown, etc.

- need to migrate the `expressions` column for existing graphs 
```
update graphs
set expressions=concat('[{"column":"', metric, '","aggregator":"', function_type,'"}]')::jsonb
where function_type <> ''
```

- @julian-highlight fyi some additions to the graphing editor UI, can update:
Screenshot 2024-11-26 at 8 52 37 PM


backend/private-graph/graph/resolver.go
```diff
@@ -3644,6 +3644,34 @@ func (r *Resolver) UpsertDiscordChannel(workspaceId int, name string) (*model.Di
 	}, nil
 }
 
+func normalizeExpressions(column *string, metricTypes []modelInputs.MetricAggregator, expressions []*modelInputs.MetricExpressionInput) ([]*modelInputs.MetricExpressionInput, error) {
```
Vadman97, line 3647: this is responsible for converting the previous multiple `metricTypes` into the new format of metric expressions?

backend/private-graph/graph/schema.graphqls
```diff
@@ -2397,18 +2401,20 @@ type Query {
...
 	): LogConnection!
 	logs_histogram(project_id: ID!, params: QueryInput!): LogsHistogram!
+	# deprecated - use `metrics` instead
 	logs_metrics(
 		project_id: ID!
 		params: QueryInput!
-		column: String!
-		metric_types: [MetricAggregator!]!
+		column: String # deprecated - use `expressions` instead
+		metric_types: [MetricAggregator!] # deprecated - use `expressions` instead
```
Vadman97, line 2409: would we be able to remove these after we migrate the old stored viz configs?

Vadman97 (approved):
super cool! local clicktest worked great

Screenshot 2024-11-27 at 12 26 06


https://www.loom.com/share/a28e16a24f554d2db4b8e57f21e5af70

backend/private-graph/graph/resolver.go
mayberryzane: yes - it's relevant for anyone using our Grafana integration as it will still be sending requests in the old format. I can make an update to the Grafana integration to use the new format after releasing this, but users may still be on older versions.

backend/private-graph/graph/schema.graphqls
mayberryzane: unfortunately no since older versions of the Grafana integration will still be using these deprecated graphql resolvers

PR #9672, opened 2024-11-27
add new fields to clickhouse writes

mayberryzane:
## Summary
- split out certain otel standard attributes into their own column and dual write (writes will be ignored in db right now)

- checked ingested traces locally without and with adding these columns to db 

- will follow up with https://github.com/highlight/highlight/pull/9673
```
> step 1: add new fields to ClickhouseTraceRow for writes
step 2: create new columns in traces / traces_sampling tables
step 3: backfill new columns for existing data
step 4: reference new columns in queries
step 5: remove these attributes from TraceAttributes
```

- no

### DIFF
--- a/backend/clickhouse/traces.go
+++ b/backend/clickhouse/traces.go
@@ -122,31 +122,61 @@ var TracesSampleableTableConfig = SampleableTableConfig{
 }
 
 type ClickhouseTraceRow struct {
-	Timestamp        time.Time
-	UUID             string
-	TraceId          string
-	SpanId           string
-	ParentSpanId     string
-	ProjectId        uint32
-	SecureSessionId  string
-	TraceState       string
-	SpanName         string
-	SpanKind         string
-	Duration         int64
-	ServiceName      string
-	ServiceVersion   string
-	TraceAttributes  map[string]string
-	StatusCode       string
-	StatusMessage    string
-	Environment      string
-	HasErrors        bool
-	EventsTimestamp  []time.Time         `json:"Events.Timestamp" ch:"Events.Timestamp"`
-	EventsName       []string            `json:"Events.Name" ch:"Events.Name"`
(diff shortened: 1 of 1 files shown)


backend/clickhouse/traces.go
```diff
@@ -122,63 +122,105 @@ var TracesSampleableTableConfig = SampleableTableConfig{
...
+		}
+	}
+	return results
+}
+
+func getHttpAttributes(attributes map[string]string) map[string]string {
+	httpAttributes := getSubAttributes(attributes, "http.")
+	delete(httpAttributes, "http.response.body")
+	delete(httpAttributes, "http.request.body")
+	delete(httpAttributes, "http.url")
```
Vadman97, line 178: makes sense, this is removing these keys from `TraceAttributes` since they will now be on the dedicated column. shouldn't we do this in `getSubAttributes` for all of the attributes we bring out? or are we just starting with these because these are the largest keys (so they will  have the most impact on being removed from TraceAttributes)

Vadman97 (approved):

backend/clickhouse/traces.go
mayberryzane: the `delete` here is just removing them from `httpAttributes` and not `TraceAttributes` - I'll be removing all of these from `TraceAttributes` as the last step in this migration but for now we need to dual write since we'll still be referencing `TraceAttributes` until these new columns are backfilled and new code to query the new columns is deployed

mayberryzane: and I'm giving these 3 their own column because they're like >70% of all ingested data. a little hacky but should have a large impact for queries that aren't referencing the request/response body

Vadman97: > I'll be removing all of these from TraceAttributes as the last step in this migration but for now we need to dual write

ahh makes sense, ty for explainng

PR #9673, opened 2024-11-27
add otel attribute columns to db

mayberryzane:
## Summary
- adds new columns to `traces` / `traces_sampling` tables

- clicktested to check that these new columns are written to and traces are queryable

```
step 1: add new fields to ClickhouseTraceRow for writes
> step 2: create new columns in traces / traces_sampling tables
step 3: backfill new columns for existing data
step 4: reference new columns in queries
step 5: remove these attributes from TraceAttributes
```

- no


backend/clickhouse/traces.go
```diff
@@ -122,63 +122,105 @@ var TracesSampleableTableConfig = SampleableTableConfig{
...
+		}
+	}
+	return results
+}
+
+func getHttpAttributes(attributes map[string]string) map[string]string {
+	httpAttributes := getSubAttributes(attributes, "http.")
+	delete(httpAttributes, "http.response.body")
+	delete(httpAttributes, "http.request.body")
+	delete(httpAttributes, "http.url")
```
Vadman97, line 178: makes sense, this is removing these keys from TraceAttributes since they will now be on the dedicated column. shouldn't we do this in getSubAttributes for all of the attributes we bring out? or are we just starting with these because these are the largest keys (so they will have the most impact on being removed from TraceAttributes)

Vadman97 (approved):

backend/clickhouse/migrations/000131_add_otel_columns.up.sql
```diff
@@ -0,0 +1,30 @@
...
+    ADD COLUMN IF NOT EXISTS ProcessAttributes Map(LowCardinality(String), String),
+    ADD COLUMN IF NOT EXISTS OsAttributes Map(LowCardinality(String), String),
+    ADD COLUMN IF NOT EXISTS TelemetryAttributes Map(LowCardinality(String), String),
+    ADD COLUMN IF NOT EXISTS WsAttributes Map(LowCardinality(String), String),
+    ADD COLUMN IF NOT EXISTS EventAttributes Map(LowCardinality(String), String),
+    ADD COLUMN IF NOT EXISTS DbAttributes Map(LowCardinality(String), String);
+ALTER TABLE traces_sampling_new_mv
+MODIFY QUERY
+SELECT *
+FROM traces;
```
Vadman97, line 30: TIL, so you need to recreate this even when the MV is `SELECT *`

backend/clickhouse/traces.go
mayberryzane: https://github.com/highlight/highlight/pull/9672#discussion_r1861275770

Vadman97: ty!

PR #9674, opened 2024-11-27
fix crash on billing page

mayberryzane:
## Summary
- new format expects data to be stored under `seriesKey`
- should change the type signature of `data` to prevent this

- clicktested locally

- no

- no


Vadman97:
thanks 🚢

PR #9681, opened 2024-12-02
Allow spaces around search operators

ccschmitz:
## Summary

Allows users to use space characters around filter operators. e.g. `span_name=gorm.Query` can now be `span_name = gorm.Query` and behave the same as before.

**Before**
Screenshot 2024-12-04 at 3 37 34 PM

**After**
Screenshot 2024-12-04 at 3 37 20 PM

Play around with different query formats and ensure the behavior of the query language works as expected (returns the correct results) and that the syntax highlighting in the UI is correct for all queries you can throw at it.

N/A - all existing queries should still work as expected, just expanding the syntax.

N/A


backend/parser/parser_test.go
```diff
@@ -50,6 +50,11 @@ func TestWildcardSearch(t *testing.T) {
 	assert.Equal(t, "SELECT * FROM t WHERE SpanName ILIKE '%asdf%' AND ServiceName ILIKE '%-graph%'", sql)
 }
 
+func TestSpacesInSearch(t *testing.T) {
+	sql, _ := buildSqlForQuery("span_name !=  KafkaWorkersOnStrike")
+	assert.Equal(t, "SELECT * FROM t WHERE NOT (toString(SpanName) = 'KafkaWorkersOnStrike')", sql)
```
Vadman97, line 55: nice :D

frontend/src/components/Search/Parser/listener.ts
```diff
@@ -110,10 +115,13 @@ export class SearchListener extends SearchGrammarListener {
 	}
 
 	exitKey_val_search_expr = (_ctx: Key_val_search_exprContext) => {
-		this.currentExpression.value = this.currentExpression.text.substring(
-			this.currentExpression.key.length +
-				this.currentExpression.operator.length,
+		// everything after the key, operator, and whitespace
+		const regex = new RegExp(
+			`^${this.currentExpression.key}\\s*${this.currentExpression.operator}\\s*(.*)$`,
```
Vadman97, line 120: not sure what the culprit is, but there seems to be a new flicker in the dropdown after a value is entered
https://www.loom.com/share/d5ef948087244c18b0da8292b6792ede

Vadman97 (approved):
looks good other than the question in the loom, thanks!

frontend/src/components/Search/Parser/listener.ts
```diff
@@ -128,6 +161,10 @@ export class SearchListener extends SearchGrammarListener {
 		this.expressions.push(this.currentExpression)
 		this.currentExpression = { ...DEFAULT_EXPRESSION }
 	}
+
+	// exitEveryRule = (ctx: ParserRuleContext) => {
+	// 	console.log('::: exitEveryRule', ctx.constructor.name, ctx.getText())
+	// }
```
ccschmitz, line 167: TODO: Cleanup!

frontend/src/components/Search/Parser/listener.ts
ccschmitz: See https://github.com/highlight/highlight/pull/9681#issuecomment-2532068336 - believe this is resolved!

PR #9707, opened 2024-12-10
escape utf8 characters in key

mayberryzane:
## Summary
- `btoa` breaking when group has utf8 characters - use `encodeURIComponent`

- clicktested locally

- no

- no


Vadman97 (approved):
works great in preview, ty for the quick fix

PR #9709, opened 2024-12-10
Extend duration we store attribution cookie

ccschmitz:
## Summary

According to the [js-cookie docs on the `expires` attribute](https://github.com/js-cookie/js-cookie?tab=readme-ov-file#expires):

> Define when the cookie will be removed. Value must be a Number which will be interpreted as days from time of creation or a Date instance. If omitted, the cookie becomes a session cookie.

This means our current cookie is being set as a session cookie. We want to store this longer than that.

Also updates our logic for storing UTM params to always store any available.

N/A

N/A

### DIFF
--- a/frontend/src/util/attribution.ts
+++ b/frontend/src/util/attribution.ts
@@ -7,7 +7,10 @@ export const setAttributionData = () => {
 	const originalReferrer = Cookies.get('referrer')
 
 	if (referrer && !originalReferrer) {
-		Cookies.set('referrer', referrer, { domain: 'highlight.io' })
+		Cookies.set('referrer', referrer, {
+			domain: 'highlight.io',
+			expires: 365,
+		})
 	}
 }
 
--- a/highlight.io/utils/attribution.ts
+++ b/highlight.io/utils/attribution.ts
@@ -80,6 +80,7 @@ export const setAttributionData = () => {
 	identify(clientID, referrer)
 	Cookies.set('referrer', JSON.stringify(referrer), {
 		domain,
+		expires: 365,
 	})
 
 	return referrer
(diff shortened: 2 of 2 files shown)


frontend/src/util/attribution.ts
```diff
@@ -7,7 +7,10 @@ export const setAttributionData = () => {
 	const originalReferrer = Cookies.get('referrer')
 
 	if (referrer && !originalReferrer) {
-		Cookies.set('referrer', referrer, { domain: 'highlight.io' })
+		Cookies.set('referrer', referrer, {
+			domain: 'highlight.io',
+			expires: 365,
```
Vadman97, line 12: i think this value needs to be unix millis? see other places we use `Cookies.set`. so something like
```
        const expires = new Date()
        expires.setTime(expires.getTime() + 15 * 60 * 1000)
        Cookies.set('referrer', referrer, {
	    domain: 'highlight.io',
	    expires,
        })
```

highlight.io/utils/attribution.ts
```diff
@@ -80,6 +85,7 @@ export const setAttributionData = () => {
 	identify(clientID, referrer)
 	Cookies.set('referrer', JSON.stringify(referrer), {
 		domain,
+		expires: 365,
```
Vadman97, line 88: same

Vadman97 (approved):

frontend/src/util/attribution.ts
ccschmitz: @Vadman97 according to the docs, I think it's days from time of creation until it expires:

> Value must be a [Number](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number) which will be interpreted as days from time of creation or a [Date](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date) instance.

PR #9710, opened 2024-12-10
fix: nextjs sdk proxy configuration

Rutik7066:
## Changes
1. The `getHighlightConfig` function also returns `configureHighlightProxy`. 
2. In the <HighlightInit/> component, set the `backendUrl` to `/highlight-events` if `configureHighlightProxy` is enabled in the next config. The properties of the next config can be accessed using the `getInfo` function from `next/config`.

/claim #9042


sdk/highlight-next/src/next-client.tsx
```diff
@@ -22,10 +22,15 @@ export function HighlightInit({
 				(hostname) => !window.location.hostname.includes(hostname),
 			)
 
+		const optionsWithBackendUrl = {
+			...highlightOptions,
+			backendUrl: '/highlight-events',
```
Vadman97, line 27: great, but we want this to only happen if we can detect that `withHighlightConfig(...)` is set up. perhaps the `withHighlightConfig` wrapper can emit a `process.env` NEXT_PUBLIC value that is picked up here?

Vadman97 (changes requested):

sdk/highlight-next/src/next-client.tsx
Rutik7066: Can't we directly do it in `withHighlightConfig(...)` itself

Rutik7066:
@Vadman97 I updated the changes. I think we can merge this now.

Rutik7066:
@Vadman97 Just pinging to know status of the pr

Rutik7066:
@Vadman97  waiting for your review...

PR #9715, opened 2024-12-11
update otel user interaction instrumentation

Vadman97:
## Summary

Fixes bug in `user-instrumentation.ts` debounce code which would incorrectly call the listener.
This was causing libraries relying on event listeners to break.

* Updates `user-instrumentation.ts` based on latest opentelemetry webjs repo.

https://www.loom.com/share/32282d20f7d348ea932b09cae20e8da9

changeset

no

### DIFF
--- a/.changeset/light-elephants-pull.md
+++ b/.changeset/light-elephants-pull.md
@@ -1,0 +1,10 @@
+---
+'highlight.run': patch
+---
+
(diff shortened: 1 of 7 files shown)


SpennyNDaJets (approved):

SpennyNDaJets (approved):
Code LGTM!

PR #9726, opened 2024-12-13
Sync timestamp charts

SpennyNDaJets:
## Summary
When hovering over one chart, we should should the same points on other charts on the dashboard. Making it easier to relate charts to each other.

Screenshot 2024-12-13 at 12 39 17 PM

1. Create multiple line charts on the same dashboard.
2. Hover over one of the line charts
- [ ] Hovers over each of the other line charts at the same timestamp

N/A

N/A

### DIFF
--- a/frontend/src/pages/Graphing/components/BarChart.tsx
+++ b/frontend/src/pages/Graphing/components/BarChart.tsx
@@ -100,6 +100,7 @@ export const BarChart = ({
 		onMouseMove,
 		onMouseUp,
 		onMouseLeave,
+		onMouseOver,
 	} = useGraphCallbacks(
 		xAxisMetric,
 		setTimeRange,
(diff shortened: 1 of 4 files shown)


frontend/src/pages/Graphing/components/BarChart.tsx
```diff
@@ -107,93 +109,102 @@ export const BarChart = ({
...
-					type="number"
-					hide={showYAxis === false}
-				/>
-
-				{showGrid && (
-					<CartesianGrid
-						strokeDasharray=""
-						vertical={false}
-						stroke="var(--color-gray-200)"
+		<span onMouseOver={onMouseOver}>
```
SpennyNDaJets, line 112: `onMouseEnter` does not always fire due to quickly moving over and blocking child elements, result in the tooltip not being displayed. Recharts charts and ResponsiveContainer does not support `onMouseOver`, so add a wrapper to detect when the mouse enters the chart

frontend/src/pages/Graphing/components/BarChart.tsx
```diff
@@ -107,93 +109,102 @@ export const BarChart = ({
...
+			<ResponsiveContainer ref={chartRef}>
+				<RechartsBarChart
+					data={data}
+					syncId={syncId}
```
SpennyNDaJets, line 116: Other change in this file - the rest is whitespace changes

ccschmitz (approved):
📈

PR #9739, opened 2025-01-07
Update Python / OTEL Copy

jay-khatri:
## Summary

Update the copy for the Python / OTEL / Logging text so it reads nicer in the tutorial / docs. 

Before this change it looked like this:
image


SpennyNDaJets (approved):

Vadman97 (approved):
🚢 rebase on `origin/main` to fix e2e test. will merge manually as this is highlight.io only change

PR #9744, opened 2025-01-07
fix node-fetch import breaking next edge runtime compatibility

Vadman97:
## Summary

`node-fetch` is not edge runtime compatible, but we only need to use the types.
Removes a runtime type assertion to remove the JS implementation requirement.

Address next.js 15 compatibility from app router wrappers. 
Fixes dependency on `@prisma/instrumentation` breaking with new prisma 6.x

Fixes HIG-5025
Fixes HIG-5056 (see https://github.com/prisma/prisma/issues/25978)

[testing with e2e next.js app edge runtime deploy and with latest sdk changes](https://nextjs-highlight-demo-vkorolik-highlight-run.vercel.app/)

Screenshot from 2025-01-14 14-18-06

changeset

no


SpennyNDaJets (approved):
Code LGTM

sdk/highlight-node/package.json
```diff
@@ -27,7 +27,7 @@
 		"access": "public"
 	},
 	"dependencies": {
-		"@prisma/instrumentation": "^5.22.0",
+		"@prisma/instrumentation": "^6.2.1",
```
ccschmitz, line 30: @Vadman97 would we actually want to make this more relaxed and lower the version to something like `>=5.0.0`? Also, do you think it makes sense to convert this to a peer dependency?

Vadman97: @ccschmitz would that require us doing a dynamic import in the SDK, conditional on the `@prisma/instrumentation` package being available?

ccschmitz: If it's a peer dependency, wouldn't this just make it less likely to run into version conflicts? I _think_ the way this would work is that if they don't have the package installed it would install it for them, and it would reach for the latest version. If they do have it installed already it would use that version. By referencing an earlier version I think we would be more lenient and avoid potential conflicts with earlier versions of the package being installed in the consumer.

Vadman97: hmm, maybe i don't understand the difference between a peer dependency of `"@prisma/instrumentation": ">=5.0.0"` and a normal dependency of `"@prisma/instrumentation": ">=5.0.0"`, as once i moved this to a peer dep i have a typescript error because `@highlight-run/node` does not have `@prisma/instrumentation` installed

`TS2307: Cannot find module @prisma/ instrumentation or its corresponding type declarations.`

ccschmitz: Oh, right... I think the `peerDependencies` behavior is different between yarn and npm. I think npm >=7 installs them by default, but yarn does not.

PR #9750, opened 2025-01-08
update kafka connect task definition

Vadman97:
## Summary

Removes the static port binding on the kafka connect worker definition
to allow multiple instances to run on the same physical EC2 host.

copied from new production deployed task definition

no, same as what is now manually deployed

no

### DIFF
--- a/deploy/kafka-connect-task.json
+++ b/deploy/kafka-connect-task.json
@@ -3,18 +3,9 @@
 	"containerDefinitions": [
 		{
 			"name": "kafka-connect",
-			"image": "173971919437.dkr.ecr.us-east-2.amazonaws.com/highlight-kafka-connect:latest",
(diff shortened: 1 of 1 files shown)


mayberryzane:
@Vadman97 fyi I increased the kafka connect ecs task to 64gb memory last night as it was OOMing

mayberryzane (approved):

Vadman97:
thanks @mayberryzane - saw that. might be better to reduce the max batch size then because that's larger than all the other tasks and makes it harder to balance the ec2 boxes. typically the service autoscales when there is more load (avoiding an OOM because there are more workers so the max batch size is never hit) but yesterday the port conflict and lack of ASG autoscaling meant we weren't able to spin up more instances, causing the OOM to be repeatedly hit.

i'll play with the batch size with this PR to keep it stable with 32gbs

PR #9795, opened 2025-01-19
load last viewed product for app.highlight.io

ysadamt:
## Summary

- Fixes HIG-5043
- `app.highlight.io` now loads last viewed product (only sessions, logs, traces, logs)

https://www.loom.com/share/2fd67afba4134b5d90052f7ee043b972?sid=b6acbddd-65ca-41a1-9892-2572e807ab04

No

No


frontend/src/routers/AppRouter/AppRouter.tsx
```diff
@@ -115,13 +124,30 @@ export const AppRouter = () => {
...
+		}
+	}, [location.pathname, setLastVisitedRoute])
+
+	useEffect(() => {
+		if (isLoggedIn) {
+			if (lastVisitedRoute && isValidLastVisitedRoute(lastVisitedRoute)) {
+				navigate(lastVisitedRoute, { replace: true })
+			}
+		}
+		// eslint-disable-next-line react-hooks/exhaustive-deps
```
ysadamt, line 148: I couldn't find a way to put `navigate` in the dependency array without breaking the functionality - I believe the `navigate` function keeps re-rendering?

I have tried putting
```
if (isLoggedIn) {
			if (lastVisitedRoute && isValidLastVisitedRoute(lastVisitedRoute)) {
				navigate(lastVisitedRoute, { replace: true })
			}
		}
```		
 in a `useCallback` but the issue still persists.
 
 Any ideas on not using `eslint-disable-next-line` would be greatly appreciated!

frontend/src/routers/AppRouter/AppRouter.tsx
```diff
@@ -115,13 +124,30 @@ export const AppRouter = () => {
...
 	const { projectId } = useNumericProjectId(previousLocation)
 	const [nextParam] = useQueryParam('next', StringParam)
 	const [configurationIdParam] = useQueryParam('configurationId', StringParam)
 	const isVercelIntegrationFlow = !!nextParam || !!configurationIdParam
 	const navigate = useNavigate()
 	const isValidProjectId = Number.isInteger(Number(projectId))
 
+	useEffect(() => {
+		if (isValidLastVisitedRoute(location.pathname)) {
+			setLastVisitedRoute(location.pathname)
```
Vadman97, line 138: this may end up saving filters into the last visited url that will be reapplied, creating a confusing experience. for instance https://app.highlight.io/1/logs/MjAyNS0wMS0yMVQxNjoxMTozNVosNGY2M2NmNTYtZWJiMi00ZTNlLTgyNzQtNWVkODlkMDIwYjU1 will be saved with the specific log line from the `view in context` button because the log is part of the path rather than query string (i noticed you are already correctly dropping the QS to avoid saving other filter params like https://preview.highlight.io/1/sessions?end_date=2025-01-08T05%3A59%3A59.999Z&start_date=2025-01-06T06%3A00%3A00.000Z). 

since this is specific to the products, perhaps we truncate the pathname as well to just save the `/sessions`, `/errors`, etc. value without a sub-path?

Vadman97 (approved):
just a nit about how the logs page url is saved, but this works great in [preview](https://preview.highlight.io/1/sessions?%7Er_preview=%7B%22action%22%3A%22start%22%2C%22mode%22%3A%22production%22%2C%22region%22%3A%22ewr%22%2C%22variantId%22%3A%2201JJ4STKZ2A2KE1F79GC4N6EBT%22%2C%22variantData%22%3A%22%7B%5C%22type%5C%22%3A%5C%22branch%5C%22%2C%5C%22branch%5C%22%3A%5C%22load-last-viewed-product%5C%22%2C%5C%22githubOwnerName%5C%22%3A%5C%22highlight%5C%22%2C%5C%22githubRepositoryName%5C%22%3A%5C%22highlight%5C%22%7D%22%7D).

PR #9796, opened 2025-01-20
Rework "Getting Started" docs to be broken down by language at the top level

ysadamt:
## Summary

Closes GRO-172

Vercel preview

No

No


Vadman97 (approved):
nice, looks great so far! let's remove the existing files so git can track the file renames, then we should be good to go

Vadman97:
awesome! should be good to go. just did some testing of the redirects and found the following that need to be added

(all of the dotnet ones)
https://www.highlight.io/docs/getting-started/backend-tracing/dotnet-4
https://www.highlight.io/docs/getting-started/backend-tracing/dotnet
https://www.highlight.io/docs/getting-started/backend-sdk/dotnet-4
...

https://www.highlight.io/docs/getting-started/backend-logging/http

we can address this in a follow up; it does look like all of those links pages work on the new version, just need to add some more redirects

PR #9798, opened 2025-01-21
avoid sending alert modification notifications for pagerduty

Vadman97:
## Summary

Creating / updating alerts would trigger a notification to pagerduty resulting in a false alarm.
Avoid sending modification alerts to `steelhouse.pagerduty.com` emails.

checking filtering logic
Screenshot 2025-01-22 at 18 33 56
Screenshot 2025-01-22 at 18 34 10

no

no

### DIFF
--- a/backend/alerts/v2/destinations/email/messages.go
+++ b/backend/alerts/v2/destinations/email/messages.go
@@ -3,6 +3,8 @@ package emailV2
 import (
 	"context"
 	"fmt"
+	"github.com/samber/lo"
+	"strings"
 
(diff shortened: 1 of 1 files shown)


ccschmitz (approved):
Code looks good, but wondering, should we be doing this for all pagerduty alerts (remove the company-specific subdomain)?

Vadman97:
great call @ccschmitz - for some reason i thought steelhouse was the general pagerduty ingestion endpoint, but looks like it's customer specific

PR #9802, opened 2025-01-21
Fix table latency on dashboards

SpennyNDaJets:
## Summary
Some dashboard tables are causing frontend performance issues. Optimize by:
- virtualize the table
- memoize data sorting and filtering
- break out row into new component to avoid re-rendering
- remove tooltips for titles

1. Create a table or view a table with lots of rows
2. Load the page
- [ ] No lagging

N/A

N/A


Vadman97 (approved):
nice, way better perf from the table https://www.loom.com/share/116b7017c6d54b2fa84734594bff661d

PR #9812, opened 2025-01-23
sql builder for graphing editor

mayberryzane:
## Summary
- sets up clickhouse readonly user w/ row policies
- transforms sql string by parsing it as a select statement and replacing columns / tables / other expressions as needed
- handles the `$time_interval` macro
- sql builder uses the monaco editor with lots of the features (line numbers, minimap) turned off
- adds the sql builder to `GraphingEditor` for highlight admins

- clicktested

- more frontend work to be done for styling, autocomplete, and error handling

- @julian-highlight - I will let you know when this is merged in


backend/alerts/predictions/predictions.go
Vadman97, line 1: unrelated fixes? just trying to understand, all good to leave them here

backend/clickhouse/clickhouse.go
```diff
@@ -37,29 +41,37 @@ func GetPostgresConnectionString() string {
...
-			log.WithContext(context.Background()).WithField("Open", stats.Open).WithField("Idle", stats.Idle).WithField("MaxOpenConns", stats.MaxOpenConns).WithField("MaxIdleConns", stats.MaxIdleConns).Debug("Clickhouse Connection Stats")
-			hmetric.Gauge(context.Background(), "clickhouse.open", float64(stats.Open), nil, 1)
-			hmetric.Gauge(context.Background(), "clickhouse.idle", float64(stats.Idle), nil, 1)
+			for _, c := range []driver.Conn{conn, connReadonly} {
+				stats := c.Stats()
+				name := lo.Ternary(c == conn, "conn", "connReadonly")
+				log.WithContext(context.Background()).WithField("Open", stats.Open).WithField("Idle", stats.Idle).WithField("MaxOpenConns", stats.MaxOpenConns).WithField("MaxIdleConns", stats.MaxIdleConns).Debugf("Clickhouse Connection Stats - %s", name)
+				hmetric.Gauge(context.Background(), "clickhouse.open", float64(stats.Open), nil, 1)
+				hmetric.Gauge(context.Background(), "clickhouse.idle", float64(stats.Idle), nil, 1)
+			}
```
Vadman97, line 62: could send a tag here
```suggestion
tags := []attribute.KeyValue{attribute.String("name", name)}
				hmetric.Gauge(context.Background(), "clickhouse.open", float64(stats.Open), tags, 1)
				hmetric.Gauge(context.Background(), "clickhouse.idle", float64(stats.Idle), tags, 1)
			}
```

backend/clickhouse/migrations/000135_create_readonly_user.up.sql
```diff
@@ -0,0 +1,25 @@
+CREATE USER IF NOT EXISTS highlight_readonly IDENTIFIED BY '' SETTINGS readonly = 2;
```
Vadman97, line 1: TIL 2 means settings can be changed but only for self

backend/clickhouse/migrations/000135_create_readonly_user.up.sql
```diff
@@ -0,0 +1,25 @@
...
+GRANT SELECT ON errors_joined_vw TO highlight_readonly_role;
+GRANT SELECT ON sessions TO highlight_readonly_role;
+GRANT SELECT ON sessions_joined_vw TO highlight_readonly_role;
+GRANT SELECT ON session_events_vw TO highlight_readonly_role;
+GRANT SELECT ON fields TO highlight_readonly_role;
+GRANT SELECT ON traces TO highlight_readonly_role;
+GRANT SELECT ON traces_sampling_new TO highlight_readonly_role;
+GRANT SELECT ON logs TO highlight_readonly_role;
+GRANT SELECT ON logs_sampling TO highlight_readonly_role;
+CREATE ROW POLICY IF NOT EXISTS error_groups_readonly ON error_groups USING ProjectID = getSetting('SQL_highlight_project_id') TO highlight_readonly_role;
```
Vadman97, line 14: do these not have to be `AS RESTRICTIVE` to limit access?

backend/clickhouse/query.go
```diff
@@ -1051,6 +1283,165 @@ func (client *Client) getSamplingStats(ctx context.Context, tables []string, pro
...
+	builderInfo, err := builderFromSql(config, input)
+	if err != nil {
+		return nil, err
+	}
+	applyBlockFilter(builderInfo.Builder, input)
+	addAttributes(config, builderInfo.Fields, input.ProjectIDs, input.Params, builderInfo.Builder, true)
+
+	sql, args := builderInfo.Builder.BuildWithFlavor(sqlbuilder.ClickHouse)
+
+	sql += fmt.Sprintf(" SETTINGS SQL_highlight_project_id=%d", input.ProjectIDs[0])
```
Vadman97, line 1296: just checking, but because this comes last, even if the user sets this in the query this would be overriden?

frontend/src/pages/Graphing/GraphingEditor.tsx
```diff
@@ -606,6 +615,8 @@ export const GraphingEditor: React.FC = () => {
 		bucketBySetting,
 		fetchedLimitMetric,
 		expressions,
+		sqlEnabled,
+		sql,
```
Vadman97, line 619: different x axis behavior

https://www.loom.com/share/3c099c860ac14787b9820a920f4a2309

Vadman97 (approved):
💪 💪 💪

Vadman97:
im sure there will be more to test / debug, but best to start playing with it in prod

backend/alerts/predictions/predictions.go
mayberryzane: converted `b.BucketMax` and `b.BucketMin` to pointers since they're not returned in the sql response - this is just handling the conversion

backend/clickhouse/migrations/000135_create_readonly_user.up.sql
mayberryzane: `PERMISSIVE` / `RESTRICTIVE` are to detemine how policies interact with other policies - in this case only one policy applies per table so there's no difference

backend/clickhouse/query.go
mayberryzane: in the end, I'm doing 3 things:
- adding `SQL_highlight_project_id` setting via the context
- making sure `SETTINGS` clauses are parsed and then ignored / not included in the result query
- returning an error for any query containing `SQL_highlight_project_id` - the `SETTINGS` clause only allows literals so as far as I can tell it's not possible to construct this dynamically (e.g. `concat('SQL_highli','ght_project_id')

frontend/src/pages/Graphing/GraphingEditor.tsx
mayberryzane: the data returned is expected in this case - interpolating / extrapolating datapoints is possible for simple queries but tricky in general, so the response will not include any buckets for which there is no data. I'm handling this on the UI side instead of returning "empty" buckets in the response - I already changed the x axis to be numeric to handle gaps in data, will change the domain start/end to be the timestamps

backend/clickhouse/migrations/000135_create_readonly_user.up.sql
mayberryzane: 👍 also updated this to `readonly = 1` with `highlight_readonly_role` as `CHANGEABLE_IN_READONLY` to limit changes to other settings

PR #9816, opened 2025-01-24
fix frontend rendering of missing webvitals

Vadman97:
## Summary

Sessions like /26425/sessions/owVTdWNf043ha1xxtUwlhN5gu8RF
would crash rendering because they were missing web vitals and the new
query path would require web vitals to be present.

[reflame](https://preview.highlight.io/26425/sessions/owVTdWNf043ha1xxtUwlhN5gu8RF)
checking web vitals loading correctly

no

no

### DIFF
--- a/backend/private-graph/graph/schema.resolvers.go
+++ b/backend/private-graph/graph/schema.resolvers.go
@@ -5782,6 +5782,10 @@ func (r *queryResolver) WebVitals(ctx context.Context, sessionSecureID string) (
 	if err != nil {
 		return nil, err
 	}
+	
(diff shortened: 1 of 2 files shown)


SpennyNDaJets (approved):
Doesn't crash which is better! Do we still want this empty event?

Screenshot 2025-01-24 at 11 38 59 AM

Vadman97:
@SpennyNDaJets good call, i'm gonna investigate this further once this is merged - there is definitely still some bug resulting in the metrics not being found in the first place. the backend change should fix the empty web vitals rendering once merged

PR #9824, opened 2025-01-24
add basic opentelemetry metrics support to highlight.run

Vadman97:
## Summary

Introduces OTLP metrics export to the highlight.run SDK.
Adds a `H.gauge` function to supercede the legacy `pushMetrics` GraphQL export.
Migrates existing auto-reported metrics to the new export format.

Screenshot from 2025-01-26 19-33-48

Screenshot from 2025-01-26 19-46-38

changeset

no

### DIFF
--- a/sdk/client/src/otel/exporter.ts
+++ b/sdk/client/src/otel/exporter.ts
@@ -2,8 +2,14 @@ import { OTLPTraceExporter } from '@opentelemetry/exporter-trace-otlp-http'
 import { OTLPMetricExporter } from '@opentelemetry/exporter-metrics-otlp-http'
 import { MAX_PUBLIC_GRAPH_RETRY_ATTEMPTS } from '../utils/graph'
 import { ExportResult, ExportResultCode } from '@opentelemetry/core'
+import { AggregationTemporalityPreference } from '@opentelemetry/exporter-metrics-otlp-http/build/src/OTLPMetricExporterOptions'
 
-type ExporterConfig = ConstructorParameters<typeof OTLPTraceExporter>[0]
+export type TraceExporterConfig = ConstructorParameters<
+	typeof OTLPTraceExporter
+>[0]
+export type MetricExporterConfig = ConstructorParameters<
+	typeof OTLPMetricExporter
+>[0]
 
 // This custom exporter is a temporary workaround for an issue we are having
 // with requests stalling in the browser using the sendBeacon API. There is work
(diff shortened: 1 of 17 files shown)


sdk/client/src/otel/user-interaction.ts
```diff
@@ -1,3 +1,5 @@
+/// <reference types="zone.js" />
```
ccschmitz, line 1: What does this do? Fix some type errors? Is this why we needed to add zone.js as a dependency?

sdk/client/src/otel/exporter.ts
```diff
@@ -17,33 +21,65 @@ type SendOnErrorCallback = Parameters<OTLPTraceExporter['send']>[2]
...
 			headers: {}, // a truthy value enables sending with XHR instead of beacon
 		})
 	}
 
-	send(
-		items: ReadableSpan[],
-		onSuccess: () => void,
-		onError: SendOnErrorCallback,
-	): void {
+	export(items: any, resultCallback: (result: ExportResult) => void) {
```
ccschmitz, line 32: Didn't realize the difference between `send` and `export`. Definitely makes sense to have this retry logic here instead of `send` - thanks!

sdk/firstload/src/index.tsx
```diff
@@ -361,6 +368,9 @@ const H: HighlightPublicInterface = {
 	},
 	metrics: (metrics: Metric[]) => {
 		try {
+			for (const m of metrics) {
+				H.gauge(m)
+			}
```
ccschmitz, line 373: Doesn't `recordMetric` already do this in its implementation? Would this result in duplicate metrics since we're calling `gauge` here and inside `recordMetric`?

ccschmitz (approved):
Looks great to me! A couple minor questions, but nothing that needs to hold this up.

ccschmitz:
Was also curious, are we planning to add support for other metric types (histogram and sum)?

sdk/client/src/otel/exporter.ts
Vadman97: this change came from the otel client-side upgrade; i believe they changed some of the API in the latest version

sdk/firstload/src/index.tsx
Vadman97: good catch!

Vadman97:
> Was also curious, are we planning to add support for other metric types (histogram and sum)?

yup! adding the rest of the metric types - will also update the notion

sdk/client/src/otel/user-interaction.ts
Vadman97: had type errors after updating the otel packages - i think they were previously importing zone.js as a transitive dep, but that's no longer the case. this seems to make typescript understand that zone.js is a dependency, but i'm not too sure. if you have an idea of how to do this better, lmk

PR #9835, opened 2025-01-27
show errors in sql editor

mayberryzane:
## Summary
- fixes a couple bugs around saving / loading graphs w/ sql
- displays backend parsing / runtime errors inside the code editor
  - if message is formatted with line and column numbers, underline only those lines. else underline the entire text
Screenshot 2025-01-27 at 12 40 33 PM
Screenshot 2025-01-27 at 12 35 46 PM

- clicktested locally

- will follow with some styling updates so the errors and suggestions tooltips don't look so out of place

- @julian-highlight updated styling in attached loom video


mayberryzane:
https://www.loom.com/share/571d4a01c36e465a993c06e85f323428

backend/clickhouse/migrations/000139_create_readonly_user.up.sql
Vadman97, line 1: line 1 is what caused the deploy to fail in CH cloud (but not locally) - maybe we make a note of that in this migration in case this happens in a future self-hosted CH version?

Vadman97 (approved):
tested locally - https://www.loom.com/share/a7c47b1163eb44a9bf9a8ce30f1ba0de
follow ups can be addressed separately
* error popover on the red squiggly takes a while to appear, have to hover for a whole second
* metrics not working
* noticed my cursor position in the editor reset to the beginning at one point while i was typing, but wasn't able to reproduce

PR #9848, opened 2025-01-29
improve sql editor autocomplete by querying keys

mayberryzane:
## Summary
- previously, was querying keys once without a query to return a list of the most popular keys
- this will re-query keys using the current word
- support metrics queries
Screenshot 2025-01-28 at 5 34 06 PM

- clicktested locally
https://www.loom.com/share/9a79b10f23af4751a912d6e3d7ce00a4

- no

- no


ccschmitz (approved):
Nice!

PR #9858, opened 2025-01-30
Onboarding followups

SpennyNDaJets:
## Summary
Make the following updates to the new onboarding flow:
- more margin for headers
- title-case headers
- update MUX icon
- cache icons to avoid flicker when selecting
- update colors on healthcheck

1. View the new onboarding docs
Note: new icons won't be cached until new landing page is deployed

N/A

N/A


frontend/src/pages/Connect/ConnectPage/index.tsx
Vadman97, line 1: not sure if we are tracking this separately, but did we also want to update the nav bar to move the `connect` button  to the top level out of the ... menu?

vercel.json
```diff
@@ -1,5 +1,16 @@
...
+	"headers": [
+		{
+			"source": "/images/(.*)",
+			"headers": [
+				{
+					"key": "Cache-Control",
+					"value": "public, max-age=3600"
+				}
+			]
+		}
```
Vadman97, line 14: looks like this doesn't fix the issue. this might also be a bad idea in terms of vercel network transfer costs (because we are fetching the image without optimization)

we could try the next.js optimized version (eg. https://www.highlight.io/_next/image?url=%2F_next%2Fstatic%2Fmedia%2Fsession-replay.abf97e7c.webp&w=3840&q=75&dpl=dpl_J79fBF8iLhW8eEuULM8LwRdrHcBM) but that still feels hacky. ideally, could we bundle these images into the frontend build so they are served from `app.highlight.io/...`?

https://www.loom.com/share/2f8ddaaa01834915aa92660d0a4d28fa

Vadman97 (changes requested):

vercel.json
Vadman97: ah, saw the note `Note: new icons won't be cached until new landing page is deployed` - makes sense since the header isn't applied yet.

Vadman97: still, i would be curious to see if it's easy to host the images directly from app.highlight.io

SpennyNDaJets: Yeah, I can try and move them to app.highlight.io, since I don't believe they are being used on the highlight.io site either

SpennyNDaJets: @Vadman97 I moved the images to app.highlight.io - its working locally, but TBD if anything needs to be done for render/reflame

highlight.io/components/QuickstartContent/QuickstartContent.tsx
```diff
@@ -140,21 +140,12 @@
...
-}
-
 export type QuickStartContent = {
 	title: string
-	metaTitle?: string
 	subtitle: string
-	logoUrl?: string
+	logoKey?: string
 	entries: Array<QuickStartStep>
+	// TODO(spenny): make this required once old quickstarts are cleaned up
```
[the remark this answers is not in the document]
SpennyNDaJets: Will fix in follow up

PR #9864, opened 2025-01-31
fix og image generation

Vadman97:
## Summary

Fixes OpenGraph image generation for highlight.io blog and docs by making
the code compatible with Next.js 15.

generating og images
image
image

no

no


SpennyNDaJets (commented):
Code makes sense to me!

Looks like there may be some additional places we import the old utils, such as `highlight.io/pages/api/og/customer/[slug].tsx` (see render deployment)

SpennyNDaJets (approved):
LGTMPS

PR #9871, opened 2025-02-03
fix project ids sent by frontend with member project level access

Vadman97:
## Summary

Corrects an issue in the new member invite modal that would send
all project IDs for a `MEMBER` invite when specific projects were selected.

Screenshot 2025-02-03 at 09 42 39
Screenshot 2025-02-03 at 09 42 57

no

no

### DIFF
--- a/frontend/src/pages/WorkspaceTeam/components/InviteMemberModal.tsx
+++ b/frontend/src/pages/WorkspaceTeam/components/InviteMemberModal.tsx
@@ -45,7 +45,6 @@ function InviteMemberModal({
 	workspaceInviteLinks?: any
 	toggleShowModal: (value: boolean) => void
 }) {
-	const emailInputRef = useRef<HTMLInputElement>(null)
 	const form = Form.useStore<{
 		email: string
 		role: AdminRole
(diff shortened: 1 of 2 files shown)


frontend/src/pages/WorkspaceTeam/components/InviteMemberModal.tsx
```diff
@@ -76,6 +88,8 @@ function InviteMemberModal({
 				? (allProjects?.map((p) => p?.id).filter(Boolean) as string[])
 				: newProjectIds
 
+		console.log('vadim', { projectIdsToSend, newAdminRole })
```
ccschmitz, line 91: Cleanup?

frontend/src/pages/WorkspaceTeam/components/InviteMemberModal.tsx
```diff
@@ -223,6 +242,42 @@ function InviteMemberModal({
...
+																.includes(
+																	p?.name ??
+																		'',
+																),
+														)
+														?.map((p) => p?.id)
+														?.filter(Boolean) as
+														| string[]
+														| undefined) ?? [],
+												)
```
ccschmitz, line 279: Makes sense, but seems like a lot of extra logic for something we should be able to grab directly from the form state on submission 🤔 - wondering if we could actually remove this `newProjectIds` and `setNewProjectIds` entirely and just use the state stored in the form rather than React state.

Don't worry about refactoring right now to get this bug fixed, but something to consider at some point.

ccschmitz (approved):
Looks good! A quick thought on the state management for the form, but def not blocking.

frontend/src/pages/WorkspaceTeam/components/InviteMemberModal.tsx
Vadman97: good call - cleaned this up!

PR #9874, opened 2025-02-03
support session and error j/k shortcuts crossing page boundary

Vadman97:
## Summary

`j` and `k` buttons would switch between next/previous sessions and error groups.
When a page boundary is encountered, the next/previous resource would not be loaded
as the frontend relies on the contents of the current page to include the resource.

Updates the logic to rely on the resource index, loading the desired page to find the necessary resource.
The logic can be summarized as follows:

```
Instead of thinking in terms of “page 1, element 5”, you maintain a single state that represents the absolute index of the selected item in the overall list. For example, if each page shows 10 items, then:
•	The first page contains indices 0–9.
•	The second page contains indices 10–19.
•	And so on.
* When the user hits the next key:
•	You increment the absolute index.
•	Then you calculate the page number based on that index.
•	Finally, you ensure that the page is loaded (or fetch it if needed) and that the correct element is highlighted.
```

[reflame preview](https://preview.highlight.io/1/sessions/tw5mCgimzXI0pHIYPlDwaF6XX6v2?%7Er_preview=%7B%22action%22%3A%22start%22%2C%22mode%22%3A%22production%22%2C%22region%22%3A%22ewr%22%2C%22variantId%22%3A%2201JK6B0GQK08K1X9HEY83ZB3CM%22%2C%22variantData%22%3A%22%7B%5C%22type%5C%22%3A%5C%22branch%5C%22%2C%5C%22branch%5C%22%3A%5C%22sup-127-fix-session-player-jk-shortcuts-not-crossing-page-boundary%5C%22%2C%5C%22githubOwnerName%5C%22%3A%5C%22highlight%5C%22%2C%5C%22githubRepositoryName%5C%22%3A%5C%22highlight%5C%22%7D%22%7D)

(results changing bug shown in loom fixed via 4d916afc074232ef82da40b574b6398a387aec71 and 6ef233c1d11e84a4982242d705c06ce319c9dde7)
https://www.loom.com/share/ca87cbf52cbc4a36b6b25ec502a6f325

no

no

### DIFF
--- a/frontend/src/components/KeyboardShortcutsEducation/KeyboardShortcutsEducation.tsx
+++ b/frontend/src/components/KeyboardShortcutsEducation/KeyboardShortcutsEducation.tsx
@@ -340,11 +340,11 @@ export const PlayerKeyboardShortcuts: ShortcutItem[] = [
 	DevToolsShortcut,
 	{
 		description: `Play the next session`,
-		shortcut: ['shift', 'N'],
+		shortcut: ['J'],
 	},
 	{
 		description: `Play the previous session`,
-		shortcut: ['shift', 'P'],
+		shortcut: ['K'],
 	},
 	{
 		description: `Decrease the playback speed`,
--- a/frontend/src/components/Search/SearchContext.tsx
+++ b/frontend/src/components/Search/SearchContext.tsx
@@ -46,6 +46,7 @@ interface SearchContext extends Partial<ReturnType<typeof useSearchTime>> {
 	resetMoreResults: () => void
 	histogramBucketSize?: DateHistogramBucketSize
 	pollingExpired: boolean
+	changeResultIndex?: (index: number) => void
 	aiMode: boolean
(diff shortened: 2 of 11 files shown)


SpennyNDaJets (changes requested):
Looking good! One thing I noticed is if there is a query on the sessions/errors page, then the query gets removed when pressing `J` or `K`, which seems like a regression

Vadman97:
> Looking good! One thing I noticed is if there is a query on the sessions/errors page, then the query gets removed when pressing `J` or `K`, which seems like a regression

doesn't look like a regression (app.highlight.io prod has this bug right now) but good catch - i'll fix it while i'm at it

Vadman97:
> Looking good! One thing I noticed is if there is a query on the sessions/errors page, then the query gets removed when pressing `J` or `K`, which seems like a regression

thanks for catching this @SpennyNDaJets - should be fixed now.

SpennyNDaJets (approved):

PR #9915, opened 2025-02-04
fix search on web vitals

mayberryzane:
## Summary
- add some logging
- fix case for search on null value

- testing in preview

- no

- no


SpennyNDaJets (approved):
LGTM

PR #9918, opened 2025-02-05
use next smallest bucket window when total time range is too small for the requested bucketing

mayberryzane:
## Summary
- when requested bucket size >= the time range, use the max supported size from `DEFAULT_TIME_PRESETS`

- clicktested locally

- no

- no


frontend/src/pages/Graphing/components/Graph.tsx
```diff
@@ -1185,6 +1186,28 @@ const Graph = ({
...
+				lastPreset = preset
+			} else {
+				break
+			}
+		}
+
+		bucketByWindow = moment
+			.duration(lastPreset.quantity, lastPreset.unit)
+			.asSeconds()
+	}
```
Vadman97, line 1209: https://www.loom.com/share/7b3be8b848b9411e891ffd7f987402a6

Vadman97 (approved):

frontend/src/pages/Graphing/components/Graph.tsx
mayberryzane: able to repro this in preview as well. I think this is just the max buckets limitation - when you change bucket by to 1 second, it only loads the last 4 minutes of data (240 max buckets), and it looks like this metric is only reported every minute causing there to be only 3/4 data points in the graph

Vadman97: oh, i misunderstood the PR (it's for the opposite case of the bucketing being too large for the time range, not for the case of the bucketing being too small to display all data)

Vadman97: 🚢

PR #9948, opened 2025-02-07
Public roadmap touchups

SpennyNDaJets:
## Summary
Update the public roadmap view to take into account the changes on the Github side of the roadmap

View the road map in the preview at  `/docs/general/roadmap` 

Screenshot 2025-02-07 at 8 34 00 PM

N/A

N/A

### DIFF
--- a/highlight.io/components/common/Roadmap/Roadmap.tsx
+++ b/highlight.io/components/common/Roadmap/Roadmap.tsx
@@ -39,24 +39,19 @@ export const Roadmap = (content: { content: RoadmapProps }) => {
 			<div className="border-[1px] border-divider-on-dark my-10 w-[300px]" />
 			<div className="flex flex-col xl:flex-row w-full gap-16 xl:gap-0">
 				<RoadmapColumn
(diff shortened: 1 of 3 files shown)


SpennyNDaJets:
> * Where is all the stuff that we've shipped? Should we have anything in there, and maybe consider grouping things we've shipped in the past into quarterly or yearly buckets?
> * Maybe unrelated to this PR and more about how we're using the GH project, but I wonder if we should worry about these groups like `OTel Support + Integrations` and `Product Polish`. I doubt users care about that stuff as much as they do the individual features like "Rust SDK" or "Mobile Session Replay." I wonder if we should reconsider the groupings or maybe just remove them?
> * Do we have a way for user-submitted issues to get into the roadmap? Wondering if we have a good mechanism for capturing user ideas and slotting them into our roadmap.

These are all great questions - I can adjust the labels to be by quarter rather than by theme. I think we need to devote more time to figuring them out, so adding it to this card as well: https://linear.app/highlight/issue/HIG-5136/public-roadmap-improvements

PR #9966, opened 2025-02-11
add basic ability to download logs

Vadman97:
## Summary

Adds a new download button to the logs count display that allows downloading the current loaded logs as a CSV.
Screenshot 2025-02-10 at 22 52 04

[reflame preview](https://preview.highlight.io/1/sessions?%7Er_preview=%7B%22action%22%3A%22start%22%2C%22mode%22%3A%22production%22%2C%22region%22%3A%22ewr%22%2C%22variantId%22%3A%2201JKSNK581E7JYEWMEFXWYFYJW%22%2C%22variantData%22%3A%22%7B%5C%22type%5C%22%3A%5C%22branch%5C%22%2C%5C%22branch%5C%22%3A%5C%22sup-184-download-filtered-logs-as-json%5C%22%2C%5C%22githubOwnerName%5C%22%3A%5C%22highlight%5C%22%2C%5C%22githubRepositoryName%5C%22%3A%5C%22highlight%5C%22%7D%22%7D)

currently only logs; will add for traces if there is a customer use-case

yes @julian-highlight

### DIFF
--- a/frontend/src/pages/Graphing/hooks/exportGraph.ts
+++ b/frontend/src/pages/Graphing/hooks/exportGraph.ts
@@ -98,7 +98,10 @@ export const exportGraph = async (
 					return col
 						? m.isAfter(moment().subtract(10, 'year'))
 							? m.format('MM/DD/YYYY HH:mm:ss')
-							: col.toString().replaceAll(/[,;\t]/gi, '|')
+							: col
+									.toString()
+									.replaceAll(/[,;\t]/gi, '|')
+									.replaceAll(/\s+/gi, ' ')
 						: ''
 				})
 				.join(','),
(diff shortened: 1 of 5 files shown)


SpennyNDaJets (approved):
Code and click test look good!

Some follow up ideas/discussions:
- do we need the `cursor` column in the downloaded csv, or should we filter out?
- do we want to download based on columns in table vs downloading all rows?
- do we want to download all logs that search would return vs only logs that are shown on the UI?

Not too opinionated, but just wanted to share some questions incase we want to follow up later

Vadman97:
> * do we want to download based on columns in table vs downloading all rows?

good points! thanks for the feedback

* removing cursor, agreed that it's not useful
* all columns is nice since then the downloader can manipulate in excel how they want
* downloading everything matching result would be better (rather than just what's rendered) but would require a substantially different implementation (doing some sort of streaming for the backend to get all results, since your query could match millions of records, but then the file would be massive either way). i think what's rendered on the frontend is a reasonable starting point, given we've only recently heard this ask from a customer

PR #9997, opened 2025-02-21
support SQL alerts

mayberryzane:
## Summary
- add sql support for alerts
  - runs user's query across all data within the alert's interval
  - shows a table instead of graph since this is not a time series, just the current interval
  - disables anomaly detection
- update styling for alert settings page to match graphing settings (draggable sidebar, query editor, etc)

- clicktested locally, ensured alerts were firing
Screenshot 2025-02-20 at 6 12 01 PM

- no, can think about adding more time series features in future for monitoring (is pretty hard in the general case though)

- @julian-highlight fyi


Vadman97 (approved):
💪

PR #10016, opened 2025-02-25
Check error retention on last error object

SpennyNDaJets:

### DIFF
--- a/backend/private-graph/graph/schema.resolvers.go
+++ b/backend/private-graph/graph/schema.resolvers.go
@@ -5469,11 +5469,11 @@ func (r *queryResolver) ErrorGroup(ctx context.Context, secureID string, useClic
 	if err != nil {
 		return nil, err
 	}
-	if eg.UpdatedAt.Before(retentionDate) {
-		return nil, e.New("no new error instances after the workspace's retention date")
-	}
 	if err := r.loadErrorGroupFrequenciesClickhouse(ctx, eg.ProjectID, []*model.ErrorGroup{eg}); err != nil {
 		return nil, err
+	}
+	if eg.LastOccurrence.Before(retentionDate) {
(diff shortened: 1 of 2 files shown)


backend/private-graph/graph/schema.resolvers.go
```diff
@@ -5471,12 +5471,12 @@ func (r *queryResolver) ErrorGroup(ctx context.Context, secureID string, useClic
 	if err != nil {
 		return nil, err
 	}
-	if eg.UpdatedAt.Before(retentionDate) {
```
Vadman97, line 5474: this PR looks good (seems like it would fix the issue), but why is the UpdatedAt column not actually updated for the error group, even though new instances are updated? is that just a performance optimization we made to not modify that field?

in the case of the [reported error](https://app.highlight.io/126377/errors/cuKLu5qfCFSCXq9e7xW4m1LEHQ1N?page=1&query=status%3DOPEN), the UI does show that the error group has today as the last error date (guessing this uses `LastOccurrence`, but the clickhouse query for the error group shows an `UpdatedAt` of `2025-02-13 20:56:15.668145`)

Vadman97 (approved):
🚢 just thinking out loud on the explanation for this, but looks good

backend/private-graph/graph/schema.resolvers.go
SpennyNDaJets: When checking the logic, I don't see where we are updating the ErrorGroup in `HandleErrorAndGroup`, but saw us creating a fingerprint and ErrorObject. Maybe it was something we cleaned up recently, but have been struggling to find the PR there 🤔 

Either way, I think since an error group can be updated by the user, it is probably best to rely on the last instance to determine the retention policy.

PR #10022, opened 2025-02-26
fix stripe payment method check

Vadman97:
## Summary

Stripe payment method check would be too strict in validating the CVC 
(some mastercard customers may fail the check even with a valid payment method).

Screenshot 2025-02-26 at 12 07 31

checking behavior in production

clear `cannot show billing banner` for affected customer

no

### DIFF
--- a/backend/pricing/pricing.go
+++ b/backend/pricing/pricing.go
@@ -1313,8 +1313,11 @@ func (w *Worker) GetBillingIssue(ctx context.Context, workspace *model.Workspace
 	for _, paymentMethod := range i.PaymentMethodList().Data {
 		paymentMethods += 1
 		if paymentMethod.Card != nil && paymentMethod.Card.Checks != nil {
-			if paymentMethod.Card.Checks.CVCCheck == stripe.PaymentMethodCardChecksCVCCheckFail {
-				log.WithContext(ctx).WithField("customer", customer.ID).Info("stripe cvc check failed")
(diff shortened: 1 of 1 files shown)


backend/pricing/pricing.go
```diff
@@ -1313,8 +1313,11 @@ func (w *Worker) GetBillingIssue(ctx context.Context, workspace *model.Workspace
 	for _, paymentMethod := range i.PaymentMethodList().Data {
 		paymentMethods += 1
 		if paymentMethod.Card != nil && paymentMethod.Card.Checks != nil {
-			if paymentMethod.Card.Checks.CVCCheck == stripe.PaymentMethodCardChecksCVCCheckFail {
-				log.WithContext(ctx).WithField("customer", customer.ID).Info("stripe cvc check failed")
+			if paymentMethod.Card.Checks.CVCCheck == stripe.PaymentMethodCardChecksCVCCheckFail && paymentMethod.Card.Checks.AddressPostalCodeCheck == stripe.PaymentMethodCardChecksAddressPostalCodeCheckFail {
```
SpennyNDaJets, line 1316: Curious if this should be an `||` vs an `&&`?

SpennyNDaJets (approved):
Just the one comment but otherwise GTG

backend/pricing/pricing.go
Vadman97: trying to make this stricter - only fail the card if the CVC check fails AND the ZIP code check fails. otherwise consider the payment method valid (since in the mastercard instance, the card is valid even tho the CVC check fails, but the zip check passes)

PR #10026, opened 2025-02-27
setup trace context propagation through kafka

Vadman97:
## Summary

Propagate the trace context via the kafka message headers in our backend.
Helps with debugging kafka processing and demoing the distributed tracing.

image

will monitor number of spans created in prod in case this affects sampling

no


Mike-KafkaUnionPresident (commented):
Please fix the test

PR #10027, opened 2025-02-27
Snooze error alerts

SpennyNDaJets:
## Summary
Snoozed errors are still alerting

Screenshot 2025-02-27 at 10 27 46 AM

1. Create an error alert that reports to Slack
2. Cause an error
- [ ] Alerted in slack
3. Snooze the error
4. Cause the error
- [ ] Not alerted in slack
5. After snooze expires
6. Cause the error
- [ ] Alerted in slack

N/A

N/A

### DIFF
--- a/backend/clickhouse/errors.go
+++ b/backend/clickhouse/errors.go
@@ -28,6 +28,7 @@ type ClickhouseErrorGroup struct {
 	Event               string
 	Status              string
 	Type                string
+	SnoozedUntil        int64
 	ErrorTagID          int64
 	ErrorTagTitle       string
 	ErrorTagDescription string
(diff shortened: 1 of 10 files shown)


backend/clickhouse/errors.go
```diff
@@ -72,6 +73,11 @@ func (client *Client) WriteErrorGroups(ctx context.Context, groups []*model.Erro
 			Status:    string(group.State),
 			Type:      group.Type,
 		}
+		if group.SnoozedUntil != nil {
+			chEg.SnoozedUntil = group.SnoozedUntil.UTC().UnixMicro()
+		} else {
+			chEg.SnoozedUntil = -1
```
SpennyNDaJets, line 79: Couldn't set to 0 because `toDateTime64('0', 6)` is not valid, so set to `-1`

backend/clickhouse/migrations/000145_reacreate_errors_joined_vw.up.sql
```diff
@@ -0,0 +1,10 @@
+DROP VIEW IF EXISTS errors_joined_vw;
+CREATE VIEW IF NOT EXISTS errors_joined_vw AS
```
SpennyNDaJets, line 2: Looks like we do not need to recreate row policies as they are on a different table, but wanted to double check

Vadman97 (commented):
lgtm, but defer to @mayberryzane

mayberryzane (approved):

PR #10049, opened 2025-03-05
remove extra marshalling/unmarshalling in pushPayload

mayberryzane:
## Summary
- there are a few places in pushPayload that we marshal/unmarshal to transform between types; however this is expensive and circumvents the type system

- clicktested locally to make sure sessions were being recorded correctly
- observed logs to check for 

- will monitor prod after deployment - able to revert and replay the sessions queue if issues

- no

### DIFF
--- a/backend/public-graph/graph/resolver.go
+++ b/backend/public-graph/graph/resolver.go
@@ -2531,20 +2531,27 @@ type PushPayloadChunk struct {
 }
 
 func (r *Resolver) ProcessCompressedPayload(ctx context.Context, sessionSecureID string, payloadID int, data string) error {
+	querySessionSpan, _ := util.StartSpanFromContext(ctx, "public-graph.pushPayload.decompress")
+
 	reader, err := gzip.NewReader(base64.NewDecoder(base64.StdEncoding, strings.NewReader(data)))
 	if err != nil {
+		querySessionSpan.Finish(err)
 		return err
 	}
 
 	js, err := io.ReadAll(reader)
 	if err != nil {
+		querySessionSpan.Finish(err)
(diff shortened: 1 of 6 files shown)


backend/public-graph/graph/resolver.go
```diff
@@ -2622,14 +2665,18 @@ func (r *Resolver) ProcessPayload(ctx context.Context, sessionSecureID string, e
 
 	projectID := sessionObj.ProjectID
 	hasBeacon := sessionObj.BeaconTime != nil
+
+	loadWorkspaceSettingsSpan, _ := util.StartSpanFromContext(ctx, "public-graph.pushPayload.loadWorkspaceSettings")
```
Vadman97, line 2669: if you return the context and propagate it, redis child spans can be associated to this parent

Vadman97 (commented):
makes sense to me! great to finally tackle this todo lol

testing locally then will approve

Vadman97 (approved):
https://www.loom.com/share/65a7a10ece5c4760a8afa4e58cfd7221

clicktest looks good, live mode seems to not be loading for me localy, but that could be my previous raw events colliding

mayberryzane:
> https://www.loom.com/share/65a7a10ece5c4760a8afa4e58cfd7221
> 
> clicktest looks good, live mode seems to not be loading for me localy, but that could be my previous raw events colliding

Live mode working for me locally, playback is choppy as if it's applying each 2 seconds' worth of events in a single batch, I don't think that would be related to these backend changes tho:
https://www.loom.com/share/722d62d2eb9a46f9b0f401c53eaff141

PR #10050, opened 2025-03-05
Safe base64 encoding

SpennyNDaJets:
## Summary
`btoa` and `atob` are erroring out when trying to encode emojis. Safely encode none ascii characters.

https://github.com/user-attachments/assets/cd9ec8ca-6282-4b35-9da4-dfadff2dfdc9

1. Create a new graph
- [ ] Able to select an event
- [ ] Able to reload the page
2. Select an event with an emoji
- [ ] Able to select event
- [ ] Able to reload the page
- [ ] Able to save graph

N/A

N/A


Vadman97 (approved):
clicktest looks great!

SpennyNDaJets:
@Vadman97 just made a couple changes to replace `+` and `  ` characters in the encoding as they were getting pulled out of the url incorrectly before they could be decoded

PR #10051, opened 2025-03-06
convert graphing editor settings to sql

mayberryzane:
## Summary
- convert settings to SQL when switching editor modes
- https://www.loom.com/share/98f6c63ac38a4af4bc65afffb580f838

- clicktested locally

- no

- no

### DIFF
--- a/frontend/src/pages/Graphing/GraphingEditor.tsx
+++ b/frontend/src/pages/Graphing/GraphingEditor.tsx
@@ -78,7 +78,10 @@ import { GraphContextProvider } from './context/GraphContext'
 import TemplateMenu from '@/pages/Graphing/TemplateMenu'
 import { Panel } from '@/pages/Graphing/components/Panel'
 import { useGraphTime } from '@/pages/Graphing/hooks/useGraphTime'
-import { DEFAULT_SQL, SqlEditor } from '@/pages/Graphing/components/SqlEditor'
+import {
+	convertSettingsToSql,
+	SqlEditor,
+} from '@/pages/Graphing/components/SqlEditor'
 import { ActionBar } from '@/pages/Graphing/components/ActionBar'
 import {
 	AlertSettings,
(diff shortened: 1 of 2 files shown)


frontend/src/pages/Graphing/GraphingEditor.tsx
```diff
@@ -712,6 +710,14 @@ export const GraphingEditor: React.FC = () => {
 		],
 	)
 
+	const setEditor = (e: Editor) => {
+		if (e === Editor.SqlEditor) {
+			setSqlInternal(convertSettingsToSql(settings, startDate, endDate))
+			setSql(convertSettingsToSql(settings, startDate, endDate))
```
SpennyNDaJets, line 716: Ignore if I am missing something, but do we have to do this twice? Or can we do the conversion once and set to a variable?

```
const convertedSql = convertSettingsToSql(settings, startDate, endDate)
setSqlInternal(convertedSql)
setSql(convertedSql)
```

frontend/src/pages/Graphing/components/SqlEditor.tsx
```diff
@@ -70,6 +71,139 @@ export const DEFAULT_ALERT_SQL = [
...
+		.join(', ')
+
+	const selectExprs = [
+		...groupingExprs,
+		...expressions.map((e) => expressionToSql(e)),
+	]
+
+	let sql = `SELECT ${selectExprs.join(',\n\t')}\nFROM ${productTypeToTable(productType)}`
+	if (query) {
+		sql += `\nWHERE ${query}`
```
SpennyNDaJets, line 196: Only edge case I saw is if there are multiple statements without explicit `AND` and maybe missing some string, but think it should be pretty clear to the user what to fix, so don't think it is blocking since may be a complicated issue

Screenshot 2025-03-06 at 10 39 57 AM

mayberryzane: good point! I think fixing this is a little tricky, can do later

frontend/src/pages/Graphing/GraphingEditor.tsx
mayberryzane: nope I copy pasted this - will fix

PR #10067, opened 2025-03-11
fix state issue in oauth sso signin

Vadman97:
## Summary

For production, app.highlight.io and pri.highlight.io are different domains and need a cross-domain cookie.

https://www.loom.com/share/5fa72e0c27d24fa5a2f89f3a4d261bb2

no

no


mayberryzane (approved):
🚢

PR #10076, opened 2025-03-13
Debug enable business dashboards

SpennyNDaJets:
## Summary
Bug reported that they cannot create a dashboard despite only having 1 dashboard - DB confirms only 1 dashboard

- add logging
- remove unnecessary queries
- update wording on frontend to say "3 dashboards" as per the backend and [pricing page](https://www.highlight.io/pricing)

1. Create a new workspace
2. Try to create a new dashboard
- [ ] Successfully create dashboard
3. Try to create a new dashboard
- [ ] Successfully create dashboard
4. Try to create another dashboard
- [ ] Cannot create dashboard
- [ ] Receive update plan with "More than 3 dashboards are not available on your current plan."

Remove unnecessary logs after debug

N/A

### DIFF
--- a/backend/private-graph/graph/schema.resolvers.go
+++ b/backend/private-graph/graph/schema.resolvers.go
@@ -79,27 +79,23 @@ import (
 func (r *allWorkspaceSettingsResolver) EnableBusinessDashboards(ctx context.Context, obj *model.AllWorkspaceSettings) (bool, error) {
 	w, err := r.isUserInWorkspaceReadOnly(ctx, obj.WorkspaceID)
 	if err != nil {
-		log.WithContext(ctx).WithField("workspace_id", obj.WorkspaceID).Info("EnableBusinessDashboards: error querying workspace")
 		return false, err
 	}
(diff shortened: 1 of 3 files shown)


backend/private-graph/graph/schema.resolvers.go
```diff
@@ -77,23 +77,24 @@ import (
 
 // EnableBusinessDashboards is the resolver for the enable_business_dashboards field.
 func (r *allWorkspaceSettingsResolver) EnableBusinessDashboards(ctx context.Context, obj *model.AllWorkspaceSettings) (bool, error) {
-	w, err := r.isUserInWorkspaceReadOnly(ctx, obj.WorkspaceID)
```
mayberryzane, line 80: I think we should keep this check - at the moment this doesn't return anything but a boolean but the function could be changed in the future without realizing there's no permission check

mayberryzane (approved):
looks good pending `isUserInWorkspaceReadOnly`

PR #10080, opened 2025-03-13
Bump version

SpennyNDaJets:
## Summary
`Publish npm packages` is failing - left out bumping the version using this PR as a guide: https://github.com/highlight/highlight/pull/9598

N/A

N/A

N/A

### DIFF
--- a/sdk/firstload/package.json
+++ b/sdk/firstload/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "highlight.run",
-	"version": "9.11.1",
+	"version": "9.12.0",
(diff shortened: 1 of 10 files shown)


sdk/firstload/package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "highlight.run",
-	"version": "9.11.1",
+	"version": "9.12.0",
```
Vadman97, line 3: doesn't look like this was done with `yarn changeset version`. why did we need to do this manually?

Vadman97 (approved):

sdk/firstload/package.json
SpennyNDaJets: I reran with that. I think I just ran `yarn changeset` before

PR #10084, opened 2025-03-14
Zane/run query

mayberryzane:
## Summary
- quality of life improvements for the SQL editor 
- add variables to suggestions
- new errors state
- update to match designs

- clicktested locally

- no

- @julian-highlight fyi


SpennyNDaJets (commented):
Looking good! One little bug I found: when running a query with `Command - S`, I'm not able to type in any text boxes in the app 🤔 

Could be my browser or something but wanted to see if you were experiencing the same thing

SpennyNDaJets (approved):
Looks like the issue is not related to PR and currently in prod - we should create a card?

PR #10087, opened 2025-03-17
fix @highlight-run/cloudflare SDK using invalid OTel class extension

Vadman97:
## Summary

The `@highlight-run/cloudflare` SDK would use an exporter that `extends OTLPTraceExporter`
which would break in the cloudflare runtime, because that required importing JS implementation for
`@opentelemetry/otlp-exporter-base` which is not cloudflare runtime compatible.

Replace the `extends` call with a class that implements the necessary methods.

Fixes HIG-5168
Fixes HIG-5167

before
Screenshot 2025-03-17 at 13 07 04

after (expected sample error from e2e cloudflare app)
Screenshot 2025-03-17 at 13 08 19

cloudflare worker proxied html
Screenshot 2025-03-17 at 13 12 23

data coming through in highlight
Screenshot 2025-03-17 at 13 14 13

new versions released in branch
`yarn changeset add; yarn changeset version`

no

### DIFF
--- a/.changeset/great-carrots-retire.md
+++ b/.changeset/great-carrots-retire.md
@@ -1,5 +1,0 @@
----
-'@highlight-run/cloudflare': minor
----
-
-fix cloudflare sdk breaking due to invalid opentelemetry class extension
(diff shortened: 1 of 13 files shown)


SpennyNDaJets (approved):
image

PR #10089, opened 2025-03-17
push payload perf

mayberryzane:
## Summary
- in production, the `public-graph.pushPayload.parseEvents.remarshalEvents` span is 30%-40% of the total processPayload time for larger messages
- this change switches every `resolver.go` json.Marshal / Unmarshal call to use an alternate library
- change sampling logic to sample in 1/1000 of processPayload messages to ingest more processPayload spans and avoid overrepresenting larger messages

- ran locally, saw an improvement in the remarshal spans
Screenshot 2025-03-17 at 2 08 48 PM

- no, can revert if issues 

- no

### DIFF
--- a/backend/public-graph/graph/resolver.go
+++ b/backend/public-graph/graph/resolver.go
@@ -4,7 +4,6 @@ import (
 	"compress/gzip"
 	"context"
 	"encoding/base64"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"hash/fnv"
@@ -19,6 +18,8 @@ import (
 	"strconv"
 	"strings"
 	"time"
+
+	"github.com/goccy/go-json"
 
 	"github.com/PaesslerAG/jsonpath"
 	"github.com/aws/smithy-go/ptr"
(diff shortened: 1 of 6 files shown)


backend/public-graph/graph/resolver.go
```diff
@@ -2654,8 +2655,8 @@ func (r *Resolver) ProcessPayload(ctx context.Context, sessionSecureID string, e
 		util.Tag("numberOfErrors", len(errors)),
 		util.Tag("numberOfEvents", len(events.Events)),
 	}
-	// sample-in payload with many events
-	if len(events.Events) > 1_000 {
+	// sample-in 1/1000 of ProcessPayload calls
+	if env.IsProduction() && isIngestedBySample(ctx, "", 1./1000) {
```
Vadman97, line 2659: should this be based on the ctx traceID?

could also do this at a higher level, like at the kafka worker processing for all messages, or all pushpayload messages

Vadman97 (approved):

backend/public-graph/graph/resolver.go
mayberryzane: ah yes it should probably use the trace id, switching this over

PR #10092, opened 2025-03-19
only update columns which have changed, fix span kind

mayberryzane:
## Summary
- updating the session metadata is the single most expensive DB operation:
Screenshot 2025-03-18 at 6 02 34 PM
- only update the columns which need updating to avoid unnecessary writes to multiple indexes

- clicktested locally to make sure sessions were being recorded properly, processed + live mode

- no, also able to revert and replay the queue if major issues

- no


Vadman97 (approved):
💪

PR #10093, opened 2025-03-19
write fields to redis/clickhouse, fewer data sync submits

mayberryzane:
## Summary
- save fields to Redis and then ClickHouse without saving to Postgres
- add `fields_by_session` table and mv for looking up fields by session id
- remove `ResumedAfterProcessedTime` updates as that is just for debugging and doesn't seem necessary
- clean up some unnecessary? logic for excluded sessions
- only submit to the data sync queue if fields have changed (this is 20-25% of push payload time)
Screenshot 2025-03-19 at 12 57 51 PM

- local clicktesting
- tested places where fields are used - session CSV report, metadata panel, etc
- tested live + resumed sessions
- tested old user properties alert

- `fields_by_session` table has already been populated in prod
- can remove `fields` and `session_fields` tables in Postgres eventually

- no


frontend/src/pages/Alerts/SessionAlert/SessionAlertPage.tsx
```diff
@@ -689,8 +642,6 @@ const SessionAlertForm = ({
 								name={formStore.names.trackProperties}
 							>
 								<Select
-									onSearch={handleTrackPropertiesSearch}
-									options={trackPropertiesSuggestions}
```
ccschmitz, line 693: Wondering how this component works the data fetching logic removed, but thinking it doesn't really matter because this code is no longer accessible, right? I wasn't able to hit this in my click testing.

ccschmitz (approved):
Code + click test LGTM!

frontend/src/pages/Alerts/SessionAlert/SessionAlertPage.tsx
mayberryzane: yeah, I decided to just remove this - the component is unusable without the field suggestions, but will continue to display any track properties that were already selected for these deprecated alerts. The old way of searching was across all field names / values which doesn't scale well with how we've been saving them. Users wanting a user / track property alert should use the new alert form.

PR #10112, opened 2025-03-25
Version unpkg package

SpennyNDaJets:
## Summary
In the docs/tutorials, use a versioned package instead of the latest version tag for safer downloads.

N/A

N/A


Vadman97 (approved):
Pinning unpkg dependencies to a specific version instead of using the latest tag prevents unexpected breakages and ensures reproducible builds.

PR #10119, opened 2025-03-27
add e2e python to node kafka example

Vadman97:
## Summary

Adds a new e2e example app that sets up a Python Kafka producer and a Node.js Kafka consumer,
demonstrating trace context propagation between the two services.

image

no

no


ccschmitz (approved):
🆒

PR #10127, opened 2025-04-03
add initial launchdarkly integraton to highlight.run

Vadman97:
## Summary

Adds initial support for a `H.registerLD()` call to register a LaunchDarkly client as an integration with Highlight.
The integration forwards LaunchDarkly client events to highlight and vice-versa.
New [docs](https://highlight-landing-okadd5l6z-highlight-run.vercel.app/docs/general/integrations/launchdarkly-integration) covering the integration.

New unit tests added.
E2E next app validation
Screenshot 2025-04-08 at 10 18 39
Screenshot 2025-04-08 at 10 18 56

changeset

no


e2e/nextjs/src/app/components/highlight-identify.tsx
```diff
@@ -1,14 +1,24 @@
...
 
 import { H } from '@highlight-run/next/client'
+import { initialize } from '@launchdarkly/js-client-sdk'
 import { useEffect } from 'react'
 
+H.init('1', {
+	debug: { clientInteractions: true, domRecording: true },
+})
+const ldClient = initialize(process.env.NEXT_PUBLIC_LAUNCHDARKLY_SDK_KEY)
+// @ts-ignore
```
kinyoklion, line 11: We may want to add a comment explaining the ignore, or add a global type definition.

docs-content/general/7_integrations/launchdarkly-integration.md
```diff
@@ -0,0 +1,34 @@
...
+H.init("<YOUR_PROJECT_ID>", {});
+const ldClient = initialize("<YOUR_LD_CLIENT_TOKEN>");
+H.registerLD(ldClient);
+```
+
+## API
+
+### `identify()`
+
+Calling [`H.identify()`](../../sdk/client.md#Htrack) will forward the data to LaunchDarkly's `identify()`.
```
kinyoklion, line 25: We talked offline some, but we may not want to do this.

e2e/nextjs/src/app/components/highlight-identify.tsx
kinyoklion: Or, if we do, then figure out the limitations and an explanation of how to format a context.
Similar to our guides for how to use an OpenFeature context: https://github.com/launchdarkly/openfeature-java-server/?tab=readme-ov-file#examples

docs-content/general/7_integrations/launchdarkly-integration.md
ccschmitz: Yeah, I avoided doing that in https://github.com/highlight/highlight/pull/10096/files#diff-8472a4bb462a3f26d4e308c51f935ec34642690e789d5dfee93989bd4fc350ac.

Vadman97: Good catch; I had [removed this from the SDK](https://github.com/highlight/highlight/blob/d7ab0fb54dd8f0aed8886d761a0f204427b11240/sdk/client/src/integrations/launchdarkly/index.ts#L124) but forgot to update the doc.

ccschmitz (approved):
Looks like there's a docs change pending and some type errors in CI, but overall, the code makes sense and looks good to me!

e2e/nextjs/src/app/components/highlight-identify.tsx
Vadman97: Looked into this more with next.js client side rendering and found a different solution with a lazy initialization in the component. Could be pulled out into a shared context, but for the test app going to leave this a bit hacky.

sdk/highlight-run/src/integrations/launchdarkly/index.ts
```diff
@@ -0,0 +1,152 @@
...
+const FEATURE_FLAG_SCOPE = 'feature_flag'
+const FEATURE_FLAG_KEY_ATTR = `${FEATURE_FLAG_SCOPE}.key`
+const FEATURE_FLAG_PROVIDER_ATTR = `${FEATURE_FLAG_SCOPE}.provider_name`
+const FEATURE_FLAG_CONTEXT_KEY_ATTR = `${FEATURE_FLAG_SCOPE}.context.key`
+const FEATURE_FLAG_VARIANT_ATTR = `${FEATURE_FLAG_SCOPE}.variant`
+const FEATURE_FLAG_SPAN_NAME = 'evaluation'
+
+const LD_INITIALIZE_EVENT = '$ld:telemetry:initialize'
+const LD_ERROR_EVENT = '$ld:telemetry:error'
+const LD_TRACK_EVENT = '$ld:telemetry:track'
```
kinyoklion, line 24: I thought I commented on this before, but I cannot seem to find it.

We may want to either use string interpolation to add to the key, or we may want to directly pass the key.

Metrics in LD are by key, so using the same key for all track events will make them indiscernible at the moment.

Vadman97: Sounds good, so for track, something like `'$ld:telemetry:track:{track event name}'` would be better?

As for the initialize / error events, do we need to use the session id / body in the key, or would sending that in the `data` suffice?

Vadman97: Updated to interpolate as described above for the track event. If I understand correctly, the metric represents a count of occurrences of an event, so the other event types should be fine.

Vadman97:
Periodic metric emission has been removed. Going to merge this to start validating on app.highlight.io and will address follow-ups separately.

PR #10129, opened 2025-04-04
support ISS SSO login initiation

Vadman97:
## Summary

Support a login flow initiated by the SSO provider such as Okta
which will request using the `iss` query string paramater.

In these cases, the oauth middleware will redirect back to the SSO provider
initiating the OAuth2 flow.

deploying to prod w okta

no

no

### DIFF
--- a/backend/private-graph/graph/auth_client.go
+++ b/backend/private-graph/graph/auth_client.go
@@ -56,9 +56,9 @@ type OAuthClient struct {
 }
 
 type OAuthAuthClient struct {
-	store                *store.Store
-	oauthClientsByID     map[string]*OAuthClient
-	oauthClientsByDomain map[string]*OAuthClient
+	store                  *store.Store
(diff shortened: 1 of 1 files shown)


backend/private-graph/graph/auth_client.go
```diff
@@ -406,6 +410,56 @@ func (c *OAuthAuthClient) handleOAuth2Callback(w http.ResponseWriter, r *http.Re
...
+			return
+		}
+
+		// validates the redirect hostname (u.Host) against known clients
+		// to protect from arbitrary redirects
+		client := c.oauthClientsByDomain[u.Host]
+		if client == nil {
+			log.WithContext(ctx).WithField("domain", u.Host).Error("no oauth client found")
+			http.Error(w, "no oauth client found", http.StatusBadRequest)
+			return
```
Vadman97, line 444: need oauthClientsByProvider rather than by domain

mayberryzane (approved):

Issue #10216, opened 2025-07-26
Document otlp_endpoint setting in Python SDK


rgov:
It seems that one needs to set `highlight_io.H(otlp_endpoint="http://localhost:4317")` with the Python SDK when deploying a hobby instance of Highlight. This option isn't documented anywhere.

https://github.com/highlight/highlight/blob/9a9b27ab7fe3f045c03d968babbec7ac1fc77eb9/e2e/python/highlight_flask/app.py#L20

rgov:
It looks like there's more to it, otherwise I just get:

```
WARNING:opentelemetry.exporter.otlp.proto.grpc.exporter:Transient error StatusCode.UNAVAILABLE encountered while exporting logs to localhost:4137, retrying in 1s.
WARNING:opentelemetry.exporter.otlp.proto.grpc.exporter:Transient error StatusCode.UNAVAILABLE encountered while exporting traces to localhost:4137, retrying in 1s.
```

PR #10228, opened 2025-08-02
Python: Allow specifying transport for OTLP/HTTP

rgov:
## Summary

This change allows specifying an OTLP/HTTP endpoint for the Python SDK. Currently, the Python SDK always assumes OTLP/gRPC. Related: #10216

* Adds an `otlp_transport` argument to the `H()` constructor which can be either `OTLPTransport.HTTP` or `OTLPTransport.GRPC`.

* Imports both HTTP and gRPC implementations of `OTLPLogExporter` et al. from the `opentelemetry` package and selects based on the provided transport option. Note `OTLPMetricExporter` has slightly different parameters.

* Detects the transport automatically based on the port number. It falls back to OTLP/GRPC for backwards compatibility.

* The default endpoint is exposed as `H.DEFAULT_OTLP_ENDPOINT` and `H.DEFAULT_OTLP_TRANSPORT`. The existing `H.OTLP_HTTP` is still present but it is confusing since this is actually an OTLP/GRPC endpoint (and is over HTTPS not HTTP).

* Tests have been updated for the new behaviors.

I ran `pytest`. I will test against an OTLP/HTTP collector endpoint soon.

No.

No.


rgov:
Added documentation for `otlp_endpoint` and `otlp_transport` so this will now close #10216.

PR #10359, opened 2026-03-29
Archived

rbxict:


rbxict:
I re-checked the current PR surface. This change only updates the vendored rrweb subtree plus the generated bundle for the Safari snapshot fix; it is not intended to publish a separate package version from this PR, so I have intentionally not added a changeset. The remaining failing Vercel status is the external team-authorization gate rather than a code/test regression on this branch.

Issue #6775, opened 2023-10-02
Performance of canvas snapshotting on Safari is very poor


Pinpickle:
**Describe the bug**

Snapshotting a canvas element in Safari 16.x takes a long time (>20ms). This leads to a frame drop every time the canvas is snapshotted.

**To Reproduce**

Go to the Codsandbox full-screen here https://jqgtld.csb.app/ 

Look at the "Snapshot" stat in the top right corner (or in the console). On my computer it is ~20ms on Safari 16.5. On Chrome, it is between 0 and 0.3ms.

1057 x 734 (retina display, so double the dimensions)

Here's the codesandbox with the editor: https://codesandbox.io/s/dazzling-roentgen-jqgtld?file=/src/App.js

**Expected behavior**

Overhead for snapshotting to be near 0

**Screenshots**

## Safari

![CleanShot 2023-10-02 at 15 23 03 png](https://github.com/highlight/highlight/assets/3238878/1bc6d193-ea1f-4fda-ae03-7ea9c9cf9143)

## Chrome

![CleanShot 2023-10-02 at 15 23 55 png](https://github.com/highlight/highlight/assets/3238878/7a3767f7-94a0-467b-9c23-bac305507d66)

## iOS

![CleanShot 2023-10-02 at 15 27 57 png](https://github.com/highlight/highlight/assets/3238878/664fc6c4-27b8-43ce-be71-b35b7553716f)

**Additional context**

Environment:

```
System:
    OS: macOS 13.4
    CPU: (8) arm64 Apple M1
    Memory: 51.78 MB / 8.00 GB
Browsers:
    Chrome: 117.0.5938.92
    Safari: 16.5
```

Provided this also affects Safari 17 (I don't know if it does), this affects _every_ Safari-based browser (including everything on iOS). This performance drop means we had to disable canvas recording for Safari on our app which really limits its usefulness.

The larger the canvas, the longer the snapshot time, from what I can tell. Changing `canvasMaxSnapshotDimension` does not appear to make a difference. Safari doesn't even respect the options parameter for `createBitmapImage`: https://developer.mozilla.org/en-US/docs/Web/API/createImageBitmap

I'm reasonably confident this comes from the call to `createBitmapImage` here:

https://github.com/highlight/highlight/blob/ed2ea183d0d11781736e2001b01e24e55c33e8cb/frontend/src/__generated/rr/rr.js#L4266-L4269

[...]

PR #10377, opened 2026-04-13
fix(rrweb): improve safari canvas snapshot performance

Kabi10:
/claim #6775

Fixes the Safari canvas snapshot performance bug where `createImageBitmap` with `resizeWidth`/`resizeHeight` options is blocking and slow (>20ms per frame) on Safari/WebKit.

## Root cause

WebKit has a known bug where `createImageBitmap(source, { resizeWidth, resizeHeight })` performs synchronous CPU work instead of using GPU acceleration, causing frame drops during canvas recording.

## Solution

**Safari-only detection** — only apply the workaround on actual Safari/WebKit browsers (not Chrome, Edge, or Chrome for iOS):

```js
static shouldUseManualBitmapResize(win) {
  const vendor = win.navigator?.vendor ?? ""
  const userAgent = win.navigator?.userAgent ?? ""
  return vendor === "Apple Computer, Inc." &&
    userAgent.includes("Safari/") &&
    !userAgent.includes("Chrome/") &&
    !userAgent.includes("CriOS/") &&
    !userAgent.includes("Chromium/") &&
    !userAgent.includes("Edg/")
}
```

**Manual 2-step resize** — draw to an intermediate canvas first, then call `createImageBitmap` without resize options (avoids the slow WebKit code path):

```js
context.imageSmoothingEnabled = true
context.drawImage(source, 0, 0, width, height)
return createImageBitmap(resizedCanvas)
```

**Graceful fallback** — if OffscreenCanvas or getContext("2d") is unavailable, falls back to standard `createImageBitmap` with resize options.

## Advantages over other approaches

- **Surgical**: targets only Safari/WebKit, no overhead on Chrome/Firefox (other PRs apply workaround to all browsers)
- **Bug fix**: initializes `useManualBitmapResize` in constructor (field was declared but never assigned — Safari path never triggered without this)
- **Image quality**: adds `context.imageSmoothingEnabled = true` before `drawImage`
- Applies to both canvas and video element sources

Fixes #6775

### DIFF
--- a/.changeset/safari-canvas-perf.md
+++ b/.changeset/safari-canvas-perf.md
@@ -1,0 +1,5 @@
+---
+"@highlight-run/frontend": patch
+---
+
+Improve canvas snapshotting performance on Safari by pre-resizing bitmaps manually.
--- a/__generated/rr/rrweb/rr.js
+++ b/__generated/rr/rrweb/rr.js
@@ -13751,6 +13751,7 @@ var CanvasManager = class {
       this.mutationCb(mutation);
     };
     this.options = options;
+    this.useManualBitmapResize = CanvasManager.shouldUseManualBitmapResize(win);
     if (recordCanvas && sampling === "all") {
       this.debug(null, "initializing canvas mutation observer", { sampling });
       this.initCanvasMutationObserver(win, blockClass, blockSelector);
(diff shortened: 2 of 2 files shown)


Kabi10:
Closing to reduce backlog. Happy to reopen if maintainers want to review.

Issue #8635, opened 2024-05-21
Update workspace settings to not use antd components


ccschmitz:

Vadman97:
figma designs: 

https://www.figma.com/design/rdBPNzn7Klk6RG1LHUCE5B/Screen-Designs?node-id=7765-357643

itsalam:
Hi, I'd like to take this on, do you need require a deadline?

Gmin2:
can  i get assigned as this is of much work 
/attempt #8635

itsalam:
@ccschmitz, Is the main purpose to remove components out of the page component, or specifically the ones posted? It seems like the Box component from: 

```import { Box } from '@highlight-run/ui/components'```

Doesn't contains any antd components, so I'm wondering if that was intentional, and if so, why?

krishna9358:
hey @Vadman97 I just saw the issue. I am frontend developer. Is it still open so that can I can work on this ?

andygoodluck:
👋 Hi team, I'd like to claim this bounty for updating workspace and project settings to not use antd components.

**My Experience:**
- 2+ years of React/TypeScript development
- Experience with component migration and refactoring
- Familiar with various UI component libraries

**My Approach:**
1. Identify all antd component usage in workspace/project settings
2. Plan replacement with alternative components
3. Migrate components while maintaining functionality
4. Test thoroughly to ensure no regressions
5. Update documentation if needed

**Estimated Timeline:** 3-5 days

I'm ready to start immediately upon approval. Looking forward to contributing!

Best regards,
Andy (@andygoodluck)

V1ki:
Update: submitted a first small PR to start chipping away at this, per the 'one PR per page' suggestion.\n\nPR: https://github.com/highlight/highlight/pull/10313\n\nNext I'll tackle the remaining antd usage in settings-related pages.

Rohankumarsingh0101:
Hi, I see multiple PRs addressing this issue. Is the bounty still active and are there specific remaining pages/components that still need migration from antd?

vikrantwiz02:
Hi @Vadman97,

I see that V1ki is currently tackling the GitHubSettingsModal in PR #10313. To avoid duplicating work while helping clear out the remaining tech debt, I ran a search on the codebase and found that antd's Select is still being imported in a settings-adjacent file:
frontend/src/pages/ErrorsV2/GitHubEnhancementSettings/GitHubEnhancementSettingsForm.tsx

[...]

PR #10487, opened 2026-05-28
Replace session alert legacy selects

zoeysandel:
/claim #8635

## Summary
- Replaced the Session Alert page's legacy Ant Design-backed `@components/Select/Select` import with `@highlight-run/ui` `Select`.
- Preserved the existing `string[]` form state shape by normalizing selected UI Select option objects back to string values before save.
- Guarded loaded multi-select form values with empty-array fallbacks so the UI Select stays in multi-select mode when optional API fields are absent.
- Removed SessionAlert-only `.ant-select` CSS overrides that are no longer used after the migration.
- Added targeted regression coverage for the select value normalization helper.

## Validation
- `npx -y prettier@3.3.3 --check frontend/src/pages/Alerts/SessionAlert/SessionAlertPage.tsx frontend/src/pages/Alerts/SessionAlert/styles.css.ts frontend/src/pages/Alerts/SessionAlert/selectUtils.ts frontend/src/pages/Alerts/SessionAlert/selectUtils.test.ts`
- `npx -y esbuild@0.19.5 frontend/src/pages/Alerts/SessionAlert/SessionAlertPage.tsx --bundle '--external:*' --loader:.tsx=tsx --loader:.ts=ts --loader:.css=empty --format=esm --outfile=/tmp/highlight-session-alert-check.mjs --log-level=error`
- `npx -y tsx@4.19.2 -e "<selectValuesToStrings assertions>"`
- `rg -n "@components/Select/Select|TextHighlighter|notFoundContent|mode=|onSearch=|optionFilterProp|labelInValue|ant-select|selectContainer" frontend/src/pages/Alerts/SessionAlert` -> no matches
- `git diff --check`

## Security
UI-only alert configuration refactor. No auth, alert mutation, notification destination payload, threshold, or backend behavior changes.

## Demo
I could not run an authenticated local Session Alert settings page in this environment. The change is scoped to the existing selector UI and was validated with focused format, parse, legacy-pattern, helper-regression, and diff checks.

Prepared with AI assistance and manually reviewed before submission.

### DIFF
--- a/frontend/src/pages/Alerts/SessionAlert/SessionAlertPage.tsx
+++ b/frontend/src/pages/Alerts/SessionAlert/SessionAlertPage.tsx
@@ -51,28 +51,10 @@ import AlertNotifyForm from '@/pages/Alerts/components/AlertNotifyForm/AlertNoti
 import AlertTitleField from '@/pages/Alerts/components/AlertTitleField/AlertTitleField'
 import analytics from '@/util/analytics'
 
+import { selectValuesToStrings } from './selectUtils'
 import * as styles from './styles.css'
 
 const SEPARATOR = ':$&'
-
-const selectValuesToStrings = (values: unknown): string[] => {
-	if (!Array.isArray(values)) {
-		return []
-	}
-
-	return values.map((value) => {
-		if (
-			value &&
-			typeof value === 'object' &&
-			'value' in value &&
-			value.value !== undefined
-		) {
-			return String(value.value)
-		}
-
-		return String(value)
-	})
-}
 
 const SessionAlertOptions = [
 	{ title: 'New Session', value: SessionAlertType.NewSessionAlert },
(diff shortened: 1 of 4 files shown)


zoeysandel:
Short demo for this #8635 slice:

Session Alert select migration demo

What it shows:
- optional alert fields load into the migrated Select as arrays, not `undefined`;
- selected UI Select values render as tags;
- save state is normalized back to the existing `string[]` alert payload shape.

The authenticated Session Alert settings page was not available in my local environment, so this is a focused static recording of the PR behavior plus the validation listed in the PR body.

zoeysandel:
@Vadman97 quick ping because GitHub would not let me send a formal review request.

This is a narrow #8635 slice for `SessionAlertPage`: legacy Select removed, `string[]` alert payload preserved, nullish multi-select values guarded, and a short demo is attached above.

PR #10541, opened 2026-06-03
refactor(settings): replace remaining antd components with custom @highlight-run/ui components

Simultech369:
This PR completes the migration away from Ant Design (`antd`) in the settings components:

- Replaced `<Checkbox>` with the custom `<SwitchButton>` component in both `AutoJoinInput.tsx` and `AutoJoinForm.tsx`, updating the handler callbacks to use value-based boolean states.
- Replaced `<Divider>` elements with spacing `<Box>` layouts in `AutoJoinInput.tsx` and `NewProjectPage.tsx`.
- Replaced the Ant Design `<Select>` dropdown with the custom `<Form.Select>` component in `GitHubSettingsModal.tsx`, mapping the repository options structure (utilizing `id`, `name`, and `value`) and wiring it up cleanly to the `formStore` bindings.

Verified that the implementation and component options structure matches correct usage patterns (such as `AdminForm.tsx`) and compiles successfully.

### DIFF
--- a/frontend/src/pages/NewProject/AutoJoinInput.tsx
+++ b/frontend/src/pages/NewProject/AutoJoinInput.tsx
@@ -1,8 +1,6 @@
 import { useAuthContext } from '@authentication/AuthContext'
 import Tooltip from '@components/Tooltip/Tooltip'
-import { Box, Text } from '@highlight-run/ui/components'
-import { Divider } from 'antd'
-import Checkbox, { CheckboxChangeEvent } from 'antd/es/checkbox'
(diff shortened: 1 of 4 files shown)


Simultech369:
I noticed this was closed while #10557 is still open. If the intent was to split this work up, that makes sense. If you'd still like the remaining settings-side antd removals, I can break them into smaller follow-up PRs and resubmit.

Simultech369:
I am cleaning up older open PRs on my account. If this is still reviewable or useful to the project, I am happy to keep it open and make changes.

Otherwise, I will close it after a short grace period to reduce queue noise.

PR #10548, opened 2026-06-05
Remove antd tooltip type from TimelineBar

sjj79475067-max:
## Summary
- remove the `antd/es/tooltip` type-only import from `TimelineBar.tsx`
- replace it with a local `TooltipPlacement` union covering the placements used by this component: `top`, `topLeft`, and `topRight`
- keep the Popover runtime behavior unchanged

/claim #8635

## Verification
- GitHub compare confirms this PR only changes `frontend/src/pages/Player/Toolbar/TimelineIndicators/TimelineBar/TimelineBar.tsx`
- touched-file check confirms the `antd/es/tooltip` import is removed


sjj79475067-max:
Verification note: this is a type-only cleanup, so there is no visual UI delta to record. The diff removes the antd/es/tooltip import from TimelineBar.tsx and keeps the same top / topLeft / topRight placement values for the existing Popover behavior.

PR #10557, opened 2026-06-05
Remove TimelineBar Ant tooltip placement type import

Simultech369:
Description

This removes the remaining antd/es/tooltip type-only import from TimelineBar.tsx by replacing TooltipPlacement with a local placement union.

The component only uses three placements: top, topLeft, and topRight, so the local type keeps the same runtime behavior while removing the Ant Design type dependency from this player timeline slice.

Validation:

Focused ESLint passed for TimelineBar.tsx
Frontend source TypeScript check passed with test files excluded
Full frontend typecheck is blocked by existing test-global typing issues unrelated to this change


Simultech369:
I am cleaning up older open PRs on my account. If this is still reviewable or useful to the project, I am happy to keep it open and make changes.

Otherwise, I will close it after a short grace period to reduce queue noise.

